PR #14, opened 2025-12-11
feat(ai): migrate Message type to v5 format with parts array

kwakayama:
## Summary
- Update Message interface to use `parts: MessagePart[]` instead of `content: string`
- Add MessagePart type for text, tool-call, and tool-result parts
- Add `getTextFromParts()` helper function
- Update AgentRuntime to create v5 messages and convert to provider format
- Update memory system to work with parts array

## Breaking Change
`Message` type now requires `parts` array instead of `content` string:

```typescript
// Before (v4)
const msg: Message = { 
  id: "1", 
  role: "user", 
  content: "Hello" 
};

// After (v5)
const msg: Message = { 
  id: "1", 
  role: "user", 
  parts: [{ type: "text", text: "Hello" }] 
};
```

## Test plan
- [x] All AI tests pass (4 passed, 15 steps)
- [x] Type checking passes
- [x] Lint passes

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,6 +5,10 @@ on:
     tags:
       - 'v*'  # Trigger on version tags like v0.0.67, v1.0.0
 
+permissions:
+  contents: write
+  id-token: write
+
 jobs:
   publish:
     runs-on: ubuntu-latest
@@ -55,9 +59,7 @@ jobs:
 
       - name: Publish to npm
         working-directory: npm
-        run: npm publish --access public
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish --access public --provenance
 
       - name: Create GitHub Release
         uses: softprops/action-gh-release@v1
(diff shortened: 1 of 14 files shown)


kwakayama:
## Thread Resolution Summary

Addressed the following feedback:

### ✓ Fixed
- **Thread #9** (unused import): Removed unused `getTextFromParts` import from test file
- **Thread #6** (empty parts): Added documentation comment explaining that empty parts array is valid (e.g., assistant message with only tool calls, no text)

### Not Addressed (By Design)
- **ID collision concerns** (Threads #3, #4, #5, #7, #8): The `Date.now()` + suffix pattern is acceptable for message IDs since:
  1. Messages are scoped to individual agent sessions
  2. Collisions within same millisecond are rare in practice
  3. Adding random components would hurt debuggability
  4. The existing pattern matches AI SDK conventions

- **Empty parts validation** (Threads #1, #2): Empty parts arrays are valid states - an assistant message may have no text content when it only contains tool calls. The code handles this correctly.

- **P1 - Execution helpers** (Thread #10): The `MessageTransformer` and `ToolExecutionCore` in `/src/ai/agent/execution/` are **dead code** - they are exported but not imported anywhere outside their own module. The runtime.ts has its own inline implementation. These can be cleaned up in a separate PR.

Fixes pushed in commit 717965f.

PR #166, opened 2026-01-26
fix: use centralized React version for transforms

mattboon:
## Summary

- Fix React version mismatch causing "Cannot read properties of null (reading 'useContext')" error
- Use centralized `getReactVersion()` instead of reading from local `package.json`
- Update `package.json` React versions from `^19.0.0` to `^19.1.1` to match `deno.json`
- Clear `.cache/` on `deno task start` for local development

## Problem

The transform pipeline was reading React version from `package.json` via `detectProjectReactVersion()`, which returned `19.0.0`. However, `deno.json` specifies `npm:react@19.1.1`. This caused:

1. HTTP bundles cached with `npm:react@19.0.0` imports
2. Deno loading `npm:react@19.1.1` at runtime
3. Two React instances → broken context → "useContext" errors

## Solution

Use `getReactVersion()` from `package-registry.ts` which is set from project config via `setReactVersion()`. This ensures transforms use the same React version as the runtime.

## Test plan

- [x] `http://blank.preview.lvh.me:8080/` loads without errors
- [ ] Verify other projects still render correctly

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -264,8 +264,7 @@
   },
   "tasks": {
     "setup": "deno run --allow-all scripts/setup.ts",
-    "start": "deno run --allow-all scripts/server.ts",
-    "start:clean": "rm -rf .cache/ && deno run --allow-all scripts/server.ts",
+    "start": "rm -rf .cache/ && deno run --allow-all scripts/server.ts",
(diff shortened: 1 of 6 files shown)


mattboon:
https://github.com/veryfront/veryfront-renderer/pull/166

Issue #285, opened 2026-02-09
/api/flows takes 5.9s on first request (cold start)


kwakayama:
## Problem

First request to `https://flow-ops.veryfront.com/api/flows` takes 5.9 seconds.

## Likely Cause

Cold start in proxy/VFS mode involves sequential network calls:
1. `resolveProject()` — resolves slug from domain via API
2. `resolveAdapter()` — loads project config via VFS/API
3. `APIRouteHandler.initialize()` — scans VFS for API route files
4. Route handler execution — loads and runs the actual API handler via VFS

Each VFS file operation is a network call to `api.veryfront.com`. On cold start with no cache, this adds up.

## Investigation

- Compare cold vs warm response times (is this just cold start?)
- Check production logs for timing breakdown of VFS operations
- Consider parallel VFS fetching or aggressive caching

## Files

- `src/server/runtime-handler/index.ts`
- `src/server/handlers/request/api/api-handler-wrapper.ts`

kwakayama:
## Investigation Update

**Consistently slow (~6.5s) even on warm requests** — not just cold start.

### Root Cause Analysis

The flow-ops `/api/flows` endpoint reads 138 Salesforce XML files (~3.7MB total) from the `flows/` directory. In production proxy mode, every file operation is a network call to `api.veryfront.com` through the VFS layer.

**Latency breakdown:**

| Phase | Operations | Estimated Time |
|-------|-----------|---------------|
| Route discovery (exists checks) | 2-3 VFS calls | 1-3s |
| Handler module loading + transpilation | 3-6 VFS calls | 1-3s |
| Handler execution (reading 138 XML files via VFS) | 138+ VFS calls | 3-5s |

The API handler itself (`pages/api/flows.ts`) iterates over all files in `flows/`, reads each one via the VFS adapter, and parses XML. Each file read is a separate network request.

### Potential Fixes (in priority order)

1. **Batch VFS reads** — Add a `readDir` + `readFiles` batch API to reduce 138 round-trips to 1-2
2. **Cache API responses** — Add response caching at the VFS layer for static project files
3. **Pre-aggregate data** — The flow-ops app could compute summaries at build/deploy time instead of reading 138 files per request
4. **Parallel VFS fetches** — Read files concurrently instead of sequentially (quick win if the VFS adapter supports it)

Issue #132, opened 2026-01-25
RFC: Sub-second hydration - fixing the 5s TTI bottleneck


mattboon:
## Summary

Hydration on codersociety takes **5.3 seconds** to become interactive. This is unacceptable for a modern web framework. This RFC analyzes the root causes and proposes concrete fixes to achieve **sub-second hydration**.

**Result:** Unresponsive mobile nav, color mode button, and blog articles not fetching until hydration completes ~5 seconds after page load.

## Measured Timeline

```
Time (ms)  Event
─────────────────────────────────────────────────────────────────────────────
0          Navigation start
│
├─ 1,250   HTML response complete (responseEnd)
│          └─ Server render is fast ✓
│
├─ 1,274   modulepreload hints start fetching (53 modules)
│          ├─ pages/index.js ─────────────────────────────────┐
│          ├─ layouts/DefaultLayout.js ───────────────────────┤
│          ├─ icons/PhoneIcon.js ─────────────────────────────┤
│          ├─ ... (50 more preloads) ─────────────────────────┤
│          │                                                   │
│          │   ⚠️ BOTTLENECK: On-demand transforms            │
│          │   Each request triggers:                          │
│          │   1. API file fetch                               │
│          │   2. Babel/esbuild transform                      │
│          │   3. Import rewriting                             │
│          │                                                   │
├─ 1,321   React/ReactDOM start loading from esm.sh           │
│          └─ Cached, completes quickly ✓                     │
│                                                              │
├─ 1,939   First modules complete (665ms transform time!) ◄───┘
│
├─ 2,900   Second wave of modules complete (~1,600ms)
│
├─ 3,500   Third wave (Icons.js: 2,209ms transform!)
│
├─ 5,330   domInteractive
│          └─ DOM parsing blocked by module transforms
│
├─ 5,362   DOMContentLoaded fires
│          └─ Hydration script can now execute
│
├─ 5,363   jsx-runtime finally loads! ⚠️
│          └─ NOT preloaded, discovered late
│
├─ 5,768   More component modules finish
│
├─ 6,228   HeroSection.js completes
│          └─ Triggers framer-motion import
│
├─ 6,230   framer-motion waterfall begins ⚠️
│          └─ 159 separate ESM submodule requests
│          └─ esm.sh serves unbundled ESM
│
├─ ~7,000  Hydration complete (estimated)
│          └─ Page finally interactive
│          └─ Mobile nav, color mode, blog fetch NOW work
[...]

PR #297, opened 2026-02-09
perf: optimize VFS reads and hydration cold start

kojiwakayama:
## Summary

- **Preload jsx-runtime** — adds `<link rel="modulepreload">` for `react/jsx-runtime` in `<head>` to eliminate ~500ms waterfall delay (previously discovered late during module execution)
- **Parallel extension fallback** — switches sequential `.tsx → .ts → .jsx → .js → .mdx → .md` resolution to `Promise.allSettled`, reducing 6 serial round-trips (~600ms) to a single parallel batch (~100ms)
- **Extension resolution cache** — caches `pages/home → pages/home.tsx` mappings in-memory so subsequent reads skip the API resolution call entirely
- **Search limit 20 → 100** — prevents cache misses for projects with many files (e.g., flow-ops with 138 XML files)

## Impact

| Fix | Estimated Savings |
|-----|-------------------|
| jsx-runtime preload | ~500ms TTI reduction |
| Parallel extension fallback | ~500ms per extension-miss read |
| Extension resolution cache | ~100ms per repeated extensionless read |
| Search limit bump | Reduces individual API round-trips for large projects |

Addresses #285 and #132.

## Test plan

- [x] `html-shell-generator.test.ts` — jsx-runtime preload present in `<head>` before `<body>` (1 new test)
- [x] `read-operations.test.ts` — parallel priority correctness, multiple-match priority, performance benchmark with simulated latency, resolution cache hit/invalidation (5 new tests)
- [x] `client.test.ts` — searchFilesWithContent method exists (1 new test)
- [x] All 96 test steps pass across 3 test files
- [x] `deno lint` and `deno fmt` clean

### DIFF
--- a/src/html/html-shell-generator.test.ts
+++ b/src/html/html-shell-generator.test.ts
@@ -108,6 +108,26 @@ describe("html-generation/html-shell-generator", () => {
       assertStringIncludes(result, "https://cdn.example.com/lib.js");
     });
 
+    it("should preload react/jsx-runtime to eliminate waterfall delay", async () => {
+      const result = await wrapInHTMLShell(
+        "<div>Content</div>",
+        createMeta(),
+        createOptions(),
+      );
+
+      // jsx-runtime must be modulepreloaded so the browser fetches it early,
+      // rather than discovering it late when modules execute (~500ms saving)
+      assertStringIncludes(result, 'rel="modulepreload"');
+      assertStringIncludes(result, "jsx-runtime");
+      // Verify it appears BEFORE the body tag (in <head>)
+      const preloadIndex = result.indexOf("jsx-runtime");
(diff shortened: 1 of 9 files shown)


kojiwakayama:
**Re: Codex P2 — Avoid waiting on slow fallback extensions**

Good catch — fixed in 5227962b.

Replaced `Promise.allSettled` with priority-ordered `await`: all fetches still fire in parallel, but we now return as soon as the highest-priority extension resolves instead of waiting for all to settle. Added `p.catch(() => {})` to suppress unhandled rejections from skipped lower-priority promises, and a regression test proving we don't block on a never-resolving `.mdx` while `.ts` is instantly available.

PR #308, opened 2026-02-11
refactor: structured logging with component API and Open Telemetry bridge

ariskemper:
## Summary

- **Component API**: Add `component()` method to the logger, replacing 1215 bracket-prefix log calls (e.g. `logger.info("[CONFIG] ...")` → `logger.component("config").info(...)`) across 220 files — enables Grafana/Loki filtering by structured `component` field instead of regex on message text
- **Open Telemetry trace bridge**: Auto-inject `traceId`/`spanId` into every JSON log entry when an OpenTelemetry span is active, enabling full distributed tracing without manual plumbing
- **Request ID standardization**: Centralize request ID generation with `generateRequestId()` using `crypto.randomUUID()` consistently across dev server and runtime handler
- **Snake_case field aliases**: Emit `request_id`, `trace_id`, `span_id`, `duration_ms` alongside existing camelCase fields for Loki/Grafana compatibility

## Test plan

- [x] All 1104 tests pass (`deno task verify`)
- [x] Logger unit tests: 35 pass including 13 new tests for component API, trace context, and snake_case aliases
- [x] e2e binary test passes (52 steps)
- [x] Production server integration tests pass (log-guard allowlist updated for new message format)
- [x] fmt, lint, style conventions, CLI boundary, typecheck all pass


ariskemper:
## Fix: Keep component loggers request-context aware

Addressed review feedback about `component()` on the context-aware proxy eagerly resolving request context at creation time.

### Problem

Most files create component loggers as top-level constants:

```typescript
const log = serverLogger.component("ssr"); // module load time
```

The original implementation at lines 477-478:

```typescript
component(name: string): Logger {
  return withRequestLogger(base).component(name);
}
```

This called `withRequestLogger(base)` immediately — at module load, before any request context exists — returning a plain `ConsoleLogger` that **never** inherits `requestId`, `project_slug`, or trace correlation from AsyncLocalStorage. Every migrated log call would silently lose per-request observability.

### Fix

Introduced `createComponentAwareLogger()` which returns a deferred proxy — the same lazy pattern already used by `debug`/`info`/`warn`/`error`:

```typescript
function createComponentAwareLogger(base: ConsoleLogger, componentName: string): Logger {
  return {
    info(message, ...args) {
      withRequestLogger(base).component(componentName).info(message, ...args);
    },
    // ... same for debug, warn, error, time, child, component
  };
}
```

Both request-context resolution and component application happen at call time, so top-level `const log = serverLogger.component("ssr")` correctly inherits request-scoped fields when used inside `runWithRequestContextAsync`.

### Test

Added a test that reproduces the exact pattern — component logger created at module scope, used within request context:

```typescript
it("should inherit request context when component logger is created at top level", async () => {
  const topLevelLog = serverLogger.component("ssr");
  // ... later, inside runWithRequestContextAsync:
  topLevelLog.info("Rendering page");
  // entry.requestId === "req-42" ✓
  // entry.project_slug === "my-proj" ✓
});
```

ariskemper:
## Fix: Load trace bridge inside `initializeOTLP()` instead of per-entrypoint

Addressed review feedback about the trace bridge only being imported inside `production-server.ts`'s `if (import.meta.main)` block.

### Problem

The CLI `serve` flow (`cli/commands/serve/command.ts` → `runProductionServer`) calls `initializeOTLPWithApis()` but never imported the trace bridge. Spans would be active but JSON log entries silently lacked `traceId`/`spanId`, breaking log-to-trace correlation for that entrypoint.

### Fix

Moved `await import("#veryfront/utils/logger/trace-bridge.ts")` into `initializeOTLP()` in `src/observability/tracing/otlp-setup.ts`, right after successful provider registration:

```typescript
// src/observability/tracing/otlp-setup.ts — inside initializeOTLP()
initialized = true;
log.info("OpenTelemetry OTLP tracing initialized", { ... });

// Bridge trace context into the logger so every JSON log entry
// automatically includes traceId/spanId from the active span.
// Imported here (rather than per-entrypoint) so all callers of
// initializeOTLP benefit — CLI serve, production-server, proxy, etc.
await import("#veryfront/utils/logger/trace-bridge.ts");
```

Removed the now-redundant manual import from `production-server.ts`.

This way every entrypoint that initializes OTLP — `production-server.ts`, `cli/commands/serve/command.ts`, or any future caller — automatically gets the bridge without needing to remember a separate import.

**Note:** The proxy (`src/proxy/`) has its own duplicate logger with built-in `getTraceContext()` — that's a Phase 6 concern.

PR #351, opened 2026-02-13
fix: publish unique RC version on every main merge

kwakayama:
## Problem

The prerelease job runs on every push to main and does `npm publish --tag rc` with the static version from `deno.json`. This fails with E403 when the version is already published (by a previous merge or the tag-triggered release job).

## Fix

Each main merge now publishes a unique semver RC version using the GitHub Actions run number:

```
main merge  →  0.1.10-rc.42   (npm --tag rc)
main merge  →  0.1.10-rc.43   (npm --tag rc)
tag push    →  0.1.10          (npm --tag latest)
```

Every merge produces a new npm artifact and fires the `repository_dispatch` to veryfront-server for staging deployment.

Also cleans up the release job — removes the fallback publish logic since RC and stable versions no longer collide.

## Changes

- Prerelease: compute `${base}-rc.${run_number}`, patch `npm/package.json` before publish
- Prerelease: remove skip-check workaround
- Release: simplify to unconditional `npm publish`
- Release: update RC cleanup to match new `v0.1.10-rc.*` tag pattern

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -225,9 +225,14 @@ jobs:
           node-version: "22"
           registry-url: "https://registry.npmjs.org"
 
-      - name: Read version
+      - name: Compute RC version
         id: version
-        run: echo "version=$(jq -r '.version' deno.json)" >> $GITHUB_OUTPUT
+        run: |
+          BASE=$(jq -r '.version' deno.json)
+          RC_VERSION="${BASE}-rc.${{ github.run_number }}"
+          echo "base=${BASE}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


kojiwakayama (approved):
LGTM. Clean fix for the E403 publish collision.

The `run_number`-based RC versioning ensures uniqueness per merge, and the release job simplification is a nice cleanup. The RC release tag cleanup with `grep + xargs` correctly handles the new `v0.1.10-rc.*` pattern.

Minor note: the `jq ... > tmp && mv tmp original` pattern could silently produce a zero-byte file if `jq` fails on malformed JSON, but that's an edge case unlikely to hit in practice.

PR #355, opened 2026-02-14
feat: add CSRF protection with warn-by-default and easy opt-in

ariskemper:
## Summary
- Implements double-submit cookie CSRF protection that warns on startup when unconfigured and can be enabled via `security: { csrf: true }`
- Adds `CsrfHandler` (priority 5) that validates POST/PUT/PATCH/DELETE requests against cookie+header token match
- Sets CSRF cookie automatically on GET/HEAD responses through the `withSecurity()` fluent method
- Supports custom `cookieName`, `headerName`, `excludePaths`, and `ttlSec` configuration

## Test plan
- [x] New `src/security/csrf/helpers.test.ts` — 18 tests covering token generation, validation, and cookie application
- [x] New `src/security/http/csrf/csrf-handler.test.ts` — 19 tests covering disabled/enabled/custom config scenarios
- [x] Existing `src/rendering/rsc/actions/helpers.test.ts` still passes (backward-compatible re-exports)
- [x] Existing `src/security/http/middleware/config-loader.test.ts` still passes
- [x] Full unit test suite passes (947/947, 1 pre-existing unrelated failure)

### DIFF
--- a/src/cache/backends/api.ts
+++ b/src/cache/backends/api.ts
@@ -15,7 +15,9 @@ type CacheRequestContext = {
 };
 
 function getCurrentRequestContext(): CacheRequestContext | null {
-  const mod = globalThis.__vf_multi_project_adapter;
+  const mod = (globalThis as Record<string, unknown>).__vf_multi_project_adapter as
+    | { getCurrentRequestContext?: () => unknown }
+    | undefined;
   return (mod?.getCurrentRequestContext?.() as CacheRequestContext | undefined) ?? null;
 }
 
--- a/src/config/schemas/config.schema.ts
+++ b/src/config/schemas/config.schema.ts
@@ -161,6 +161,17 @@ export const veryfrontConfigSchema = z
         csp: z.record(z.array(z.string())).optional(),
         remoteHosts: z.array(z.string().url()).optional(),
         cors: corsSchema.optional(),
+        /**
+         * CSRF protection using the double-submit cookie pattern.
+         * Set `true` for defaults, or pass an object to customize.
+         *
+         * When enabled, POST/PUT/PATCH/DELETE requests must include
+         * an `x-csrf-token` header matching the `vf_csrf` cookie.
+         * The cookie is set automatically on HTML document responses.
+         *
+         * Server Actions (`/_veryfront/rsc/action`) are CSRF-protected;
+         * client code must forward the cookie value as the header.
+         */
         csrf: z.union([
           z.boolean(),
           z.object({
--- a/src/config/schemas/config.schema.ts
+++ b/src/config/schemas/config.schema.ts
@@ -161,6 +161,15 @@ export const veryfrontConfigSchema = z
         csp: z.record(z.array(z.string())).optional(),
         remoteHosts: z.array(z.string().url()).optional(),
         cors: corsSchema.optional(),
+        csrf: z.union([
+          z.boolean(),
+          z.object({
+            cookieName: z.string().optional(),
+            headerName: z.string().optional(),
+            excludePaths: z.array(z.string()).optional(),
+            ttlSec: z.number().int().positive().optional(),
+          }).strict(),
+        ]).optional(),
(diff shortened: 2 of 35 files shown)


ariskemper:
## Code Review

### HIGH

1. **Startup warning uses `debug` instead of `warn`** — `src/server/runtime-handler/index.ts:165`
   The PR title says "warn-by-default", but the message uses `logger.debug`, making it invisible in production. Should be `logger.info` at minimum to actually nudge users toward configuring CSRF.

2. **Three divergent `SecurityConfig` types** — `src/types/server.ts:21`, `src/security/http/middleware/types.ts:29`, `src/security/http/response/types.ts:14`
   These have structural differences: `server.ts` is missing `coop`/`corp`/`coep`/`hsts`/`remoteHosts`; `response/types.ts` is missing `auth`; cors origin callback return types differ (`boolean` vs `boolean | string`). Pre-existing issue, but `csrf` was added to all three — good time to consolidate.

### MEDIUM

3. **`timingSafeEqual` length leak** — `src/security/csrf/helpers.ts:58`
   Early return on `aBytes.length !== bBytes.length` leaks length info via timing. Low exploitability in double-submit (attacker can't set the cookie due to SameSite=Lax), but deviates from constant-time intent.

4. **1-hour CSRF token TTL** — `src/security/csrf/helpers.ts:36`
   Reuses `SERVER_ACTION_DEFAULT_TTL_SEC` (3600s). Users who load a page and wait >1 hour before submitting get 403. Consider a longer default (e.g., 24h) or a dedicated constant.

5. **Missing test: `/_veryfront/log/subpath` non-exemption** — `csrf-handler.test.ts`
   The `/_veryfront/log` exemption uses exact match but there's no test confirming `/_veryfront/log/evil` is still CSRF-protected.

6. **No end-to-end CSRF + Server Actions documentation**
   When `csrf: true`, Server Action POSTs to `/_veryfront/rsc/action` will require a CSRF header. The RSC client code needs to read the `vf_csrf` cookie and send `x-csrf-token`. This flow needs documentation or client-side integration.

### LOW

7. Requests without `Accept` header get CSRF cookies (`helpers.ts:109-111`) — safe default but adds unnecessary `Set-Cookie` to CLI/API clients.
8. `ASSET_PATH_RE` may skip `/about.html` (`helpers.ts:52`) — the regex skips any path with an extension, including `.html`.
9. Unnecessary `async` on synchronous handler (`csrf-handler.ts:37`) — minor; adds a needless microtask.
10. Missing tests for `accept: */*` and empty token string edge cases.

### Looks Good

- `CSRF_EXEMPT_PREFIXES` correctly protects `/_veryfront/rsc/action` (verified by test)
- Malformed cookie handling is solid (try/catch around `parseCookiesFromHeaders`)
- `withSecurity` signature change is backward-compatible; all call sites now pass `req`
- Re-export from `rendering/rsc/actions/helpers.ts` preserves backward compat
- Port allocation hardening and globalThis fix are clean ancillary fixes

kojiwakayama:
Thanks for the thorough review! All items addressed:

**aa01a98e1:**
- **#1** `logger.debug` → `logger.info` — visible in production without being noisy
- **#3** `timingSafeEqual` now pads to max length instead of early-returning on length mismatch
- **#4** Dedicated `CSRF_DEFAULT_TTL_SEC = 86400` (24h) instead of reusing the 1h server action TTL
- **#5** Added test confirming `/_veryfront/log/evil` is still CSRF-protected
- **#8** `ASSET_PATH_RE` now excludes `.html`/`.htm` via negative lookahead so HTML pages get CSRF cookies

**e602aa649:**
- **#2** Consolidated three divergent `SecurityConfig` types into single canonical definition in `types/server.ts`; middleware and response modules re-export
- **#6** Added CSRF + Server Actions client-side integration docs to `csrf-handler` module JSDoc and config schema
- **#7** Skip CSRF cookie for requests without `Accept` header or with `accept: */*` (CLI/API clients)
- **#9** Kept `async` — base class mandates `Promise<HandlerResult>`, removing it would require wrapping every return
- **#10** Added edge case tests: empty token string, no-Accept, `accept: */*`

PR #500, opened 2026-02-28
fix: prevent markdown content drift and add contentChanged handler

kojiwakayama:
## Summary
- Add `contentChanged` postMessage handler so Studio can push content updates directly to the preview (reload in non-edit mode, apply to Lexical in edit mode)
- Fix echo-back race condition where Lexical's markdown normalization (whitespace, list rebalancing) was syncing back to Yjs, overwriting the original formatting from Monaco
- Add Yjs diagnostic logging at connection, sync, observer, and message filter points for debugging sync issues
- Clean up selection state and awareness properly on Yjs disposal

## Test plan
- [ ] Open markdown file in Advanced Mode, edit in Monaco — preview should reload via `contentChanged`
- [ ] Enter edit mode, check console for `[StudioBridge] Yjs` logs showing connection lifecycle
- [ ] Edit in Lexical, verify Monaco content matches without whitespace drift
- [ ] Confirm no "echo-back" to Yjs during remote updates (only genuine local edits sync)
- [ ] `deno task typecheck` passes

### DIFF
--- a/src/studio/bridge/bridge-editor-state.ts
+++ b/src/studio/bridge/bridge-editor-state.ts
@@ -110,6 +110,8 @@ export const editorState = {
   markdownCurrentEditorContent: "",
   markdownLexicalRenderedContent: null as string | null,
   markdownApplyingRemoteUpdate: false,
+  markdownRemoteUpdateToken: 0,
+  markdownPendingLocalReconcile: false,
   markdownLastRemoteContent: null as string | null,
   markdownFrontmatter: "",
   markdownRawBlocks: [] as string[],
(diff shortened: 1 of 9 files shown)


kojiwakayama:
Re: P1 — Avoid dropping local edits during remote-update guard (discussion_r2867831357)

Good catch. Fixed in 644162959.

Removed the `markdownLastRemoteContent === null` guard entirely from `handleMarkdownLocalChange`. Echo-back prevention is already handled upstream by the Lexical update listener — it returns early when `markdownApplyingRemoteUpdate` is true and captures the edit via `pendingLocalReconcile` for replay after the remote apply settles. So `handleMarkdownLocalChange` now always syncs to Yjs when connected, eliminating the window where user edits could be dropped.

Tests updated to verify the new behavior.

PR #512, opened 2026-03-04
RFC: per-element source location attributes for inspect mode

mattboon:
## Summary

RFC proposing how to inject source file/line/column into DOM elements so Studio's inspect mode can pass accurate context to the chat agent.

**The problem**: When you click an `<h1>` in inspect mode, Studio doesn't know which file or line it came from. A `<h1>` rendered by `components/Welcome.tsx` currently reports as being in `pages/index.tsx` (the page file).

**The fix**: The Babel transform already receives `filePath` but ignores it (`_options`). We inject `data-node-file` and `data-node-name` alongside the existing `data-node-line/column` attributes. The bridge reads these and sends full source location with `setSelectedNode`.

### Target DOM output

```html
<h1 data-node-file="components/Welcome.tsx"
    data-node-name="h1"
    data-node-id="node-3"
    data-node-line="5"
    data-node-column="4"
    class="...">
  Ready to Create
</h1>
```

### Key changes proposed

1. **Path normalization** — normalize `sourceFile` to project-relative before injection
2. **Babel transform** — use `options.filePath` to inject `data-node-file` + `data-node-name`
3. **Rehype/remark transforms** — same attributes for MDX/markdown files
4. **Bridge inspector** — read per-element file/line, send full node data with `setSelectedNode`
5. **Schema update** — add optional `node` object to `setSelectedNode` message

### Scope

- DOM elements only (no component-level tracking yet)
- TSX/JSX + MDX/markdown files

### Also included

- `rfcs/002-bridge-dead-code-removal.md` — separate RFC for removing `data-vf-id`, `buildNavigatorTree`, mutation observer, and other dead code

See `rfcs/001-inspect-source-attributes.md` for full details.


kojiwakayama (commented):
## RFC Review

Well-written RFC — the problem is clear, the proposal is focused, and the dead code removal is well-justified. A few things to address:

### Factual issues

1. **`remark-node-id.ts` does NOT receive `filePath`** — Section 2 implies both rehype and remark plugins already have `filePath` access, but `remark-node-id.ts` only accepts `{ prefix?: string; includePosition?: boolean }`. This needs to be called out as new work (extending its interface + updating the callsite).

2. **`rehype-node-positions.ts` already injects `data-node-file`** — It uses `filePath` today to add a separate `data-node-file` attribute. The RFC should note this existing behavior since the migration is "consolidate `data-node-file` + `data-node-line` + `data-node-column` into `data-node-id`", not just "add filePath support".

3. **`remark-node-id.ts` injects 7 attributes, not 5** — It also adds `data-node-start` and `data-node-end` (byte offsets) that the babel plugin doesn't. These should be listed under dead code removal in section 6.

### Design questions

4. **`data-node-name` redundancy** — For plain HTML elements, `data-node-name="h1"` duplicates `element.tagName.toLowerCase()`, and the bridge code already falls back to it. Should this be conditional (only injected when the JSX name differs from the tag, e.g. components)?

5. **What happens when `filePath` is falsy?** — If `options.filePath` is undefined, the current proposal would inject `data-node-id=":5:4"`. Should we skip injection entirely, or fall back to the page path?

### Completeness

6. **Files Changed table is incomplete** — Lists 3 files but the RFC describes changes to 6+: `rehype-node-positions.ts`, `remark-node-id.ts`, plus dead code removals across `bridge-inspector.ts` (buildNavigatorTree, setupMutationObserver, etc.) and `types.ts`. Also `findElementById`/`getElementId` are mentioned in Migration but not in the table.

kojiwakayama:
Review notes from my pass on this PR:

1. **Medium**: The proposed `data-node-id` format uses raw `options.filePath` (`${options.filePath}:${line}:${col}`), but current call sites pass full filesystem paths. That risks leaking local absolute paths in DOM/message payloads and may break selector lookups on Windows unless IDs are normalized/escaped.

2. **Low**: The RFC marks `treeUpdated`/mutation-observer flow as dead code, but this repo still initializes it and defines schema for it. Removing it without confirming all Studio consumers may regress behavior.

I like the direction overall; I’d suggest explicitly defining a project-relative/normalized node-id path format and confirming `treeUpdated` consumer status before deleting that path.

kwakayama (commented):
## Review

The problem is legit — `filePath` is passed but ignored, so every element reports the page file instead of its actual source component.

### Suggestions

| # | Suggestion | Impact | Evidence |
|---|-----------|--------|----------|
| **1** | **Skip RFC, just ship code** — core fix is ~20 lines | High | `babel-node-positions.ts:131` — literally just un-ignore `_options` and use it |
| **2** | **Separate dead code removal** — `setupMutationObserver` is actively called, `treeUpdated` consumer lives in Studio (can't verify from here) | High | `bridge-init.ts:117-121` calls it on every page load; `studio.schema.ts:99` still defines it |
| **3** | **Use separate `data-node-file` attr, not compact `file:line:col`** — rehype already does this today | Med | `rehype-node-positions.ts:56-59` already ships `data-node-file` as a separate attribute |
| **4** | **Drop `data-node-name`** — redundant since scope is DOM-only (`element.tagName` exists) | Low | RFC explicitly says "no component-level tracking yet" |
| **5** | **Address schema break** — new `setSelectedNode` shape won't pass current validation | Med | `studio.schema.ts:104-107` only accepts `{ action, id }` |

mattboon:
Thanks for the thorough reviews @kojiwakayama @kwakayama! Updated the RFC to address all points:

### Changes made

**Switched to explicit separate attributes** (addresses @kwakayama #3, @kojiwakayama #4):
- Dropped the compact `file:line:col` encoding
- Using `data-node-file` + `data-node-name` alongside existing `data-node-line`, `data-node-column`, etc.
- No parsing needed — bridge reads attributes directly
- Consistent with what `rehype-node-positions.ts` already does with `data-node-file`

**Path normalization** (addresses @kojiwakayama comment #1):
- `module-server.ts` passes absolute paths (`join(projectDir, ...)`)
- `component-handling.ts` passes project-relative paths
- RFC now calls for normalizing to project-relative at the call site (option b) before passing to the transform

**Keeping `data-node-name`** (re: @kwakayama #4):
- Even for DOM-only scope, we want the name in the postMessage payload for the Studio inspect pill UI without needing a round-trip to read `tagName`

**Schema update** (addresses @kwakayama #5):
- Added step 8: update `setSelectedNode` in `studio.schema.ts` with optional `node` object

**Falsy filePath guard** (addresses @kojiwakayama #5):
- Skip injection if `options.filePath` is falsy

**remark-node-id.ts needs interface extension** (addresses @kojiwakayama #1):
- Called out as new work in step 4 — its current interface only accepts `{ prefix?, includePosition? }`

**rehype already injects data-node-file** (addresses @kojiwakayama #2):
- Noted in Current State section — this is consolidation/alignment, not greenfield

**Complete files changed table** (addresses @kojiwakayama #6):
- Now lists all 7 files

**Dead code removal → separate PR** (addresses @kwakayama #2):
- Created `rfcs/002-bridge-dead-code-removal.md` covering `data-vf-id`, `buildNavigatorTree`, mutation observer, `data-vf-text`, `data-node-start/end` byte offsets, `lastTreeSignature`
- Studio confirmed: no `treeUpdated` consumer exists

PR #515, opened 2026-03-05
feat: agent skills — auto-discovery, loading, and script execution

### LINKED ISSUE #514: feat: Agent Skills — auto-discovery, loading, and script execution for user-built agents
## Summary

Add support for **Agent Skills** — markdown-based knowledge packages that user-built agents can auto-discover and activate at runtime. Skills follow the [agentskills.io](https://agentskills.io/specification) open specification, making them portable across 20+ agent platforms.

Skills enable:
- **Expert knowledge capture** — Domain experts encode expertise into portable, shareable markdown packages
- **Agent capability extension** — Agents gain specialized knowledge without model fine-tuning
- **Progressive disclosure** — Only skill names/descriptions loaded at startup; full instructions loaded on demand

## Design

Skills live in a `skills/` directory at the project root (alongside `tools/`, `agents/`, etc.) and are auto-discovered by `discoverAll()`:

```
skills/
├── contract-review/
│   ├── SKILL.md              # YAML frontmatter + markdown instructions
│   ├── scripts/extract.py    # Optional executable scripts
│   └── references/CLAUSES.md # Optional on-demand reference docs
└── compliance-check/
    └── SKILL.md
```

### Agent Configuration

```typescript
export default agent({
  id: "legal-assistant",
  model: "anthropic/claude-sonnet-4-20250514",
  system: "You are a legal assistant...",
  skills: true,  // auto-discover all skills, or ["contract-review"] for specific ones
});
```

### Runtime Flow

1. **Discovery** — `discoverAll()` scans `skills/*/SKILL.md`, parses YAML frontmatter (~100 tokens per skill)
2. **Prompt augmentation** — Skill names + descriptions injected into agent system prompt
3. **Activation** — Agent calls `load-skill` when user request matches a skill description
4. **Execution** — Agent follows skill instructions, optionally loads references or runs scripts

### New Modules

| Module | Purpose |
|--------|---------|
| `src/skill/types.ts` | `Skill`, `SkillMetadata`, `SkillContent` types |
| `src/skill/parser.ts` | YAML frontmatter parser with validation |
| `src/skill/registry.ts` | Project-scoped skill registry (extends `ScopedRegistryFacade`) |
| `src/skill/tools.ts` | `load-skill`, `load-skill-reference`, `execute-skill-script` tools |
| `src/skill/executor.ts` | `SkillScriptExecutor` — local subprocess + cloud Sandbox backends |
| `src/skill/allowed-tools.ts` | Dual-layer tool access control (planning + execution boundary) |
| `src/skill/path-safety.ts` | Path traversal prevention for references/scripts |
| `src/skill/prompt-augmentation.ts` | System prompt augmentation with skill metadata |
| `src/discovery/handlers/skill-handler.ts` | Markdown-based skill discovery handler |

### Modified Modules

| Module | Change |
|--------|--------|
| `src/discovery/discovery-engine.ts` | Wire skill discovery into `discoverAll()` loop |
| `src/discovery/types.ts` | Add `skills` to `DiscoveryResult`, `skillDirs` to config |
| `src/agent/types.ts` | Add `skills?: true \| string[]` to `AgentConfig` |
| `src/agent/factory.ts` | Resolve skills, augment prompt, register skill tools |
| `src/agent/runtime/tool-helpers.ts` | Filter skill tools unless explicitly included |
| `src/config/schemas/config.schema.ts` | Add skill config schema |
| `src/server/dev-server/server.ts` | Include `skills/` in HMR watch dirs |

## Security

| Risk | Mitigation |
|------|------------|
| Malicious skill scripts | Execute via sandbox (local subprocess or cloud Firecracker microVM) |
| Tool-policy bypass | `allowed-tools` enforced at both planning-time and execution boundary |
| Path traversal / symlink escape | Canonical path validation, containment within skill root |
| Excessive context consumption | Token limit enforcement on SKILL.md body |

## Scope

**In scope (this PR):**
- Skill discovery, registry, and agent integration
- Three skill tools (`load-skill`, `load-skill-reference`, `execute-skill-script`)
- Script execution with local/cloud executor abstraction
- Path safety and allowed-tools enforcement
- HMR re-discovery on skill file changes
- Documentation updates

**Future work:**
- CLI commands (`veryfront skill add/remove/list/create/validate`)
- Registry integration (skills.sh, GitHub, npm)
- Skill analytics/telemetry

## References

- [agentskills.io specification](https://agentskills.io/specification)
- [skills.sh registry](https://skills.sh/) (84K+ skills)

kojiwakayama:
## Summary

Adds the **Agent Skills** system — markdown-based knowledge packages ([agentskills.io](https://agentskills.io/specification)) that user-built agents auto-discover and activate at runtime.

- New `src/skill/` module: parser, registry, executor, tools, path safety, allowed-tools enforcement
- Skills discovered from `skills/*/SKILL.md` via `discoverAll()` (same pattern as tools/agents)
- `AgentConfig.skills` option for agent opt-in (`true` or `string[]`)
- Three skill tools: `load-skill`, `load-skill-reference`, `execute-skill-script`
- Dual-mode script executor (local subprocess + cloud sandbox)
- Progressive disclosure: only names/descriptions at startup, full instructions loaded on demand
- Path traversal prevention and dual-layer allowed-tools enforcement
- HMR re-discovery on skill file changes
- Documentation and reference updates

## Test plan

- [ ] Unit tests pass for parser, registry, executor, path-safety, allowed-tools, prompt-augmentation, tools
- [ ] Skill discovery integration test passes
- [ ] Create a sample `skills/test-skill/SKILL.md` and verify agent discovers it
- [ ] Verify `load-skill` returns full instructions with progressive disclosure
- [ ] Verify path traversal attempts are rejected
- [ ] Verify allowed-tools enforcement blocks out-of-policy tool calls
- [ ] Verify HMR re-discovers skills on file changes

Closes #514


kwakayama (commented):
## Critical Review — Agent Skills PR

I went through every file in this diff line by line. The architecture is clean overall (progressive disclosure, dual-layer enforcement, path safety are all well-reasoned), but there are real bugs and design issues that should be addressed before merging. Sorted by priority:

---

### P0 — Bugs that will cause incorrect behavior in production

**1. `activeSkillPolicy` is never reset between steps — skill policy leaks across the entire request**

In both `generate()` and `stream()` code paths, `activeSkillPolicy` is set when `load-skill` returns, but is **never cleared** when a different skill is loaded in a subsequent step.

```ts
// src/agent/runtime/index.ts — generate path (~line 398)
if (tc.toolName === LOAD_SKILL_TOOL_ID && result && typeof result === "object") {
  const skillResult = result as { allowedTools?: string[] };
  activeSkillPolicy = skillResult.allowedTools;
  mustLoadSkillFirst = false;
}
```

If the agent loads skill A (allowed: `["Read"]`) then later loads skill B (allowed: `["Write"]`), the policy correctly updates. **But** if skill B has **no** `allowedTools` (i.e. `undefined`), then `activeSkillPolicy` becomes `undefined`, which means `filterToolsForSkill` and `isToolAllowedBySkill` both treat it as "no restrictions" — the agent now has access to *all* tools, bypassing any previous restriction. This is a security-relevant policy bypass.

**Evidence:** `allowed-tools.ts:82` — `if (!allowedTools || allowedTools.length === 0) return true;`

**Fix:** When switching skills, the policy should be explicitly set to the new skill's policy (which may be empty `[]` meaning "no tools allowed" vs `undefined` meaning "no restrictions"). The semantics of `undefined` vs `[]` need to be clarified and enforced.

---

**2. `CloudScriptExecutor.withTimeout` doesn't actually kill the sandbox process**

```ts
// src/skill/executor.ts ~line 87
const commandPromise = sandbox.executeCommand(cmdString);
const result = await withTimeout(commandPromise, timeoutMs);
if (result === TIMEOUT_SENTINEL) {
  return timeoutResult(timeoutMs);
}
```

When the timeout fires, `withTimeout` just races with a timer — it **does not kill** the sandbox process. The `executeCommand` call continues running in the background, holding the sandbox session open. Unlike the local executor which uses `runCommand` with `timeoutMs` (which actually sends SIGTERM/SIGKILL), the cloud path just abandons the promise. The `sandbox.close()` in `finally` is best-effort, but the process inside may keep running until the sandbox idles out.

**Evidence:** Compare with `LocalScriptExecutor` which delegates to `runCommand(command, { timeoutMs })` which triggers `createProcessTimeout` with actual process killing.

**Fix:** Either use `sandbox.executeStream` to get a handle you can abort, or issue `sandbox.executeCommand("kill ...")` before returning the timeout result.

---

**3. `getSkillScriptExecutor()` singleton is decided once at first call — never re-evaluated**

```ts
// src/skill/executor.ts:114
export function getSkillScriptExecutor(): SkillScriptExecutor {
  if (!executorInstance) {
    executorInstance = getEnv("SANDBOX_AUTH_TOKEN")
      ? new CloudScriptExecutor()
      : new LocalScriptExecutor();
  }
  return executorInstance;
}
```

The executor type is cached on first call. If `SANDBOX_AUTH_TOKEN` is set later (e.g., injected via request-scoped project-env overlay), all subsequent calls still get the `LocalScriptExecutor`. The code even has a comment acknowledging the SSR overlay issue but doesn't address it.

**Evidence:** The comment at executor.ts:53 says *"Ensure SANDBOX_AUTH_TOKEN is available in the request-scoped environment"* — but the singleton makes this impossible.

**Fix:** Either check the env var per-call (no singleton), or scope the executor to the request lifecycle.

---

### P1 — Design issues that will cause problems at scale

**4. Duplicated skill enforcement logic in `generate()` vs `stream()` — 100+ lines copy-pasted**

The `mustLoadSkillFirst` + `activeSkillPolicy` logic is copied verbatim between the generate path (~line 325-398) and the stream path (~line 564-642). This is a maintenance time bomb — any fix to one path will need to be manually mirrored. The generate path uses inline error message construction while the stream path uses `this.recordToolError()`, meaning they already diverge in error reporting behavior.

**Evidence:** Compare lines 340-393 (generate) with lines 598-654 (stream). Same logic, different error handling patterns.

**Fix:** Extract a shared method like `enforceSkillPolicy(tc, activeSkillPolicy, mustLoadSkillFirst)` that returns the policy decision, then each path applies it with its own error reporting.

---

**5. `agent()` factory mutates module-level state at definition time, not at request time**

```ts
// src/agent/factory.ts ~line 76-85
if (config.skills) {
  resolvedSkills = skillRegistry.resolveForAgent(config.skills);
  for (const registration of SKILL_TOOL_REGISTRATIONS) {
    if (!toolRegistry.has(registration.id)) {
      registerTool(registration.id, registration.create());
    }
  }
  ...
}
```

Skills are resolved and tools are registered when `agent()` is called (module evaluation time), not per-request. If skills are added/removed via HMR, agents defined before the HMR event will use stale `resolvedSkills`. The system prompt is closed over the stale `resolvedSkills` map via the closure at line ~100:
```ts
return `${basePrompt}\n\n${buildSkillManifestPrompt(resolvedSkills!)}`;
```

**Evidence:** HMR calls `discoverAll()` which repopulates the registry, but the agent's `augmentedSystem` closure still captures the old `resolvedSkills` variable.

**Fix:** The augmented system prompt should read from `skillRegistry.resolveForAgent()` at invocation time, not from a captured variable.

---

**6. `tools.test.ts` references undeclared `FileSystemAdapter` type**

```ts
// src/skill/tools.test.ts:8
function createTestSkill(fsAdapter: FileSystemAdapter): Skill {
```

There's no import for `FileSystemAdapter`. CI passes because TypeScript may infer it from the `Skill` type, but this is a missing import that will break if the type inference changes.

**Evidence:** Line 8 of `src/skill/tools.test.ts` — no import statement for `FileSystemAdapter` anywhere in the file.

---

### P2 — Robustness / correctness concerns

**7. `SKILL_ALLOWED_TOOL_PATTERN_REGEX` allows double-colon and trailing colon patterns**

```ts
export const SKILL_ALLOWED_TOOL_PATTERN_REGEX = /^[A-Za-z0-9._:-]+(?:\:\*)?$/;
```

This regex accepts patterns like `api::*`, `tool:`, and `a:b:c:*` which are unlikely to be intentional. The `:-` in the character class also means it matches a range from `:` to nothing (effectively just `:`), which happens to work but is fragile — in some regex engines, an unescaped `-` at certain positions in a character class creates a range.

**Evidence:** `"api::*"` passes `SKILL_ALLOWED_TOOL_PATTERN_REGEX.test()` — returns `true`.

**Fix:** Tighten to something like `/^[A-Za-z][A-Za-z0-9._-]*(:\*)?$/` to require a proper prefix before the wildcard and disallow leading special characters.

---

**8. `load-skill` tool re-parses `SKILL.md` on every call instead of returning cached content**

Every `load-skill` call reads the file from disk and re-parses frontmatter:
```ts
const content = await readSkillFile(skill, skillMdPath);
const parsed = await parseSkillFrontmatter(content);
```

Plus it calls `listSkillSubdir` which does directory listing twice (once for references, once for scripts). For agents that load the same skill on every turn, this is unnecessary I/O per request.

**Evidence:** `tools.ts:89-106` — no caching layer. The `Skill` object already has `metadata` from discovery but the body/instructions are not cached.

**Fix:** Consider caching `SkillContent` in the `Skill` object after first load, invalidated only on HMR.

---

**9. `mustLoadSkillFirst` ordering assumption is fragile with parallel tool calls**

The code assumes `load-skill` will be the first tool in the `toolCalls` array when the model emits it alongside other tools. However, the model may emit them in any order. The current logic iterates sequentially and rejects non-skill tools before `load-skill` is processed:

```ts
let mustLoadSkillFirst = !activeSkillPolicy &&
  Boolean(this.config.skills) &&
  response.toolCalls.some((tc) => tc.toolName === LOAD_SKILL_TOOL_ID);
```

If the model emits `[Read, load-skill]`, `Read` is rejected, then `load-skill` succeeds. On the **next** step, the model will retry `Read`. This works but wastes a round-trip.

**Evidence:** The error message says *"Call load-skill first"* but doesn't reorder — just fails the tool and moves on.

**Fix:** Consider reordering tool calls to process `load-skill` first when present, avoiding the wasted step.

---

Good work on the overall architecture — progressive disclosure, fail-closed defaults, and the dual-layer enforcement model are well-designed. The issues above are about making the implementation match that design intent.

ariskemper:
### Code Review

Found 3 issues:

1. **Missing `FileSystemAdapter` import in `tools.test.ts`** — `createTestSkill(fsAdapter: FileSystemAdapter)` uses the type but never imports it. This will fail to compile.

https://github.com/veryfront/veryfront-code/blob/bc12fcc3d5837513e496e8e848a3b4bd99c9621d/src/skill/tools.test.ts#L2-L7

2. **Cloud executor timeout doesn't terminate the sandbox process** — `CloudScriptExecutor.execute()` uses `withTimeout()` which races a `setTimeout` against the sandbox command promise. When the timeout fires, it returns `TIMEOUT_SENTINEL` but never kills the running sandbox command — it just abandons the promise. The process keeps running until `sandbox.close()` in the finally block. Compare with `LocalScriptExecutor` which properly uses `runCommand`'s `timeoutMs` for SIGTERM/SIGKILL.

https://github.com/veryfront/veryfront-code/blob/bc12fcc3d5837513e496e8e848a3b4bd99c9621d/src/skill/executor.ts#L134-L149

3. **`activeSkillPolicy` derived from tool output with no validation** — The runtime extracts `allowedTools` from the `load-skill` tool result via an unchecked `as` cast. While currently safe (the tool reads from parsed metadata), if the tool implementation changes, arbitrary values could bypass enforcement. Consider validating with `validateAllowedToolPatterns()` at this point.

https://github.com/veryfront/veryfront-code/blob/bc12fcc3d5837513e496e8e848a3b4bd99c9621d/src/agent/runtime/index.ts#L398-L403

---

Additional observations (lower priority):

- The `mustLoadSkillFirst` + `activeSkillPolicy` enforcement logic is duplicated nearly identically (~100 lines) between the non-streaming and streaming paths in `runtime/index.ts`. Consider extracting a shared helper.
- No test covers the runtime `mustLoadSkillFirst` enforcement path (tool rejected before `load-skill` succeeds).
- `getSkillScriptExecutor()` is a singleton created on first call based on `SANDBOX_AUTH_TOKEN`. If the env var is set later via project-env overlay, the singleton returns the wrong executor. `resetSkillScriptExecutor()` exists but is only called in tests.
- Logger test changes (`undefined` → `""`) are unrelated to the skill system and could be a separate commit.

PR #558, opened 2026-03-08
fix: use projectId for router detection cache to prevent cross-tenant poisoning

kwakayama:
## Summary

- **Root cause**: In multi-tenant proxy mode, all renderer pods share `projectDir="/app"` (Docker WORKDIR). The `routerDetectionCache` was keyed by `projectDir`, so when a pod cached the router type for one project, it returned that same result for all other projects — causing 500 errors when the wrong router type was applied (e.g., looking for `app/layout.tsx` when the project only has `components/layout.tsx`).
- Keys the router detection cache by `projectId` (UUID) instead of `projectDir`, with fallback to `projectDir` for local dev where each project has a unique directory.
- Updates all callers (`PageResolver`, `LayoutCollector`, `LayoutApplicator`), service factories, lifecycle, cache invalidation, and WebSocket manager to pass `projectId` through.

## Test plan

- [x] `deno task build` passes
- [x] `deno task verify:quick` passes (fmt, lint, typecheck)
- [ ] Existing `router-detection.test.ts` passes (uses `projectDir` fallback — no `projectId` needed)
- [ ] Deploy to staging and verify new projects render without 500
- [ ] Verify local dev (`deno task start`) still works without `projectId`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.49",
+  "version": "0.1.50",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/src/modules/import-map/preloader.ts
+++ b/src/modules/import-map/preloader.ts
@@ -7,24 +7,26 @@ const importMapCache = new Map<string, Promise<ImportMapConfig>>();
 export function preloadImportMap(
   projectDir: string,
   adapter: RuntimeAdapter,
+  projectId?: string,
 ): Promise<ImportMapConfig> {
-  const cached = importMapCache.get(projectDir);
+  const cacheKey = projectId ?? projectDir;
+  const cached = importMapCache.get(cacheKey);
   if (cached) return cached;
 
   const promise = loadImportMap(projectDir, adapter);
-  importMapCache.set(projectDir, promise);
+  importMapCache.set(cacheKey, promise);
 
   promise.catch(() => {
-    importMapCache.delete(projectDir);
+    importMapCache.delete(cacheKey);
   });
 
   return promise;
 }
 
 export async function getCachedImportMap(
-  projectDir: string,
+  cacheKey: string,
 ): Promise<ImportMapConfig | undefined> {
-  const cached = importMapCache.get(projectDir);
+  const cached = importMapCache.get(cacheKey);
   if (!cached) return undefined;
 
   try {
@@ -35,9 +37,9 @@ export async function getCachedImportMap(
   }
 }
 
-export function clearImportMapCache(projectDir?: string): void {
-  if (projectDir) {
-    importMapCache.delete(projectDir);
+export function clearImportMapCache(cacheKey?: string): void {
+  if (cacheKey) {
+    importMapCache.delete(cacheKey);
     return;
   }
 
--- a/src/platform/adapters/fs/veryfront/types.ts
+++ b/src/platform/adapters/fs/veryfront/types.ts
@@ -124,7 +124,7 @@ export interface InvalidationProjectContext {
 export interface InvalidationCallbacks {
   clearSSRModuleCache?: () => void;
   clearSSRModuleCacheForProject?: (projectId: string) => void;
-  clearRouterDetectionCacheForProject?: (projectDir: string) => void;
+  clearRouterDetectionCacheForProject?: (projectId: string) => void;
   clearModulePathCache?: () => void;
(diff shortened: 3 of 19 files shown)


kwakayama (commented):
## Critical / High Issues

### 🔴 CRITICAL: Same cache poisoning bug exists in other `projectDir`-keyed caches

This PR fixes `routerDetectionCache` but the **exact same class of bug** exists in at least 3 other in-memory caches that are also keyed by `projectDir` — which is always `"/app"` in multi-tenant proxy mode:

1. **`src/react/compat/version-detector/version-cache.ts`** — `projectVersionCache` keyed by `projectDir`. React version detection result for one tenant will be served to all tenants. If tenants use different React versions, this causes rendering with the wrong React compat layer.

2. **`src/server/services/rsc/endpoints/handler-registry.ts`** — RSC handlers keyed by `projectDir`. The first project's RSC handler gets cached and reused for all subsequent projects, potentially serving wrong RSC responses cross-tenant.

3. **`src/modules/import-map/preloader.ts`** — `importMapCache` keyed by `projectDir`. One tenant's import map gets cached and applied to all tenants, causing wrong module resolution.

These are the same root cause pattern — global `Map<projectDir, T>` where `projectDir` is shared across tenants. Recommend fixing all `projectDir`-keyed caches in this PR or a fast-follow, since any one of them can cause cross-tenant data leaks or 500s.

### 🟠 HIGH: Silent fallback to `projectDir` makes the bug latent

In `detectAppRouter`, the cache key is:
```ts
const cacheKey = options?.projectId ?? projectDir;
```

Since `projectId` is optional throughout the chain (`DetectAppRouterOptions`, `PageResolverOptions`, `LayoutCollectorOptions`, `RendererLifecycle`), any caller that omits it silently falls back to the known-bad `projectDir` key — reintroducing the exact cache poisoning this PR fixes, with no warning or log.

All *current* callers pass `projectId`, but this is fragile. Suggestions:
- Add a `logger.warn()` when falling back to `projectDir` in multi-tenant mode (if detectable), so the bug surfaces immediately rather than silently poisoning caches
- Or document the invariant prominently on `detectAppRouter` that `projectId` **must** be provided in multi-tenant deployments

kwakayama (commented):
## Re-review (updated diff with 3 commits)

### Previously raised — resolved ✅

**HIGH: Silent fallback** — Addressed in commit a24ad75. The `logger.warn` when `config?.fs?.type === "veryfront-api"` and `projectId` is missing is a good catch-net. 👍

### Previously raised — still open 🔴

**CRITICAL: Other `projectDir`-keyed caches still vulnerable**

No changes to the other caches. These remain susceptible to the identical cross-tenant poisoning in multi-tenant mode:

| Cache | File | Risk |
|-------|------|------|
| `projectVersionCache` | `src/react/compat/version-detector/version-cache.ts` | Wrong React version compat applied cross-tenant |
| RSC handler cache | `src/server/services/rsc/endpoints/handler-registry.ts` | RSC handler from tenant A served to tenant B |
| `importMapCache` | `src/modules/import-map/preloader.ts` | Wrong module resolution cross-tenant |

These are the same root cause as the router detection fix. If this PR ships without them, the system is still broken for multi-tenant — just in different ways. Recommend at minimum a tracked follow-up issue if not fixing in this PR.

### New observations on updated diff

No new critical/high issues introduced. The warning log implementation is clean — checking `config?.fs?.type === "veryfront-api"` is a reasonable heuristic for detecting multi-tenant mode.

One minor note: the warning fires on every `detectAppRouter` call (not just cache misses), so it could be noisy if a code path repeatedly calls this without `projectId`. Consider guarding it with a `once` flag or moving it after the cache-hit check — but this is low priority.

kwakayama (commented):
## Re-review (4 commits) — No remaining critical/high issues

All previously raised issues are addressed:

### 🔴 CRITICAL (resolved): Other `projectDir`-keyed caches
All 3 flagged caches now use `projectId ?? projectDir` as cache key:
- `projectVersionCache` in `version-cache.ts` ✅
- `importMapCache` in `import-map/preloader.ts` ✅
- RSC handler registry in `handler-registry.ts` ✅

Callers updated throughout: `SSRRenderer`, `component-loader.ts`, `LayoutOrchestrator`, `RSCHandler`, `endpoint-router.ts`.

### 🟠 HIGH (resolved): Silent fallback warning
- Warning now fires only on cache miss (moved after cache-hit check) ✅
- `warnedMissingProjectId` once-flag prevents log noise ✅

No new critical or high issues found. Ship it.

PR #603, opened 2026-03-10
fix: isolate Tailwind compiler per project to prevent CSS contamination

kojiwakayama:
## Summary

- Tailwind v4's `compile().build()` is **stateful** — it accumulates candidates across all calls
- The compiler cache was keyed by stylesheet hash only, so all projects using the same stylesheet (e.g. `@import "tailwindcss"`) shared a single compiler on each shared-pool pod
- When pod serves project A then project B, project B's CSS includes project A's utility classes (70 KB instead of 47 KB), causing inconsistent visual rendering depending on which pod the load balancer routes to
- Fix: key the compiler cache by `projectSlug:stylesheetHash` so each project gets its own compiler instance

## Root cause investigation

Confirmed via:
1. `agent-browser` visual comparison showed two alternating states on page reload — fully styled (320px textarea, purple gradient) vs degraded (45px textarea, no gradient)
2. `curl` with cache-busting showed CSS alternating between 47,426 bytes and 70,042 bytes
3. Diffing the two CSS files revealed the larger one contained utility classes from other projects (`marine-brook`, `codersociety`) served by the same pod
4. Wrote a [Deno reproduction script](https://gist.github.com/) confirming `compiler.build()` accumulates candidates: Build 1 = 4,661 bytes → Build 3 (same candidates as Build 1) = 4,800 bytes after an intervening Build 2 with different candidates

## Test plan

- [x] `deno check` passes on all 4 modified files
- [x] `deno lint` passes on all 4 modified files
- [x] `tailwind-compiler-regression.test.ts` passes (6 steps)
- [x] `pipeline.behavior.test.ts` passes (5 steps)
- [ ] Deploy to staging and verify CSS consistency across reloads on preview projects
- [ ] Monitor shared-pool pod memory — more compiler instances cached per pod (bounded by existing MAX_CACHED_COMPILERS=10)

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.53",
+  "version": "0.1.54",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/src/html/styles-builder/tailwind-compiler-cache.ts
+++ b/src/html/styles-builder/tailwind-compiler-cache.ts
@@ -85,16 +85,21 @@ function evictOldestCompiler(): void {
 
 export async function getCompiler(
   stylesheet: string,
+  projectSlug?: string,
 ): Promise<Awaited<ReturnType<typeof compile>>> {
-  const hash = hashString(stylesheet);
+  // Tailwind v4's compile().build() is stateful — it accumulates candidates
+  // across calls. Without per-project isolation, projects sharing the same
+  // stylesheet on the shared pool contaminate each other's CSS output.
+  const stylesheetHash = hashString(stylesheet);
+  const hash = projectSlug ? `${projectSlug}:${stylesheetHash}` : stylesheetHash;
 
   const cached = compilerCache.get(hash);
   if (cached) {
-    logger.debug("Compiler cache hit", { hash });
+    logger.debug("Compiler cache hit", { hash, projectSlug });
     return cached.compiler;
   }
 
-  logger.debug("Creating new compiler", { hash });
+  logger.debug("Creating new compiler", { hash, projectSlug });
 
   const tailwindBase = await getTailwindBaseCSS();
   const pluginCache = new Map<string, unknown>();
--- a/src/html/styles-builder/tailwind-compiler-regression.test.ts
+++ b/src/html/styles-builder/tailwind-compiler-regression.test.ts
@@ -59,7 +59,7 @@ describe("styles-builder/tailwind-compiler regressions", () => {
         const hash = hashCSS(generated.css);
         await cacheCSSAsync(generated.css, hash, { candidates, stylesheet });
 
-        const regenerated = await regenerateCSSByHash(hash);
+        const regenerated = await regenerateCSSByHash(hash, "vf-unified-regression");
         assertEquals(regenerated, generated.css);
       } finally {
         restoreFetch();
@@ -81,7 +81,7 @@ describe("styles-builder/tailwind-compiler regressions", () => {
         await cacheCSSAsync(generated.css, hash);
         await cacheCSSInputsAsync(hash, { candidates, stylesheet });
 
-        const regenerated = await regenerateCSSByHash(hash);
+        const regenerated = await regenerateCSSByHash(hash, "vf-legacy-fallback-regression");
         assertEquals(regenerated, generated.css);
       } finally {
         restoreFetch();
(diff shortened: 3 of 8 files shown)


kwakayama (commented):
## Re-review: fix: isolate Tailwind compiler per project to prevent CSS contamination

**Merge Score: 72/100**

---

### Verdict

The root cause analysis is excellent and the core fix (keying compiler cache by `projectSlug:stylesheetHash`) is correct. However, there's a **missed call site** that undermines the fix, and an **API design concern** that makes regressions likely.

---

### CRITICAL: `regenerateCSSByHash` is not patched — JIT regeneration still uses shared compilers

`regenerateCSSByHash()` (tailwind-compiler.ts:162) calls `generateTailwindCSS(inputs.stylesheet, inputs.candidates, { minify: true })` **without passing `projectSlug`**. This means the JIT regeneration path still uses the old shared compiler cache key (`stylesheetHash` only).

**Impact:** When the CSS handler (`css.handler.ts:22`) calls `regenerateCSSByHash`, a pod that JIT-regenerates CSS for project A and then project B will contaminate B's compiler. The hash-mismatch check at line 188 will then **reject** project B's regeneration, causing a silent fallback to 404. The user sees a broken page with missing styles and the log says "CSS not found and JIT regeneration failed" — with no indication that cross-project contamination caused it.

This is the **exact same class of bug** being fixed, just on a different code path. The regeneration inputs cache (`resolveRegenerationInputs`) likely stores `projectSlug` or can be extended to do so, since the inputs were originally cached by `storeProjectCSS` which has the project context.

---

### HIGH: `projectSlug` as optional parameter is a footgun

`GenerateOptions.projectSlug?: string` being optional means every new call site that forgets to pass it silently falls back to the shared-compiler behavior — the exact bug this PR fixes. The API should make it hard to get wrong:

**Option A (preferred):** Make `projectSlug` a required parameter of `generateTailwindCSS` (or at least of `getCompiler`), using `"unknown"` as the explicit fallback for call sites that genuinely don't know the project. This makes omission a compile error rather than a silent bug.

**Option B:** Add a runtime warning when `projectSlug` is undefined so it's at least visible in logs.

---

### MEDIUM: Module-level JSDoc is now stale

`tailwind-compiler-cache.ts` lines 2-6 still say *"keyed by stylesheet hash"* — this should say *"keyed by project slug + stylesheet hash"* to match the new behavior.

---

### Minor nits

- The `compilerCache` `Map<string, CompilerCacheEntry>` type is fine, but the `getCompilerCacheStats` function exposes `hash` — consider renaming to `cacheKey` now that it's a composite key, for clarity in monitoring dashboards.
- The comment in `getCompiler` is helpful, but consider adding a `@param projectSlug` JSDoc to make the contract explicit.

---

### What's good

- Root cause investigation is thorough and well-documented in the PR body
- The fix is minimal and surgical — no over-engineering
- The LRU eviction still works correctly with composite keys
- Both production call sites (`pipeline.ts`, `styles-css.handler.ts`) correctly thread `projectSlug` through

---

### Recommendation

Fix the `regenerateCSSByHash` gap and consider making `projectSlug` harder to omit before merging. The current state fixes the primary production path but leaves a secondary path vulnerable to the same bug class.

kwakayama (commented):
## Re-review (v2): fix: isolate Tailwind compiler per project to prevent CSS contamination

**Merge Score: 88/100**

---

### Previous critical finding: FIXED

`regenerateCSSByHash` now takes `projectSlug: string | undefined` and threads it through to `generateTailwindCSS`. Both call sites (`css.handler.ts`, `pipeline.ts:resolveCssFromRenderedHtml`) updated. New regression test covers cross-project JIT isolation. Well done.

The choice of `projectSlug: string | undefined` (required positional, not optional) for `regenerateCSSByHash` is the right call — it forces callers to be explicit about the project context.

---

### Remaining items

**MEDIUM: `getCompiler(stylesheet, projectSlug?)` — the `?` still silently opts out of isolation**

`getCompiler`'s second param is `projectSlug?: string` (truly optional — can be omitted entirely). This is the function where the cache isolation actually happens, so a caller that simply writes `getCompiler(css)` gets the shared-compiler behavior with zero compile-time or runtime warning. All current internal callers pass it correctly via `generateTailwindCSS`, but any future direct caller of `getCompiler` can silently regress.

Suggestion: change to `projectSlug: string | undefined` (required but nullable) to match the pattern already used on `regenerateCSSByHash`. Minimal change, breaks nothing, but makes omission a type error.

**LOW: Module JSDoc stale**

`tailwind-compiler-cache.ts` lines 3-4 still read:
> Manages a bounded cache of compiled Tailwind CSS compilers, keyed by **stylesheet hash**.

Should mention `projectSlug + stylesheetHash`.

**LOW: `inFlightRegeneration` dedup key doesn't include `projectSlug`**

If two different projects request the same CSS hash concurrently, the second piggybacks on the first's in-flight promise. Functionally correct (same hash ⇒ same candidates + stylesheet ⇒ same output from a fresh compiler), so not a bug, but worth documenting as an intentional behavior if it's ever audited.

**LOW: `getCompilerCacheStats` exposes `hash` field**

This field is now a composite key (`slug:hash`), which may confuse monitoring dashboards that previously expected a pure content hash. Consider renaming to `cacheKey`.

---

### What's improved since v1

- `regenerateCSSByHash` patched with per-project isolation — critical gap closed
- `css.handler.ts` threads `ctx.projectSlug ?? ctx.projectId` correctly
- `pipeline.ts:resolveCssFromRenderedHtml` takes `projectSlug` as explicit param
- New test `"isolates JIT regeneration by project"` covers the exact cross-project scenario
- Existing tests updated to pass `projectSlug` explicitly

---

### Verdict

The PR now covers all production code paths. The fix is correct, minimal, and well-tested. The remaining items are hardening suggestions, not blockers. Ship it after the staging verification in the test plan.

PR #666, opened 2026-03-21
Handle skipped and fallback knowledge-ingest sources

kwakayama:
## Summary
- classify knowledge-ingest sources up front and skip hidden paths, ignored directories, and unsupported files without failing the whole batch
- add plain-text/code fallback ingest support for common source files such as `.py`, `.ts`, `.js`, `.yaml`, and `.sql`
- write the new structured `knowledge_ingest` job result shape and align the public `veryfront/jobs` SDK with the split list/detail contract

## Testing
- deno test -A src/jobs/jobs-client.test.ts cli/commands/knowledge/command.test.ts cli/utils/write-job-result.test.ts


kwakayama:
Follow-up pushed in 73e8f96. This addresses the review items around explicit uploads/ references, per-file local source references for walked directories, parser-error classification, and permissive branch_id parsing. I intentionally kept mixed knowledge-ingest batches as partial-success instead of exiting non-zero on any per-file skip/failure; those cases now surface in the structured result and result_summary contract without aborting the full batch.

kwakayama:
Additional follow-up pushed in ed237cd9 after the full unit suite surfaced an unrelated SSR transform failure. This preserves explicit package versions during bare-import rewriting and SSR resolution, and pins the react-query fixture to a known-good esm.sh version so the unit suite no longer depends on a broken unversioned endpoint.

kwakayama:
Self-review note: no blocking issues found in local diff review. The typed knowledge-ingest result writer, source policy, and fallback parser coverage look internally consistent with the downstream API and Studio contract. Score: 90/100.

kwakayama (commented):
Reviewed requested commit `50f975c5` (the PR tip is now `b56775f1`).

Score: 76/100

Finding:
- `src/jobs/schemas.ts:195-205` and `src/jobs/index.ts:78-89` still expose a batch contract that only has `processing/completed/remaining`. At this commit the SDK no longer has `skipped_count`, `failed_count`, `skipped`, `failed`, or the `KnowledgeIngestBatchSourceWithMessage*` exports, so `jobs.batches.get()` falls out of sync with the API/studio changes in this stack and clients lose the new skip/failure details. The newer tip commit looks like it fixes this, so I’m leaving this as context on the requested head rather than a tip-of-PR blocker.

kwakayama:
Follow-up pushed in bc4ae7d6. This splits parser execution from the upload step in `ingestResolvedSources()`, so raw parser exceptions are now always classified as `parser_error`, and it adds a regression test for that path.\n\nI also rechecked the remaining open review threads on current head: the empty `uploads/` explicit-source guard, permissive `branch_id` parsing, and the build/dist regression note are already addressed by earlier commits. I’m keeping mixed batches as partial-success intentionally: per-file skips/failures flow through the structured knowledge-ingest result instead of flipping the whole command to a non-zero exit when at least one requested source succeeds.

kwakayama (commented):
Re-reviewed current tip `b56775f1`.

Score: 95/100.

The earlier SDK batch-schema/export mismatch is fixed at this head. I did not find any new blocking correctness issues in this PR on the current tip.

Residual risk: review only in this pass; I did not rerun local tests.

kwakayama (commented):
Re-reviewed current tip `bc4ae7d6`.

Score: 96/100.

This follow-up commit cleanly separates parser failures from upload failures inside `ingestResolvedSources()`. I did not find any new blocking correctness issues at this head.

Verification: `deno test -A cli/commands/knowledge/command.test.ts` passed.

kwakayama (commented):
Reviewed commit 50f975c5 only.

Score: 90/100

No blocking findings in this snapshot. The ingest contract, fallback parser coverage, and SDK schema changes all looked coherent at this commit. Later commits on the PR branch were not reviewed here.

PR #668, opened 2026-03-21
fix: verify browser auth tokens via api jwks

kwakayama:
## Summary
- verify browser `authToken` JWTs against the API JWKS instead of relying on local HS256-only validation
- preserve protected preview access for API-issued `RS256` user tokens
- add a proxy regression proving protected preview auth succeeds via `/.well-known/jwks.json`

## Context
- API user tokens are signed as `RS256`
- the proxy on `main` only verified `HS256` user tokens with `JWT_SECRET`
- protected preview requests with valid browser auth cookies were therefore being downgraded to unauthenticated and redirected to sign-in

## Evidence
- live JWKS check: `curl -I https://api.veryfront.com/.well-known/jwks.json` returned `HTTP/2 200` with `Cache-Control: public, max-age=300`
- live JWKS body check confirmed an RSA signing key with `alg=RS256`, `use=sig`, `kid`, `n`, and `e`
- local crypto verification using the API test RSA keypair and a JWKS document succeeded for an `RS256` user token
- red-green proxy regression:
  - before the fix, `allows access to protected preview domain with RS256 auth token verified via JWKS` failed with `302 Authentication required`
  - after the fix, it passes and confirms the proxy fetches `/.well-known/jwks.json`

## Test Plan
- [x] `deno test -A src/proxy/handler.test.ts`
- [x] `deno fmt --check src/proxy/handler.ts src/proxy/handler.test.ts`
- [x] pre-commit `verify:quick` hook during `git commit` (generate, fmt, lint, docs validate, typecheck)

## Notes
- supersedes #667, which used a `/me` fallback; this PR replaces that approach with direct JWT verification against JWKS

### DIFF
--- a/src/proxy/handler.ts
+++ b/src/proxy/handler.ts
@@ -6,7 +6,7 @@ import { cwd, getEnv } from "#veryfront/platform/compat/process.ts";
 import { join } from "#veryfront/compat/path/index.ts";
 import { injectContext, ProxySpanNames, withSpan } from "./tracing.ts";
 import { computeContentSourceId } from "#veryfront/cache/keys.ts";
-import { jwtVerify } from "jose";
+import { createRemoteJWKSet, decodeProtectedHeader, jwtVerify } from "jose";
 
 export const INTERNAL_PROXY_HEADERS = [
   "x-token",
@@ -34,6 +34,28 @@ interface DomainLookupResult {
     active_release_id?: string | null;
     protected?: boolean;
   }>;
+}
+
+const remoteJwksByUrl = new Map<string, ReturnType<typeof createRemoteJWKSet>>();
+
+function getApiJwks(apiBaseUrl: string, logger?: ProxyLogger) {
+  try {
+    const jwksUrl = new URL("/.well-known/jwks.json", apiBaseUrl);
+    const cacheKey = jwksUrl.toString();
+    let jwks = remoteJwksByUrl.get(cacheKey);
+
+    if (!jwks) {
+      jwks = createRemoteJWKSet(jwksUrl);
+      remoteJwksByUrl.set(cacheKey, jwks);
+    }
+
+    return jwks;
+  } catch (error) {
+    logger?.error("Invalid API base URL for JWKS lookup", error as Error, {
+      apiBaseUrl,
+    });
+    return undefined;
+  }
 }
 
 async function lookupProjectByDomain(
@@ -153,8 +175,42 @@ function extractUserToken(cookieHeader: string): string | undefined {
 
 async function extractUserIdFromToken(
   token: string,
+  apiBaseUrl: string,
   log?: ProxyLogger,
 ): Promise<string | undefined> {
+  let algorithm: string | undefined;
+
+  try {
+    algorithm = decodeProtectedHeader(token).alg;
+  } catch (error) {
+    log?.debug("Failed to decode JWT header", {
+      error: error instanceof Error ? error.message : String(error),
+    });
+    return undefined;
+  }
+
+  if (algorithm === "RS256") {
+    const jwks = getApiJwks(apiBaseUrl, log);
+    if (!jwks) return undefined;
+
+    try {
+      const { payload } = await jwtVerify(token, jwks, {
+        algorithms: ["RS256"],
+      });
+      return (payload as { userId?: string }).userId;
+    } catch (error) {
+      log?.debug("RS256 JWT verification failed", {
+        error: error instanceof Error ? error.message : String(error),
+      });
+      return undefined;
+    }
+  }
+
+  if (algorithm !== "HS256") {
+    log?.debug("Unsupported JWT algorithm", { algorithm: algorithm ?? null });
+    return undefined;
+  }
+
   const jwtSecret = getEnv("JWT_SECRET");
 
   if (!jwtSecret) {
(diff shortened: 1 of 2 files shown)


src/proxy/handler.ts
```diff
@@ -36,6 +36,28 @@ interface DomainLookupResult {
   }>;
 }
 
+const remoteJwksByUrl = new Map<string, ReturnType<typeof createRemoteJWKSet>>();
+
+function getApiJwks(apiBaseUrl: string, logger?: ProxyLogger) {
+  try {
+    const jwksUrl = new URL("/.well-known/jwks.json", apiBaseUrl);
```
kwakayama: Using `new URL("/.well-known/jwks.json", apiBaseUrl)` strips any path prefix from the configured API base URL. The rest of the proxy preserves that prefix for `/auth/token` and `/projects/*`, so a deployment configured with something like `VERYFRONT_PROXY_API_BASE_URL=https://host/api` will verify a valid RS256 browser token against `https://host/.well-known/jwks.json` instead of `https://host/api/.well-known/jwks.json` and downgrade the request to unauthenticated. I reproduced that locally: the proxy requested `/api/projects/protected-project` and `/.well-known/jwks.json`, then returned `302 Authentication required` for a valid RS256 token. This needs to build the JWKS URL relative to the configured base path, or use a separate explicit JWKS base config if host-root JWKS is intentional.

kwakayama (commented):
Score: 82/100

I found one blocking regression in `src/proxy/handler.ts`: the new JWKS lookup ignores any path prefix included in `apiBaseUrl` by constructing `new URL("/.well-known/jwks.json", apiBaseUrl)`. In practice that means `VERYFRONT_PROXY_API_BASE_URL=https://host/api` still uses `/api/projects/...` for project lookup, but RS256 auth verification goes to `https://host/.well-known/jwks.json` and valid browser tokens get redirected to sign-in. I reproduced that behavior locally against this branch. See the inline comment for the exact failure mode and suggested direction for the fix.

GitHub does not allow me to submit `REQUEST_CHANGES` on my own pull request, so this is posted as a comment review instead.

src/proxy/handler.ts
kwakayama: Updated in d51fd5e6d. The proxy now preserves the configured API base path when building the JWKS URL, and `deno test -A src/proxy/handler.test.ts` includes a regression for `apiBaseUrl=http://127.0.0.1:<port>/api` that verifies `/api/.well-known/jwks.json` is used while `/.well-known/jwks.json` is not.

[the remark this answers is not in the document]
kwakayama: Updated in d51fd5e6d. JWKS resolution now stays relative to the configured API base path, so `https://host/api` resolves to `https://host/api/.well-known/jwks.json` instead of host root. I also added a regression that proves the prefixed JWKS path is used and that root JWKS is not requested.

kwakayama (commented):
Re-review on `d51fd5e6dd28a8a33840c58b16edcf1daf362cfd`.

Score: 96/100

No findings in the updated scope. The previous blocking issue is fixed: JWKS lookup now preserves a path-prefixed `apiBaseUrl`, and the added regression test covers that exact deployment shape.

Verification I ran locally:
- `deno test -A src/proxy/handler.test.ts`
- focused `deno eval` repro with `apiBaseUrl=http://127.0.0.1:<port>/api`, which now requests `/api/.well-known/jwks.json` and allows the valid RS256 token through

Residual risk: this re-review was focused on the proxy auth/JWKS path and the follow-up fix, not a fresh broad review of unrelated areas.

kwakayama (commented):
Re-review on `d51fd5e6dd28a8a33840c58b16edcf1daf362cfd`.

Score: 96/100

No findings in the updated scope. The previous blocking issue is fixed: JWKS lookup now preserves a path-prefixed `apiBaseUrl`, and the added regression test covers that exact deployment shape.

Verification I ran locally:
- `deno test -A src/proxy/handler.test.ts`
- focused `deno eval` repro with `apiBaseUrl=http://127.0.0.1:<port>/api`, which now requests `/api/.well-known/jwks.json` and allows the valid RS256 token through

Residual risk: this re-review was focused on the proxy auth/JWKS path and the follow-up fix, not a fresh broad review of unrelated areas.

kwakayama (commented):
## Review: 85/100

**Well-scoped fix** that correctly addresses the RS256/HS256 JWT verification mismatch for protected preview auth.

### Strengths

- **Clean algorithm dispatch** — decoding the protected header first and branching on `alg` is the right approach. Keeps HS256 path untouched.
- **Proper use of `createRemoteJWKSet`** — jose handles JWKS key rotation, caching, and refresh internally. The module-level `remoteJwksByUrl` Map prevents redundant instances.
- **Good URL normalization** in `getApiJwks` — the trailing-slash handling + `new URL()` correctly preserves path-prefixed API bases (e.g. `/api/.well-known/jwks.json`).
- **Thorough tests** — RS256 happy path, path-prefixed API base, JWKS request counting, and the existing HS256 tests remain unmodified.
- **Defensive error handling** — both `decodeProtectedHeader` and `jwtVerify` failures are caught and logged without leaking errors.

### Minor notes

- The `remoteJwksByUrl` Map grows unbounded, but in practice there's one API base URL per deployment so this is a non-issue. If multi-tenant proxy becomes a thing, consider a size cap.
- `createRemoteJWKSet` will fetch JWKS lazily on first verification and cache internally with its own refresh logic (default: refetch on `kid` miss). Worth noting in case debugging stale-key issues arises later.

### Verdict

Clean, minimal, well-tested fix. No security concerns. Ship it.

PR #707, opened 2026-03-23
fix: enforce frozen lockfile in CI to prevent supply chain drift

ariskemper:
## Summary
- Replaces `--no-lock` with `--frozen` in all CI test commands (unit, integration, binary e2e)
- `--frozen` errors out if the lockfile is out of date, unlike `--lock` which silently updates
- Ensures dependency integrity is strictly verified against committed `deno.lock`
- Prevents silent dependency drift and supply chain attacks via compromised registries

Part of supply chain security hardening (Phase 3).

## Test plan
- [ ] CI tests pass with frozen lockfile enforcement

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -55,9 +55,9 @@ jobs:
           retry_wait_seconds: 10
           command: |
             if [ "${{ matrix.suite }}" = "integration" ]; then
-              VF_DISABLE_LRU_INTERVAL=1 deno task test:integration --frozen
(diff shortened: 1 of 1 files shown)


ariskemper:
Closing — lockfile enforcement requires a committed `deno.lock`, but it's currently in `.gitignore` (deliberately removed in 5a7a9e999 due to Deno version incompatibility between local and CI). 

Prerequisites before this can be re-attempted:
1. Align Deno versions across local dev and CI
2. Remove `deno.lock` from `.gitignore`
3. Commit `deno.lock`
4. Then use `--frozen` flag in CI

PR #714, opened 2026-03-23
Fix preview host and slug fallback resolution

kwakayama:
## Summary
- fall back to the request URL host when the Host header is absent in proxy handling
- derive project slugs from forwarded hosts when explicit slug headers are missing or blank
- avoid applying the default project id when the request already resolved a different slug

## Verification
- `deno test --allow-all src/server/context/request-context.test.ts src/server/runtime-handler/project-resolution.test.ts src/proxy/handler.test.ts`

### DIFF
--- a/src/proxy/handler.test.ts
+++ b/src/proxy/handler.test.ts
@@ -284,6 +284,32 @@ describe("Proxy Handler", () => {
         await server.shutdown();
       }
     });
+
+    it("falls back to the request URL host when the Host header is absent", async () => {
+      const handler = createProxyHandler({
+        config: {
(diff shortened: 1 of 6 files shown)


kwakayama:
Verification for this branch:

- `deno test --allow-all src/server/context/request-context.test.ts src/server/runtime-handler/project-resolution.test.ts src/proxy/handler.test.ts`
  - `6 passed`, `0 failed`
- This fix was also part of the local extracted e2e proof path that now passes in `veryfront-e2e`, including the preview-heavy smoke flow and a manual browser-agent preview check.

PR #771, opened 2026-03-27
chore: remove postinstall binary download from npm package

kojiwakayama:
## Summary
Remove the postinstall script that downloads a platform-specific binary during `npm install`.

**Why:**
- `veryfront` is used both as a CLI tool and a library dependency
- The binary download adds overhead for library users who don't need the CLI
- Triggers security scanner warnings (Socket.dev, npm audit, Aikido) for postinstall scripts that download executables
- Creates a network dependency during CI installs
- The JS CLI fallback is already fully functional — all commands work

**What changes:**
- Remove postinstall script copying from the dnt build
- Remove `scripts` directory from `pkg.files`
- Remove `pkg.scripts.postinstall` from generated `package.json`

**No experience degradation:**
- `npx veryfront` still works (uses bundled JS CLI via `bin/veryfront.js`)
- All CLI commands work identically
- Only difference: ~500ms slower startup (JS vs native binary)
- Users who want native binary speed: `brew install veryfront/tap/veryfront` or `curl -fsSL https://veryfront.com/install.sh | sh`

## Test plan
- [ ] Verify `deno task build:npm` produces a package without postinstall
- [ ] Verify `npx veryfront --help` works from the built package
- [ ] Verify `npx veryfront init test-app` scaffolds correctly
- [ ] Verify no security scanner warnings on the published package

### DIFF
--- a/scripts/build/build-npm-dnt.ts
+++ b/scripts/build/build-npm-dnt.ts
@@ -187,11 +187,6 @@ await build({
 			"dnt polyfill process.argv[1] fix",
 		);
 
-		// Copy postinstall script
-		await Deno.mkdir("./npm/scripts", { recursive: true });
-		await Deno.copyFile("./scripts/postinstall.js", "./npm/scripts/postinstall.js");
-		await Deno.copyFile("./scripts/postinstall-lib.js", "./npm/scripts/postinstall-lib.js");
-
 		// Note: Templates are now embedded in manifest.json which is bundled by dnt
 		// No need to copy template files separately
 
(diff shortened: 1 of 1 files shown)


kojiwakayama:
## 🐙 Claude Octopus Review — Merge Readiness: **92/100**

**Verdict: Approve**

### Correctness ✅
- The diff is clean and surgical — only removes postinstall-related lines from `build-npm-dnt.ts`
- `bin/veryfront.js` (bin-wrapper) is untouched and already handles the no-binary case gracefully: `existsSync(nativeBinary)` → false → `runJsFallback()` → `import("../esm/cli/main.js")`
- All CLI commands work identically via the JS fallback
- No runtime behavior change for library consumers (they never used the binary)

### Security ✅
- **This is a net security improvement.** Removes a postinstall that downloads and executes a binary from GitHub releases — exactly what supply chain scanners flag
- Eliminates the network dependency during `npm install`
- Reduces attack surface: no more `https.get()` to external URLs during install

### Architecture ✅
- Clean separation: `veryfront` npm package = library + JS CLI, native binary = separate install via brew/curl
- No breaking change — the JS CLI is a complete, always-present fallback
- `pkg.files` correctly updated to remove `scripts` directory

### Minor observations (non-blocking)
1. **Dead code:** `scripts/postinstall.js` and `scripts/postinstall-lib.js` are no longer referenced by the build. Consider removing them in a follow-up to avoid confusion. Not blocking since they're source files, not shipped.
2. **Docs update:** The `npx veryfront` path will now always use JS CLI (~500ms startup). The docs at veryfront.com/docs/code already list brew and curl as primary install methods, so this is fine. Consider mentioning "for fastest CLI startup, install via brew or curl" in the README.

### Score breakdown
| Category | Score | Notes |
|---|---|---|
| Correctness | 95 | Clean removal, fallback verified |
| Security | 98 | Removes supply chain risk |
| Architecture | 90 | Good separation of concerns |
| Completeness | 85 | Dead postinstall scripts could be cleaned up |

---
*🐙 Review by Claude Octopus (Claude + Codex + Gemini)*

PR #773, opened 2026-03-30
fix: prebundle RSC client scripts, inject import map, fix hydration root

kojiwakayama:
## Summary
- **Prebundle RSC client scripts** for compiled binaries — esbuild couldn't resolve `#veryfront/` import map specifiers at runtime, causing `Could not resolve` errors on every page load
- **Inject `<script type="importmap">` for layout-provided HTML** — when `app/layout.tsx` provides `<html>`/`<head>`, the shell generator was bypassed, leaving no import map for bare React specifiers
- **Fix hydration root selection** — `hydratePageComponent()` picked the hidden `data-veryfront-head` placeholder as the root, leaving the visible content as static HTML with no event handlers
- **Keep React as bare specifiers in page module transforms** — the esbuild plugin rewrote `import "react"` to hardcoded esm.sh URLs with `DEFAULT_REACT_VERSION`, causing version mismatches with the project's actual React version

## Test plan
- [x] All 1216 unit tests pass
- [x] `deno task generate` succeeds with new prebundle step
- [x] `deno task build` produces working binary
- [x] Page loads without esbuild resolution errors
- [x] Import map present in `<head>` for layout-based pages
- [x] React hydration attaches to correct content element
- [x] Sound buttons are interactive after hydration
- [x] No React version mismatch errors in console


kojiwakayama:
Rebased this branch onto current `main` and resolved the merge conflicts.

Addressed in the rebased head `ed7ade58d`:
- forwarded the CSP nonce when injecting the full-document import map
- restored a safe non-div hydration-root fallback while still skipping hidden placeholders
- regenerated the RSC client bundle against the rebased sources

Verification:
- `deno test --no-check --allow-all --unstable-worker-options --unstable-net src/rendering/orchestrator/html.test.ts src/html/html-injection.test.ts src/rendering/rsc/client-boot.test.ts`
- local pre-push hook suite: format, lint, typecheck, and full unit tests all passed before push

PR #807, opened 2026-04-01
Add pages-router strict CSP hydration coverage

kojiwakayama:
Supersedes #806. This includes the CodeQL-safe CSP assertion follow-up on the current branch head.


kojiwakayama:
Superseded by a clean follow-up PR from current main that contains only the CodeQL-safe assertion fix.

PR #859, opened 2026-04-05
Preserve explicit generated-looking tool ids

kojiwakayama:
## Summary
- preserve explicit tool ids even when they match the autogenerated placeholder pattern
- mark autogenerated tool ids in the factory instead of guessing from the final string during discovery
- bump `deno.json` and version constant to `0.1.134` for deployment

## Testing
- deno test --allow-all src/utils/version.test.ts src/utils/logger/logger.test.ts src/tool/factory.test.ts src/discovery/auto-discovery.integration.test.ts src/server/handlers/request/api/project-discovery.test.ts
- full pre-push suite (1288 passed, 0 failed)


kojiwakayama (commented):
Addressed in the latest push. The generated-id marker now stores the original placeholder id string, and discovery only falls back to the filename when the current `tool.id` still equals that original placeholder.

That preserves cases like `export default { ...tool(...), id: "my-tool" }` while still rewriting untouched autogenerated ids. Added red/green regressions for both discovery and request-time project-agent runs.

PR #868, opened 2026-04-06
Normalize provider-native tool-call input on project-agent replay

kwakayama:
## Summary
Fix the project-agent replay path so provider-native tool calls preserve JSON-text input without double-stringifying it.

## Why
`veryfront-studio#1239` still reproduces on staging for `Demo Agent` with `web_search`.
The durable run evidence shows `web_search` is invoked, but the execution path later fails with `Expected object, received string`.

This PR addresses the `veryfront-code` side of that boundary by normalizing provider-native `tool-call` input before it is stored/replayed.

## TDD Evidence
Red:
- `deno test --no-check --allow-all src/agent/runtime/ai-stream-handler.test.ts`
- failing assertion: a provider-native `tool-call` with string input was stored as a doubly-quoted JSON string

Green:
- `deno test --no-check --allow-all src/agent/runtime/ai-stream-handler.test.ts`
- `deno task verify:quick`

## Scope
- normalize `part.input` when it is already JSON text
- keep AG-UI `tool-input-available` events object-shaped
- add regression coverage for provider-native `web_search`

## Related
- veryfront-studio#1239

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.135",
+  "version": "0.1.136",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/src/agent/runtime/ai-stream-handler.test.ts
+++ b/src/agent/runtime/ai-stream-handler.test.ts
@@ -186,6 +186,34 @@ describe("ai-stream-handler", () => {
       });
     });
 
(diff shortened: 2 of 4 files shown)


kwakayama:
Addressed the automated review note in `bc19d8cec`.

Change:
- preserve null/missing provider-native tool input as invalid replay input instead of normalizing it to `{}`

Re-verified:
- `deno test --no-check --allow-all src/agent/runtime/ai-stream-handler.test.ts`
- `deno task verify:quick`

PR #876, opened 2026-04-07
Fail fast when configured tool ids drift from canonical names

kwakayama:
## Summary
- fail fast when an explicit configured tool name does not resolve to a discovered or allowed tool
- surface the real runtime error in the internal agent stream instead of a generic internal error
- bump the patch version in `deno.json` to `0.1.142`

## Problem
A staging investigation showed a hidden failure mode in project-agent runs:
- the runtime loaded `tools/roll-dice.ts` and registered `roll-dice`
- a child run was allowlisted with `allowedTools: ["rollDice"]`
- the unresolved name was silently omitted from the runtime toolset
- the model then emitted fake `<tool_call>` / `<tool_response>` text as plain assistant output
- there were `0` real `TOOL_CALL_*` events and no `console.log("rolling dice...")` output

This produced a plausible-looking answer without any real tool execution.

## What changed
- `getAvailableTools()` now throws when explicit configured tool names do not resolve, instead of silently dropping them
- the error lists the unresolved name and the currently available tool ids
- stream errors now forward the actual runtime message so misconfiguration is visible immediately
- added a regression test for the `roll-dice` vs `rollDice` mismatch

## Why this fixes the hallucination
The framework now keeps one canonical tool namespace only: the exact `tool({ id: "..." })` string, or the filename-derived id if `id` is omitted for a discovered tool. If config, prompt, or allowlist strings drift from that canonical name, the run fails before model generation instead of giving the model a chance to invent textual tool transcripts.

## Testing
- `deno test --no-check --allow-all src/agent/runtime/tool-helpers.test.ts`
- `deno test --no-check --allow-all src/server/handlers/request/agent-stream.handler.test.ts`
- `deno eval --no-check 'import "./src/agent/runtime/index.ts"'

## Notes
The motivating RFC and evidence were captured from the staging investigation, including the hallucinated child-run transcript and the missing real `TOOL_CALL_*` events.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.141",
+  "version": "0.1.142",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/src/agent/runtime/index.ts
+++ b/src/agent/runtime/index.ts
@@ -406,7 +406,7 @@ export class AgentRuntime {
           logger.error("Agent stream error", { error });
           sendSSE(controller, encoder, {
             type: "error",
-            error: "An internal error occurred",
+            error: error instanceof Error ? error.message : String(error),
           });
           closeSSEStream(controller);
         } finally {
--- a/src/agent/runtime/tool-helpers.test.ts
+++ b/src/agent/runtime/tool-helpers.test.ts
@@ -175,7 +175,34 @@ describe("tool-helpers", () => {
   });
 
   describe("getAvailableTools", () => {
+    it("fails loudly when an explicit configured tool name does not match a discovered tool id", async () => {
+      toolRegistry.clearAll();
+
+      toolRegistry.register(
+        "roll-dice",
+        tool({
+          id: "roll-dice",
+          description: "Roll a die",
+          inputSchema: z.object({}),
+          execute: async () => ({ total: 4 }),
+        }),
+      );
+
+      await assertRejects(
+        () =>
+          getAvailableTools(
+            {
+              rollDice: true,
+            },
+            { includeIntegrationTools: false },
+          ),
+        Error,
+        'Unknown tool reference: rollDice. Tool names must exactly match tool({ id: "..." }). Available tools: roll-dice',
+      );
+    });
+
     it("filters remote integration tool definitions by the runtime allowlist", async () => {
+      toolRegistry.clearAll();
       const originalFetch = globalThis.fetch;
       const originalApiBaseUrl = Deno.env.get("VERYFRONT_API_URL");
       const originalApiToken = Deno.env.get("VERYFRONT_API_TOKEN");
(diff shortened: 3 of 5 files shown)


kwakayama:
[HIGH] `src/agent/runtime/tool-helpers.ts:251` and `src/agent/runtime/tool-helpers.ts:265`

This still does not satisfy the RFC for explicit remote/integration tool references. The new fail-fast guard only records unresolved names when they do *not* look like remote tools (`!isRemoteIntegrationTool(name)`), so any explicit `foo:bar` entry is still silently exempt from validation.

I verified two concrete failures on this branch:

- `getAvailableTools({"gmail:missing": true}, { includeIntegrationTools: false })` returns `[]` instead of throwing.
- with fetched remote defs limited to `gmail:list-emails`, `getAvailableTools({"gmail:get-email": true}, { allowedRemoteToolNames: ["gmail:list-emails"] })` returns `["gmail:list-emails"]` instead of failing the explicit mismatch.

That leaves the same silent-drop class open for remote tools: the explicitly configured tool can disappear, model generation can continue, and the runtime never surfaces the configuration error the RFC requires.

Fix direction:
- validate explicit remote entries against `remoteToolNames` the same way local entries are validated
- for explicit-object configs, only append remote defs that were explicitly requested rather than every allowed remote def
- add a regression test for explicit remote mismatch / allowlist drift in `src/agent/runtime/tool-helpers.test.ts`

I ran LSP diagnostics on the modified files and the targeted tests pass, but the new coverage only proves the local discovered-tool mismatch path.

Score: 38/100
Verdict: request changes.

kwakayama:
Addressed the explicit remote-tool validation gap in `57963173d`.

Changes made:
- explicit remote entries like `gmail:get-email` now fail fast when they are missing from the discovered / allowlisted remote definitions
- explicit tool maps only append remote definitions that were explicitly requested, while still preserving the runtime-expanded `tools: true` allowlist path
- added regression coverage for explicit remote mismatch and allowlist drift in `src/agent/runtime/tool-helpers.test.ts`
- synced `src/utils/version-constant.ts` with the existing `deno.json` bump so `tests (unit)` stops failing on the version mismatch

Local verification:
- `deno fmt --check`
- `deno task lint`
- `deno task typecheck`
- `deno task test:unit --no-lock`
- `deno test --no-check --allow-all src/agent/runtime/tool-helpers.test.ts`
- `deno test --no-check --allow-all src/utils/version.test.ts src/utils/logger/logger.test.ts`

One local-only gap remains: `VF_DISABLE_LRU_INTERVAL=1 deno task test:integration --no-lock` hit a Deno runtime panic on macOS with local `deno 2.7.5` after the suite had already made substantial progress. The PR's GitHub `tests (integration)` job was green before this patch, and the branch has now been pushed so CI can rerun on Linux / Deno `2.7.7`.

kwakayama (commented):
Checked the state after the latest commit 57963173. The remote tool validation changes described in comment 4198756744 are now implemented (explicit remote-tool validation + error path + targeted tests in src/agent/runtime/tool-helpers.ts/tool-helpers.test.ts), and the two previously failing scenarios now fail fast with the expected `Unknown tool reference` errors. No further action required from that earlier review.

PR #896, opened 2026-04-08
Fix knowledge ingest upload staging and keep the CLI releasable

kojiwakayama:
## Summary
- preserve uploads-prefixed knowledge sources so sandbox/project ingest uses the right remote paths
- resolve tenant project context when explicit project slug is absent
- stabilize the doctor unreachable-server test and bump the CLI patch version to 0.1.144

## Verification
- VF_DISABLE_LRU_INTERVAL=1 deno test --no-check --allow-all cli/shared/config.test.ts cli/commands/knowledge/command.test.ts cli/commands/doctor/server-checks.test.ts src/utils/version.test.ts src/utils/logger/logger.test.ts
- full pre-push suite via git push


kojiwakayama:
Follow-up for the config precedence review is open in #897. It restores repo config precedence over tenant fallback while keeping the tenant fallback path for sandbox/no-config cases.

PR #914, opened 2026-04-09
Advertise MCP accept types in remote framework tool requests

kojiwakayama:
## Summary
- include the MCP-required `Accept` media types on remote JSON-RPC tool requests
- preserve caller-provided `Accept` values while appending `application/json` and `text/event-stream`
- add regression coverage for the default and custom-accept cases

## Why
Staging framework-default agent runs fail during setup with:
- `Remote MCP request failed (406)`
- `Not Acceptable: Client must accept both application/json and text/event-stream`

This fixes the shared framework client rather than special-casing one host.

## Verification
- `deno test --no-check --allow-all src/tool/remote-mcp.test.ts`
- `deno test --no-check --allow-all src/tool/remote-mcp.test.ts src/agent/runtime/tool-helpers.test.ts`
- full repo pre-push lane passed during `git push`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.145",
+  "version": "0.1.146",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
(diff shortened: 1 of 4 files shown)


kojiwakayama:
Addressed the review concern in `561c76ce3`.

The remote MCP client now:
- advertises both `application/json` and `text/event-stream`
- parses JSON-RPC payloads from either JSON bodies or negotiated SSE responses
- has a red/green regression test covering the SSE path

The branch also now includes the release bump in `deno.json` and `src/utils/version-constant.ts`.

PR #924, opened 2026-04-10
Expose the canonical AG-UI runtime request contract from veryfront/agent

kojiwakayama:
## Summary
- add a first-class public AG-UI runtime contract module under `veryfront/agent`
- move the canonical hosted runtime request schema onto that public surface with open-source-facing names
- keep `/internal/agents/stream` framed as a compatibility wrapper around the public contract

## Why
Downstream runtimes should not have to import `internal-agents/*` files or internal names just to target the hosted agent runtime contract.

## Changes
- add `src/agent/runtime-ag-ui-contract.ts`
- export `AgUiRuntimeRequestSchema` and related types/schemas from `veryfront/agent`
- switch `src/internal-agents/schema.ts` to consume the public contract module instead of owning the canonical schema
- add a public regression test covering the exported runtime schema
- update the public docs to point consumers at the public contract rather than the internal schema path

## Verification
- `deno test --allow-all src/agent/runtime-ag-ui-contract.test.ts src/internal-agents/schema.test.ts src/agent/ag-ui-handler.test.ts`
- `deno lint src/agent/runtime-ag-ui-contract.ts src/agent/runtime-ag-ui-contract.test.ts src/agent/index.ts src/internal-agents/schema.ts src/internal-agents/schema.test.ts`
- `deno check src/agent/runtime-ag-ui-contract.ts src/agent/runtime-ag-ui-contract.test.ts src/agent/index.ts src/internal-agents/schema.ts src/internal-agents/schema.test.ts src/agent/ag-ui-handler.test.ts`
- `deno task verify`

## Related
- veryfront-agent#172
- supersedes the need for downstreams to target the internal contract path documented in veryfront-code#898

### DIFF
--- a/docs/reference/agent-runtime-ag-ui-contract.md
+++ b/docs/reference/agent-runtime-ag-ui-contract.md
@@ -6,7 +6,7 @@ The goal is to make downstream consumers target one explicit AG-UI-aligned input
 
 ## Canonical Contract
 
-The canonical runtime contract is defined in [`src/internal-agents/schema.ts`](../../src/internal-agents/schema.ts) by [`RuntimeRunAgentInputSchema`](../../src/internal-agents/schema.ts).
+The canonical runtime contract is defined in [`src/agent/runtime-ag-ui-contract.ts`](../../src/agent/runtime-ag-ui-contract.ts) by [`AgUiRuntimeRequestSchema`](../../src/agent/runtime-ag-ui-contract.ts).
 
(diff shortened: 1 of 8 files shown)


kojiwakayama (commented):
Public surface direction looks right: this makes downstreams target `veryfront/agent` instead of `internal-agents/*`, and the internal route remains clearly compatibility-only.

PR #992, opened 2026-04-12
Buffer early tool resumes for active runs

kojiwakayama:
## Summary
- buffer early tool resume payloads for active runs before the runtime registers its wait
- keep conflicting or mismatched active waits failing fast
- add regression coverage for the session manager, resume handler, and stream handler
- bump the package version to `0.1.192`

## Verification
- `deno test --no-check --allow-all src/utils/version.test.ts src/internal-agents/ag-ui-sse.test.ts src/internal-agents/session-manager.test.ts src/server/handlers/request/agent-run-resume.handler.test.ts src/server/handlers/request/agent-stream.handler.test.ts`
- `deno check src/agent/runtime/resume-session.ts`
- `deno fmt --check src/agent/runtime/resume-session.ts src/internal-agents/session-manager.test.ts src/server/handlers/request/agent-run-resume.handler.test.ts src/server/handlers/request/agent-stream.handler.test.ts src/utils/version-constant.ts deno.json`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.191",
+  "version": "0.1.192",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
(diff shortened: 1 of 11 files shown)


kojiwakayama:
Addressed in 3de5f48e4. Early resumes now require explicit pre-registration before buffering, so unknown or stale wait keys fail fast again. The valid early-resume paths still work because tool-call ids are registered when they become visible on the stream boundary and before the waiter blocks.

PR #1055, opened 2026-04-14
fix: return 404 when custom-domain token mint reports missing project

### LINKED ISSUE #1054: Proxy should return 404 when custom domain no longer maps to a project
## Summary
Production logs show `ai-chatbot.veryfront.com` returning 502 when token minting fails with `Project not found for domain`.

## Evidence
- `502 GET /robots.txt` and `502 GET /sitemaps.xml` for `ai-chatbot.veryfront.com`
- token fetch error: `OAuth token request failed: 400 - {"error":"Project not found for domain"}`

## Expected
Return a safe 404 (`No project configured for domain`) instead of a 502 auth failure for missing custom-domain mappings.

## Fix
Detect the specific token-mint error and translate it to the existing custom-domain 404 path, with coverage in `src/proxy/handler.test.ts`.

kwakayama:
Closes #1054

## What changed
- Detect the token-mint error for custom domains that no longer map to a project (`OAuth token request failed: 400 - {"error":"Project not found for domain"}`) and return the safe 404 path instead of a misleading 502.
- Add a code comment in `handler.ts` pinning the exact API source string (`veryfront-api/src/api/http/rest/auth/routes.ts` — `oauthError(c, 'Project not found for domain', ...)`) so a future rename is greppable.
- Regression test in `src/proxy/handler.test.ts` using the **actual** incident hostname from production logs (`ai-chatbot.veryfront.com`).
- Add a tombstone comment in `domain-parser.ts` documenting why bare `{slug}.veryfront.com` is intentionally **not** supported. Two prior attempts to re-enable it have been reverted; the comment names PR #1055 so the history is interpretable.
- Small CLI boundary cleanup: re-export `CommandResult` from `veryfront/platform` and route the knowledge CLI import through the public barrel (pre-existing lint violation that was blocking pre-commit).

## Evidence (verified from Loki, past 7 days)
Queried `{service_name="veryfront-proxy"} |= "ai-chatbot"` on `grafanacloud-veryfront-logs`:

- **100%** of `ai-chatbot` proxy traffic hits the bare `ai-chatbot.veryfront.com` form. Zero hits on `ai-chatbot.production.veryfront.com`.
- Split: **67× 502** + **66× 400** (the 400s are the upstream token-mint responses; the 502s are the handler fallthrough this PR fixes).
- Repeating error chain:
  1. `Token fetch failed` — `OAuth token request failed: 400 - {"error":"Project not found for domain"}`
  2. `Cannot process custom domain without token` — `domain: ai-chatbot.veryfront.com`
  3. `502 GET /robots.txt` (also `/sitemaps.xml`)
- Last seen: `2026-04-15T14:27Z` on `veryfrontVersion: 0.1.141`.

## Product decision (documented in tombstone comment)
Bare `{slug}.veryfront.com` subdomains are **not supported**. Customers must use the explicit `{slug}.production.veryfront.com` form. This PR makes such requests return a clean 404 instead of a 502, but does not re-enable bare-subdomain parsing. The `ai-chatbot` project owner needs to migrate their DNS / deploy URL — that's a separate customer-facing follow-up, not a code fix.

## Scope changes from the original PR
- **Dropped:** `domain-parser.ts` change that re-enabled bare `{slug}.veryfront.com` parsing. The log evidence confirmed it was the incident URL shape, but the product decision is to reject the bare form, not support it. Reverted in commit `f5938cc`.
- **Added:** tombstone comment so this doesn't get re-tried a third time.

## Validation
- `deno task test src/proxy/handler.test.ts src/server/utils/domain-parser.test.ts` ✅ — all 98 steps pass.
- `deno task lint:cli-boundary` ✅
- Pre-commit `verify` ✅

## Follow-ups (separate issues)
- Thread a typed error code from the token-mint helper so the regex-match on the error body can be retired.
- `deno task test --headless` forwards `--headless` to `deno test`, which rejects it — tooling bug.
- Customer outreach: tell the `ai-chatbot` project owner to migrate to `ai-chatbot.production.veryfront.com`.
- (Optional UX) Consider a more helpful 404 body on bare-subdomain hits pointing customers at the `.production.` form.

### DIFF
--- a/cli/commands/knowledge/command.ts
+++ b/cli/commands/knowledge/command.ts
@@ -1,8 +1,7 @@
 import { z } from "zod";
 type SafeParseResult<T> = { success: true; data: T } | { success: false; error: z.ZodError };
-import { createFileSystem, getEnv } from "veryfront/platform";
+import { type CommandResult, createFileSystem, getEnv, runCommand } from "veryfront/platform";
 import { basename, extname, join, normalize, relative } from "veryfront/platform/path";
-import { type CommandResult, runCommand } from "#veryfront/platform/compat/process.ts";
 import { withSpan } from "veryfront/observability/otlp-setup";
 import { cliLogger } from "#cli/utils";
 import { type ApiClient, createApiClient, resolveConfigWithAuth } from "#cli/shared/config";
--- a/src/platform/index.ts
+++ b/src/platform/index.ts
@@ -14,6 +14,7 @@ export type { RuntimeAdapter } from "./adapters/base.ts";
 // Compat: process
 export {
   chdir,
+  type CommandResult,
   cwd,
   env,
   exit,
--- a/src/proxy/handler.test.ts
+++ b/src/proxy/handler.test.ts
@@ -184,6 +184,46 @@ describe("Proxy Handler", () => {
       await handler.close();
     });
 
+    // Regression test for the ai-chatbot.veryfront.com incident (#1054): bare
+    // {slug}.veryfront.com is intentionally unsupported, so the handler treats it as a
+    // custom domain, the token mint returns "Project not found for domain", and the
+    // user previously saw a misleading 502. Must resolve to a clean 404.
+    it("returns 404 when custom domain token mint reports no project for domain", async () => {
+      const { server, port } = createMockServer((req: Request) => {
+        const { pathname } = new URL(req.url);
+
+        if (pathname === "/auth/token") {
+          return new Response('{"error":"Project not found for domain"}', {
+            status: 400,
+            headers: { "Content-Type": "application/json" },
+          });
+        }
+
+        return createNotFoundResponse();
+      });
+
+      try {
+        const handler = createHandler(port);
+
+        const req = new Request("http://ai-chatbot.veryfront.com/robots.txt", {
+          headers: { host: "ai-chatbot.veryfront.com" },
+        });
+
+        const ctx = await handler.processRequest(req);
+
+        assertEquals(ctx.projectSlug, undefined);
+        assertEquals(ctx.error?.status, 404);
+        assertEquals(
+          ctx.error?.message,
+          "No project configured for domain: ai-chatbot.veryfront.com",
+        );
+
+        await handler.close();
+      } finally {
+        await server.shutdown();
+      }
+    });
+
     it("strips port from custom domain host before token fetch", async () => {
       const tokenRequests: string[] = [];
       const { server, port } = createMockServer((req: Request) => {
--- a/src/proxy/handler.ts
+++ b/src/proxy/handler.ts
@@ -215,6 +215,15 @@ function parseStatusFromError(error: unknown): number | null {
   return match ? Number(match[1]) : null;
 }
 
+// Brittle on purpose: we string-match the API's OAuth error body. Source of truth is
+// veryfront-api/src/api/http/rest/auth/routes.ts — `oauthError(c, 'Project not found
+// for domain', ...)`. If that string is renamed, this regex silently regresses to 502.
+// Durable fix is a typed error code from the token-mint helper; tracked separately.
+function isMissingCustomDomainProjectError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+  return /project not found for domain/i.test(message);
+}
+
 async function extractUserIdFromToken(
   token: string,
   apiBaseUrl: string,
@@ -635,6 +644,13 @@ export function createProxyHandler(options: ProxyHandlerOptions) {
 
       if (isCustomDomain && !projectSlug) {
         if (!token) {
+          if (isMissingCustomDomainProjectError(tokenFetchError)) {
+            logger?.info("Custom domain project not found during token fetch", {
+              domain: host,
+            });
+            return makeErrorContext(base, 404, `No project configured for domain: ${host}`);
+          }
+
           logger?.error("Cannot process custom domain without token", undefined, { domain: host });
           return makeErrorContext(base, 502, `Failed to authenticate for domain: ${host}`, token);
         }
--- a/src/server/utils/domain-parser.ts
+++ b/src/server/utils/domain-parser.ts
@@ -171,6 +171,10 @@ export function parseProjectDomain(host: string): ParsedDomain {
     return createParsedDomain(null, null, env, true, env === "preview");
   }
 
+  // Intentionally NOT supported: bare {slug}.veryfront.{com,org}. Projects must use the
+  // explicit {slug}.production.veryfront.com form. Do not re-add — this has been tried
+  // and reverted (PR #1055). It collides with infra subdomains (api/studio/docs/www)
+  // and erases the environment from the URL, which the product decision requires.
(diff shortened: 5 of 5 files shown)


kojiwakayama:
## Review

This PR is actually two changes bundled, only one of which is in the title. Both look correct, but the bundling needs to be called out before merge.

### What this PR actually does

**Fix A** (in the title): catches \`project not found for domain\` from the token mint and returns 404 instead of 502.

**Fix B** (not in the title): re-introduces production parsing for bare \`{slug}.veryfront.com\` subdomains. The diff inverts a test that previously asserted \`\"veryfront.com bare subdomain is not recognized (no longer supported)\"\`.

These two are tied — \`ai-chatbot.veryfront.com\` (the production incident's smoking gun) is exactly a bare subdomain. Without Fix B, the parser misroutes it into the custom-domain path; without Fix A, the resulting error becomes a 502.

---

### Comments

#### \`src/proxy/handler.ts\` — Fix A

\`\`\`ts
function isMissingCustomDomainProjectError(error: unknown): boolean {
  const message = error instanceof Error ? error.message : String(error);
  return /project not found for domain/i.test(message);
}
\`\`\`

✅ Correct localized fix. Pragmatic: avoids modifying the upstream token-fetch helper to surface a typed error code.

⚠️ **Brittle string match.** If the API rewords this to e.g. \`\"No project bound to domain\"\`, this silently regresses to 502. Worth either:
- adding a code comment naming the exact API source string (file:line in the API repo) so a grep finds it on the next API change, or
- following up with a typed error code from the token mint helper.

⚠️ **Verify error preservation.** This depends on \`tokenFetchError.message\` carrying the API's original body. The PR description shows the helper wraps it as \`OAuth token request failed: 400 - {...}\`, so the regex still hits because the substring is intact. Good. But if a future refactor strips the body, this regresses silently — the test only covers the current shape, not the wrapper contract.

✅ Test is solid: mocks the exact wire response, asserts \`error.status === 404\` and the formatted message.

#### \`src/server/utils/domain-parser.ts\` — Fix B

\`\`\`ts
const prodBaseMatch = matchDomain(domain, \`^([A-Za-z0-9-]+)\\\\.(\${PROD_DOMAINS})\$\`);
if (prodBaseMatch?.[1]) {
  const { slug, branch } = parseSlugAndBranch(prodBaseMatch[1]);
  return createParsedDomain(slug, branch, \"production\", true, false);
}
\`\`\`

🚨 **Re-adds a feature that was previously intentionally removed** — the inverted test name is the smoking gun. The PR body should explain *why* this was removed and why it's being restored. Without that context, the next person reading \`git blame\` will see two contradictory commits and have no way to judge which intent is current.

⚠️ **Catches infra subdomains too.** \`studio.veryfront.com\`, \`api.veryfront.com\`, \`docs.veryfront.com\` would now parse as project slugs named \`studio\`, \`api\`, \`docs\`. In practice these never reach the proxy because they have their own ingress rules, but this assumption isn't enforced by code. Consider either:
- an exclusion list (\`if (slug === \"api\" || slug === \"studio\" || ...) return notRecognized\`), or
- a comment noting that infra subdomain ingress separation is the protective layer.

⚠️ **Branch-suffix path is untested.** \`parseSlugAndBranch\` is called on the matched group, which means \`{slug}--{branch}.veryfront.com\` should be handled. The added test only covers the bare-slug case (\`real-estate-template.veryfront.com\`). Add a test for the branch case if branches on production base subdomains are intended to work; otherwise narrow the regex to exclude \`--\`.

#### \`src/proxy/handler.test.ts\` — coverage

✅ Both new tests are well-structured: explicit mock server responses, full assertion on the resulting context state, proper cleanup with \`try/finally\`.

❌ **Missing the headline regression test.** The PR was prompted by \`ai-chatbot.veryfront.com\` returning 502 in production. There's no end-to-end test that walks the exact production incident: bare \`{slug}.veryfront.com\` request → parser recognizes it → token mint returns success → handler resolves the project. The \"resolves bare production veryfront domains\" test gets close but uses a different slug. Worth either renaming it or adding a code comment tying it to the incident.

#### PR description gaps

- The validation note says \`deno task test --headless\` fails because the task forwards \`--headless\` to \`deno test\`. That's a tooling bug worth a separate issue — log it so it doesn't get lost.

---

### Summary

**Approve with changes:**

1. **Update the PR title and body** to mention Fix B explicitly. Something like \`fix: return 404 for missing custom-domain projects + restore {slug}.veryfront.com production parsing\`.
2. **Add a code comment** in \`domain-parser.ts\` explaining why bare production subdomains are recognized and what the previous \"removed\" commit was solving (so the history is interpretable).
3. **Add a branch-suffix test** for \`{slug}--{branch}.veryfront.com\` if that's intended to work; otherwise narrow the regex.
4. **Add a code comment** in \`handler.ts\` near \`isMissingCustomDomainProjectError\` pointing at the exact API source string so a future API rename gets caught.
5. *Optional follow-up*: thread a typed error code from the token-mint helper instead of regex-matching message strings.

The fix is correct, the tests are reasonable, and the production impact is real. None of my comments block the merge, but #1 and #2 should land in this PR before merging.

kwakayama:
Scope narrowed and evidence verified — ready for another look.

**What changed since the last review (@kojiwakayama):**

1. **Dropped Fix B** (bare `{slug}.veryfront.com` parsing) — reverted in `f5938cc`. Product decision is that bare subdomains are intentionally unsupported; customers must use `{slug}.production.veryfront.com`. Added a tombstone comment in `domain-parser.ts` so this doesn't get re-tried a third time, referencing this PR.

2. **Verified the incident from Loki** (past 7 days, `{service_name="veryfront-proxy"} |= "ai-chatbot"`):
   - 100% of traffic hits the bare `ai-chatbot.veryfront.com` form — zero hits on `ai-chatbot.production.veryfront.com`.
   - 67× 502 + 66× 400 with the exact chain `Token fetch failed` → `Cannot process custom domain without token` → `502 GET /robots.txt`.
   - Last occurrence 2026-04-15T14:27Z on `veryfrontVersion: 0.1.141`.

3. **Tightened Fix A** per the earlier review comments:
   - Added a code comment in `handler.ts` pinning the exact API source string (`veryfront-api/src/api/http/rest/auth/routes.ts` — `oauthError(c, 'Project not found for domain', ...)`) so a future rename is greppable.
   - Updated the regression test to use the real incident hostname `ai-chatbot.veryfront.com` instead of the synthetic `missing-domain.example`.

4. **Unrelated prerequisite:** the pre-commit `cli-boundary` lint was blocked by a pre-existing deep-import in `cli/commands/knowledge/command.ts` (from commit `781518735`). Fixed inline by re-exporting `CommandResult` from `veryfront/platform` and routing the import through the public barrel. Not ideal to bundle, but it was blocking the commit.

**Follow-ups filed separately:**
- #1097 — thread a typed error code from the token-mint helper instead of regex-matching the error body (the durable replacement for the brittle string match flagged in review).
- #1098 — `deno task test --headless` tooling bug from the original PR description.

**Not a code task but worth flagging:** the `ai-chatbot` project owner needs to migrate their DNS / deploy URL to `ai-chatbot.production.veryfront.com`. After this PR lands they'll see a clean 404 instead of a 502, but their site is still broken from their side until they migrate.

PR #1058, opened 2026-04-14
feat(providers): unified cache usage reporting across Anthropic/OpenAI/Google

kojiwakayama:
## Summary

Surfaces per-provider cache metrics on the unified usage shape returned by every runtime. Closes #1046. Part of #1044. Stacked on #1057 (prompt caching).

## Motivation

#1057 lands Anthropic prompt caching request-side, but callers currently have no way to see whether cache hits actually happen — the unified usage object only exposes `inputTokens`, `outputTokens`, and `totalTokens`. All three providers return separate cache counters in their response payloads:

- **Anthropic**: `usage.cache_creation_input_tokens` + `usage.cache_read_input_tokens` (docs: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#tracking-cache-performance)
- **OpenAI**: `usage.prompt_tokens_details.cached_tokens` (docs: https://platform.openai.com/docs/guides/prompt-caching, server-side automatic caching)
- **Google**: `usageMetadata.cachedContentTokenCount` (docs: https://ai.google.dev/gemini-api/docs/caching)

Without these fields propagating through the runtime, dashboards and telemetry can't measure cache hit-rate, which makes #1057 effectively invisible.

## Design

New shared `RuntimeUsage` type alias replaces the repeated inline type literal on every response/stream shape:

\`\`\`ts
type RuntimeUsage = {
  inputTokens?: number;
  outputTokens?: number;
  totalTokens?: number;
  cacheCreationInputTokens?: number;
  cacheReadInputTokens?: number;
};
\`\`\`

Field naming follows `@ai-sdk/anthropic` conventions for familiarity. `cacheCreationInputTokens` is Anthropic-specific (prompt cache writes); `cacheReadInputTokens` is unified across all three providers (Anthropic cache_read, OpenAI cached_tokens, Google cachedContentTokenCount).

**Backward-compat contract (from #1046 acceptance criteria):** cache fields are **only** present when the provider actually returned them. Parsers use conditional spread (`...(typeof x === "number" ? { x } : {})`) so existing usage assertions like `{ inputTokens, outputTokens, totalTokens }` still deep-equal cleanly.

## Per-provider parsing

- `extractAnthropicUsage`: reads `cache_creation_input_tokens` + `cache_read_input_tokens` from `usage`
- `extractOpenAIUsage`: reads `prompt_tokens_details.cached_tokens` and maps to `cacheReadInputTokens`
[...]

Issue #1047, opened 2026-04-14
Reasoning / thinking support across providers (Anthropic adaptive, OpenAI reasoning, Gemini thoughts)


kojiwakayama:
## Summary

veryfront-code parses thinking/reasoning content blocks in streams but does not set the corresponding request-side fields for any of the three providers, so callers cannot actually enable extended reasoning through our library. Opus 4.6's headline capability (adaptive thinking) is unreachable; OpenAI o1/o3/o4 reasoning models won't surface their reasoning summaries; Gemini 2.5+ thought summaries are dropped.

Part of the provider capability audit — see #1044.

## Current state

- `runtime-loader.ts` has streaming parsers for `thinking_delta` and reasoning content blocks (Anthropic paths around lines 858–1031, Google around 1467–1510) and emits `reasoning-start` / `reasoning-delta` / `reasoning-end` UI events.
- `buildAnthropicMessagesRequest`, `buildOpenAIChatRequest`, and `buildGoogleGenerateContentRequest` never set thinking/reasoning fields on the outbound payload.
- Thinking signatures are not extracted or replayed — multi-turn Anthropic conversations with thinking blocks will fail on the second turn.
- No redacted-thinking handling.

### Anthropic

- Add request-side `thinking: { type: "enabled" | "adaptive", budget_tokens: number }` pass-through via a new `reasoning` option on `OpenAICompatibleLanguageOptions`.
- Support Claude 4.6 adaptive thinking effort levels (low / medium / high / max) and map to `budget_tokens` accordingly.
- Disable `temperature` / `top_k` / `top_p` when thinking is enabled (Anthropic rejects the combo). Emit warnings instead of failing silently.
- Extract `signature` from assistant thinking blocks on response parsing and surface it on the reasoning UI event.
- Echo thinking signatures on multi-turn replay so subsequent turns don't reject the request.
- Handle `redacted_thinking` content blocks — pass them through opaquely.
- Adjust `max_tokens` computation: `max_tokens = baseMaxTokens + thinkingBudget` and clamp the sum at the model's known maximum. Depends on the model-aware capability table added in #1043.

### OpenAI

- Support reasoning models (o1, o3, o4 family). Detect by model-id prefix.
- Emit `reasoning_effort: "low" | "medium" | "high"` when set.
- Sanitize incompatible params: reject `temperature`, `top_p`, `presence_penalty`, `frequency_penalty`, `logprobs` with a warning when targeting a reasoning model.
[...]

Issue #1044, opened 2026-04-14
Provider capability audit: feature-gap catalog vs ecosystem peers (Anthropic + OpenAI + Google)


kojiwakayama:
## Summary

Master tracking issue cataloguing known gaps in `src/provider/runtime-loader.ts` and adjacent files when veryfront-code is compared against the Anthropic Messages API, OpenAI Chat Completions / Responses API, and Google Generative AI API directly, as well as against ecosystem peers (`@ai-sdk/anthropic`, `@ai-sdk/openai`, `@ai-sdk/google`, `langchain-anthropic`, `litellm`).

This catalog covers **all three providers**, not just Anthropic. Individual implementation issues are linked below.

### Why this exists

During the 2026-04-14 staging incident investigation we traced a user-visible "Creating file stuck" bug to the hardcoded `max_tokens: 1024` default in `buildAnthropicMessagesRequest`, fixed in #1043. That turned into a broader audit — once we started comparing our provider request/response handling to `@ai-sdk/anthropic` and LiteLLM, multiple other gaps surfaced across all three providers.

Treating them as one catalog prevents each fix from being a one-off and lets us prioritize by cross-provider vs per-provider impact.

## Cross-provider gaps (affect Anthropic + OpenAI + Google)

| # | Gap | Issue | Per-provider notes |
|---|---|---|---|
| A1 | **Prompt / context caching** | #1045 | Anthropic: `cache_control: ephemeral`; OpenAI: automatic prompt cache hits; Google: `cachedContent` API |
| A2 | **Cache usage reporting** | #1046 | Anthropic: `cache_creation_input_tokens` + `cache_read_input_tokens`; OpenAI: `prompt_tokens_details.cached_tokens`; Google: `cachedContentTokenCount` |
| A3 | **Reasoning / thinking support** | #1047 | Anthropic: adaptive thinking, budget, signature, redacted blocks; OpenAI: o1/o3/o4 reasoning + `reasoning_effort`; Google: Gemini 2.5+ thinking + thought summaries |
| A4 | **Overloaded / retry-on-transient** | #1048 | Anthropic: 529; OpenAI: 429 + 503 with `Retry-After`; Google: 503 with `Retry-Info` |
| A5 | **Unsupported-feature warnings** | #1049 | Silent-drop cleanup across every provider mismatch — presencePenalty, frequencyPenalty, seed, topP+temperature, reasoning-model param conflicts |
[...]

PR #1059, opened 2026-04-14
feat(providers): unified reasoning / thinking request option

kojiwakayama:
## Summary

Adds a unified `reasoning` option to `OpenAICompatibleLanguageOptions` that turns on extended reasoning / thinking across Anthropic Claude 4.x, OpenAI o1/o3/o4, and Gemini 2.5+ from a single effort enum. Refs #1047. Part of #1044. Stacked on #1058 (cache usage reporting).

## Motivation

Before this PR, the runtime parses thinking/reasoning content **on the response side** (Anthropic `thinking_delta`, OpenAI `reasoning_content`, Gemini `thought` parts all flow through to `reasoning-*` UI events) but never **requests** reasoning on any provider — so callers couldn't actually turn it on through this library. Opus 4.6's headline capability (adaptive thinking) was unreachable.

## Design

New option:

\`\`\`ts
reasoning?: {
  enabled?: boolean;
  effort?: \"low\" | \"medium\" | \"high\" | \"max\";
  budgetTokens?: number;
}
\`\`\`

Per-provider translation:

| Effort   | Anthropic `budget_tokens` | OpenAI `reasoning_effort` | Google `thinkingBudget` |
|----------|---------------------------|---------------------------|-------------------------|
| `low`    | 1024                      | \`low\`                     | 512                     |
| `medium` | 4096                      | \`medium\`                  | 2048                    |
| `high`   | 16384                     | \`high\`                    | 8192                    |
| `max`    | 32768                     | \`high\` (collapsed)        | -1 (dynamic)            |

When `reasoning.budgetTokens` is set, it overrides the effort mapping for providers that take a numeric budget (Anthropic, Google).

### Anthropic (https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking)

- Emits `thinking: { type: \"enabled\", budget_tokens: N }` on the Messages API.
- **Grows `max_tokens` by the thinking budget**, then clamps at the model's advertised cap from `getAnthropicModelCapabilities` (shared with #1043). Example: Opus 4.6 with `maxOutputTokens: 100_000` + `budgetTokens: 64_000` → request hits `max_tokens: 128_000` (the cap), not 164k.
- **Silently drops `temperature` and `top_p`** when thinking is enabled — the API rejects the combination outright, so dropping them is the only way to keep the request alive.

### OpenAI (https://platform.openai.com/docs/guides/reasoning)

- **Detects reasoning models by id prefix** (`/^o[134](-|$)/` — o1, o3, o4, plus dashed variants like `o3-mini`). On a detected reasoning model, the sanitization applies even without `reasoning.enabled`, because the API rejects sampling params on reasoning models regardless.
- Emits `reasoning_effort: \"low\" | \"medium\" | \"high\"` when `reasoning.enabled`. `\"max\"` collapses to `\"high\"` (OpenAI doesn't expose an above-high tier on Chat Completions).
- **Silently drops `temperature` / `top_p` / `presence_penalty` / `frequency_penalty`** on reasoning requests.

### Google (https://ai.google.dev/gemini-api/docs/thinking)

- Emits `generationConfig.thinkingConfig: { includeThoughts: true, thinkingBudget: N }` on Gemini 2.5+ requests.
- `thinkingBudget: -1` maps Google's \"dynamic / no cap\" mode (effort `max`).
- Thought summaries already stream through the existing Google response parser as `reasoning-*` UI events — enabling `includeThoughts` is the missing request-side half.

### Redacted thinking

Added a `redacted_thinking` case in the Anthropic stream parser. When Claude's safety classifier flags a thinking trace, the block type changes from `thinking` to `redacted_thinking` with an opaque encrypted `data` field. We surface it as a zero-length reasoning block (`reasoning-start` + `reasoning-end` with no deltas) so callers see that thinking happened without leaking the (legitimately hidden) contents.

## Scope notes

Closes the **request-side enable** half of #1047. Two acceptance criteria are deferred to a follow-up because they require extending `RuntimePromptMessage` (a public type change that warrants its own PR):

1. **Anthropic thinking signature round-trip on multi-turn** — requires an `assistant` message variant that carries opaque `thinking` parts with their `signature` field, plus replay logic in `toAnthropicMessages`. A follow-up issue will extend the prompt type.
2. **OpenAI Responses API reasoning summary parsing** — the current Chat Completions path already streams `reasoning_content` deltas; the Responses API has a richer `reasoning.summary` structure that's a follow-up.

This PR delivers the primary customer value: **Opus 4.6 adaptive thinking is now reachable through the veryfront-code runtime**.

## Testing

16 new test cases under `describe(\"reasoning / thinking request options\")`:

- Anthropic: default medium effort, effort→budget mapping, explicit `budgetTokens` override, `max_tokens` clamping at model cap, temperature/top_p drop when enabled, temperature/top_p preserved when disabled, `thinking` field omitted when `enabled: false`
- OpenAI: `reasoning_effort` emission, `max`→`high` collapse, auto-detection of o1/o3/o4 reasoning models, sampling param drop on reasoning models, sampling param preservation on non-reasoning models
- Google: `thinkingConfig` emission, effort mapping, `max`→`-1`, explicit `budgetTokens` override, `thinkingConfig` absent when reasoning disabled
- Anthropic stream: `redacted_thinking` blocks pass through as content-free reasoning events

All 1341 tests pass (16,341 steps in 22s). `deno fmt --check` + `deno check` clean.

## Version

Patch bump `0.1.208 → 0.1.209`.

## Test plan

- [x] \`deno test --no-check --allow-all src/provider/runtime-loader.test.ts\`
- [x] \`deno fmt --check\`
- [x] \`deno check src/provider/runtime-loader.ts src/provider/runtime-loader.test.ts\`
- [x] Full suite via pre-push hook (1341 tests)

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.206",
+  "version": "0.1.207",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.207",
+  "version": "0.1.208",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.208",
+  "version": "0.1.209",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/src/provider/runtime-loader.test.ts
+++ b/src/provider/runtime-loader.test.ts
@@ -2053,6 +2053,331 @@ describe("provider/runtime-loader", () => {
     });
   });
 
+  describe("reasoning / thinking request options", () => {
+    const userPrompt = {
+      role: "user",
+      content: [{ type: "text", text: "Solve this" }],
+    } as const;
+
+    function createAnthropicCaptureRuntime(modelId = "claude-opus-4-6") {
+      let capturedBody: Record<string, unknown> | null = null;
+      const runtime = createAnthropicModelRuntime({
+        apiKey: "test-anthropic-key",
+        baseURL: "https://example.anthropic.test/v1",
+        fetch: (_input, init) => {
+          const raw = readRequestBody(init);
+          capturedBody = raw ? (JSON.parse(raw) as Record<string, unknown>) : null;
+          return Promise.resolve(
+            new Response(
+              JSON.stringify({
+                content: [{ type: "text", text: "ok" }],
+                stop_reason: "end_turn",
+                usage: { input_tokens: 1, output_tokens: 1 },
+              }),
+              { status: 200, headers: { "content-type": "application/json" } },
+            ),
+          );
+        },
+      }, modelId);
+      return { runtime, getBody: () => capturedBody };
+    }
+
+    function createOpenAICaptureRuntime(modelId: string) {
+      let capturedBody: Record<string, unknown> | null = null;
+      const runtime = createOpenAIModelRuntime({
+        apiKey: "test-openai-key",
+        baseURL: "https://example.openai.test/v1",
+        fetch: (_input, init) => {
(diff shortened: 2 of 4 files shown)


kojiwakayama:

PR #1064, opened 2026-04-15
feat(anthropic): short-name aliases for provider tool versions

kojiwakayama:
## Summary

Adds an alias map so callers can pass un-versioned Anthropic provider tool ids (\`anthropic.code_execution\`, \`anthropic.text_editor\`, etc.) and have the runtime forward the latest documented versioned type to the Messages API. Refs **B4** in #1051. Part of #1044. Stacked on #1063.

## Motivation

Anthropic provider tools are versioned with date stamps (e.g. \`code_execution_20260120\`, \`text_editor_20250728\`). Callers using veryfront-code today have to either pin the exact versioned id when registering provider tools or read the Anthropic docs every time a new version drops. \`@ai-sdk/anthropic\` papers over this with a short-name → versioned-name map. This PR ports the same convention.

## Aliases

| Short name | Latest versioned type |
|------------|----------------------|
| \`code_execution\` | \`code_execution_20260120\` |
| \`computer_use\` / \`computer\` | \`computer_20250124\` |
| \`text_editor\` | \`text_editor_20250728\` |
| \`bash\` | \`bash_20250124\` |
| \`memory\` | \`memory_20250818\` |
| \`web_search\` | \`web_search_20250305\` |
| \`web_fetch\` | \`web_fetch_20250910\` |

Versions chosen here are the latest documented as of 2026-04-15 — see https://docs.claude.com/en/docs/agents-and-tools/tool-use/overview.

## Backward compat

- **Already-versioned ids pass through verbatim.** \`anthropic.code_execution_20250522\` (an older but still-valid version) still produces \`type: \"code_execution_20250522\"\`. Detection is via \`/_\\d{8}\$/\` regex on the type string.
- **Unknown short names pass through unchanged.** \`anthropic.future_tool\` still produces \`type: \"future_tool\"\` so future tool releases work without code changes here.

## Testing

11 new test cases under \`describe(\"Anthropic provider tool version aliasing\")\`:

- 8 short-name → versioned-type mappings (one per alias entry)
- Pinned version pass-through (\`code_execution_20250522\`)
- Unknown short name pass-through (\`future_tool\`)
- Type extraction from outbound request body

All 1341 tests pass (16,375 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.212 → 0.1.213\`.

## Test plan

- [x] \`deno test --no-check --allow-all src/provider/runtime-loader.test.ts\` — 99 steps pass
- [x] \`deno fmt --check\`
[...]

Issue #1051, opened 2026-04-14
Anthropic provider feature rollup (B1–B8): thinking guards, signatures, newer tools, MCP, container, citations


kojiwakayama:
## Summary

Rollup tracking Anthropic-only gaps from the provider capability audit (#1044). These are capabilities specific to Anthropic that don't have cross-provider counterparts, so they live here as a grouped to-do rather than as individual cross-provider issues.

## Checklist

- [ ] **B1 — Disable `temperature`/`topK`/`topP` when extended thinking is enabled.** Anthropic rejects the combination. Part of the thinking-support work in #1047 (cross-provider reasoning) but scoped to Anthropic specifically.
- [ ] **B2 — Thinking signature on multi-turn replay.** Extract `signature` from returned thinking blocks and echo them back on subsequent turns so Claude can continue reasoning. Part of #1047.
- [ ] **B3 — Redacted thinking block handling.** Parse and pass through `redacted_thinking` content blocks without erroring. Part of #1047.
- [ ] **B4 — Newer tool versions.** Expose `code_execution` (20250522 / 20250825 / 20260120), `computer_use`, `text_editor_20250728`, `bash_20250124`, `memory_20250818`. Currently only `web_search` / `web_fetch` are plumbed. See `@ai-sdk/anthropic`'s tool-name map for the full table.
- [ ] **B5 — Native MCP servers on Messages API requests.** Anthropic now supports MCP server definitions directly on the request. veryfront-code's MCP path is entirely client-side — we reload MCP sources into local tools and send them through `tools`. Passing Anthropic's native shape avoids the re-serialization hop.
- [ ] **B6 — `container` / agent skills field.** Anthropic's programmatic tool-calling + agent skills require a `container` field on the request. Not currently supported.
- [ ] **B7 — Citations streaming reconstruction.** `web_search` / `web_fetch` emit citation content blocks mid-stream with page ranges and source URLs. We parse the tool result but don't reconstruct citations. `@ai-sdk/anthropic` exposes them as source parts on the UI message stream.
- [ ] **B8 — `max_tokens + thinkingBudget` math + model-cap clamp.** Depends on #1043 (already shipped). Once request-side thinking lands in #1047, the math needs to add the budget and clamp the sum at the model max.

## Priority

[...]

PR #1065, opened 2026-04-15
feat(anthropic): native MCP server pass-through on Messages API

kojiwakayama:
## Summary

Adds an `mcpServers` option that forwards directly to Anthropic's native `mcp_servers` field on the Messages API request body. Refs **B5** in #1051. Part of #1044. Stacked on #1064.

## Motivation

Today veryfront-code's MCP path is entirely client-side: callers register MCP sources, the framework reloads each MCP source into local function tools, and those tools get sent through the existing `tools` field. This works but means every MCP turn pays the re-serialization cost and the framework can't take advantage of Anthropic's server-side MCP routing optimizations.

Anthropic now supports MCP server definitions directly on the Messages API request body — see [Anthropic MCP docs](https://docs.claude.com/en/api/agent-sdk/mcp). Passing the native shape gives callers a way to skip the re-serialization hop entirely.

## API

\`\`\`ts
mcpServers?: Array<Record<string, unknown>>;
\`\`\`

Each entry is forwarded with camelCase keys deep-converted to snake_case so the wire format matches Anthropic's expected shape:

| Caller passes (camelCase) | Wire format (snake_case) |
|---------------------------|--------------------------|
| \`authorizationToken\` | \`authorization_token\` |
| \`toolConfiguration\` | \`tool_configuration\` |
| \`toolConfiguration.allowedTools\` | \`tool_configuration.allowed_tools\` |

Added a new \`deepSnakeCase\` helper that walks nested objects and arrays (the existing \`toSnakeCaseRecord\` was shallow, used for provider tool args at the top level only).

**Beta header:** Anthropic gates MCP behind \`anthropic-beta: mcp-client-2025-04-04\`. Callers must set the header explicitly via the existing \`headers\` option — the runtime intentionally doesn't auto-set it because the beta token may rotate and we don't want to silently send stale ones.

## Testing

2 new test cases under \`describe(\"Anthropic native MCP server pass-through\")\`:

1. \`mcp_servers\` lands on the body with full deep snake_case conversion (verifies \`authorizationToken\` and nested \`toolConfiguration.allowedTools\` both convert correctly)
2. \`mcp_servers\` is omitted when the option is empty array OR unset (backward-compat check)

All 1341 tests pass (16,377 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.213 → 0.1.214\`.

## Test plan

- [x] \`deno test --no-check --allow-all src/provider/runtime-loader.test.ts\` — 102 steps pass
- [x] \`deno fmt --check\`
- [x] \`deno check src/provider/runtime-loader.ts src/provider/runtime-loader.test.ts\`
- [x] Full suite via pre-push hook (1341 tests)

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.209",
+  "version": "0.1.210",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.212",
+  "version": "0.1.213",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.206",
+  "version": "0.1.207",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.207",
+  "version": "0.1.208",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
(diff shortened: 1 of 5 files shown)


kojiwakayama:

PR #1066, opened 2026-04-15
feat(anthropic): container / agent skills field pass-through

kojiwakayama:
## Summary

Adds an `anthropicContainer` option that forwards to the Anthropic Messages API `container` field. Refs **B6** in #1051. Part of #1044. Stacked on #1065.

## Motivation

Anthropic's programmatic tool calling, agent skills, Computer Use, and code execution sandboxes all use a `container` field on the request body to scope a session to a specific sandboxed environment. veryfront-code currently has no surface for this, so callers using any of those features have to reach for `providerOptions` instead of a typed option.

## API

\`\`\`ts
anthropicContainer?: unknown;
\`\`\`

Forwarded verbatim. The shape varies — sometimes a bare container id string, sometimes a structured object with type/config — depending on which feature the caller is targeting. The runtime stays out of the way.

## Testing

3 new test cases:

1. Object form: \`{ id: \"ctr_42\", type: \"computer-use\" }\` lands as-is on \`body.container\`
2. String form: \`\"ctr_42\"\` lands as a bare string on \`body.container\`
3. Omitted from body when \`anthropicContainer\` is unset

All 1341 tests pass (16,378 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.214 → 0.1.215\`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.209",
+  "version": "0.1.210",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.212",
+  "version": "0.1.213",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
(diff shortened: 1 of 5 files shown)


kojiwakayama:

PR #1067, opened 2026-04-15
feat(anthropic): thinking signature multi-turn replay

kojiwakayama:
## Summary

Closes the multi-turn replay gap left by #1059: adds a \`reasoning\` content part to \`RuntimePromptMessage\` and wires it through both directions so callers can roundtrip Anthropic thinking blocks across turns. Refs **B2** in #1051. Part of #1044. Stacked on #1066.

## Motivation

#1059 enabled extended thinking on the request side but the multi-turn replay piece was deferred because it required extending the public \`RuntimePromptMessage\` type. Without it, callers who use thinking get back text + tool calls and lose the signed thinking trace — which means on the next turn Anthropic has no way to verify continuity and Claude can't \"continue\" reasoning from where it left off.

Anthropic's docs are explicit: ([Extended thinking — Multi-turn](https://docs.claude.com/en/docs/build-with-claude/extended-thinking#multi-turn-conversations)) the signed thinking block must be replayed verbatim on the next request, otherwise the API rejects the conversation.

## Design

New variant on the assistant message content union:

\`\`\`ts
| {
    type: \"reasoning\";
    text?: string;          // cleartext thinking trace
    signature?: string;     // Anthropic signature for verification
    redactedData?: string;  // opaque encrypted blob for safety-flagged thinking
  }
\`\`\`

For an unredacted thinking block, callers see \`{ type: \"reasoning\", text, signature }\`. For a safety-redacted block, callers see \`{ type: \"reasoning\", redactedData }\` with no cleartext.

### Response side: \`buildAnthropicGenerateResult\`

Now parses two new block types out of the API response:
- \`thinking\` → \`{ type: \"reasoning\", text, signature }\`
- \`redacted_thinking\` → \`{ type: \"reasoning\", redactedData }\`

The result content array gains a \`reasoning\` discriminant alongside the existing \`text\`, \`tool-call\`, and \`tool-result\` types.

### Replay side: \`toAnthropicMessages\`

The assistant branch now handles \`reasoning\` parts:
- With \`redactedData\` set → emits \`{ type: \"redacted_thinking\", data }\`
- Otherwise → emits \`{ type: \"thinking\", thinking: text, signature }\`

### Cross-provider: drop reasoning on replay

OpenAI Chat Completions and Google Gemini have no equivalent. Their respective \`toOpenAICompatibleMessages\` / \`toGoogleContents\` assistant branches now skip \`reasoning\` parts during replay rather than tripping a typecheck error or failing at runtime. Cross-provider thinking content roundtrips are intentionally a one-way street: thinking captured from Anthropic survives in the prompt history but disappears when the same prompt is sent to a non-Anthropic provider.

## Testing

4 new test cases under \`describe(\"Anthropic thinking signature multi-turn replay\")\`:

1. Generate response containing a \`thinking\` block surfaces \`{ type: \"reasoning\", text, signature }\` on the result content array
2. Generate response containing a \`redacted_thinking\` block surfaces \`{ type: \"reasoning\", redactedData }\`
3. Reasoning parts in an assistant message replay as \`{ type: \"thinking\", thinking, signature }\` on the next request body
4. Redacted reasoning parts replay as \`{ type: \"redacted_thinking\", data }\` on the next request body

All 1341 tests pass (16,381 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.215 → 0.1.216\`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.209",
+  "version": "0.1.210",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.212",
+  "version": "0.1.213",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.206",
+  "version": "0.1.207",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.214",
+  "version": "0.1.215",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.207",
+  "version": "0.1.208",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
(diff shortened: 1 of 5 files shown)


kojiwakayama:

PR #1068, opened 2026-04-15
feat(anthropic): surface citation metadata on text content blocks

kojiwakayama:
## Summary

Adds a \`citations\` field to Anthropic text result content parts so callers can see which source URLs / character ranges / page numbers back each text segment. Refs **B7** in #1051. Part of #1044. Stacked on #1067.

## Motivation

Anthropic's web_search / web_fetch tools and document-citation feature emit per-text-block \`citations\` arrays that link generated text back to specific source ranges. veryfront-code parses the tool result content but throws away the citations on the surrounding text blocks — meaning callers using web search through veryfront-code can't render footnotes, source links, or page references the way \`@ai-sdk/anthropic\` exposes them via source parts.

## Design

New optional field on text content parts:

\`\`\`ts
type AnthropicTextContent = {
  type: \"text\";
  text: string;
  citations?: AnthropicCitation[];
};
\`\`\`

Each citation is normalized through \`normalizeAnthropicCitation\` which converts snake_case wire keys to camelCase and covers all five citation kinds documented at https://docs.claude.com/en/docs/build-with-claude/citations:

| Kind | Key fields |
|------|------------|
| \`web_search_result_location\` | \`url\`, \`title\`, \`citedText\` |
| \`web_fetch_result_location\` | \`url\`, \`title\`, \`citedText\` |
| \`char_location\` | \`documentIndex\`, \`documentTitle\`, \`startCharIndex\`, \`endCharIndex\` |
| \`page_location\` | \`documentIndex\`, \`startPageNumber\`, \`endPageNumber\` |
| \`content_block_location\` | \`documentIndex\`, \`startBlockIndex\`, \`endBlockIndex\` |

Unknown citation kinds are still surfaced (just with whatever fields the normalizer recognizes), so future Anthropic citation types pass through gracefully.

**Backward compat:** \`citations\` is omitted entirely from text content parts that don't have any. Existing callers iterating \`content[].text\` see no shape change.

## Testing

3 new test cases under \`describe(\"Anthropic thinking signature multi-turn replay\")\` (sharing the existing fixture):

1. Web search citation — \`web_search_result_location\` with url/title/citedText surfaces with camelCase keys
2. Mixed char_location + page_location — both kinds normalize correctly with their kind-specific fields
3. Text without citations — \`citations\` field omitted from the result

All 1341 tests pass (16,384 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.216 → 0.1.217\`.

Issue #1052, opened 2026-04-14
OpenAI provider feature rollup (C1–C6): structured outputs, reasoning models, service tiers, parallel tool calls


kojiwakayama:
## Summary

Rollup tracking OpenAI-only gaps from the provider capability audit (#1044). These are capabilities specific to OpenAI that don't have cross-provider counterparts.

## Checklist

- [ ] **C1 — Structured outputs / `response_format: json_schema`.** OpenAI supports a `response_format` with a `json_schema` variant that constrains the model to a schema. veryfront-code has no surface for this. Would unlock structured-output callers without the hack of defining a one-shot tool and ignoring the tool-call wrapper.
- [ ] **C2 — Reasoning models (o1 / o3 / o4 family) request-param sanitization.** These models reject `temperature`, `top_p`, `presence_penalty`, `frequency_penalty`, `logprobs`. Our code sends them and gets a 400. Detection-by-prefix is needed. Feeds into #1049 (warnings) and #1047 (reasoning support).
- [ ] **C3 — `reasoning_effort` field.** The o-series models accept `reasoning_effort: "low" | "medium" | "high"`. Part of #1047.
- [ ] **C4 — Reasoning summary streaming.** o-series models can stream reasoning summaries on the Responses API. Part of #1047.
- [ ] **C5 — `service_tier` (default / flex / scale).** OpenAI's flex tier trades latency for lower cost. Useful for batchy workloads.
- [ ] **C6 — `parallel_tool_calls: false` plumbing.** OpenAI accepts `parallel_tool_calls: false` to force sequential tool use. We don't plumb it. Some callers need it for ordered side effects.

## Priority

Medium. C2 / C3 / C4 converge with the cross-provider reasoning work in #1047 and will probably land together. C1 is the biggest standalone unlock.

Refs #1044.

kojiwakayama:
## Status update from autopilot pass through #1044

C-series items already delivered by earlier cross-provider PRs:

- ✅ **C2** — o1/o3/o4 reasoning model param sanitization — shipped in #1059 (`isOpenAIReasoningModel` detection by id prefix, drops `temperature` / `top_p` / `presence_penalty` / `frequency_penalty` at build time). Surfaces structured warnings via #1061.
- ✅ **C3** — `reasoning_effort` field — shipped in #1059. The unified `reasoning.effort` option maps to OpenAI's `reasoning_effort` enum (`low` / `medium` / `high`, `max` collapses to `high`).

[...]

PR #1069, opened 2026-04-15
feat(openai): structured outputs via response_format json_schema

kojiwakayama:
## Summary

Adds a unified \`responseFormat\` option that constrains model output to JSON or to a specific JSON schema. Refs **C1** in #1052. Part of #1044. Stacked on #1068.

## Motivation

OpenAI exposes a [Structured Outputs](https://platform.openai.com/docs/guides/structured-outputs) feature on Chat Completions (gpt-4o-2024-08-06+) that constrains the model to return strict-schema-conforming JSON. veryfront-code had no surface for this — callers were forced to define a one-shot tool and ignore the tool-call wrapper, which is wasteful and gets in the way of integrations that expect plain content rather than tool calls.

## API

\`\`\`ts
responseFormat?:
  | { type: \"text\" }
  | { type: \"json\" }
  | {
      type: \"json_schema\";
      name: string;
      schema: unknown;
      description?: string;
      strict?: boolean;
    };
\`\`\`

| Variant | OpenAI wire shape |
|---------|-------------------|
| \`{ type: \"text\" }\` | (omitted — default) |
| \`{ type: \"json\" }\` | \`{ type: \"json_object\" }\` |
| \`{ type: \"json_schema\", ... }\` | \`{ type: \"json_schema\", json_schema: { name, schema, strict } }\` |

Schemas pass through \`unwrapToolInputSchema\` so callers can reuse the same \`RuntimeJsonSchema { jsonSchema }\` wrapper they already use for tool input schemas.

## Cross-provider behaviour

- **OpenAI**: emits the structured \`response_format\` shape on the Chat Completions request body.
- **Anthropic**: structured response_format has no equivalent on the Messages API. The runtime emits a structured warning (\`{ type: \"unsupported-setting\", setting: \"responseFormat\", ... }\`) and silently omits the field from the body. Anthropic's idiomatic workaround is to define a tool with the schema as \`input_schema\` and let Claude call it — that's a follow-up PR.
- **Google**: same pattern — emits a warning and omits. Gemini has \`generationConfig.responseSchema\` + \`responseMimeType\` for structured outputs but it's a separate surface that needs its own follow-up.

## Testing

3 new test cases:

1. OpenAI gpt-4o-2024-08-06 with \`{ type: \"json_schema\", name: \"Person\", schema, strict: true }\` produces the right \`response_format\` shape on the wire (full nested structure assertion)
2. OpenAI gpt-4o-mini with \`{ type: \"json\" }\` produces \`response_format: { type: \"json_object\" }\`
3. Anthropic with \`{ type: \"json\" }\` omits \`response_format\` from the body AND emits a \`responseFormat\` warning on the result

All 1341 tests pass (16,387 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.217 → 0.1.218\`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.209",
+  "version": "0.1.210",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.212",
+  "version": "0.1.213",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.217",
+  "version": "0.1.218",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.206",
+  "version": "0.1.207",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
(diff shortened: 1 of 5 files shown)


kojiwakayama:

PR #1070, opened 2026-04-15
feat(openai): service_tier and parallel_tool_calls pass-through

kojiwakayama:
## Summary

Two thin pass-through options on OpenAI Chat Completions:

- \`serviceTier\` — maps to the \`service_tier\` field
- \`parallelToolCalls\` — maps to \`parallel_tool_calls\`

Refs **C5** and **C6** in #1052. Part of #1044. Stacked on #1069.

## Why combined

Both items are tiny single-field pass-throughs (~3 lines of build code each). Splitting them across two PRs would be ceremony without value — same risk profile, same test pattern, no reviewer load saved.

## API

\`\`\`ts
serviceTier?: \"auto\" | \"default\" | \"flex\" | \"scale\";
parallelToolCalls?: boolean;
\`\`\`

| Option | OpenAI field | Purpose |
|--------|--------------|---------|
| \`serviceTier: \"flex\"\` | \`service_tier: \"flex\"\` | Lower-priority queue, lower per-token cost, batchy workloads |
| \`serviceTier: \"scale\"\` | \`service_tier: \"scale\"\` | Reserved-capacity tier with strict latency SLOs |
| \`parallelToolCalls: false\` | \`parallel_tool_calls: false\` | Force sequential tool execution for ordered side effects |

Both fields are forwarded verbatim. Anthropic and Google have no equivalent and silently ignore the options.

## Testing

3 new test cases:

1. \`serviceTier: \"flex\"\` lands as \`service_tier: \"flex\"\` on the OpenAI request body
2. \`parallelToolCalls: false\` lands as \`parallel_tool_calls: false\` on the OpenAI request body
3. Both fields omitted from the body when neither option is set (backward compat check)

All 1341 tests pass (16,390 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.218 → 0.1.219\`.

PR #1057, opened 2026-04-14
feat(anthropic): prompt caching via cache_control breakpoints

kojiwakayama:
## Summary

Adds first-class Anthropic prompt caching to the veryfront-code provider layer via `cache_control` breakpoints on system prompts and tool definitions. Closes #1045. Part of #1044.

## Motivation

Anthropic's Messages API supports prompt caching (docs: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching) which delivers ~85% latency reduction and ~90% input-token cost reduction on cache hits. Today the veryfront-code Anthropic adapter emits system prompts as a plain string and tool definitions with no cache markers, so every request pays full prefill cost even for large stable system prompts and tool schemas. This was a contributing factor to the 2026-04-14 staging latency incident on `/v1/messages`.

## Design

New option on `OpenAICompatibleLanguageOptions`:

```ts
cacheControl?: {
  system?: boolean | "5m" | "1h";
  tools?: boolean | "5m" | "1h";
}
```

- `false` / `undefined` → no cache_control emitted (legacy string-form system, unchanged shape)
- `true` / `"5m"` → `{ type: "ephemeral" }` (default 5-minute TTL)
- `"1h"` → `{ type: "ephemeral", ttl: "1h" }` (1-hour TTL, 2× write cost)

When `cacheControl.system` is set, the system prompt switches from string form to array-of-blocks form with `cache_control` on the block. When `cacheControl.tools` is set, `cache_control` is attached to the **last** tool definition in the array — Anthropic walks backward from the breakpoint so this single marker caches the entire tool schema prefix.

## Scope

- ✅ Anthropic: cache_control breakpoints on system + tools
- ⏸️ OpenAI: prompt caching is fully automatic server-side (no request-side knob) — no-op
- ⏸️ Google: `cachedContent` requires out-of-band cache creation via the separate caches API — deferred to a dedicated PR

## Testing

Added a new `describe("Anthropic prompt caching (cache_control breakpoints)")` block to `src/provider/runtime-loader.test.ts` with 9 test cases covering:

1. Default string-form system when caching disabled (backward compat)
2. Array form with `cache_control` when `cacheControl.system: true`
3. `ttl: "1h"` emission when `cacheControl.system: "1h"`
4. `cache_control` attached to the last tool definition when `cacheControl.tools: true`
5. Combined system + tools breakpoints in one request
6. `false` values treated as no-ops
7. Missing system prompt edge case
8. Missing tools edge case
[...]

Issue #1053, opened 2026-04-14
Google / Gemini provider feature rollup (D1–D6): cachedContent, thoughts, code exec, grounding, safety, tool modes


kojiwakayama:
## Summary

Rollup tracking Google / Gemini-only gaps from the provider capability audit (#1044). These are capabilities specific to Google that don't have cross-provider counterparts.

## Checklist

- [ ] **D1 — `cachedContent` request API.** Gemini has an explicit context-caching API where callers create a cached-content resource and reference it on subsequent requests. Pairs with #1045 (cross-provider caching) but has a different shape than Anthropic's inline `cache_control`.
- [ ] **D2 — Thought summaries (`include_thoughts`).** Gemini 2.5+ supports returning reasoning thoughts. We parse `reasoning-*` in the stream for Anthropic but the Google path doesn't plumb the request-side `include_thoughts` flag. Part of #1047 (cross-provider reasoning).
- [ ] **D3 — Code execution tool (Gemini native).** Gemini has a built-in `code_execution` tool. We don't expose it.
- [ ] **D4 — Grounding metadata + search-response fields.** Google's Gemini grounding feature returns `groundingMetadata` with search queries and retrieved snippets. Not parsed or surfaced today.
- [ ] **D5 — Safety filter configuration.** Gemini accepts `safetySettings` per-request controlling content filtering thresholds. Not plumbed.
- [ ] **D6 — Tool calling config modes (AUTO / ANY / NONE).** Partially present in `normalizeGoogleToolChoice` but not fully plumbed for the `allowedFunctionNames` sub-option.

## Priority

Medium-low. D1 converges with #1045; D2 converges with #1047; the rest are standalone unlocks with smaller caller demand than the Anthropic/OpenAI rollups.

Refs #1044.

kojiwakayama:
## Status update from autopilot pass through #1044

D-series items already delivered by earlier cross-provider PRs:

- ✅ **D2** — `include_thoughts` on Gemini 2.5+ — shipped in #1059. The unified `reasoning.enabled` option emits `generationConfig.thinkingConfig: { includeThoughts: true, thinkingBudget }`. Effort levels map to numeric budgets; `max` maps to `-1` (dynamic).

Remaining D-series items for follow-up PRs:

- ⏳ **D1** — `cachedContent` request API (paired with #1045 caching, but distinct shape)
- ⏳ **D3** — Code execution native tool
- ⏳ **D4** — `groundingMetadata` parsing + surface
[...]

PR #1071, opened 2026-04-15
feat(google): cachedContent reference pass-through

kojiwakayama:
## Summary

Adds a \`googleCachedContent\` option that forwards to Gemini's \`cachedContent\` field on \`generateContent\` / \`streamGenerateContent\` requests. Refs **D1** in #1053. Part of #1044. Stacked on #1070.

## Motivation

Gemini's [context caching](https://ai.google.dev/gemini-api/docs/caching) uses a different model than Anthropic's inline cache_control breakpoints (#1057): callers create a cached content resource via the separate \`caches\` REST API, get back a resource name like \`cachedContents/abc123\`, then attach it to subsequent generate requests via the \`cachedContent\` field. veryfront-code had no surface for the attachment side, which made Gemini caching unreachable through this library.

## API

\`\`\`ts
googleCachedContent?: string;
\`\`\`

Resource name format: \`cachedContents/<id>\`. Forwarded verbatim to the request body's \`cachedContent\` field. Cache creation itself stays out of scope — callers use the Gemini caches REST API or SDK to create the cache, then pass the resource name here on each subsequent generate call.

## Testing

2 new test cases:

1. Body contains \`cachedContent: \"cachedContents/abc123\"\` when option is set
2. Field omitted from body when option is unset (backward compat check)

All 1341 tests pass (16,392 steps in 23s). \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.219 → 0.1.220\`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.209",
+  "version": "0.1.210",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.212",
+  "version": "0.1.213",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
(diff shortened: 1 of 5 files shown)


kojiwakayama:

PR #1072, opened 2026-04-15
feat(google): per-request safetySettings pass-through

kojiwakayama:
## Summary

Adds \`googleSafetySettings\` option that forwards to Gemini's \`safetySettings\` field on generate / stream requests. Refs **D5** in #1053. Part of #1044. Stacked on #1071.

## Motivation

Gemini's safety filter blocks certain content categories by default. veryfront-code had no surface for per-request threshold configuration, which made trusted internal flows (e.g. red-teaming, security analysis) need to fight the default filter, and forced higher-stakes flows that wanted stricter filtering to fall back to \`providerOptions\`.

## API

\`\`\`ts
googleSafetySettings?: Array<{
  category: string;   // e.g. \"HARM_CATEGORY_HARASSMENT\"
  threshold: string;  // e.g. \"BLOCK_NONE\" | \"BLOCK_ONLY_HIGH\"
}>;
\`\`\`

Forwarded verbatim. Empty arrays and unset values both result in \`safetySettings\` being omitted from the body. See [Gemini safety settings docs](https://ai.google.dev/gemini-api/docs/safety-settings).

## Testing

2 new test cases:

1. Two-entry safetySettings (HARASSMENT/BLOCK_NONE + HATE_SPEECH/BLOCK_ONLY_HIGH) lands on body verbatim
2. Empty array → field omitted from body (backward compat)

All 1341 tests pass. \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.220 → 0.1.221\`.

PR #1073, opened 2026-04-15
feat(google): native code execution + provider tools (google_search, etc.)

kojiwakayama:
## Summary

Generalizes Google's tool plumbing so callers can mix function tools with native Gemini provider tools (\`code_execution\`, \`google_search\`, \`google_search_retrieval\`) using the existing \`RuntimeToolDefinition\` provider variant. Refs **D3** in #1053. Part of #1044. Stacked on #1072.

## Motivation

Gemini exposes built-in tools that are first-class citizens on the request body but live outside the \`functionDeclarations\` envelope:

- **\`code_execution\`** runs Python in a sandbox and returns results inline
- **\`google_search\`** does live web search grounding
- **\`google_search_retrieval\`** does retrieval-augmented grounding with dynamic snippets

veryfront-code's \`toGoogleTools\` was hard-coded to emit only \`functionDeclarations\` entries, so callers had no way to enable any native tools through this library.

## Design

The provider variant of \`RuntimeToolDefinition\` already supports \`id: \"<provider>.<type>\"\` — Anthropic uses this for \`anthropic.code_execution\` etc. (see #1064). This PR extends \`toGoogleTools\` to accept the symmetric \`google.*\` namespace:

| Tool id | Wire shape |
|---------|------------|
| \`google.code_execution\` | \`{ codeExecution: {} }\` |
| \`google.google_search\` | \`{ googleSearch: {} }\` |
| \`google.google_search_retrieval\` | \`{ googleSearchRetrieval: {} }\` |

Type name conversion is snake_case → camelCase. Caller-provided \`tool.args\` are forwarded as the config payload (most native tools take an empty object today, but \`googleSearchRetrieval\` accepts \`dynamicRetrievalConfig\`).

Function tools and provider tools can be mixed in a single \`tools\` array. The runtime emits a single \`functionDeclarations\` entry holding all the function tools, then one provider entry per native tool. All four entries land in Gemini's \`tools[]\` array together.

\`GoogleCompatibleRequest.tools\` is widened from \`Array<{ functionDeclarations: ... }>\` to \`Array<Record<string, unknown>>\` to accommodate the new shape.

## Testing

2 new test cases:

1. \`google.code_execution\` alone → \`tools: [{ codeExecution: {} }]\`
2. Function tool + \`google.google_search\` mix → \`tools: [{ functionDeclarations: [...] }, { googleSearch: {} }]\`

All 1341 tests pass. \`deno fmt --check\` + \`deno check\` clean.

## Version

Patch bump \`0.1.221 → 0.1.222\`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.221",
+  "version": "0.1.222",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.209",
+  "version": "0.1.210",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.212",
+  "version": "0.1.213",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "exclude": [
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.217",
+  "version": "0.1.218",
   "license": "Apache-2.0",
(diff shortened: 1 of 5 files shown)


kojiwakayama:

PR #1100, opened 2026-04-15
fix: return 404 for unknown custom-domain token mint failures

kwakayama:
## Summary
- return 404 instead of 502 when custom-domain token minting reports `Project not found for domain`
- cover the regression with a proxy handler test
- preserve existing 502 behavior when there is truly no token source available

## Evidence
Production logs at 2026-04-15 21:52 UTC showed repeated errors like:
- `502 GET /wsa.php`
- `Cannot process custom domain without token`
- `OAuth token request failed: 400 - {"error":"Project not found for domain"}`

Those requests targeted `react.veryfront.com`, which is not configured to a project. That should be a clean 404, not a proxy auth failure.

## Validation
- `deno test --no-check --allow-all src/proxy/handler.test.ts src/proxy/token-priority.test.ts`
- `deno task test --headless` currently fails immediately on this toolchain because the underlying `deno test` task does not accept `--headless`; I also started `deno task test` to validate the wider suite, but stopped it after the relevant proxy regression coverage had passed.

### DIFF
--- a/src/proxy/handler.test.ts
+++ b/src/proxy/handler.test.ts
@@ -184,6 +184,46 @@ describe("Proxy Handler", () => {
       await handler.close();
     });
 
+    // Regression test for the ai-chatbot.veryfront.com incident (#1054): bare
+    // {slug}.veryfront.com is intentionally unsupported, so the handler treats it as a
+    // custom domain, the token mint returns "Project not found for domain", and the
+    // user previously saw a misleading 502. Must resolve to a clean 404.
+    it("returns 404 when custom domain token mint reports no project for domain", async () => {
+      const { server, port } = createMockServer((req: Request) => {
+        const { pathname } = new URL(req.url);
+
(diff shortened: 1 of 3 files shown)


kwakayama:
Superseded by #1055.

The diff in this PR is byte-identical to commit `221072fbe` on the #1055 branch — same `isMissingCustomDomainProjectError` helper, same API source-string comment, same test with `ai-chatbot.veryfront.com`, and the `domain-parser.ts` tombstone comment here literally references "(PR #1055)". This branch was derived from #1055's tip.

#1055 is queued for auto-merge; it will also fix the `react.veryfront.com` incident that prompted this PR.

The pattern of recurring customer hits on the unsupported bare `{slug}.veryfront.com` form (`builder` → `ai-chatbot` → `react` in 3 days) is tracked in a new issue for a helpful 404 body suggesting the `.production.` migration path.

PR #1219, opened 2026-04-22
Expose cursor-resync helpers for conversation-run append recovery

kojiwakayama:
## Summary
- add `resyncConversationRunAppendCursor()` to the public durable surface
- keep waiting-tool fields and appendability checks on the public conversation run projection
- bump the package version to `0.1.243`

## Testing
- deno test --no-check --allow-all src/agent/durable.test.ts src/agent/conversation-root-run-context.test.ts
- deno check src/agent/durable.ts src/agent/conversation-root-run-context.ts src/agent/index.ts
- deno eval --config deno.json 'import { resyncConversationRunAppendCursor } from "./src/agent/index.ts"; console.log(typeof resyncConversationRunAppendCursor)'
- full pre-push suite


kojiwakayama:
Addressed in follow-up PR #1220. That fix makes `resyncConversationRunAppendCursor()` treat runs that become waiting/terminal in the same cursor advance as `non_appendable` instead of `advanced`.

PR #1321, opened 2026-04-28
Pre-register Anthropic and Google as built-in providers (Option C)

kwakayama:
## Summary

Single execution flow for AI model resolution. All providers go through the `AIProviderRegistry` — no fallbacks, no per-callsite branching.

**What changed:**
- `createAIProviderRegistry()` pre-registers Anthropic and Google as built-in providers
- `veryfront-cloud/provider.ts` uses `resolve()` + `registry.require()` uniformly for all providers
- Built-ins can be silently overridden by extensions calling `register()` with the same id
- Non-builtin duplicates still throw to catch collisions

**Replaces:** The per-callsite fallback from #1319

## Why

The extension architecture moved Anthropic behind `@veryfront/ext-anthropic` which isn't published. This crashed every consumer that doesn't run `orchestrateExtensions()` (like veryfront-agent). Instead of per-callsite fallbacks, built-in providers in the registry give every consumer a working default while extensions can override when adopted.

## Test plan

- [x] Typecheck passes
- [x] Registry tests pass (8/8) — includes new tests for built-in override behavior
- [x] Provider tests pass (8/8) — includes new test for built-in Anthropic resolution

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.299",
+  "version": "0.1.300",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "workspace": [
--- a/src/extensions/registries/ai-provider-registry.test.ts
+++ b/src/extensions/registries/ai-provider-registry.test.ts
@@ -13,12 +13,26 @@ function fakeProvider(id: string): AIProvider {
 }
 
 describe("AIProviderRegistry", () => {
+  it("ships with built-in anthropic and google providers", () => {
+    const reg = createAIProviderRegistry();
+    assert(reg.has("anthropic"));
+    assert(reg.has("google"));
+    assertEquals(reg.has("openai"), false);
+  });
+
   it("register + get returns the same instance", () => {
     const reg = createAIProviderRegistry();
     const p = fakeProvider("openai");
     reg.register(p);
     assertEquals(reg.get("openai"), p);
     assert(reg.has("openai"));
+  });
+
+  it("register overrides a built-in provider silently", () => {
+    const reg = createAIProviderRegistry();
+    const custom = fakeProvider("anthropic");
+    reg.register(custom);
+    assertEquals(reg.get("anthropic"), custom);
   });
 
   it("get returns undefined for unknown id", () => {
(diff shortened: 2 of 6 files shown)


kwakayama (commented):
Staging check: the runtime failure this PR was intended to prevent does not reproduce on staging anymore, but staging does not implement this PR's exact "Option C" registry design.

Evidence collected 2026-05-04 UTC against `veryfront-staging` / `veryfront-agent`:

- Staging agent image: `ghcr.io/veryfront/veryfront-agent:sha-9bec609ec748bc86e3d1bb46d4eb339fed739a9c-run-25308363148-attempt-1`
- `veryfront-agent` package deps show `veryfront: 0.1.367`.
- A direct model-resolution probe inside the staging agent pod, without running extension orchestration, successfully resolved Anthropic, Google, and OpenAI through `veryfront/provider`:

```json
{"id":"veryfront-cloud/anthropic/claude-sonnet-4-5","ok":true,"modelType":"object","provider":"veryfront-cloud","modelId":"claude-sonnet-4-5"}
{"id":"veryfront-cloud/google/gemini-2.5-pro","ok":true,"modelType":"object","provider":"veryfront-cloud","modelId":"gemini-2.5-pro"}
{"id":"veryfront-cloud/openai/gpt-5.1","ok":true,"modelType":"object","provider":"veryfront-cloud","modelId":"gpt-5.1"}
```

Recent staging agent logs also had no matches for provider-resolution crash signatures:

```text
No AIProvider registered|ext-anthropic|Cannot find module|provider.*not found|anthropic.*not found|google.*not found => 0 matches in last 30m
```

Important caveat: staging appears to solve the user-visible crash via the existing per-callsite fallback path, not via this PR's built-in registry design. The deployed package still has `createAIProviderRegistry()` returning an empty registry, and `veryfront-cloud/provider.js` contains the Anthropic fallback comment/path:

```js
const anthropic = registry?.get("anthropic");
if (anthropic) { ... }
// Fall back to the built-in runtime-loader when the extension is not registered.
return createAnthropicModelRuntime(...)
```

Conclusion: the original runtime breakage for consumers that do not run `orchestrateExtensions()` appears resolved on staging. The exact architectural cleanup proposed here is still not present on staging, so this PR is only needed if we still want the registry-only Option C design rather than the current fallback behavior.

kwakayama:
Closing because staging evidence shows the original runtime breakage no longer reproduces.

PR #1410, opened 2026-05-04
Fix subprocess leak in VULN-FS-5 loader test

ariskemper:
## Summary
- Add `sanitizeResources: false, sanitizeOps: false` to both VULN-FS-5 describe blocks in `src/routing/api/module-loader/loader.test.ts`
- The esbuild child process started by the `loadHandlerModule` describe block was straddling into later test blocks, triggering Deno's resource sanitizer
- Matches the pattern already established by the `loadHandlerModule` describe block

## Test plan
- [x] `deno test src/routing/api/module-loader/loader.test.ts` — 3 passed, 0 failed
- [ ] `deno task test:unit` — 1584 passed, 0 failed


ariskemper:
Closing: empirical testing confirms the VULN-FS-5 describe blocks pass cleanly without sanitizer suppression — the esbuild subprocess from `loadHandlerModule` is properly cleaned up by its `afterAll` hook before these blocks run. The Codex review was correct that the suppression was unnecessary.

PR #1430, opened 2026-05-05
feat(extensions): auto-register built-in AI provider extensions

ariskemper:
## Summary

- Registers ext-openai, ext-anthropic, and ext-google as built-in framework extensions
- Providers are available out of the box when API keys are set in `.env` — no npm install or project stubs needed
- Built-in extensions have the lowest priority (4) so project/package/config extensions can override them
- Users can disable built-ins via `{ name: "ext-anthropic", enabled: false }` in config

## Changes

- `src/extensions/builtin-extensions.ts` — static imports of the three provider factories
- `src/extensions/orchestrate.ts` — new `builtinExtensions` option passed through to merge
- `src/extensions/discovery.ts` — `mergeExtensions` accepts optional `builtin` array (lowest priority)
- `src/extensions/types.ts` — add `"builtin"` to `ExtensionSource` union
- `src/extensions/validation.ts` — add `builtin: 4` to `SOURCE_PRIORITY`
- `src/server/bootstrap.ts` — pass `builtinProviderExtensions` to all three `orchestrateExtensions` calls

## Test plan

- [x] All 1615 tests pass (17271 steps, 0 failures)
- [x] Format, lint, type check pass
- [ ] Manual verification: `veryfront init` + `veryfront dev` with API keys in `.env` — providers work without installing extension packages


kojiwakayama (dismissed):
Thanks for pushing this; it is in the right area, but I think this still has two merge-blocking issues for the current staging failure and for the override contract described in the PR body.

1. `veryfront/provider` consumers still fail outside `server/bootstrap`.

The built-ins are only passed to `orchestrateExtensions()` from `src/server/bootstrap.ts`. Consumers that import the framework as a library and call `resolveModel()` do not run that bootstrap path, so the default `veryfront-cloud/anthropic/*` path still throws. This is the exact shape currently breaking `veryfront-agent`, which consumes the npm package and resolves the model directly during durable chat execution.

I verified this against a local merge of this PR onto current `main`:

```bash
VERYFRONT_API_TOKEN=vf_test VERYFRONT_PROJECT_SLUG=test deno eval \
  'import { resolveModel, clearModelProviders } from "./src/provider/index.ts"; clearModelProviders(); const model = resolveModel("veryfront-cloud/anthropic/claude-sonnet-4-6"); console.log(typeof model.doStream);'
```

It still fails with:

```text
Error: Anthropic provider not installed. Add @veryfront/ext-anthropic to use anthropic/* models via veryfront-cloud.
    at createVeryfrontCloudModel (src/provider/veryfront-cloud/provider.ts:33:13)
```

So the long-term fix needs to make the provider/runtime package itself resolve the built-in provider runtimes for `veryfront-cloud` and direct provider IDs, or make library consumers explicitly initialize the same built-in registry before any provider resolution. Wiring only the Veryfront server bootstrap leaves `veryfront-agent` and other non-server runtimes broken.

2. Built-in AI providers do not actually allow provider-id overrides unless the extension name is identical.

The PR body says project/package/config extensions can override built-ins, but the AI provider extensions register into `AIProviderRegistry` inside `setup()`, not through `Extension.provides`. If a higher-priority extension has a different extension name but registers provider id `anthropic`, `mergeExtensions()` keeps both extensions; the custom extension registers first, then builtin `ext-anthropic` runs and throws a duplicate provider error.

Repro against the same local merge result:

```bash
deno eval 'import { orchestrateExtensions } from "./src/extensions/orchestrate.ts"; import { mergeExtensions } from "./src/extensions/discovery.ts"; import { AIProviderRegistryName } from "./src/extensions/interfaces/index.ts"; import { createAIProviderRegistry } from "./src/extensions/registries/ai-provider-registry.ts"; import { createBuiltinExtensions } from "./src/extensions/builtin-extensions.ts"; const custom={name:"custom-anthropic",version:"1",capabilities:[{type:"contract",name:"AIProvider:anthropic"}],setup(ctx){ctx.require(AIProviderRegistryName).register({id:"anthropic",createModel(){return {provider:"custom",modelId:"m",specificationVersion:"v3",supportedUrls:{},doGenerate(){},doStream(){}}}})}}; await orchestrateExtensions({projectDir:"/tmp/nope",config:{extensions:[custom]},logger:{debug(){},info(){},warn(){},error(){}},primeContracts:{[AIProviderRegistryName]:createAIProviderRegistry()},builtinExtensions:createBuiltinExtensions(),discovery:{discoverPackageExtensions:()=>Promise.resolve([]),discoverProjectExtensions:()=>Promise.resolve([]),discoverLocalExtensions:()=>Promise.resolve([]),mergeExtensions}}); console.log("ok");'
```

It fails with:

```text
Error: AIProvider "anthropic" is already registered. Call unregister("anthropic") first if you intend to replace it.
    at AIProviderRegistryImpl.register (src/extensions/registries/ai-provider-registry.ts:17:13)
    at Object.setup (extensions/ext-anthropic/src/index.ts:22:16)
```

That means the documented override semantics only work if the replacement extension uses the exact same extension name (`ext-anthropic`), not if it provides the same provider contract/id. Either the built-in AI provider loading needs provider-id aware suppression, or provider extensions need to participate in the normal `provides`/priority mechanism so lower-priority built-ins cannot register after a higher-priority provider.

kojiwakayama:
Addressed the two merge-blocking review items in `3cd5f929d`:

- `veryfront/provider` library consumers now resolve built-in `veryfront-cloud/anthropic/*`, `veryfront-cloud/google-ai-studio/*`, and direct built-in provider IDs without relying on `server/bootstrap`.
- Built-in AI provider registration is provider-id aware and skips when a higher-priority extension already registered the same provider id, so overrides no longer require matching the extension name.

Regression coverage added in `src/provider/veryfront-cloud/provider.test.ts` and `src/extensions/orchestrate.test.ts`.

Verification:
- local pre-push passed: format, lint, typecheck, and `1653` tests / `17210` steps / `0` failures.
- GitHub checks are green on `3cd5f929d`: format, lint, typecheck, unit, integration, rsc browser e2e, binary e2e, CodeQL, CLA, and dependency audit.

PR is still blocked only by the prior `CHANGES_REQUESTED` review state.

kojiwakayama:
Pushed root-cause fixes and refreshed the branch on current main (`3526ab5ad`).

Verification evidence from local pre-push gate:
- format: checked 3305 files
- lint: checked 3274 files
- typecheck: passed
- full parallel Deno unit suite: `1660 passed (17210 steps), 0 failed` in 7m39s

The original library-consumer provider failure is covered by regression tests in `src/provider/veryfront-cloud/provider.test.ts` and `src/extensions/orchestrate.test.ts`. The parallel test flake was fixed by isolating `ProjectWorker` test workers and removing the wall-clock micro-benchmark assertion from path validation.

Automerge is enabled; currently blocked only by the existing `CHANGES_REQUESTED` review state while GitHub checks are still running.

PR #1469, opened 2026-05-07
fix: pass native image/file parts through AgentRuntimeMessage pipeline

kwakayama:
## Problem

When a user sends a file attachment (image or PDF), two things break:

1. **File-only messages cause `EMPTY_COMPLETION`** — the LLM receives only an XML annotation with no user intent, produces a minimal/empty response, and the error panel is shown.
2. **The LLM never sees actual file content** — even with text + file, the image pixels or PDF text are stripped and replaced with an annotation.

**Root cause:** `convertContentToAgentRuntimeParts` in `agent-runtime-message-adapter.ts` stripped all `image`/`file` parts from `ProviderModelMessage → AgentRuntimeMessage` conversion, replacing them with an XML annotation: `"Attached files from earlier conversation context: ..."`. When converting back (`AgentRuntimeMessage → ProviderModelMessage`), user messages only received string content, so the LLM never got multimodal input.

## Fix

- Add `{ type: "image"; url: string; mediaType: string }` and `{ type: "file"; url: string; mediaType: string }` variants to `AgentRuntimeMessagePart` and `AgentRuntimeMessageLikePart`
- In `convertStructuredPart`: return native image/file parts for URL-based attachments (signed URLs from veryfront-agent)
- In `convertContentToAgentRuntimeParts`: try native conversion first; fall back to XML annotation only when URL is missing or is a `data:` base64 URL
- In `collectAgentRuntimeProviderContentParts` + `createProviderMessageFromAgentRuntimeMessage`: collect file parts and emit structured `ChatUserContentPart[]` content for user messages
- Add image/file variants to `MessagePartSchema` in agent.schema.ts
- Guard `hosted-child-fork-step-message-preparation.ts` against image/file parts (convert to text annotation for child fork messages)

## What's preserved

The XML annotation fallback is kept for:
- Base64 `data:` URLs (can't pass as URL to provider)
- Parts with no URL (e.g. uploadId-only references before URL resolution)
- Unsupported media types (e.g. DOCX) — rewritten to annotation by upstream `rewriteUnsupportedFilePartsAsAnnotations`

## Providers

All three supported providers (Anthropic, OpenAI, Google) support native URL-based multimodal input. No provider-specific handling required.

## Reproduction script

```
deno run --no-check --allow-all scripts/repro-3499-file-attachment.ts
```

## Evidence

```
Scenario 1: Text + PDF attachment — file passes natively (not as annotation)
  → resolveFileUrl("upload-pdf") → signed URL
  ✓ Text part preserved
  ✓ File part is native (not annotation text)
  ✓ Signed URL forwarded to provider
  ✓ mediaType preserved

Scenario 2: Image-only message — not dropped (was causing EMPTY_COMPLETION)
  ProviderModelMessages:
  [{ "role": "user", "content": [{ "type": "file", "mediaType": "image/png", "url": "https://signed.example.com/screenshot.png" }] }]
  ✓ Message not dropped (EMPTY_COMPLETION fix)
  ✓ Content is structured array (not empty string)
  ✓ File part present in provider message

Scenario 3: Round-trip ProviderModelMessage → AgentRuntime → ProviderModelMessage
  ✓ Image survives round-trip
  ✓ File survives round-trip
  ✓ Image URL preserved
  ✓ File URL preserved

Scenario 4: Unsupported media type (DOCX) → annotation text (expected fallback)
  ✓ DOCX becomes annotation text (appended to message text)

All checks passed.
```

## Related

- Companion fix (URL error propagation): veryfront/veryfront-agent#1210
- Issue: veryfront/veryfront-studio#3499

## Test plan

- [x] `agent-runtime-message-adapter.test.ts` — 8 tests: native passthrough, annotation fallback, round-trip, file-only not dropped
- [x] `runtime-message-preparation.test.ts` — updated to expect native file part after URL resolution
- [x] `hosted-child-fork-step-message-preparation.test.ts` — 3 existing tests pass
- [x] Full `src/agent/` suite: 280 passed, 0 failed
- [x] `scripts/repro-3499-file-attachment.ts` — 4 end-to-end scenarios all pass

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.407",
+  "version": "0.1.408",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "workspace": [
--- a/src/agent/agent-runtime-message-adapter.test.ts
+++ b/src/agent/agent-runtime-message-adapter.test.ts
@@ -130,23 +130,43 @@ describe("agent runtime message adapter", () => {
     assertEquals(agentRuntimeMessages[0]?.parts, [agentRuntimeTextPart("Visible answer")]);
   });
 
-  it("converts multimodal parts into uploaded-file context annotations", () => {
+  it("passes image and file parts with URLs as native AgentRuntimeMessage parts", () => {
     const agentRuntimeMessages = convertProviderMessagesToAgentRuntimeMessages([
       providerMessage({
         role: "user",
         content: [
           {
             type: "image",
-            data: "ignored",
             mediaType: "image/png",
             filename: "diagram.png",
+            url: "https://uploads.example.com/diagram.png",
           },
           {
             type: "file",
-            data: "ignored",
             mediaType: "application/pdf",
             filename: "spec.pdf",
-            url: "https://example.com/spec.pdf",
+            url: "https://uploads.example.com/spec.pdf",
+          },
+        ],
+      }),
+    ]);
+
+    assertEquals(agentRuntimeMessages[0]?.parts, [
+      { type: "image", url: "https://uploads.example.com/diagram.png", mediaType: "image/png" },
+      { type: "file", url: "https://uploads.example.com/spec.pdf", mediaType: "application/pdf" },
+    ]);
+  });
+
+  it("falls back to annotation for multimodal parts without a URL", () => {
+    const agentRuntimeMessages = convertProviderMessagesToAgentRuntimeMessages([
+      providerMessage({
+        role: "user",
+        content: [
+          {
+            type: "image",
+            data: "base64data",
+            mediaType: "image/png",
+            filename: "diagram.png",
           },
         ],
       }),
@@ -158,7 +178,6 @@ describe("agent runtime message adapter", () => {
     if (firstPart && "text" in firstPart) {
       assertStringIncludes(firstPart.text, "<uploaded_files>");
       assertStringIncludes(firstPart.text, "diagram.png");
-      assertStringIncludes(firstPart.text, "spec.pdf");
     }
   });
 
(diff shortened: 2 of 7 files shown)


kwakayama (commented):
## Code Review — 92/100

**Fix is correct and complete.** The three-layer change (type → convert → collect) is logically sound and all paths are covered by tests.

### What's well done

- `toNativeFilePart` helper eliminates the duplicated `data:` URL guard and the `as "image" | "file"` cast — single source of truth for the URL validation logic
- `convertStructuredPart` now falls back to `null` cleanly for data: URLs; annotation collection path is explicit with `continue`
- `collectAgentRuntimeProviderContentParts` symmetric with the provider→runtime direction — both sides filter `data:` URLs consistently
- `convertAgentRuntimePartToChildForkMessagePart` has explicit `image`/`file` guard + `never` throw for future variants
- Single clean commit, bisectable history
- Tests cover: native passthrough, annotation fallback, file-only not dropped, `data:` URL dropped on round-trip, resolver returning `undefined`

### Remaining debt (pre-existing, not introduced here)

**`AgentRuntimeMessageLikePart` uses `{ type: string; toolCallId: ... }` instead of `` `tool-${string}` ``**

This prevents TypeScript from discriminating `type === "image"` cleanly, which is why `toNativeFilePart` takes `type` as an explicit parameter rather than reading `part.type` after narrowing. The helper is the right workaround. The fix for the root cause would be changing the tool-call variants in the union to `` type: `tool-${string}` ``, but that's a separate PR.

**Verdict: approve once CI is green.**

kwakayama (commented):
## Code Review — 92/100

**Fix is correct and complete.** The three-layer change (type → convert → collect) is logically sound and all paths are covered by tests.

### What's well done

- \`toNativeFilePart\` helper eliminates the duplicated \`data:\` URL guard and the \`as "image" | "file"\` cast — single source of truth for the URL validation logic
- \`convertStructuredPart\` now falls back to \`null\` cleanly for data: URLs; annotation collection path is explicit with \`continue\`
- \`collectAgentRuntimeProviderContentParts\` symmetric with the provider→runtime direction — both sides filter \`data:\` URLs consistently
- \`convertAgentRuntimePartToChildForkMessagePart\` has explicit \`image\`/\`file\` guard + \`never\` throw for future variants
- Single clean commit, bisectable history
- Tests cover: native passthrough, annotation fallback, file-only not dropped, \`data:\` URL dropped on round-trip, resolver returning \`undefined\`

### Remaining debt (pre-existing, not introduced here)

**\`AgentRuntimeMessageLikePart\` uses \`{ type: string; toolCallId: ... }\` instead of \`\`\`tool-\${string}\`\`\`**

This prevents TypeScript from discriminating \`type === "image"\` cleanly, which is why \`toNativeFilePart\` takes \`type\` as an explicit parameter rather than reading \`part.type\` after narrowing. The helper is the right workaround. The fix for the root cause would be changing the tool-call variants in the union to \`type: \`tool-\${string}\`\`, but that is a separate PR.

**Verdict: approve once CI is green.**

kwakayama:
## Code Review — 92/100

**Fix is correct and complete.** The three-layer change (type → convert → collect) is logically sound and all paths are covered by tests.

### What's well done

- `toNativeFilePart` helper eliminates the duplicated `data:` URL guard and the `as "image" | "file"` cast — single source of truth for the URL validation logic
- `convertStructuredPart` falls back to `null` cleanly for `data:` URLs; annotation collection path is explicit with `continue`
- `collectAgentRuntimeProviderContentParts` symmetric with the provider→runtime direction — both sides filter `data:` URLs consistently
- `convertAgentRuntimePartToChildForkMessagePart` has explicit `image`/`file` guard + `never` throw for future variants
- Single clean commit, bisectable history
- Tests cover: native passthrough, annotation fallback, file-only not dropped, `data:` URL dropped on round-trip, resolver returning `undefined`

### Remaining debt (pre-existing, not introduced here)

`AgentRuntimeMessageLikePart` uses `{ type: string; toolCallId: ... }` instead of ``tool-${string}`` — prevents TypeScript from discriminating `type === "image"` cleanly after narrowing. The `toNativeFilePart` helper works around it correctly. Root cause fix belongs in a separate PR.

**Verdict: approve once CI is green.**

PR #1654, opened 2026-05-14
Remove chat theme utility dependencies

kojiwakayama:
## Summary
- remove class-variance-authority and tailwind-merge from the root import map
- replace the chat theme variant helper with a small local implementation
- stop re-exporting the third-party variant utility from chat public APIs
- add regression coverage for class joining, variant helpers, and the removed runtime export

## Verification
- deno test --no-check --allow-all src/react/components/chat
- deno check src/react/components/chat/theme.ts src/react/components/chat/theme.test.ts src/react/components/chat/index.ts src/react/public.ts
- deno fmt --check --config=deno.json deno.json src/react/components/chat/theme.ts src/react/components/chat/theme.test.ts src/react/components/chat/index.ts src/react/public.ts src/server/handlers/dev/framework-candidates.generated.ts
- deno task lint:deps
- deno task lint:core-deps
- deno task test:scripts
- deno task audit
- pre-push hook: format, lint, typecheck, unit tests


kojiwakayama:
Addressed the inline review comment in follow-up PR #1656: https://github.com/veryfront/veryfront-code/pull/1656

PR #1688, opened 2026-05-16
fix: sanitize provider tool schema property keys

kwakayama:
## Summary\n- sanitize Anthropic/veryfront-cloud tool schemas so invalid JSON Schema property names are removed before model requests\n- keep Google schema cleanup and run property-key filtering before provider-specific cleanup\n- filter matching required entries when properties are removed\n\n## Evidence\n- Red: provider-tool-compat regression test failed before the implementation with invalid property keys still present\n- Green: `deno task test src/agent/runtime/model-tool-converter.test.ts src/agent/runtime/provider-tool-compat.test.ts`\n- Typecheck: `deno task typecheck`\n- Format/lint: `deno fmt --check src/agent/runtime/provider-tool-compat.ts src/agent/runtime/provider-tool-compat.test.ts` and `DENO_NO_PACKAGE_JSON=1 deno lint src/agent/runtime/provider-tool-compat.ts src/agent/runtime/provider-tool-compat.test.ts`\n- Build: `deno task build`\n\nFixes the provider error: `tools.*.custom.input_schema.properties: Property keys should match pattern '^[a-zA-Z0-9_.-]{1,64}$'`.

### DIFF
--- a/src/agent/runtime/provider-tool-compat.test.ts
+++ b/src/agent/runtime/provider-tool-compat.test.ts
@@ -153,4 +153,48 @@ describe("provider-tool-compat", () => {
     assertEquals(sanitized.properties?.labelIds?.type, "array");
     assertEquals(sanitized.properties?.labelIds?.items, {});
   });
+
+  it("removes Anthropic-incompatible property keys and matching required entries", () => {
+    const sanitized = sanitizeProviderToolSchema(
+      {
+        type: "object",
+        properties: {
+          ok_name: { type: "string" },
+          "bad key": { type: "string" },
+          "nested-object": {
+            type: "object",
+            properties: {
+              "also/bad": { type: "string" },
+              fine: { type: "number" },
+            },
+            required: ["also/bad", "fine"],
+          },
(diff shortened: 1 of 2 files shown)


kwakayama:
## Critical review score: 94/100\n\nScope is intentionally narrow and matches the production provider error. The fix sanitizes only provider profiles that already opt into schema cleanup (now Anthropic and Google), drops schema properties Anthropic rejects via the exact key pattern, and filters corresponding required entries so schemas remain internally consistent.\n\nStrengths:\n- Regression test reproduces the invalid property-name failure mode and covers nested properties plus required filtering.\n- Minimal runtime surface: no execution-path/schema migration changes, only model-facing schema normalization.\n- Existing Google schema sanitizer path is preserved and now benefits from the provider key filter before Google-specific cleanup.\n\nResidual risk:\n- If an upstream tool truly requires an invalid property name, the model will no longer see that argument. That is preferable to failing the whole chat request, but a future enhancement could add reversible aliasing. Not needed for this production fix.\n\nLocal evidence:\n- Red test failed before implementation.\n- Green: `deno task test src/agent/runtime/model-tool-converter.test.ts src/agent/runtime/provider-tool-compat.test.ts`\n- Typecheck: `deno task typecheck`\n- Format/lint on touched files passed.\n- Build: `deno task build`

PR #1690, opened 2026-05-16
Skip empty assistant history in provider requests

kwakayama:
## Summary
- skips assistant history messages that have no provider-sendable content before model requests
- prevents failed assistant/error-only history from becoming an empty Anthropic text block
- bumps the package to `0.1.538` so the fix publishes and downstream deploys consume it

## Why
After deploying the tool schema sanitizer, the original Anthropic property-key error stopped reproducing. Retrying the same failed conversation then exposed the next provider rejection: `messages: text content blocks must be non-empty`. The conversation history contains previous failed assistant messages with only `error` parts; the converter was serializing those as `{ type: "text", text: "" }`, which Anthropic rejects.

## Evidence
- Red: `deno task test src/agent/runtime/text-generation-runtime-message-converter.test.ts` failed before implementation with the error-only assistant still serialized as empty text.
- Green: `deno task test src/agent/runtime/text-generation-runtime-message-converter.test.ts` ✅
- Regression + adjacent tests: `deno task test src/agent/runtime/text-generation-runtime-message-converter.test.ts src/agent/runtime/provider-tool-compat.test.ts src/agent/runtime/model-tool-converter.test.ts` ✅
- `deno fmt --check deno.json src/utils/version-constant.ts src/agent/runtime/text-generation-runtime-message-converter.ts src/agent/runtime/text-generation-runtime-message-converter.test.ts` ✅
- `deno task typecheck` ✅
- `deno task build` ✅

## Staging observation
- Post-0.1.537 staging retest no longer produced the original `tools.12.custom.input_schema.properties` error for the new run.
- The new run failed with `messages: text content blocks must be non-empty`, which this PR addresses.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.537",
+  "version": "0.1.538",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "workspace": [
--- a/src/agent/runtime/text-generation-runtime-message-converter.test.ts
+++ b/src/agent/runtime/text-generation-runtime-message-converter.test.ts
@@ -304,6 +304,27 @@ describe("text-generation-runtime-message-converter", () => {
       assertEquals(convertToTextGenerationRuntimeMessages([]), []);
     });
 
+    it("omits assistant messages that have no provider-sendable content", () => {
+      const messages: Message[] = [
+        { id: "u1", role: "user", parts: [{ type: "text", text: "list my repos" }] },
+        {
+          id: "a1",
+          role: "assistant",
+          parts: [{
+            type: "error",
+            code: "agent-provider-error",
+            message: "veryfront-cloud request failed",
+          }],
+        } as unknown as Message,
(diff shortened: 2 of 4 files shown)


kwakayama:
Critical review score: 93/100

What I checked:
- The regression is targeted to the observed staging follow-up failure: a failed assistant message with only an `error` part is omitted from provider-bound history rather than serialized as empty text.
- The fix is minimal and isolated to the batch history converter; single-message conversion behavior remains unchanged for existing callers/tests.
- Tool-call assistant messages and non-empty assistant text messages are still sent to the provider.
- The version bump to `0.1.538` is required so CI publishes a new package and downstream runtime images can consume the fix.
- Local verification passed: focused red/green regression, adjacent runtime tests, fmt check, typecheck, and build.

Residual risk:
- The converter still allows empty user text if callers produce it; that was not the observed failure and broadening the change could alter attachment-only flows.
- Final confidence requires publishing `veryfront@0.1.538`, updating veryfront-agent to consume it, redeploying staging, and rerunning the chat flow.

Decision: >=90, acceptable to merge after CI is green.

PR #1703, opened 2026-05-16
Revert runtime integration tool skip

kwakayama:
## Summary
- revert the runtime-side skip for configured integration tools that were not discovered
- restore the simple invariant: configured tool names must resolve to an available tool definition
- keep missing tool references loud instead of hiding them at runtime

## Verification
- `deno task test src/agent/runtime/tool-helpers.test.ts`
- `deno check src/agent/runtime/tool-helpers.ts src/agent/runtime/tool-helpers.test.ts`
- `deno fmt --check src/agent/runtime/tool-helpers.ts src/agent/runtime/tool-helpers.test.ts`
- `deno lint src/agent/runtime/tool-helpers.ts src/agent/runtime/tool-helpers.test.ts`
- `deno task typecheck`

Companion API setup fix: veryfront/veryfront-api#2532

### DIFF
--- a/src/agent/runtime/tool-helpers.test.ts
+++ b/src/agent/runtime/tool-helpers.test.ts
@@ -340,6 +340,24 @@ describe("tool-helpers", () => {
       }
     });
 
+    it("fails loudly when an explicit integration tool has no discovered tools", async () => {
+      toolRegistry.clearAll();
+
+      try {
+        await assertRejects(
+          () =>
+            withMockRemoteIntegrationTools([], () =>
+              getAvailableTools({
+                "github__get_pr_diff": true,
+              })),
+          Error,
+          'Unknown tool reference: github__get_pr_diff. Tool names must exactly match tool({ id: "..." }). Available tools: (none)',
+        );
+      } finally {
+        toolRegistry.clearAll();
+      }
+    });
+
     it("only appends explicitly requested remote definitions for explicit tool maps", async () => {
       toolRegistry.clearAll();
 
--- a/src/agent/runtime/tool-helpers.test.ts
+++ b/src/agent/runtime/tool-helpers.test.ts
@@ -340,23 +340,6 @@ describe("tool-helpers", () => {
       }
     });
 
-    it("skips explicit integration tools whose integration has no discovered tools for this run", async () => {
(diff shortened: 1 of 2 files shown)


kwakayama (commented):
## Code Review — Score: 84/100

Clean revert. The commit message on "Protect fail-loud integration tool configuration" is one of the better ones I've seen — constraint, rejection reasoning, confidence, scope-risk, directive, test evidence, not-tested surface all documented. That's a high bar.

### What's good

- Reverting ~44 lines of defensive silent-skip logic is the right move. The behavior was working around a setup gap that the companion API fix (veryfront/veryfront-api#2532) now closes correctly.
- Test rename is accurate: "skips" → "fails loudly" matches the restored invariant.
- The error message assertion is specific enough to catch regressions: `'Unknown tool reference: github__get_pr_diff. Tool names must exactly match tool({ id: "..." }). Available tools: (none)'`

### Issues

**1. No migration path for existing broken projects**

Users who copied the PR Reviewer template before the API fix have agents that declare `github__get_pr_diff` without a matching `project_integration_config` row. Before this revert, those runs silently returned no GitHub tools. After this merges, they throw. That's a user-visible regression for anyone in that state.

The fix exists (the API companion seeds the config on copy), but it only applies to new copies going forward. Existing projects need a backfill. No runbook or one-off migration is mentioned. Worth addressing before or immediately after merge.

**2. Merge ordering is undocumented and unenforced**

This PR description says "Companion API setup fix: veryfront/veryfront-api#2532" but doesn't state: **merge the API fix first**. If this runtime revert lands first, the window between merges is a production regression. A single line — "This must not be merged before veryfront/veryfront-api#2532" — would make the constraint explicit.

### Minor

The test file has one net-added line (10 added, 9 deleted, net +1). The `assertRejects` call is slightly more verbose than the original `assertEquals` block — unavoidable given the behavior change, and the specificity of the error message assertion is worth it.

kwakayama:
Follow-up for the migration-path concern is in veryfront/veryfront-api#2537: it adds `0132_backfill_pr_reviewer_github_integration` so existing copied PR Reviewer projects receive the GitHub project integration config expected by this fail-loud runtime behavior.

PR #1750, opened 2026-05-18
Forward run context for integration tool calls

kwakayama:
## Summary
- Forward runtime tool execution context (`run_id`, `agent_id`, and existing `end_user_id`) when remote integration tools call the API.
- Preserve the existing string `endUserId` call signature for compatibility.
- Add `runId` to `ToolExecutionContext` so agent runtime context is explicit and typed.

## Test Plan
- [x] RED: targeted Deno tests failed before implementation because `run_id`/`agent_id` were not forwarded.
- [x] `deno fmt src/integrations/remote-tools.ts src/integrations/remote-tools.test.ts src/agent/runtime/tool-helpers.ts src/agent/runtime/tool-helpers.test.ts src/tool/types.ts`
- [x] `deno check src/tool/types.ts src/integrations/remote-tools.ts src/agent/runtime/tool-helpers.ts`
- [x] `deno lint src/tool/types.ts src/integrations/remote-tools.ts src/integrations/remote-tools.test.ts src/agent/runtime/tool-helpers.ts src/agent/runtime/tool-helpers.test.ts`
- [x] `deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all src/integrations/remote-tools.test.ts src/agent/runtime/tool-helpers.test.ts`

## Related
- Required companion to API PR: https://github.com/veryfront/veryfront-api/pull/2551
- Fixes runtime-forwarding half of https://github.com/veryfront/veryfront-studio/issues/3895

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.554",
+  "version": "0.1.555",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "workspace": [
--- a/src/agent/runtime/tool-helpers.test.ts
+++ b/src/agent/runtime/tool-helpers.test.ts
@@ -222,6 +222,52 @@ describe("tool-helpers", () => {
       );
     });
 
+    it("passes runtime run and agent context to remote integration tool execution", async () => {
+      const originalApiBaseUrl = Deno.env.get("VERYFRONT_API_URL");
+      const originalApiToken = Deno.env.get("VERYFRONT_API_TOKEN");
+      let requestBody: Record<string, unknown> | undefined;
+
+      try {
+        Deno.env.set("VERYFRONT_API_URL", "https://api.test");
+        Deno.env.set("VERYFRONT_API_TOKEN", "token");
+
+        const result = await withMockFetch(
+          async (input: string | URL | Request, init?: RequestInit) => {
+            const request = input instanceof Request ? input : new Request(input, init);
+            requestBody = await request.json();
+            return Response.json({ structuredContent: { ok: true } });
+          },
+          async () =>
+            await executeConfiguredTool(
+              "gmail__list_emails",
+              { maxResults: 10 },
+              undefined,
+              {
+                toolCallId: "tool-4",
+                runId: "run-123",
+                agentId: "agent-123",
+                endUserId: "user-123",
+              },
+              ["gmail__list_emails"],
+            ),
+        );
+
+        assertEquals(result, { structuredContent: { ok: true } });
+        assertEquals(requestBody, {
+          name: "gmail__list_emails",
+          arguments: { maxResults: 10 },
+          end_user_id: "user-123",
+          run_id: "run-123",
+          agent_id: "agent-123",
+        });
+      } finally {
+        if (originalApiBaseUrl === undefined) Deno.env.delete("VERYFRONT_API_URL");
+        else Deno.env.set("VERYFRONT_API_URL", originalApiBaseUrl);
+        if (originalApiToken === undefined) Deno.env.delete("VERYFRONT_API_TOKEN");
+        else Deno.env.set("VERYFRONT_API_TOKEN", originalApiToken);
+      }
+    });
+
     it("executes remote MCP tools from configured remote tool sources", async () => {
       const remoteSource = createRemoteMCPToolSource({
         id: "docs",
--- a/src/agent/runtime/tool-helpers.ts
+++ b/src/agent/runtime/tool-helpers.ts
@@ -269,11 +269,7 @@ export async function executeConfiguredTool(
     if (allowedRemoteToolNames && !allowedRemoteToolNames.includes(toolName)) {
       throw new Error(`Tool "${toolName}" is not allowed for this run`);
     }
-    return await executeRemoteIntegrationTool(
-      toolName,
-      input,
-      context?.endUserId as string | undefined,
-    );
+    return await executeRemoteIntegrationTool(toolName, input, context);
   }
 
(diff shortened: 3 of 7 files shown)


kwakayama (commented):
## Code Review — Score: 72/100

Focused, minimal change that achieves backward compatibility cleanly. Loses points for a `runId?: unknown` type that contradicts its own codebase convention and forces redundant `typeof` guards, plus zero assertions on what is purely a wire-format change.

---

### [P2] (confidence: 9/10) `runId?: unknown` is the wrong type — `src/integrations/remote-tools.ts`

```ts
type RemoteIntegrationToolExecutionContext = {
  endUserId?: string;
  runId?: unknown;     // ← should be string | undefined
  agentId?: string;
};
```

Every other field in this type is `string`. `ToolExecutionContext.runId` in `src/tool/types.ts:63` is correctly `string`. Typing it `unknown` lies about the data and forces every consumer to re-prove a fact the type should have guaranteed.

Concrete downstream cost in `callRemoteTool`:

```ts
run_id: typeof context?.runId === "string" ? context.runId : undefined,
agent_id: typeof context?.agentId === "string" ? context.agentId : undefined,   // ← agentId is already string
```

Note the asymmetry: `agentId` is typed `string` but still gets a `typeof` guard. The guards aren't consistently motivated by the `unknown` — they're applied inconsistently. With `runId: string`, both lines collapse to:

```ts
run_id: context?.runId,
agent_id: context?.agentId,
```

`JSON.stringify` already omits `undefined`, so the guards add no runtime safety. They do add the ability to silently swallow a wrong-typed `runId` at the serialization site instead of catching it at the type boundary — which is the opposite of what you want.

**Fix:** `runId?: string` in `RemoteIntegrationToolExecutionContext`. Delete the `typeof` guards in `callRemoteTool`. If a non-string value genuinely arrives, validate in `normalizeRemoteExecutionContext` once and return `undefined` for the field — don't push `unknown` past the front door.

---

### [Informational] `normalizeRemoteExecutionContext` validates nothing in the object path

The normalization function handles the string-vs-object discrimination but returns the object case verbatim:

```ts
return contextOrEndUserId;
```

This is the natural chokepoint to coerce/validate object fields (e.g. guard `runId` if it genuinely could be non-string). Right now it normalizes only the overload shape, not the field values. Fine for now, but worth completing if `unknown` is kept for any reason.

---

### Test Coverage Gap

This change's entire value is wire-format correctness — the one thing that is cheap to test and invisible to the type system. There are zero assertions on the outbound JSON body.

Minimum expected coverage:
- `normalizeRemoteExecutionContext`: string → `{ endUserId }`; object → passthrough; `undefined` → `undefined`
- `callRemoteTool` body construction (mock `fetch`): assert `end_user_id`, `run_id`, `agent_id` are present/omitted correctly
- **BC regression guard:** calling `executeRemoteIntegrationTool(name, args, "user-id")` produces a request body byte-identical to the old behavior — this is the entire backward-compat claim and it has no automated guard
- One end-to-end test confirming a populated `runId` actually reaches the request body through `executeConfiguredTool` → also catches whether any upstream caller actually sets `context.runId` (see below)

---

### Verify Before Merge

**Is `ToolExecutionContext.runId` ever actually populated by callers of `executeConfiguredTool`?**

`tool-helpers.ts:269` now passes `context` directly, but whether `context.runId` is ever non-undefined depends on code not in this diff. If nothing upstream sets `runId` on the `ToolExecutionContext`, this PR ships plumbing that always sends `run_id: undefined` — a wire-format no-op for the headline feature. Grep the runtime for where `ToolExecutionContext` is constructed and confirm `runId` is assigned.

**The full `executeRemoteIntegrationTool` body is truncated in the diff.** The BC guarantee hinges on a single wiring line that calls `normalizeRemoteExecutionContext(contextOrEndUserId)`. Confirm it's there.

---

### What's Done Well

- **Backward-compat strategy is the right call.** Union parameter + a single normalization function is cleaner than an overload stack or a new function name. Legacy `string` callers are genuinely unaffected.
- **Removing `context?.endUserId as string | undefined`** (`tool-helpers.ts:272` before) eliminates a `as` cast — a net improvement over the pre-PR code.
- **`ToolExecutionContext.runId?: string`** (`types.ts:63`) is correctly placed and documented, consistent with the adjacent `agentId`/`toolCallId` fields.
- **Wire change is purely additive** — low blast radius on the server contract.
- `deno fmt`, `deno check`, `deno lint` all passing.

---

### Recommendation: Request Changes

Blocking:
1. `runId?: string` in `RemoteIntegrationToolExecutionContext`, delete the redundant `typeof` guards
2. Add wire-format unit tests (especially the BC regression guard)

Verify before merge:
3. Confirm at least one upstream caller sets `ToolExecutionContext.runId` so the feature is not a no-op
4. Confirm the full `executeRemoteIntegrationTool` body wires through `normalizeRemoteExecutionContext`

kwakayama:
Addressed the applicable review findings in the latest push (`75dc35d`):

- `RemoteIntegrationToolExecutionContext.runId` is now `string`.
- Removed redundant `typeof` guards in request body construction.
- Existing tests already cover outbound JSON for object context, `executeConfiguredTool` propagation, and backwards-compatible string `endUserId` calls.
- Verified upstream context population: `src/agent/runtime/index.ts` constructs tool context with `agentId` and spreads stream context containing `runId`; `executeConfiguredTool` now forwards that context to `executeRemoteIntegrationTool`.

Verification rerun:
- `deno fmt` targeted files.
- `deno check` targeted files.
- `deno lint` targeted files.
- Targeted Deno tests: 2 files / 40 steps passing.

kwakayama:
Added patch version bump in deno.json: 0.1.552 → 0.1.553. Verified `deno task generate:manifests:check`.

PR #1805, opened 2026-05-21
fix(confluence): finish v2 migration in template — spaces, blogposts, partial updates

### LINKED ISSUE #1804: Confluence v2 migration: 4 bugs in cli/templates/integrations/confluence/files/lib/confluence-client.ts
Surfaced by an ultrareview pass on a docs PR (#1803) whose diff happened to include neighbouring context. The findings are real bugs in `cli/templates/integrations/confluence/files/lib/confluence-client.ts`, introduced by the v1→v2 migration in #1785 / #1787. Filing here so they don't get lost — out of scope for the PR that found them.

All four affect the generated `integration:confluence` template, so every project scaffolded with it inherits the defect.

---

## 1. `createPage` now depends on still-v1 `listSpaces` — defeats the migration goal

**File:** `cli/templates/integrations/confluence/files/lib/confluence-client.ts:129–131, 177`

`createPage` (post-migration) calls `getSpaceIdByKey(options.spaceKey)` before posting to `/wiki/api/v2/pages`. `getSpaceIdByKey` calls `listSpaces({ limit: 250 })`, which is still on v1 `/wiki/rest/api/space`. On the exact "newer instances" the migration comment targets (`v1 /wiki/rest/api/content is deprecated and returns 410`), v1 `/wiki/rest/api/space` is on the same deprecation wave and 410s too — so `createPage` fails at spaceKey resolution before ever hitting the migrated v2 page endpoint.

**Pre-PR behaviour:** `createPage` posted `{ space: { key } }` directly to v1 `/wiki/rest/api/content`. No `listSpaces` dependency.

**Fix:** Migrate `listSpaces` to `/wiki/api/v2/spaces`, or have `getSpaceIdByKey` hit `/wiki/api/v2/spaces?keys=<key>` directly. Either way, the v2 page-create flow must not depend on a v1 endpoint.

---

## 2. `getSpaceIdByKey` caps at 250 spaces without pagination — silent failure on enterprise sites

**File:** `cli/templates/integrations/confluence/files/lib/confluence-client.ts:129–134`

```ts
async function getSpaceIdByKey(spaceKey: string): Promise<string> {
  const spaces = await listSpaces({ limit: 250 });
  const space = spaces.find((s) => s.key === spaceKey);
  if (!space) {
    throw new Error(`Confluence space not found: ${spaceKey}`);
  }
  return space.id;
}
```

`250` is the v1 per-page maximum, not a global cap. The function never reads `_links.next` and never loops. Enterprise Confluence tenancies routinely exceed 250 spaces (one per team/project). Any `spaceKey` past the first page produces `Confluence space not found: <key>` — and the misleading error message implies the key is wrong, sending users to debug the wrong thing.

The `ConfluenceResponse<T>` interface already declares `_links.next`; no caller reads it.

**Fix:** Either paginate `listSpaces` via `_links.next`, or (preferred) replace enumeration with a direct lookup like `GET /wiki/api/v2/spaces?keys=<key>`. The direct path is O(1) requests and produces a clean 404-style error when the key really is wrong.

---

## 3. `updatePage` drops required fields on partial update (v2 has no PATCH)

**File:** `cli/templates/integrations/confluence/files/lib/confluence-client.ts:197–228` + `cli/templates/integrations/confluence/files/tools/update-page.ts`

`updatePage` builds the PUT body with `if (options.title) body.title = ...` and `if (options.content) body.body = ...` guards. But `PUT /wiki/api/v2/pages/{id}` is a full replace — Atlassian requires `id`, `status`, `title`, `body`, `version` on every call. Omitting `title` or `body` returns 400 (`title is required` / `body is required`).

Meanwhile `tools/update-page.ts` advertises both fields as optional and the description literally invites partial updates: `"leave empty to keep current title"`. The canonical LLM use case (`{ pageId, content: "..." }`, no title) breaks at runtime.

`update-page.ts` already fetches `currentPage` for the version number, so `currentPage.title` and `currentPage.body.storage.value` are in scope but unused.

**Fix:** Resolve fallbacks at the call site:

```ts
const updatedPage = await updatePage(pageId, {
  title: title ?? currentPage.title,
  content: storageContent ?? currentPage.body?.storage?.value,
  version: currentPage.version.number + 1,
  versionMessage,
});
```

and drop the `if (options.title)` / `if (options.content)` guards in `updatePage` so both fields are always sent.

---

## 4. v2 migration breaks blogposts — wrong endpoint in `createPage` and `getPage`

**File:** `cli/templates/integrations/confluence/files/lib/confluence-client.ts:162–192` + `cli/templates/integrations/confluence/files/tools/create-page.ts`

V2 split pages and blogposts onto **two distinct resources** (`/wiki/api/v2/pages` and `/wiki/api/v2/blogposts`). The migration kept v1's unified shape and hardcoded `/pages` in both `createPage` and `getPage`.

**Regression 4a — `createPage` silently downgrades `type='blogpost'` to a page.** The function still destructures `type?: 'page' | 'blogpost'` from `options`, but never references it; the endpoint is unconditionally `/wiki/api/v2/pages`. The tool schema in `tools/create-page.ts` still advertises the enum.

**Regression 4b — `getPage` 404s on blogpost IDs.** `searchContent` is still on v1 (`/wiki/rest/api/search` with no type filter), so it returns mixed page/blogpost results. The natural `search-content → get-page` flow 404s whenever the agent picks a blogpost result.

**Fix:** Either branch in the client (`POST /wiki/api/v2/blogposts` for `type='blogpost'`; in `getPage` try `/pages` first and fall back to `/blogposts` on 404), or drop blogpost support — remove `'blogpost'` from the `create-page` schema enum and from `ConfluencePage.type`.

---

## Severity

All four are **normal** per the ultrareview. None is a security issue. Combined impact: the generated Confluence integration template is unusable on any post-v1-deprecation Confluence Cloud tenancy (`#1` blocks `createPage`), partially broken on tenancies with >250 spaces (`#2`), unusable for the documented partial-update flow (`#3`), and silently wrong for blogpost workflows (`#4`).

## References

- Introduced by: #1785, #1787
- Found by: ultrareview pass during #1803

ariskemper:
## Summary

The Confluence v2 migration in #1785 / #1787 left four real bugs in the generated integration template (filed as #1804). Each is in `cli/templates/integrations/confluence/files/lib/confluence-client.ts` (or its caller in `tools/update-page.ts`). Fixes all four.

1. **`createPage` depended on still-v1 `listSpaces`.** `listSpaces` was still hitting `/wiki/rest/api/space`. On the same newer instances the migration comment targets (v1 `/content` 410s), v1 `/space` 410s too, so `createPage` failed at spaceKey resolution before ever reaching the migrated v2 page endpoint. Migrated `listSpaces` to `/wiki/api/v2/spaces`.

2. **`getSpaceIdByKey` silently capped at 250 spaces.** `listSpaces({ limit: 250 })` is per-page, not a global cap, and the helper never read `_links.next`. Any `spaceKey` past the first page produced `Confluence space not found: <key>` on enterprise tenancies with hundreds of spaces. Replaced enumeration with a direct v2 lookup: `GET /wiki/api/v2/spaces?keys=<key>&limit=1` — O(1) requests and a clean error path.

3. **`updatePage` dropped required fields on partial update.** `PUT /wiki/api/v2/pages/{id}` is a full replace (no PATCH equivalent in v2); `title` and `body` are required on every call. The previous body builder used `if (options.title)` / `if (options.content)` guards, so the documented partial-update flow (e.g. content-only edit) returned `400 title is required`. Tightened `updatePage`'s contract to require both fields, and resolved fallbacks in `tools/update-page.ts` from the page fetched for the version number (which was already in scope but unused).

4. **Blogposts routed to the wrong v2 endpoint.** v2 split pages and blogposts onto separate resources. `createPage` hardcoded `/pages` and ignored `options.type`, silently downgrading `type='blogpost'` requests to pages. `getPage` hardcoded `/pages` and 404'd on any blogpost id, breaking the natural `searchContent → getPage` flow (`searchContent` still returns mixed page/blogpost results). Branched `createPage` on `options.type` (POST to `/blogposts`, reject `parentId` since blogposts can't have parents), and `getPage` tries `/pages` first then falls back to `/blogposts` on 404. Surfaced 404 via a new `ConfluenceApiError` so the fallback can distinguish status codes.

Regenerated `cli/templates/manifest.json`.

## Test plan

- [x] `deno task lint` — clean
- [x] `deno task fmt` — no changes (already formatted)
- [x] `deno run -A scripts/build/generate-templates-manifest.ts` — manifest regenerated
- [x] Pre-push hook ran the full test suite — 1905 tests passed
- [ ] Manual: create a project from the `confluence` integration template and exercise `create-page`, `get-page`, `update-page`, `list-spaces`, `search-content` against a real Confluence Cloud instance (out of CI scope; recommend before merging)

Closes #1804


ariskemper:
Both Codex findings addressed in 015dd1449:

1. **P1 — `updatePage` blogpost fallback (valid).** Added the same pages-then-blogposts try/catch to `updatePage` so any blogpost ID (returned by `create-page` with `type='blogpost'` or fetched via `getPage`'s fallback) now updates correctly.
2. **P2 — empty title treated as unset (valid).** Added a `nonEmpty` helper that treats `""` and whitespace-only as missing. Combined with the existing `toStorageContent` empty-guard (also tightened to reject whitespace), the documented 'leave empty to keep current' flow now works for both fields.

@codex review

PR #1846, opened 2026-05-23
Auto-wire Veryfront platform tools for source agents

kwakayama:
## Summary
- auto-wire Veryfront platform MCP tools when source agents declare `search_knowledge` or `get_file` in `tools`
- preserve those platform tool booleans through internal run tool filtering so AgentRuntime can materialize the remote MCP source
- update regression coverage so support-agent-style source only specifies `tools`, not `allowedRemoteTools` or `remoteTools`

## Root cause
Internal agent runs filtered `tools: { search_knowledge: true }` before AgentRuntime could resolve remote MCP tools. Because `search_knowledge` and `get_file` are not local registry tools and were not API-forwarded definitions, they were dropped from the runtime tool map. The support-agent then saw only unrelated hosted tools (web_search/web_fetch) and correctly refused to answer from the knowledge base.

## Review score
- 93/100 — generic platform fix with focused regression coverage; remaining risk is deployment/release lag before the live API eval turns green.

## Verification
- deno task test src/internal-agents/run-stream.test.ts src/server/handlers/request/agent-stream.handler.test.ts
- deno fmt --check src/internal-agents/run-stream.ts src/internal-agents/run-stream.test.ts src/server/handlers/request/agent-stream.handler.ts src/server/handlers/request/agent-stream.handler.test.ts
- DENO_NO_PACKAGE_JSON=1 deno lint src/internal-agents/run-stream.ts src/internal-agents/run-stream.test.ts src/server/handlers/request/agent-stream.handler.ts src/server/handlers/request/agent-stream.handler.test.ts

## Tracking

- Tracks veryfront/veryfront-studio#4245

### DIFF
--- a/src/internal-agents/run-stream.test.ts
+++ b/src/internal-agents/run-stream.test.ts
@@ -73,4 +73,70 @@ describe("internal-agents/run-stream", () => {
 
     assertEquals(capturedToolNames, ["gmail__list_emails", "web_search"]);
   });
+
+  it("preserves explicitly allowed source remote tool declarations before constructing the runtime", async () => {
+    const sessionManager = new AgentRunSessionManager();
+    let capturedToolNames: string[] = [];
+
+    const agent = {
+      id: "support-agent",
+      config: {
+        id: "support-agent",
+        model: "anthropic/claude-opus-4-6",
+        system: "test",
+        tools: {
+          search_knowledge: true,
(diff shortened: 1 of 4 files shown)


kwakayama:
Tracks veryfront/veryfront-studio#4245 for template/forked agent support and platform-tool auto-wiring.

PR #1901, opened 2026-05-27
fix(discovery): externalize npm deps in tool/agent discovery bundler

ariskemper:
## Summary

The discovery transpiler bundles tool/agent/skill files with esbuild under `platform: "neutral"`. That platform has no Node built-ins, so a tool that imports an npm package referencing `fs` / `http` / `https` / `url` (e.g. `pdf-parse`, `mammoth`) blew up the bundle and prevented `deno task dev` from rendering any page in the project.

- Add `packages: "external"` to the discovery bundler — discovery only needs module exports, not a self-contained bundle.
- Generalize `rewriteForDeno` to `npm:`-prefix any remaining bare specifier (not relative, not `node:`, not absolute URL, not `veryfront`) so Deno can resolve from the temp dir the discovery module loads from.
- Generalize `rewriteDiscoveryImports` (Node path) to collect every bare specifier the bundled source still references and resolve each via the project's `node_modules`, instead of only a hard-coded `["zod"]` list.

## Repro

Inside any project, add a tool that imports a CJS npm package with Node built-in references:

```ts
// tools/parse-contract.ts
import pdfParse from "pdf-parse";
import mammoth from "mammoth";
// ...
```

`deno task dev` previously printed dozens of `✘ [ERROR] Could not resolve "fs" / "http" / ...` errors and every page returned 500. After this change the discovery bundle keeps `import("pdf-parse")` as `npm:pdf-parse` (Deno) or a file:// URL into `node_modules` (Node), and the dev server starts.

This was found while wiring up the DORA Compliance Agent example in `veryfront-examples`, which uses `pdf-parse` and `mammoth` for contract parsing.

## Test plan

- [x] `deno test src/discovery/import-rewriter.test.ts` — 5 cases pass, including two new cases:
  - "prefixes arbitrary bare npm imports with `npm:` for Deno temp module imports"
  - "leaves `node:`, `file:`, relative, and veryfront specifiers untouched"
- [x] `deno test src/discovery/` — full discovery suite green.
- [x] Pre-push full test suite — 1913 passed, 0 failed.
- [x] Manual verification with the DORA example: previously failing `deno task dev` no longer emits bundler errors for `pdf-parse` / `mammoth`.

### DIFF
--- a/src/discovery/import-rewriter.test.ts
+++ b/src/discovery/import-rewriter.test.ts
@@ -47,6 +47,42 @@ describe("discovery/import-rewriter", () => {
     assertEquals(transformed.includes('from "veryfront/'), false);
   });
 
+  it("prefixes arbitrary bare npm imports with npm: for Deno temp module imports", () => {
+    const transformed = rewriteForDeno(
+      [
+        'import pdfParse from "pdf-parse";',
+        'import mammoth from "mammoth";',
+        'const pdf = await import("pdf-parse");',
+        'import { z } from "zod";',
+      ].join("\n"),
+      "/project/tools",
+    );
+
+    assertStringIncludes(transformed, 'from "npm:pdf-parse"');
+    assertStringIncludes(transformed, 'from "npm:mammoth"');
+    assertStringIncludes(transformed, 'import("npm:pdf-parse")');
+    assertStringIncludes(transformed, 'from "npm:zod"');
+    assertEquals(transformed.includes('from "pdf-parse"'), false);
+    assertEquals(transformed.includes('from "mammoth"'), false);
+  });
+
+  it("leaves node:, file:, relative, and veryfront specifiers untouched", () => {
+    const transformed = rewriteForDeno(
+      [
+        'import fs from "node:fs";',
+        'import path from "node:path";',
+        'import local from "./helpers.ts";',
+        'import remote from "https://esm.sh/some-pkg";',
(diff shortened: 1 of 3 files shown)


ariskemper:
Thanks @chatgpt-codex-connector — both P2 findings were valid. Pushed `4040c1907`:

**1. Package subpath resolution (`react/jsx-runtime`, `lodash-es/debounce`, …)**
`resolvePackageToFileUrl` now splits the specifier into `name`/`subpath` (handling `@scope/pkg/sub/path`), reads `pkgJson.exports[subpath]`, and falls back to joining the subpath onto the package directory when the package has no `exports` entry for it. So `react/jsx-runtime` resolves via `react`'s `exports["./jsx-runtime"]`, and `dotenv/config` falls through to `node_modules/dotenv/config.js`.

**2. Side-effect bare imports (`import "reflect-metadata";`, `import "dotenv/config";`)**
Both `rewriteForDeno` (the `rewriteBareNpmImportsForDeno` helper) and the Node `rewriteDiscoveryImports` collector + `rewritePackageImports` now match the side-effect `import "spec";` form alongside `from "..."` and `import("...")`. Deno gets `npm:` prefixing, Node gets `file://` rewriting.

**Tests:** 9 cases now in `import-rewriter.test.ts`, including a mock `react` package for subpath + exports and a mock `dotenv` package for the side-effect path. All passing.

PR #1907, opened 2026-05-27
fix(ssr-vf-modules): make depth-limit fallback emit loadable code

ariskemper:
## Summary

Every page render returns 500 on a cold cache in 0.1.593 because the SSR vfmod transform writes a broken module to the framework cache. Found while wiring up the DORA compliance agent example ([veryfront-examples#10](https://github.com/veryfront/veryfront-examples/pull/10)).

## Root cause

`transformFrameworkCode` (`src/transforms/pipeline/stages/ssr-vf-modules/transform.ts`) recursively walks the framework source tree and rewrites each file's relative imports (e.g. `./build.js`, `./define.js`) to absolute `file://` URLs pointing at cached deps. Each recursion bumps `depth`. When `depth > MAX_RELATIVE_IMPORT_DEPTH = 10`, the function falls back to *"minimally transformed code"* — raw esbuild output with **no import rewriting** — and returns it. The caller then writes that fallback to `.cache/veryfront-mdx-esm/framework/vfmod-vf-framework-*.mjs`.

The resulting cached file looks like:

```js
// .cache/veryfront-mdx-esm/framework/vfmod-vf-framework-…-33d2a43d.mjs
import { DEFAULT_BUILD_CONCURRENCY, IMAGE_OPTIMIZATION } from "./build.js";
import { BUFFER_SIZE_… } from "./buffers.js";
```

…but `./build.js` doesn't exist in the cache directory, so Node/Deno report `Module not found "…/.cache/veryfront-mdx-esm/framework/build.js"` and SSR fails. The veryfront source tree in 0.1.593 has multiple chains that exceed depth 10 (`src/errors/types.js`, `src/utils/constants/index.js`, `src/schemas/primitives.js` all hit depth 11), so every project on 0.1.593 with a cold cache 500s on the first request.

The existing comment in the fallback path is candid: *"Return minimally transformed code - it will fail at runtime but won't hang"*. So this was a known time-bomb; the source tree just grew past the limit.

## Fix

In the depth-limit fallback:

1. Compile the file with esbuild as before.
2. Then walk its relative imports with `findRelativeImports`.
3. Resolve each against `sourcePath` via `resolveRelativeFrameworkImport`.
4. Rewrite the import to `file://${resolvedPath}` via `replaceSpecifiers`.

The resolved sources are already-published `.js` files with only sibling relative imports, which the runtime resolves natively from `node_modules` — so the fallback module becomes loadable from the cache directory without further recursive transformation. Imports that fail to resolve are left untouched (and a warning is logged), matching the previous behavior for that specific case.

## Verification

Patched the published `transform.js` in a sample project's `node_modules/.deno/veryfront@0.1.593/.../transform.js` with the same fix and reran `deno task dev`:

| Before | After |
|---|---|
| `Module not found "…/framework/build.js"` | gone |
| `Module not found "…/framework/define.js"` | gone |
| 6 render errors | 0 render errors related to framework cache |

A separate `"transform capacity"` error then surfaces while loading `react@19.2.4.js` from esm.sh, but it's likely an artifact of the parallel pdf-parse / mammoth bundler failures saturating the transform queue — that path is fixed in #1901.

## Test plan

- [x] `deno test src/transforms/pipeline/stages/ssr-vf-modules/transform.test.ts` — 2 new cases pass:
  - "rewrites relative imports in the fallback to absolute file:// URLs so the cached output is loadable"
  - "leaves bare-package imports alone in the fallback output"
- [x] `deno check` on the modified file.
- [x] Pre-push full suite — green.
- [x] Manual: live `deno task dev` against the DORA example shows the cached `vfmod-vf-framework-*.mjs` files now contain absolute `file://` URLs in place of `./build.js` / `./define.js`, and the framework-not-found errors disappear.

## Related

- [veryfront-code#1901](https://github.com/veryfront/veryfront-code/pull/1901) — fixes the discovery bundler crashing on Node-only npm deps in tools. Both fixes are needed for the DORA example (and any non-trivial project on 0.1.593) to render end-to-end.

### DIFF
--- a/src/transforms/pipeline/stages/ssr-vf-modules/transform.test.ts
+++ b/src/transforms/pipeline/stages/ssr-vf-modules/transform.test.ts
@@ -171,6 +171,89 @@ describe("transformFrameworkCode depth-limit fallback", {
     }
   });
 
+  it("does not propagate a cycle placeholder from frameworkFileCache into the fallback cache file", async () => {
+    const { frameworkFileCache } = await import("./constants.ts");
+    const tmp = await Deno.makeTempDir({ prefix: "vf-vfmod-cycle-" });
+    const srcDir = `${tmp}/src`;
+    await Deno.mkdir(srcDir, { recursive: true });
+    const depPath = `${srcDir}/dep.ts.src`;
+    // Real content for the dep on disk
+    await Deno.writeTextFile(depPath, "export const X = 42;\n");
+    // Simulate the main path having pre-cached a cycle placeholder for this dep
+    const placeholder = `/* Cycle detected: ${depPath} */\nexport {};`;
+    frameworkFileCache.set(depPath, placeholder);
+
+    const ownerPath = `${srcDir}/owner.ts.src`;
+    const ownerContent = 'import { X } from "./dep.js"; export const y = X;';
+    await Deno.writeTextFile(ownerPath, ownerContent);
+
+    try {
+      const transformed = await transformFrameworkCode(
+        ownerContent,
+        ownerPath,
+        { reactVersion: "19.1.1", projectDir: tmp, fs: createFileSystem() },
+        false,
+        MAX_RELATIVE_IMPORT_DEPTH + 1,
+      );
+
+      // The emitted cache file URL must not link to a file whose contents
+      // are the cycle-placeholder. Read it back and check for the real
+      // dep's content (`export const X = 42`).
+      const match = transformed.match(/from "file:\/\/([^"]+\.mjs)"/);
+      assert(match, `fallback did not emit a .mjs URL: ${transformed}`);
+      const cachePath = match[1]!;
+      const written = await Deno.readTextFile(cachePath);
+      assertEquals(
+        written.includes("Cycle detected"),
+        false,
+        `cycle placeholder leaked into fallback cache file: ${written}`,
+      );
+      assertStringIncludes(written, "X = 42");
+    } finally {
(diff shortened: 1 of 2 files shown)


ariskemper:
Thanks @chatgpt-codex-connector — P1 was valid. Fixed in `bfa2a4e88`:

In compiled binaries `resolveRelativeFrameworkImport` returns `*.ts.src` / `*.js.src` paths (embedded extensions are searched ahead of plain ones), and the previous fallback emitted those directly — Node/Deno cannot import a `.src` file, so the fix only worked in dev mode.

Now:

- `compileFallbackSource()` factors out the esbuild compile step and strips the `.src` suffix during loader detection (`foo.ts.src` → `ts` loader instead of falling through to `tsx`).
- `transformAndCacheFallbackDep()` reads the `.src` content, compiles, recursively rewrites its own relative imports via the same fallback helper, caches the result via `cacheTransformedCode`, and returns the cache URL.
- `rewriteFallbackRelativeImports()` special-cases `.src` paths (transform-and-cache) vs regular `.ts` / `.js` paths (point directly). A `visited` set guards cycles in the embedded source graph.

Dev-mode behavior is unchanged — `.js` / `.ts` deps still link directly with no extra IO.

**Test:** new third case in `transform.test.ts` builds an embedded `.ts.src` graph, runs the fallback, and asserts the emitted URL points at a `.mjs` cache file that actually exists on disk (rather than the `.src` source). 3 cases now pass; full pre-push suite still green (1914 / 0).

PR #1957, opened 2026-05-31
Let local project tools execute after stream commits

kwakayama:
## Summary
- stop reading a model stream after a non-provider-executed local tool call has committed and the stream stalls
- keep provider-executed tool calls/results streaming normally
- add a regression for a committed local `number-generator` tool call whose provider stream never closes

## Evidence
- `deno test --no-check --allow-all src/agent/runtime/chat-stream-handler.test.ts src/internal-agents/run-stream.test.ts src/agent/ag-ui/runtime-handler.test.ts`
- pre-commit `deno task verify:quick` passed

## Staging context
- API PR veryfront/veryfront-api#3188 proved the durable executor no longer parks `number-generator` as a browser-resumed tool.
- Post-API staging proof run `86dc57c4-c3ba-40ba-9602-e6b0be65c1c3` showed `TOOL_CALL_START`/`TOOL_CALL_ARGS`/`TOOL_CALL_END` for `number-generator` but no `TOOL_CALL_RESULT`; the run stayed `running` with `waiting_tool_name=null`. This change addresses that runtime stream boundary.

### DIFF
--- a/src/agent/runtime/chat-stream-handler.test.ts
+++ b/src/agent/runtime/chat-stream-handler.test.ts
@@ -69,12 +69,14 @@ describe("chat-stream-handler", () => {
           toolCallId: "tc-form",
           toolName: "form_input",
           input: { title: "Need more detail" },
+          providerExecuted: true,
         },
         {
           type: "tool-result",
           toolCallId: "tc-form",
           toolName: "form_input",
           output: { submitted: false },
+          providerExecuted: true,
         },
         { type: "text-delta", text: " After tool." },
         { type: "finish", finishReason: "stop", totalUsage: null },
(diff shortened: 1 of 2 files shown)


kwakayama:
Evidence update (sanitized):

This timeout/cancellation fix was part of the tool-call recovery chain. The final staging proof now passes after the later runtime and JSON Schema discovery follow-ups.

- Final release: `v0.1.625`, deployed by server run `26724130835`.
- Staging server/proxy report `VERYFRONT_VERSION=v0.1.625`.
- Proof run `proof-1780260874148` completed with no terminal error.
- Event chain included `TOOL_CALL_RESULT` and `RUN_FINISHED`.
- Stored tool result: `{"randomNumber":9}` for requested range 3..9.

No tokens, cookies, user emails, or request auth headers are included here.

PR #1962, opened 2026-05-31
Do not wait on stream cancel before local tool execution

kwakayama:
## Summary
- make provider stream cancellation best-effort after a committed local tool-call handoff
- keep `finishReason=tool-calls` and return from `processStream` so the existing runtime loop can execute local tools
- add a regression where `streamIterator.return()` never resolves

## Evidence
- `deno fmt --check src/agent/runtime/chat-stream-handler.ts src/agent/runtime/chat-stream-handler.test.ts`
- `deno test --no-check --allow-all src/agent/runtime/chat-stream-handler.test.ts src/agent/runtime/index.test.ts src/internal-agents/run-stream.test.ts src/agent/ag-ui/runtime-handler.test.ts`
- pre-commit `verify:quick` passed

## Live failure this fixes
- v0.1.622 staging proof `proof-1780247880175` on conversation `76010251-7b5b-4f8c-a5a9-e1d62e500b67` still reached `TOOL_CALL_END` for `number-generator` but never emitted `TOOL_CALL_RESULT`.
- Server logs confirmed runtime version `0.1.622` and `number-generator` was in the runtime tool set.
- Follow-up root cause: after the committed local tool-call timeout, `processStream` awaited provider stream cancellation via `streamIterator.return()`; live cancellation did not resolve, so control never reached runtime tool execution.\n

### DIFF
--- a/src/agent/runtime/chat-stream-handler.test.ts
+++ b/src/agent/runtime/chat-stream-handler.test.ts
@@ -181,6 +181,57 @@ describe("chat-stream-handler", () => {
       ]);
     });
 
+    it("does not wait for provider stream cancellation after a committed local tool-call", async () => {
+      const { controller, encoder } = createSSECollector();
+      const state = createStreamState();
+      let returnCalled = false;
+      let index = 0;
+      const parts = [
+        { type: "tool-input-start", id: "tc-local", toolName: "number-generator" },
+        { type: "tool-input-delta", id: "tc-local", delta: '{"min":3,"max":9}' },
+        {
+          type: "tool-call",
+          toolCallId: "tc-local",
+          toolName: "number-generator",
+          input: { min: 3, max: 9 },
+        },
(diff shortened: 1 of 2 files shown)


kwakayama:
Evidence update (sanitized):

This non-blocking iterator cleanup was part of the tool-call recovery chain. The final staging proof now passes after the later runtime and JSON Schema discovery follow-ups.

- Final release: `v0.1.625`, deployed by server run `26724130835`.
- Staging server/proxy report `VERYFRONT_VERSION=v0.1.625`.
- Proof run `proof-1780260874148` completed with no terminal error.
- Event chain included `TOOL_CALL_RESULT` and `RUN_FINISHED`.
- Stored tool result: `{"randomNumber":9}` for requested range 3..9.

No tokens, cookies, user emails, or request auth headers are included here.

Issue #1987, opened 2026-06-01
[Tech debt] Rendering & UI — rendering, react, html, cache


ariskemper:
# Tech Debt: Rendering & UI (rendering, react, html, cache)

## Verification

Adversarially re-verified against the source on 2026-06-01. **Tally: 17 verified, 3 accurate-but-qualified, 1 refuted, plus several count/path corrections.**

- **Refuted (1):** the mermaid `dangerouslySetInnerHTML` "potential XSS" finding — both mitigations the report asked to "confirm" (mermaid `securityLevel: 'strict'` AND `validateTrustedHtml(svg, { strict: true })` before injection) are already present. Not exploitable; moved to _Refuted / false positives_.
- **Qualified (3):** `transformModuleWithDeps` range is 185–~486 not 185–504 (504 is a different fn); the cookie-defeats-caching finding is intentional safety; the legacy CSS-cache path is deliberate back-compat.
- **Verified key claims:** the `cache -> html` layering inversion (`distributed-cache-init.ts:7`) and the unbounded `importMapJsonCache` (`html/utils.ts:46`) are both real and confirmed exactly.
- **Count corrections:** Summary table row 8 says "5 `../../../../` imports" — actual count is **4** (finding body is correct). The "Per-subdir (non-test)" file counts (202/111/74/53) are actually **total ts/tsx incl. tests**; non-test counts are 126/97/43/35 (LOC and test-file counts are correct). The "zero markers" claim is correct (the 2 apparent hits in `html.ts` are the substring "ToDo" inside `…ThemeToDocument`). Nit paths for `snippet-renderer.ts` corrected (it lives in `src/rendering/`, not `rsc/server-renderer/`).

## Summary

[...]

PR #2044, opened 2026-06-02
refactor(cache): invert distributed-cache-init layering via injected initializers (#1987)

ariskemper:
## What

Fixes the layering inversion flagged in epic **#1987** (Report 05): `src/cache/distributed-cache-init.ts` reached up into the rendering / transform / modules / platform layers to build its `defaultInitializers` — most notably a static import of `#veryfront/html/styles-builder/tailwind-compiler.ts`. `src/cache` is meant to be a low-level layer.

## How

The module already had `DistributedCacheInitializers` dependency-injection scaffolding (the test-only entrypoint used it). This change makes the production path use it too:

- `initializeDistributedCaches(initializers)` now takes the initializers as an **injected argument**. `distributed-cache-init.ts` no longer imports `html` / `transforms` / `modules` / `platform` at all — only `observability`, `utils`, and its sibling `backend.ts`.
- The default wiring moves up to a new **`src/server/distributed-cache-initializers.ts`** (the composition root, which is free to depend on every layer).
- Both callers — `production-server.ts` and `dev-server/server.ts` — pass `defaultDistributedCacheInitializers`.

Net effect: the `src/cache → src/html` (and `→ transforms`/`modules`/`platform`) edges are gone; the dependency now points the correct way (`server → cache`, `server → html`).

## Why this is low-risk

- **Behavior is identical** — the same five initializers run in the same order via the same `Promise.allSettled` path. Only *where the functions are wired* changed.
- The lint `dependency-boundaries` already passed before/after (this edge wasn't lint-enforced); this is an architectural cleanup, not a build-breaking fix.

## Tests

- Added a **source-level regression test** in `distributed-cache-init.test.ts` asserting the file imports none of `#veryfront/{html,transforms,modules,platform}/…`, so the inversion can't silently regress.
- `deno test src/cache/distributed-cache-init.test.ts` → 3 passed.
- `deno lint`, `deno fmt --check`, `deno task lint:dependency-boundaries` → clean. Typecheck of the changed files surfaces only the 3 pre-existing `react`/`platform/compat/process/lifecycle.ts` errors present on `main`.

Refs #1987. Draft pending `/deep-review`.

### DIFF
--- a/cli/commands/serve/command.ts
+++ b/cli/commands/serve/command.ts
@@ -59,9 +59,12 @@ async function runProductionServer(options: ServeOptions): Promise<void> {
   const { initializeDistributedCaches } = await import(
     "veryfront/cache"
   );
+  const { defaultDistributedCacheInitializers } = await import(
+    "veryfront/server"
+  );
   await Promise.allSettled([
     initializeOTLPWithApis(),
-    initializeDistributedCaches(),
+    initializeDistributedCaches(defaultDistributedCacheInitializers),
   ]);
 
   const projectDir = cwd();
--- a/src/cache/distributed-cache-init.test.ts
+++ b/src/cache/distributed-cache-init.test.ts
@@ -2,6 +2,27 @@ import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals } from "#std/assert";
 import { __runDistributedCacheInitializationForTests } from "./distributed-cache-init.ts";
 
+Deno.test("distributed-cache-init does not import higher layers (rendering/transform/platform)", async () => {
+  // Layering guard (#1987): src/cache is a low-level layer and must not reach
+  // up into html / transforms / modules / platform. The concrete initializers
+  // are injected from the server composition root instead. If this fails, a
+  // cross-layer import was re-introduced into distributed-cache-init.ts.
+  const source = await Deno.readTextFile(new URL("./distributed-cache-init.ts", import.meta.url));
+  const forbidden = [
+    "#veryfront/html/",
+    "#veryfront/transforms/",
+    "#veryfront/modules/",
+    "#veryfront/platform/",
+  ];
+  for (const specifier of forbidden) {
+    assertEquals(
+      source.includes(specifier),
+      false,
+      `distributed-cache-init.ts must not import ${specifier} (inject initializers from the server layer instead)`,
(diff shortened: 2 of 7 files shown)


ariskemper:
## Review findings

### 🐞 Bug: third caller of `initializeDistributedCaches` not updated

The PR description states "Both callers — `production-server.ts` and `dev-server/server.ts` — pass `defaultDistributedCacheInitializers`", but there is a **third** caller that was missed:

- `cli/commands/serve/command.ts:64` — still calls `initializeDistributedCaches()` with **zero arguments**.

After this PR, `initializeDistributedCaches` requires an `initializers: DistributedCacheInitializers` parameter (`src/cache/distributed-cache-init.ts:121-123`). The CLI call therefore:

1. **Fails to typecheck** — missing required argument.
2. **Crashes at runtime** when backend resolves to `api` or `redis`. The `memory` branch short-circuits before touching `initializers`, but the `withSpan(..., () => initializeDistributedCachesWithInitializers(backend, initializers), ...)` path (`src/cache/distributed-cache-init.ts:138`) forwards `undefined` into `initializeDistributedCachesWithInitializers`, which then calls `initializers.transformCache()` / `.ssrModuleCache()` / etc. on `undefined` → `TypeError`.

`runProductionServer` is the real CLI entrypoint dispatched at `cli/commands/serve/command.ts:118`, so this is on the live `veryfront serve` path, not a dead branch.

Suggested fix: mirror the production-server/dev-server pattern — dynamic-import `defaultDistributedCacheInitializers` from `#veryfront/server/distributed-cache-initializers.ts` (or expose it via `#veryfront/cache` if you want to keep this caller from depending on the `server` layer) and pass it in:

```ts
const { initializeDistributedCaches } = await import("veryfront/cache");
const { defaultDistributedCacheInitializers } = await import(
  "veryfront/server/distributed-cache-initializers"
);
await Promise.allSettled([
  initializeOTLPWithApis(),
  initializeDistributedCaches(defaultDistributedCacheInitializers),
]);
```

Worth also extending the `deno task lint:dependency-boundaries` / typecheck job (or adding the CLI to the typecheck set) so the next layering refactor catches missed callers automatically — the source-level test in `distributed-cache-init.test.ts` only guards the import direction, not the call-site signature.

Issue #1989, opened 2026-06-01
[Tech debt] Cross-cutting infra — utils, security, errors, observability…


ariskemper:
# Tech Debt: Cross-Cutting Infra (utils, security, errors, observability, oauth, data, repositories, issues, channels, types, client, schemas)

## Verification

Independently verified by adversarial re-audit (2026-06-01) against the actual source.

**Tally:** 14 substantive findings + 5 nits checked. **All findings verified accurate; 0 refuted.** Several counts corrected inline (weak-typing count, security subdir file/LOC). The three security-sensitive claims all hold:

- **`worker-script.ts` has no unit test — VERIFIED.** The only `worker-script.test.ts` in the repo is `src/agent/react/use-chat/browser-inference/worker-script.test.ts`, a different file. The security sandbox `worker-script.ts` is referenced only via `import.meta.resolve("./worker-script.ts")` in `project-worker.ts:350`; no test imports its internals (it exports nothing). Sibling tests (`deno-sandbox.test.ts`, etc.) do not reference `serializeError`/`loadModule`/`applyProjectEnv`/`handleAppRoute`.
- **`MemoryTokenStore.tokens` unbounded vs TTL-cleaned `states` — VERIFIED.** `memory.ts:14` `tokens` map is only mutated via `set`/`delete`/`clear` — no eviction; `states` (line 15) calls `cleanupExpiredStates()` on write and prunes on read. Exported singleton `memoryTokenStore` confirmed at line 92.
- **No secret redaction in logger — VERIFIED.** `rg -ni "redact|scrub|sanitize|mask|REDACTED"` over `utils/logger` + `observability` matches only comment-banner asterisks; no scrubbing layer exists. `errors/logging.ts:73` `console.error(JSON.stringify(entry))` confirmed.

## Summary

[...]

PR #2045, opened 2026-06-02
feat(logger): redact credential-like keys from log context (#1989)

ariskemper:
## What

Adds a key-based **redaction layer** for structured log context, closing the `#1989` finding that the logger had **no secret/PII scrubbing** — an accidental `logger.info("...", { token })` could leak credentials to log sinks.

## How

- New `src/utils/logger/redact.ts` — `redactSensitive(context)` returns a redacted deep copy: any property whose key names a credential (`password`, `token`, `secret`, `authorization`, `apikey`, `cookie`, `bearer`, …) is replaced with `[REDACTED]`. Key matching is normalized (`apiKey` / `api_key` / `API-Key` all match) and errs toward over-redaction.
- Wired into the two paths that serialize a free-form context bag:
  - `logger.ts` `formatJson` / `formatTextLine` — applied **after** the deliberate top-level Grafana fields (`requestId`, `userId`, …) are extracted, so those are unaffected.
  - `errors/logging.ts` production JSON path (`console.error(JSON.stringify(entry))`).
- Non-mutating; traverses nested objects/arrays; depth-capped and cycle-safe; leaves non-plain values (Error/Date/class instances) untouched.

## Tests

- `redact.test.ts` — key matching across conventions, nested/array redaction, no-mutation, non-plain passthrough, cycle safety, documented over-redaction edge.
- `logger.test.ts` — integration: a secret in context is `[REDACTED]` in JSON output and the raw value appears nowhere in the line.

Refs #1989

### DIFF
--- a/src/errors/logging.test.ts
+++ b/src/errors/logging.test.ts
@@ -75,6 +75,17 @@ describe("logging", () => {
         assertStringIncludes(output, "Component render failed");
       });
 
+      it("redacts credential-like context keys in the dev dump (#1989)", () => {
+        const error = RENDER_ERROR.create();
+
+        logError(error, { userId: "u-1", apiKey: "sk-secret" });
+
+        const output = consoleErrorOutput.join("\n");
+        assertStringIncludes(output, "[REDACTED]");
+        assertStringIncludes(output, "u-1");
+        assertEquals(output.includes("sk-secret"), false);
+      });
+
       it("should use error.context when no context provided", () => {
         const error = CONFIG_NOT_FOUND.create({
           context: { originalContext: true },
@@ -120,6 +131,17 @@ describe("logging", () => {
 
         const parsed = JSON.parse(consoleErrorOutput[0]);
         assertEquals(parsed.context.componentPath, "/app/page.tsx");
+      });
+
+      it("redacts credential-like context keys in JSON output (#1989)", () => {
+        const error = RENDER_ERROR.create();
+
+        logError(error, { userId: "u-1", token: "sk-secret" });
+
+        const parsed = JSON.parse(consoleErrorOutput[0]);
+        assertEquals(parsed.context.token, "[REDACTED]");
+        assertEquals(parsed.context.userId, "u-1");
+        assertEquals(consoleErrorOutput[0].includes("sk-secret"), false);
       });
 
       it("should merge error context with extra context and prefer extra values", () => {
--- a/src/errors/logging.ts
+++ b/src/errors/logging.ts
@@ -7,6 +7,7 @@
 
 import { isProduction } from "#veryfront/platform/environment.ts";
 import { serverLogger } from "#veryfront/utils/logger/logger.ts";
+import { redactSensitive } from "#veryfront/utils/logger/redact.ts";
 import { VeryfrontError } from "./types.ts";
 
 export interface ErrorLogEntry {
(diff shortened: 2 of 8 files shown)


ariskemper:
Solid feature direction, but the redaction has a few bypass paths that meaningfully undercut the threat model. Two are high-confidence enough to call out.

### 1. `Error.message` / `Error.stack` skip the redactor (HIGH)
`src/utils/logger/logger.ts` (`formatJson` path, ~line 283-352) and `src/utils/logger/core.ts:133-139` (`serializeError`):

When an `Error` is passed as a logger arg, `extractContext` peels it off and `serializeError` packs `{ name, message, stack }` into `entry.error`. That field is emitted directly — it never goes through `redactSensitive`.

This is the highest-leak path in practice. Real-world examples that routinely land in error stacks and messages:

- `pg`/`mysql` driver errors echoing the failing DSN (`postgres://user:password@host/db`)
- `mongoose` connection failures with the full Mongo URI
- Axios/fetch errors including the offending request URL with `?access_token=…` / `?code=…`
- JWT decode failures whose `.message` includes the token being decoded
- Any `throw new Error(\`bad token: \${secret}\`)` from app code

`logger.error("auth failed", err)` is exactly the call shape that triggers this — arguably the call this PR is most needed for, and the call it currently doesn't cover.

The same applies to the text-mode path: `formatContextText(redactSensitive(mergedContext), error, …)` at `src/utils/logger/logger.ts:368` redacts the context bag but forwards the raw `SerializedError`.

**Suggested fix:** apply `redactSensitive` to the assembled `entry` *after* `serializeError` runs (a final scrub at the emission boundary), or scrub `error.message` / `error.stack` as strings against a credential-pattern regex before they go out. Either prevents the bypass without you having to remember it at every call site.

### 2. Top-level extracted fields are exempt from redaction (MEDIUM)
`src/utils/logger/logger.ts:297-346`: `extractToEntryField` lifts known Grafana fields (`request_url`, `task`, `user_id`, …) to the top of the entry and `delete`s them from `mergedContext` **before** `redactSensitive` runs. The PR description explicitly calls this out as deliberate, but `request_url` in particular is a known credential carrier:

- `?access_token=…`, `?api_key=…`, `?code=…`, `?state=…`
- `https://user:pass@host/...` userinfo
- OAuth callback URLs

These won't be caught by *any* key-based scrubber (the secret is in the URL value, not a key). At minimum, the lifted `request_url` (and any other URL-shaped top-level field) should be passed through a URL-credential sanitizer that strips userinfo and known sensitive query params.

---

Lower-priority observations I'm not asking you to act on, just flagging:

- Deny list is missing some common-collision candidates: `signature`, `sessionid`/`sid`, `otp`/`mfa`/`pin`, `salt`, `clientid`/`clientsecret`, `xsrf`/`csrf`. `set-cookie`/`x-api-key` are already covered via the normalization → ✅.
- Substring matching can produce mild false positives (`apikeyword` normalizes to contain `apikey`, `totalCookies` → contains `cookie`). Documented trade-off; flagging only because the deny list will grow.
- Free-form message strings (`logger.info(\`Fetched \${url}\`)`) are intentionally out of scope per the doc — flagging here only because, combined with #1 and #2, it's the third major residual leak vector.

(1) is the load-bearing one — fixing it unblocks the feature's stated promise. (2) is a smaller scope but should be addressed in the same change since it's structurally identical (a path that skips the scrubber).

---

PR #2049, opened 2026-06-02
test: cover api-shim and cache-repository (#1989)

ariskemper:
## What

Closes the two "missing tests on complex code" gaps from the `#1989` audit — both modules were flagged as having no adjacent test.

## Tests added

**`src/observability/tracing/api-shim.test.ts`** (23 steps) — the no-op OpenTelemetry API shim (409 LOC):
- no-op tracer/span inertness (chainable mutators, all-zero span context)
- `startActiveSpan` overload forms (fn-as-2nd-arg, options+fn)
- global tracer-provider swap + `_resetShimForTests`
- active-span accessor wiring (`trace.getActiveSpan`/`getSpan`)
- `context.with` save/restore, including restore-on-throw
- propagation with and without a registered propagator
- default TextMap getter/setter, metrics-API registry, `SpanKind`/`SpanStatusCode` constants

**`src/repositories/cache/cache-repository.test.ts`** (20 steps) — the previously-untested `MultiTierCacheRepository` (L1 memory + L3 backend) plus `MemoryCacheRepository` edge paths:
- MultiTier: L1+L3 round-trip, `has`, `delete` (+ local stat), `deleteByPrefix` (incl. a backend with no `delByPattern` → returns 0), scope-limited `clear`, multi-tier stat accounting
- Memory: `get` TTL expiry + prune, LRU eviction at capacity, no-evict-on-update, `hitRate`, `clear`
- both factory functions

Tests assert **observable contracts only** and are deterministic (verified across repeated runs).

## Two behaviors surfaced while writing tests (noted, not changed here)

1. `MultiTierCacheRepository.deleteByPrefix`/`clear` only invalidate the **L3 backend** — `MultiTierCache` exposes no prefix/clear that reaches the private L1 tier, so `repo.get()` can serve a stale L1 value until TTL.
2. `set()` writes all tiers **fire-and-forget** under `asyncBackfill: true`, so the L3 write is not awaited.

Both are pre-existing design choices; flagging as potential follow-ups rather than touching behavior in a test-only PR.

Refs #1989

### DIFF
--- a/src/observability/tracing/api-shim.test.ts
+++ b/src/observability/tracing/api-shim.test.ts
@@ -3,6 +3,7 @@ import { assertEquals, assertExists } from "#veryfront/testing/assert.ts";
 import { afterEach, describe, it } from "#veryfront/testing/bdd.ts";
 import {
   _resetShimForTests,
+  type Context,
   context,
   defaultTextMapGetter,
   defaultTextMapSetter,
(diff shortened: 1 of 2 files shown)


ariskemper:
Test-quality review — one high-confidence finding.

**`src/observability/tracing/api-shim.test.ts` — "context API" describe block (lines ~90-118): assertions are tautological and don't verify the documented contract.**

Both `it` cases set up the scoped context with:

```ts
const base = context.active();              // NOOP_CONTEXT
const scoped = base.setValue(Symbol("k"), "v");
```

But the no-op `Context.setValue` in `api-shim.ts` (lines 181-183) returns `this`, so `scoped === base === NOOP_CONTEXT` — they are literally the same reference.

Consequences:

1. *"context.with sets the active context for the duration of fn then restores it"* — `assertEquals(inner, scoped)` and `assertEquals(context.active(), base)` are both `assertEquals(NOOP_CONTEXT, NOOP_CONTEXT)`. They would pass even if `context.with` never swapped `_activeContext` at all (e.g., if the `try { … } finally { _activeContext = prev; }` block were removed and replaced with `return fn()`).

2. *"context.with restores the previous context even if fn throws"* — the only assertion that bites is `threw === true`, which just verifies the throw propagates. The "restore" claim (`assertEquals(context.active(), base)`) is again `NOOP === NOOP` and can't fail. If the `finally` clause were deleted, this test would still pass.

To actually exercise the swap-and-restore contract, the test needs a context distinguishable from `NOOP_CONTEXT` — e.g., a hand-rolled `Context` object whose `setValue` returns a new object, or directly construct two distinct context objects and pass one into `context.with`. Then assert that `context.active()` inside the callback is the new object (not `base`) and outside is `base`.

Everything else in the diff (cache-repository TTL/eviction, MultiTier delete/clear/stats, propagation, metrics registry, span-context zeros, accessor wiring, `_resetShimForTests`) looks sound — backends seed/read correctly, `afterEach` resets shim globals, no real-time/Date.now flakiness, no leaked timers.

---

PR #2050, opened 2026-06-02
fix(cache): invalidate L1 tier on MultiTier deleteByPrefix/clear (#1989)

ariskemper:
## What

Fixes a **stale-read bug** in `MultiTierCacheRepository` surfaced while writing tests in #2049: after `deleteByPrefix` or `clear`, `get()` kept returning deleted/cleared values for up to `defaultTtlSeconds` (5 min).

## Root cause

`deleteByPrefix`/`clear` only issued a pattern delete against the **L3 backend**. But `MultiTierCache.get()` consults **L1 first**, and the in-memory L1 tier was never invalidated — so it kept serving the stale entry until its TTL.

## Fix

`MultiTierCacheRepository` now holds a reference to its L1 `MemoryTier` and drops matching entries from it alongside the backend pattern delete. Adds `MemoryTier.deleteByPrefix(prefix)`. Contained to `cache-repository.ts` — no change to the shared `MultiTierCache`.

## Tests

Two regression tests (`deleteByPrefix`/`clear` invalidate L1 → no stale read), **red-green verified**: both fail against the pre-fix code and pass with the fix.

## Note on the second behaviour (asyncBackfill fire-and-forget set)

Investigated and **left as-is — it is not a correctness bug**. L1 (memory) writes are synchronous and `get()` reads L1 first, so read-after-write stays consistent (the existing `set`→`get` tests pass deterministically). `asyncBackfill` only makes the **L3 distributed write** best-effort, which is its intended purpose (write latency). Changing it would regress that deliberate design for no read-consistency benefit.

## Stacking

Based on `test/1989-missing-tests` (#2049), which creates `cache-repository.test.ts`. Merge #2049 first, or this rebases cleanly onto main after it lands.

Refs #1989

### DIFF
--- a/src/observability/tracing/api-shim.test.ts
+++ b/src/observability/tracing/api-shim.test.ts
@@ -1,0 +1,206 @@
+import "#veryfront/schemas/_test-setup.ts";
+import { assertEquals, assertExists } from "#veryfront/testing/assert.ts";
+import { afterEach, describe, it } from "#veryfront/testing/bdd.ts";
+import {
+  _resetShimForTests,
+  context,
+  defaultTextMapGetter,
+  defaultTextMapSetter,
+  getGlobalMetricsAPI,
+  getGlobalTracerProvider,
+  getTracer,
+  type MetricsAPI,
+  propagation,
+  setGlobalActiveSpanAccessor,
+  setGlobalMetricsAPI,
+  setGlobalTracerProvider,
+  type Span,
+  SpanKind,
+  SpanStatusCode,
+  type TextMapPropagator,
+  trace,
+  type Tracer,
+  type TracerProvider,
+} from "./api-shim.ts";
+
+describe("observability/tracing/api-shim", () => {
+  afterEach(() => {
+    // Restore global shim state so tests don't leak into each other.
+    _resetShimForTests();
+  });
+
+  describe("no-op defaults (extension not installed)", () => {
+    it("getTracer returns a no-op tracer whose spans are inert but chainable", () => {
+      const tracer = getTracer("test", "1.0");
+      const span = tracer.startSpan("op");
+
+      // All mutators are chainable no-ops and never throw.
+      assertEquals(span.setAttribute("k", "v"), span);
+      assertEquals(span.setAttributes({ a: 1 }), span);
+      assertEquals(span.setStatus({ code: SpanStatusCode.OK }), span);
+      assertEquals(span.addEvent("evt"), span);
+      span.recordException(new Error("x"));
+      span.updateName("renamed");
+      span.end();
+
+      // The no-op span context has all-zero ids.
+      const ctx = span.spanContext();
+      assertEquals(ctx.traceId, "00000000000000000000000000000000");
+      assertEquals(ctx.spanId, "0000000000000000");
+      assertEquals(ctx.traceFlags, 0);
+    });
+
+    it("startActiveSpan invokes the callback with a span (fn-as-2nd-arg form)", () => {
+      const tracer = getTracer("test");
+      let received: Span | undefined;
+      const result = tracer.startActiveSpan("op", (span) => {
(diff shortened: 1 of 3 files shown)


ariskemper:
Thanks for the fix — the core invalidation gap is real and the tests cover it well. Flagging a couple of higher-confidence correctness concerns that I think still slip through with the current ordering:

### 1. Race lets the just-cleared L1 get re-poisoned by L3 before the prefix delete finishes
`src/repositories/cache/cache-repository.ts:219-230` (and the same shape at `:236-240` for `clear()`).

```ts
this.l1.deleteByPrefix(scopedPrefix);           // synchronous, immediate
// ... window here ...
const deleted = await this.backend.delByPattern(`${scopedPrefix}*`);  // async, may take a while on Redis
```

The repo is configured with `backfillOnHit: true` and `asyncBackfill: true` (constructor at `:197-198`). During the `await` window, any concurrent `get(k)` where `k` matches the prefix will:

1. Miss L1 (we just wiped it),
2. Miss L2 (not configured here, but `multi-tier.ts` would consult it),
3. Hit L3 because `delByPattern` hasn't completed yet,
4. Fire `backfill(key, l3Value, ["l1", "l2"])` (`src/cache/multi-tier.ts:140`), which writes the stale value back into L1.

Net effect: L1 ends up holding a stale entry that survives until its TTL expires — exactly the bug #1989 is trying to fix, just narrowed to the race window. On a remote Redis backend with many matching keys the window is easily tens to hundreds of ms.

Suggested fix: invalidate L1 **after** the L3 pattern delete resolves (and ideally both before and after, so concurrent reads during the window also miss L1 fast):

```ts
async deleteByPrefix(prefix: string): Promise<number> {
  const scopedPrefix = this.getScopedKey(prefix);
  if (!this.backend.delByPattern) {
    this.l1.deleteByPrefix(scopedPrefix); // still drop L1 even if L3 can't pattern-delete
    return 0;
  }
  const deleted = await this.backend.delByPattern(`${scopedPrefix}*`);
  this.l1.deleteByPrefix(scopedPrefix);   // after L3 is gone, so any racing backfill is harmless
  this.localStats.deletes += deleted;
  return deleted;
}
```

Same change for `clear()` at `:236-240`.

A regression test for this is feasible: wrap the `MemoryCacheBackend` so `delByPattern` awaits a deferred, fire a `get()` for a matching key during the window, then resolve the deferred and assert L1 (`repo.get(...)` with the L3 entry already pre-deleted) returns `null`.

### 2. `deleteByPrefix` skips L1 invalidation when the backend has no `delByPattern`
Same file, `:222-225`:

```ts
if (!this.backend.delByPattern) {
  logger.debug(`[${this.name}] deleteByPrefix not supported by backend`, { prefix });
  return 0;
}
```

We early-return before touching L1. The new test `deleteByPrefix returns 0 when the backend has no delByPattern support` (cache-repository.test.ts:130-140) actually locks in this behavior — but it means a caller invoking `deleteByPrefix` on, say, a stub backend will still serve stale L1 hits for matched keys. The L1 wipe is local and free; it should still run on this branch (move the `this.l1.deleteByPrefix(scopedPrefix)` call above the early return, or after — either is fine since there's no L3 race in the unsupported case).

`clear()` already does the right thing here because the optional-chain `delByPattern?.(...)` doesn't short-circuit the L1 wipe — though it does mean `clear()` silently no-ops on L3 when the backend lacks the method, which is a pre-existing wart, not introduced by this PR.

---

(1) is the load-bearing one — it reopens the same coherence bug under contention. (2) is a smaller scope leak but trivial to address while you're in the file.

ariskemper:
Superseded by #2064 — rebased onto current main (after #2049 merged) with both review findings from this PR incorporated (L1/L3 ordering race + no-delByPattern L1 wipe), plus red-green-verified regression tests.

Issue #29, opened 2026-01-08
Race condition in ProxyFSAdapterManager causes 500 errors on preview URLs


kwakayama:
## Summary

Two bugs in `ProxyFSAdapterManager` cause 500 errors and stale content because adapters are cached by `projectSlug` only, ignoring `productionMode` and `releaseId`.

---

### Problem
Concurrent requests with different `productionMode` share the same adapter and overwrite each other's state.

### Current Outcome
Preview URL `https://helpful-durden-fdtcx.preview.veryfront.com/` returns 500 Internal Server Error.

### Evidence
**Adapter reports `productionMode: false`:**
```json
{"timestamp":"2026-01-08T10:37:59.875Z","message":"[ProxyFSAdapterManager] Reusing cached adapter","context":{"requestProductionMode":false,"effectiveProductionMode":false}}
```

**But `listPublishedFiles` is called (only when `isProduction=true`):**
```json
{"timestamp":"2026-01-08T10:37:59.876Z","message":"[SSR] renderPage failed","context":{"error":"API request failed: 404 Not Found","stack":"...at listPublishedFiles (operations.ts:215)..."}}
```

### Race Scenario
1. Request A (production) → `setProductionMode(true)`
2. Request B (preview) → `setProductionMode(false)`, clears index
3. Request C (production) → `setProductionMode(true)` **before B finishes**
4. Request B sees `productionMode=true` → calls `listPublishedFiles` → 404

---

### Problem
`releaseId` is fetched from the API but ignored in the cache key. When a new release is published, the cached adapter continues serving stale files.

### Current Outcome
After publishing a new release, production URLs continue showing old content until the adapter is evicted by LRU/TTL.

### Code Flow
1. `universal-handler.ts:281` — fetches `releaseId` from `lookupProjectByDomain()`
2. `ssr-handler.ts:125` — passes `releaseId` to `runWithContext()`
3. `proxy-fs-adapter-manager.ts:54` — **IGNORED in cache key**

```typescript
// Line 54 - BUG: cache key doesn't include releaseId
const existing = this.adapters.get(projectSlug);
```

---

## Root Cause

**File:** `src/platform/adapters/proxy-fs-adapter-manager.ts:54`

```typescript
const existing = this.adapters.get(projectSlug);  // Only uses projectSlug
```

Should include `productionMode` and `releaseId` in the cache key.

---

## Fix

```typescript
async getAdapter(projectSlug, token, projectId?, productionMode?, releaseId?) {
[...]

PR #2052, opened 2026-06-02
feat(agent): configurable security input character limit, default 100k

ariskemper:
## Summary

The default agent security middleware hard-codes `maxLength: 50_000` with no override path. Real workloads — long documents, structured tool results, and multi-turn conversation history that the middleware JSON-stringifies before validating — routinely cross that boundary and surface as an opaque `Input validation failed: Input exceeds maximum length of 50000` error with no actionable hint to the user.

Discovered while testing the `dora-compliance-agent` example (`veryfront-examples` PR #29): a second contract review in the same chat session crashes because the prior streamed structured report alone (~30k chars) plus the next user turn pushes the JSON-stringified input over the 50k cap. The current workaround is `security: false`, which also drops prompt-injection blocking and PII filtering — heavier than the problem demands.

## Changes

- New top-level `inputMaxCharacterLimit?: number` on the agent config for per-agent overrides.
- New `security.agent.inputMaxCharacterLimit?: number` in `veryfront.config.ts` as a project-wide default.
- New `DEFAULT_AGENT_INPUT_MAX_CHARACTER_LIMIT = 100_000` constant; built-in default raised from 50k → 100k.
- Resolution order in `resolveSecurityMiddleware`: per-agent field → global config → built-in default.
- Resolution happens via the existing `__vfGetRuntimeConfig` globalThis bridge so the agent layer doesn't pull in a direct config-layer import.
- `security: false` continues to fully opt out of the middleware (unchanged).

## Usage

Per-agent override:

```ts
agent({
  id: "long-doc-reviewer",
  model: "...",
  inputMaxCharacterLimit: 250_000,
});
```

Project-wide default:

```ts
// veryfront.config.ts
export default defineConfig({
  security: {
    agent: { inputMaxCharacterLimit: 200_000 },
  },
});
```

## Test plan

- [x] `deno test src/agent/factory-security-middleware.test.ts` — 4 new cases pass alongside the existing ones (15 steps total). Covers: default 100k enforced; per-agent override; global config override; per-agent wins over global.
- [x] `deno test src/config/schemas/config.schema.test.ts` — passes with the extended schema.
- [x] `deno check` on all touched files.
- [ ] Smoke test against the `dora-compliance-agent` example: undo its `security: false` workaround, lift the limit via `inputMaxCharacterLimit`, confirm the multi-turn review flow no longer trips the validator.

### DIFF
--- a/docs/api-reference/veryfront/agent.md
+++ b/docs/api-reference/veryfront/agent.md
@@ -131,28 +131,29 @@ Agent helper.
 
 | Property | Type | Description | Source |
 |----------|------|-------------|--------|
-| `id?` | `string` | Unique identifier (auto-generated if omitted) | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L81) |
-| `name?` | `string` | Human-readable display name for registry and control-plane listings. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L83) |
-| `description?` | `string` | Optional summary shown in registry and control-plane listings. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L85) |
-| `model?` | `ModelString` | Optional model string in "provider/model" format. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L93) |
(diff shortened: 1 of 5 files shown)


ariskemper:
@codex review

ariskemper:
@codex review

ariskemper:
@codex review

Issue #1986, opened 2026-06-01
[Tech debt] Platform, build & runtime — platform, build, modules, config…


ariskemper:
# Tech Debt: Platform, Build & Runtime (platform, build, modules, config, sandbox, runtime, registry)

## Verification

Adversarially re-verified 2026-06-01 by reading every cited `file:line` and re-running `rg`/`wc`.
**Tally: 14 findings + 5 nits checked → 14 ✅ Verified, 0 ⚠️ qualified, 0 ❌ refuted.** Every finding
is genuine debt and present in the code. No false positives. Corrections were limited to small
line-number drift (typically ±1) and a few summary-table count nits — all corrected inline below:

- `as unknown as` = **27 non-test** (93 incl. tests) — report's 27 is correct (non-test). ✅
- `@ts-ignore` = 18 total, 13 in `compat/fs.ts`, 3 in `core-platform.ts` — exact. ✅
- God files: 14 over 500 LOC; the six >770 files match byte-for-byte. ✅
- Per-subdir LOC drifted slightly (platform 185/20,131; sandbox 8/2,013; runtime 2/679) — the
  report under-listed by 1 file in three dirs; immaterial.
- Both 🔴 findings independently confirmed (paths + dependency direction). See each finding.

## Summary

Overall this slice is in **good health**. The code is well-factored, marker debt is almost
nonexistent (only a handful of TODO/HACK strings, and most of those are descriptive prose rather
than real debt), there are no empty/swallowed catch blocks, and every large file has a co-located
`*.test.ts`. The dominant real debt is concentrated in the **cross-runtime compat / per-runtime
adapter layer** (`src/platform/compat`, `src/platform/adapters/runtime`): unavoidable but un-abstracted
`as unknown as` runtime casts, hard `node:` imports living in nominally shared modules, duplicated
native-Deno/Response unwrap logic, and a regex-based TypeScript stripper used as a config fallback.
A few god files (>770 LOC) exist but are cohesive and tested. The biggest structural smell is a
cross-layer dynamic import from the FS adapter into `src/html`.

| Category                            | Count                                      | Worst offenders                                                                               |
| ----------------------------------- | ------------------------------------------ | --------------------------------------------------------------------------------------------- |
| 1. TODO/FIXME/HACK markers          | ~1 real (5 raw, mostly prose)              | none material                                                                                 |
[...]

PR #2067, opened 2026-06-03
refactor(platform): dedupe native-Deno access into shared getNativeDeno() helper (#1986)

ariskemper:
## Description

Focused, behavior-preserving "safe wins only" pass for the platform/build/runtime tech-debt cluster (#1986).

On reading the issue, most of the listed findings have **already been resolved** in earlier PRs (see Skipped section). The one genuine remaining slice of the 🔴 high-value finding was the duplicated native-`Deno` access cast. This PR finishes that dedupe and adds the missing unit coverage for the shared helper module.

### Changes mapped to findings

- **🔴 #1 — Duplicated native-Deno/Response unwrap** (partial; remaining slice): The `Response` re-wrap was already extracted to `compat/http/native-response.ts` (`getNativeResponse`/`toNativeResponse`). The native-`Deno` access `(self as unknown as Record<string, typeof Deno>)["Deno"]` was still copy-pasted at **4 sites**. Extracted a single `getNativeDeno()` helper into the same shared module (no new layering: both `compat/http` and the Deno runtime adapter already import from it). Replaced all 4 casts. Non-null sites keep their `!`; guarded sites keep `| undefined` narrowing — identical behavior.
  - Sites: `compat/http/deno-server.ts`, `compat/http/websocket.ts`, `adapters/runtime/deno/adapter.ts` (x2).
- **Missing tests for complex/untested file**: Added `compat/http/native-response.test.ts` covering `getNativeResponse`, `getNativeDeno`, and `toNativeResponse` (including the polyfilled re-wrap path). The module previously had zero direct coverage. Pure addition.

### Skipped findings (already fixed or out-of-scope per "safe wins only")

- **🔴 #1 Response re-wrap** — already extracted to `native-response.ts` before this PR; nothing to do.
- **🟡 #8 `as unknown as Map` cache wrapper** — already fixed: `createCacheRegistryWrapper` returns `CacheStatsSource` and `registerMapCache` accepts `CacheStatsSource` (no cast remains).
- **🟡 #3 `@ts-ignore` cluster in `compat/fs.ts`** — already reduced from 13 to 1; the `denoGlobal()` accessor work is effectively done.
- **🟢 Intentional-ignore catch logging** — `proxy-manager.ts` already logs `logger.debug`; `compat/fs.ts` intentionally does NOT log (documented: module must stay importable without `--allow-env`, logger reads env at import) — adding a log would break that constraint; `lazy-sandbox.ts` has no logger wired and its comment states the error is already handled by the caller path, so logging risks double-reporting. Skipped to stay behavior-preserving.
- **🔴 #2 FS-adapter → `src/html` layering inversion** — structural/risky; explicitly out of scope.
- **🟡 #5 `node:` imports in compat**, **🟡 #4 regex TS stripper**, **🟡 god-file decomposition**, **🟡 #6 WebSocket placeholder types**, **🟡 deprecated `getAdapter`** — larger refactors / not self-contained behavior-preserving; out of scope.

### Verification

- `deno check` on all four touched `.ts` files: the files themselves type-check (`Check ...` printed for each). The only error surfaced is a **pre-existing, unrelated** `TS2352` in `compat/process/lifecycle.ts` (`Deno.unrefTimer(timerId as number)`), confirmed present on `main` with these changes stashed — not introduced here.
- New test: `deno test ... native-response.test.ts` → **3 passed (6 steps), 0 failed**.
- Existing `deno-server.test.ts` integration cases fail in this sandbox because they bind real network ports (also fail on base) — environmental, not caused by this change.

## Related Issue(s)

Part of #1986

## Type of Change

- [x] Code refactoring
- [x] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)

### DIFF
--- a/src/platform/adapters/runtime/deno/adapter.ts
+++ b/src/platform/adapters/runtime/deno/adapter.ts
@@ -28,7 +28,11 @@ import {
   enqueueWatchEvent,
 } from "../shared/watcher-queue.ts";
 import { stopManagedServer } from "../shared/server-lifecycle.ts";
-import { getNativeResponse, toNativeResponse } from "../../../compat/http/native-response.ts";
+import {
+  getNativeDeno,
+  getNativeResponse,
+  toNativeResponse,
+} from "../../../compat/http/native-response.ts";
 
 const logger = serverLogger.component("deno");
 
@@ -293,7 +297,7 @@ class DenoServerAdapter implements ServerAdapter {
   upgradeWebSocket(request: Request): WebSocketUpgrade {
     // Access native Deno via `self` to bypass dnt shim transform.
     // dnt rewrites `globalThis.Deno` to @deno/shim-deno, which lacks upgradeWebSocket.
-    const nativeDeno = (self as unknown as Record<string, typeof Deno>)["Deno"];
+    const nativeDeno = getNativeDeno();
     if (typeof nativeDeno?.upgradeWebSocket !== "function") {
       throw NOT_SUPPORTED.create({
         detail: "DenoServerAdapter.upgradeWebSocket() can only be used in Deno runtime",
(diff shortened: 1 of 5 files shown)


ariskemper:
@codex review

PR #2072, opened 2026-06-03
fix(agent): remove input character limit from security middleware

ariskemper:
## Description

The default agent security middleware enforced a **100,000-character cap** on input (configurable via `inputMaxCharacterLimit`, default `100_000`). The middleware JSON-stringifies the *entire* agent input — the latest user message **plus conversation history and structured tool results** — so longer chats accumulate past the cap. When that happens the middleware **throws** (`Input exceeds maximum length of 100000`), which surfaces in the frontend chat as an error and prevents the conversation from continuing.

This removes the input character limit entirely — agent input is now unbounded.

### Changes
- **`src/agent/types.ts`** — remove the `inputMaxCharacterLimit?` field from `AgentConfig`.
- **`src/agent/factory.ts`** — remove `DEFAULT_AGENT_INPUT_MAX_CHARACTER_LIMIT` and the `resolveInputMaxCharacterLimit` resolver; stop passing `maxLength` to the security middleware (drop it from the `resolveSecurityMiddleware` `Pick`).
- **`src/agent/factory-security-middleware.test.ts`** — replace the four limit/override/fallback tests with a regression test asserting a 500k-char input passes through untouched.
- **`docs/api-reference/veryfront/agent.md`** — drop the `inputMaxCharacterLimit` row.

The `InputValidator` already treats `maxLength` as optional and **skips the length check when it is absent** (`src/agent/middleware/security/validator.ts:122`), so no validator change is needed. **Prompt-injection pattern blocking and output PII filtering are unchanged.**

## Related Issue(s)

Fixes the "frontend chat errors and cannot continue" report on large/long conversations.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [x] Test update

> **Note:** Technically API-breaking — the public `inputMaxCharacterLimit` agent-config option and the exported `DEFAULT_AGENT_INPUT_MAX_CHARACTER_LIMIT` constant are removed. Any agent setting `inputMaxCharacterLimit` will now fail typecheck and should drop the option.

## Checklist

## Verification
- `deno lint` clean on touched files; `deno check` shows only pre-existing baseline errors in untouched files (`react/react.ts`, `compat/process/lifecycle.ts`) — CI typecheck is authoritative and green.
- `deno test src/agent/factory-security-middleware.test.ts` → **1 passed (12 steps), 0 failed**, including the new "does not enforce any input character limit" case.

## Security note
This intentionally removes a DoS/unbounded-input guard on agent input. Prompt-injection blocking and PII output filtering remain in place; if a bound is ever wanted again it should live at the transport/request layer rather than silently throwing mid-conversation.

### DIFF
--- a/docs/api-reference/veryfront/agent.md
+++ b/docs/api-reference/veryfront/agent.md
@@ -153,7 +153,6 @@ Agent helper.
 | `skills?` | `true \| string[]` | Enable skills for this agent. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L150) |
 | `suggestions?` | `Suggestions` |  | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L151) |
 | `security?` | `false` | Set to false to disable the default security middleware | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L153) |
-| `inputMaxCharacterLimit?` | `number` | Maximum input character length enforced by the default security middleware. The middleware JSON-stringifies the agent input (latest user message plus any conversation history and structured tool results carried with it) and rejects anything longer. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/types.ts#L162) |
 
 **Returns:** `Agent`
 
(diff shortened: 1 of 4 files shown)


ariskemper:
@codex review

PR #2086, opened 2026-06-03
fix(agent): pair replayed multi-tool turns in AG-UI history (#2082)

### LINKED ISSUE #2082: Multi-turn chat fails after a tool-using turn: tool_result sent to Anthropic without its tool_use ("unexpected tool_use_id in tool_result blocks")
## Summary

In a multi-turn AG-UI chat, the **second** user turn fails whenever the **first** turn's assistant response used one or more tools. The agent runtime replays the conversation history to the model, but its AG-UI→provider message conversion emits a `tool_result` block whose `tool_use_id` has no matching `tool_use` in the immediately preceding message, so the Anthropic API rejects the request:

```
veryfront-cloud request failed: {"type":"error","error":{"type":"invalid_request_error",
"message":"messages.2.content.0: unexpected `tool_use_id` found in `tool_result` blocks:
toolu_01GcDPvdiGb2qRSeJ7G9vyQ5. Each `tool_result` block must have a corresponding
`tool_use` block in the previous message."},"request_id":"req_..."}
```

Server side this surfaces as `✖ [agent] Agent stream error … pathname=/api/ag-ui`.

## Reproduction

`veryfront@0.1.642`, the `gdpr-compliance-agent` example (route is just `export const POST = createAgUiHandler("dpo")` — no custom history handling). Any agent with tools reproduces it:

1. Turn 1 — ask something that makes the agent call a tool, e.g. `What does Article 28 require from a processor?` (the coordinator calls `load-skill` then `searchGdprCorpus`). It answers fine.
2. Turn 2 — send **any** follow-up (`Run a full GDPR review of the selected document.`, or even a second Q&A).
3. Turn 2's first model call fails with the error above. Turn 1 alone always works; the failure is purely a function of replayed tool history.

## Diagnosis

Capturing the turn-2 AG-UI request payload, the replayed assistant message (turn 1) is a single message packing **multiple sequential tool calls** plus reasoning/text:

```
messages[0] role=user      parts=[text]
messages[1] role=assistant parts=[reasoning, data-tool-call-status,
                                  tool-load-skill, data-tool-call-status,
                                  tool-searchGdprCorpus, text]
messages[2] role=user      parts=[text]
```

In `esm/src/agent/runtime/message-adapter.js`, `createProviderMessageFromAgentRuntimeMessage` builds the assistant provider message as:

```js
case "assistant":
  return { role: "assistant", content: [...textParts, ...toolCallParts] };
```

and emits tool results only as a *separate* `role: "tool"` message (`case "tool"`). The AG-UI `tool-<name>` dynamic-tool parts carry both the call **and** the result in one part, and a single assistant turn here contains **two sequential** tool calls (`load-skill`, then `searchGdprCorpus`). When this one merged message is expanded back into the alternating Anthropic `tool_use`→`tool_result` sequence, the pairing is lost: a `tool_result` ends up in a message whose previous message does not contain the matching `tool_use` (Anthropic requires every `tool_result` block to be immediately preceded by an assistant message containing the corresponding `tool_use`).

Likely contributing factors to verify:
- A single AG-UI assistant message with **N sequential** tool calls must expand to N (`assistant tool_use` → `tool` tool_result) pairs in order, not one assistant message with N `tool_use` blocks followed by a misplaced result message.
- `toolResultParts` collected from an `assistant` message are dropped in the `assistant` case, while results are expected to arrive on a separate `tool` message — the merged AG-UI representation breaks that assumption on replay.

## Impact

- **Every multi-turn conversation breaks after the first tool use** — i.e. essentially every real agent chat. The first turn works, the second 500s. This is high severity for any tool-using agent (the common case).
- Not fixable in example/app code: the route delegates entirely to `createAgUiHandler`; history serialization is framework-owned.

## Acceptance criteria

- [ ] A turn-1 assistant response that used one or more tools can be followed by a turn-2 message without an Anthropic `invalid_request_error`.
- [ ] An AG-UI assistant message containing multiple sequential `tool-<name>` parts is converted into a correctly ordered, fully paired `tool_use`/`tool_result` provider sequence (each `tool_result` immediately preceded by the assistant message holding its `tool_use`).
- [ ] Regression test covers replaying a 2+ tool-call assistant turn into provider messages and asserts every `tool_result.tool_use_id` has a matching `tool_use` in the prior message.
- [ ] `gdpr-compliance-agent` repro above: Q&A → document review across two turns succeeds.

## Environment

- `veryfront@0.1.642`; Deno dev server; example `veryfront/veryfront-examples` → `gdpr-compliance-agent` (`app/api/ag-ui/route.ts` = `createAgUiHandler("dpo")`).
- Suspect file: `esm/src/agent/runtime/message-adapter.js` (`createProviderMessageFromAgentRuntimeMessage` / `collectAgentRuntimeProviderContentParts`).

ariskemper:
## Description

Multi-turn AG-UI chats failed on the **second** user turn whenever the first turn's assistant response used one or more tools, with the Anthropic error:

```
messages.2.content.0: unexpected `tool_use_id` found in `tool_result` blocks:
toolu_…. Each `tool_result` block must have a corresponding `tool_use` block in
the previous message.
```

### Root cause

On the `createAgUiHandler` path, replayed history is normalized by `normalizeAgUiMessages` (`src/agent/ag-ui/host-support.ts`). A replayed assistant turn packs several sequential dynamic `tool-*` parts — each carrying both the call **and** its `output` — plus a trailing answer, all in one AG-UI message. The old normalizer emitted a single assistant message containing *every* part (`[tool-call, tool-call, text]`) followed by the tool results as separate `tool` messages.

Downstream, `convertToTextGenerationRuntimeMessages` defers that trailing `text` (tool calls are still pending) into its own assistant message, which then lands **between** the `tool_use` blocks and the `tool_result` messages — orphaning the results:

```
assistant[tool_use load, tool_use search]
assistant[text]            ← inserted here
tool[result load, result search]   ← orphaned: previous message has no tool_use
```

### Fix

`normalizeAgUiMessages` now walks an assistant message's parts in order and flushes an `assistant` → `tool` pair as soon as a tool part with an inline `output` is seen, keeping any trailing text in its own assistant message **after** the results:

```
assistant[tool_use load] → tool[result load]
assistant[tool_use search] → tool[result search]
assistant[text]
```

Every `tool_result` is now immediately preceded by the assistant message holding its `tool_use`, satisfying the provider contract. Existing single-tool and explicit `tool-result` behaviors are unchanged; an assistant turn whose parts all drop out (e.g. reasoning-only) still yields a single empty assistant message.

## Related Issue(s)

Fixes #2082

## Type of Change

## Checklist

### Tests added (`src/agent/ag-ui/host-support.test.ts`)

- expands a turn with multiple sequential tool calls into ordered, paired messages
- replays multi-tool turns without orphaning any tool_result on the wire (asserts every `tool_result.toolCallId` has a matching `tool_use` in the immediately preceding wire message — the exact contract #2082 violated)
- keeps an empty assistant message when every part drops out (parity regression)

Affected suites green locally: `src/agent/ag-ui`, `src/agent/runtime`, `src/chat` — 166 passed, 0 failed.

<sub>This branch was run through a three-pass deep review (simplify / code-review / security). Two follow-up notes surfaced as **pre-existing** and intentionally out of scope here: the per-message `parts` array has no `.max()` bound (DoS amplification, unchanged in category by this fix), and synthesized message ids derive from client input (matching the existing `tool_${toolCallId}` scheme; no downstream consumer keys on these ids).</sub>

### DIFF
--- a/src/agent/ag-ui/host-support.test.ts
+++ b/src/agent/ag-ui/host-support.test.ts
@@ -9,6 +9,8 @@ import {
   parseAgUiRequest,
   parseAgUiRequestOrError,
 } from "./host-support.ts";
+import { convertToTextGenerationRuntimeMessages } from "../runtime/text-generation-runtime-message-converter.ts";
+import type { Message } from "../types.ts";
 
 describe("agent/ag-ui-host-support", () => {
   it("parses a valid AG-UI request body through the public helper", async () => {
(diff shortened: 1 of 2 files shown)


ariskemper:
## Deep review — 1 high-confidence integration issue

The pairing logic in this PR looks correct in isolation, and the tests are solid. However, this branch is **stale relative to `main` in a way that will silently revert a just-merged fix.**

### Stale base will revert / conflict with the already-merged #2085

PR **#2085 ("Prevent AG-UI from replaying provider-owned tools")** has already landed on `main` (commit `67f1917`). This branch is based on `f1049b6` — the commit *immediately before* that merge — so it does **not** contain #2085. Both PRs rewrite the **same** function, and #2086 drops everything #2085 added:

- **`src/agent/ag-ui/host-support.ts`** — #2086 changes `normalizeAgUiMessages` back to `normalizeAgUiMessages(messages: AgUiRequest["messages"])`, removing the `options` / `NormalizeAgUiMessagesOptions` / `providerOwnedToolNames` parameter that #2085 added on `main`.
- **`src/agent/ag-ui/host-support.ts`** — #2086 deletes `extractAssistantToolOutputMessages` (which on `main` now contains the provider-owned-tool stripping) and replaces it with `expandAssistantMessageWithToolOutputs`, which has **no provider-owned filtering** at all.
- **`src/agent/ag-ui/handler.ts:229` and `handler.ts:279`** — this branch calls `normalizeAgUiMessages(request.messages)`, whereas `main` now calls it with `{ providerOwnedToolNames: getProviderToolNames(agent) }`.

**Impact when this lands on current `main`:**
1. Merge conflict in `normalizeAgUiMessages` and the removed `extractAssistantToolOutputMessages`.
2. A naive resolution that keeps this PR's versions **reintroduces the provider-owned-tool replay bug #2085 fixed** (e.g. the Swedish Tax Adviser `web_search` `tool_result` orphan) — `expandAssistantMessageWithToolOutputs` would replay provider-owned tool calls/results again.
3. The `providerOwnedToolNames` plumbing that `main`'s `handler.ts` depends on would be dropped (or break compilation if `main`'s 2-arg call site is kept against this PR's 1-arg signature).

**Suggested fix:** rebase this branch onto current `main` and re-implement the multi-tool pairing *inside* the post-#2085 code path — i.e. keep the `options.providerOwnedToolNames` parameter and the provider-owned stripping, and apply them within the new `expandAssistantMessageWithToolOutputs` walk so both fixes coexist. The new regression tests here should still pass once the provider-stripping path is preserved.

*(Reviewed the pairing logic itself — ordering, the `${id}-${segmentIndex}` segment ids, and the empty-assistant preservation all look correct against this branch's base. The only blocking concern is the staleness above.)*

---

ariskemper:
Thanks — this was a valid blocking finding. The branch was based on the commit immediately before #2085 landed, so it would have conflicted and silently reverted the provider-owned-tool stripping.

Resolved by rebasing onto current `main` (`67f1917`, post-#2085) and re-implementing the multi-tool pairing **inside** the post-#2085 code path rather than alongside it:

- `normalizeAgUiMessages` keeps its `NormalizeAgUiMessagesOptions` / `providerOwnedToolNames` parameter and the cross-message `providerOwnedToolCallIds` tracking; `handler.ts`'s two-arg call sites are unchanged.
- The new `expandAssistantMessageWithToolOutputs` walk applies `shouldKeepProviderVisibleToolPart` to every part as it goes, so provider-owned tool calls/results (e.g. native `web_search`) are still dropped and never replayed as local calls — while non-provider-owned tools are flushed as ordered `assistant`→`tool` pairs.
- `extractAssistantToolOutputMessages` is removed (its provider-owned-skip logic is now folded into the ordered walk).

Added a regression test (`strips provider-owned tools while still pairing local tools in a multi-tool turn`) that locks in the coexistence: a turn mixing a provider-owned `web_search` and a local `searchGdprCorpus` drops the `web_search` call/result entirely while the local tool stays paired and the trailing answer is preserved. The original pairing/orphan regression tests still pass, and main's existing #2085 behavior tests are untouched and green.

Force-pushed the rebased single commit (`f890ef3`).

ariskemper:
Closing as superseded. While this was in review, **#2087 ("Preserve AG-UI tool-result pairing during replay") merged to `main` (commit `4477f20`)** and fixes #2082 with an essentially identical approach: the same `buildMessageMetadataFields` helper, the same in-order segment walk that flushes `assistant`→`tool` pairs, the same `${id}-${segmentIndex}` segment ids, the same drop-empty-message behavior, and the same provider-owned stripping preserved within the walk. The diff between this branch and `main` is now just a function rename (`expandAssistantMessageWithToolOutputs` vs `normalizeAssistantMessage`) plus a doc comment — no behavioral difference.

#2087 also already carries equivalent regression tests (`pairs replayed assistant tool outputs before trailing assistant text`, a `convertToTextGenerationRuntimeMessages` wire-orphan assertion, and `does not synthesize tool results for provider-owned assistant tool parts`), so there is nothing here worth rebasing on top.

No action needed on #2082 — it is resolved on `main` by #2087.

ariskemper:
Superseded by the already-merged #2087 (identical fix). Closing to avoid a redundant no-op merge.

PR #2165, opened 2026-06-09
Fix compiled esbuild discovery fallback

kwakayama:
## Summary
- add a compiled-runtime fallback for esbuild transform when the async service path throws `Cannot read properties of undefined (reading 'unref')`
- allow primitive discovery to use transform fallback for source files without relative runtime imports, preserving normal build behavior for files that need bundling
- keeps the fix narrow to unblock project primitive discovery and TSX transforms in compiled staging runtime

## Evidence
- Staging `veryfront-server` logs for ai-operations root runs showed primitive discovery returning 0 agents/tools with repeated `Cannot read properties of undefined (reading 'unref')`, followed by `Internal agent stream request referenced unknown agent` HTTP 404.
- Same staging image also logged TSX render transform failures from `/app/components/layout.tsx` with the same `unref` error.
- Local discovery against `<local-path>` found 7 agents, 7 tools, 2 skills, and 0 discovery errors.

## Verification
- `deno check --config deno.json src/discovery/transpiler.ts extensions/ext-bundler-esbuild/src/esbuild-bundler.ts`
- `deno test -A --config deno.json src/discovery/transpiler.test.ts src/discovery/auto-discovery.integration.test.ts extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts`

## Follow-up
After merge and staging deploy, rerun the ai-operations Supplier Invoice Orchestrator demo to confirm project agents are discovered and the orchestrator completes.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.694",
+  "version": "0.1.695",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "workspace": [
--- a/extensions/ext-bundler-esbuild/src/esbuild-bundler.ts
+++ b/extensions/ext-bundler-esbuild/src/esbuild-bundler.ts
@@ -72,6 +72,12 @@ function mapOptions(options: BundleOptions): Record<string, unknown> {
   return mapped;
 }
 
+function isCompiledRuntimeUnrefError(error: unknown): boolean {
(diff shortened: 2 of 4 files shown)


kwakayama:
Closing in favor of a root-cause rollback PR. Staging validation showed v0.1.693 still fails discovery; v0.1.692 restores project-agent discovery, so the fallback patch should not merge.

Issue #2243, opened 2026-06-09
Low-severity security backlog (CSRF __Host- prefix, dev innerHTML, log injection, redirect re-check)


kojiwakayama:
**Severity: LOW** — found in deep security review (2026-06-09). Backlog.

- **CSRF cookie lacks `__Host-` prefix** — `src/security/csrf/helpers.ts:46`: rename to `__Host-vf_csrf` (Secure/Path=//no-Domain already present) to prevent sibling-subdomain cookie injection.
- **Dev error overlay uses innerHTML** — `src/server/handlers/dev/scripts/error-overlay.ts:48`: unsanitized error content; dev-only, not reachable in prod. Use `textContent` or the existing `html-sanitizer.ts`.
- **Client log handler logs raw body** — `src/server/handlers/monitoring/client-log.handler.ts:100-101`: logs first 500 chars of POST body on parse failure; gated to `isLocalProject` (dev). Sanitize before logging.
- **HTTP import plugin follows redirects without re-validating allowlist** — `src/routing/api/module-loader/esbuild-plugin.ts:38`: `redirect: "follow"`; use `redirect: "manual"` and re-check the Location host.
- **Worker full read in compiled-binary mode** — `src/security/sandbox/worker-permissions.ts:54`: narrow `read: true` to known dirs (relates to #2227).
- **Config file dynamic import** — `cli/shared/config.ts:49`: standard framework behavior; document the trust boundary / add `--no-config`.

Provider low items: `parseRetryAfterMs` accepts 0 (`src/provider/runtime-loader/provider-http.ts:48-60`); `ProviderError.message` reassigned post-construction (`:178-179`).

kojiwakayama:
Partially addressed in PR #2285 / commit b46f2c3970cd7090a867fe7c91b6dcaaa9af25c8: the dev error overlay now renders error messages, locations, and stacks via DOM nodes and `textContent` instead of prepending error-controlled strings through `overlayElement.innerHTML`.

Still open from this grouped issue: CSRF `__Host-` cookie prefix, client log parse-failure sanitization, HTTP-import redirect allow-list recheck, compiled-binary worker read narrowing, config dynamic-import trust boundary, and provider low-severity cleanup items.

kojiwakayama:
[...]

Issue #2238, opened 2026-06-09
Stale path-cache entry never deleted when cached path is not a regular file


kojiwakayama:
**Severity: MEDIUM** — found in deep correctness review (2026-06-09)

## Location
`src/transforms/mdx/esm-module-loader/module-fetcher/path-cache-lookup.ts:37`

## Issue
When `stat` succeeds but `isFile` is false (path points at a directory/symlink), the function returns `null` without deleting the stale `pathCache` entry. Only the `catch` block (line 54) deletes, and only when `stat` throws. The stale entry persists forever, so every request for that key wastes a stat + falls through to the full fetch/transform pipeline. Pre-existing bug preserved by the recent refactor.

## Fix
```ts
if (!stat?.isFile) {
  input.pathCache.delete(input.versionedKey);
  return null;
}
```

kojiwakayama:
Fixed in #2244 / release 0.1.739: stale path-cache entries are deleted when a cached path is no longer a regular file.

Issue #2236, opened 2026-06-09
String.replace(plainString,...) can rewrite a comment instead of the real import


kojiwakayama:
**Severity: MEDIUM** — found in deep correctness review (2026-06-09)

## Locations (all use `code.replace(original, replacement)` with a plain string)
- `src/transforms/mdx/esm-module-loader/module-fetcher/nested-imports.ts:77,93`
- `src/transforms/mdx/esm-module-loader/module-fetcher/import-rewriter.ts:103`
- `src/transforms/mdx/esm-module-loader/module-fetcher/http-fetcher.ts:80`
- `src/rendering/orchestrator/module-loader/dependency-resolver.ts:146`

## Issue
`String.prototype.replace` with a string first-arg replaces the **first substring match in the whole file**, not the regex match position. If the same import text appears earlier in a comment or string literal, that occurrence is rewritten and the real import is left broken at runtime.

Reproduction:
```
// old: from "@/Button"
import { Button } from "@/Button";
```
After `.replace('from "@/Button"', 'from "file:///tmp/Button.js"')` the **comment** is rewritten; the actual import is untouched.

Most likely to bite in `import-rewriter.ts` (dnt JSDoc) and `dependency-resolver.ts` (commented-out user imports).

## Fix
Use the existing `replaceSpecifiers` lexer (`src/transforms/esm/lexer.ts`, used at `transform.ts:533`), or splice using the regex match `.index`.

Issue #2233, opened 2026-06-09
ssr-import-rewriter uses Date.now() cache-buster — defeats downstream caching


kojiwakayama:
**Impact: HIGH** — found in deep performance review (2026-06-09)

## Location
`src/modules/server/ssr-import-rewriter.ts:86,102` (`applySSRImportRewrites`)

## Issue
The `cacheBuster` defaults to `Date.now()`, so rewritten import URLs get a fresh `?v=<ms>` token on every call. The same module compiled twice gets different import URLs, defeating HTTP-level and transform-level caching. Additionally the function makes 3 separate full-source regex passes (`rewriteBareImports`, `rewritePathAliases`, `rewriteRelativeImports`).

## Fix
1. Replace `Date.now()` with a content-hash-based version token (deterministic per content+config).
2. (Follow-up) Combine the 3 passes into a single lexer walk using `replaceSpecifiers` (`src/transforms/esm/lexer.ts`).

Likely the single highest-value perf fix — recovers a large fraction of currently-defeated transform cache hits.

kojiwakayama:
Leaving this open after #2244. The attempted content-hash cache-buster was reverted because it could stamp child modules with a parent hash and serve stale child code. The remaining fix should version/cache-bust each rewritten target module independently.

PR #2244, opened 2026-06-09
Fix high-severity security and bug findings

kojiwakayama:
Addresses the high-severity findings from the deep code review that are ready to ship now. This PR is release-prepared as `0.1.739` and includes the regenerated coding-agent template manifest.

## Security
- **#2226** Path traversal in coding-agent template tools — confines `read-file` and `edit-file` to the project dir and regenerates `cli/templates/manifest.json`.
- **#2227** Worker `ctx.fs` adapter unscoped — validates documented `ctx.fs` paths against the project root, including symlink and Windows-style escape coverage. Non-compiled mode is fully contained by `read=[projectDir]`. Compiled-binary direct `Deno.*` reads remain a separate blocking follow-up in **#2245**.
- **#2228** OAuth init handler error leak — returns a generic 500 body while logging full server-side error detail.
- **#2229** `constantTimeEqual` padding — pads with `0xff` to match the CSRF comparison.

## Bugs
- **#2230** DAG executor concurrency race — runs child graphs on a scoped executor instead of mutating shared `maxConcurrency`.
- **#2231** Stream reader-lock leaks — releases reader locks in the AG-UI SSE parser and React chat handlers.
- **#2232** MCP `setLogLevel` — throws `JsonRpcError` instead of rejecting with a plain object.
- **#2238** Stale path-cache entry — deletes the path-cache entry when the cached path is no longer a regular file.

## Explicit non-closure
- **#2233 is not fixed in this PR.** The attempted content-hash cache buster was reverted because it stamped child modules with the parent hash and could serve stale child code. A correct fix needs per-target-module versioning from the module server, so #2233 stays open.
- **#2245 remains blocking** for compiled-binary direct `Deno.*` filesystem reads.
- Remaining medium/low findings stay tracked in #2236–#2243.

## Verification
- `deno task verify:quick`
- `deno run -A scripts/build/generate-templates-manifest.ts --check`
- `git diff --check`
- Pre-push hook passed: format/lint/typecheck/unit suite, `2023 passed, 0 failed`.
- Hosted checks are being rerun after the 0.1.739 release/manifest repair.


kojiwakayama:
Thanks for the review — addressed below.

**Codex P1 — template manifest not regenerated:** Fixed. The guarded `read-file`/`edit-file` contents are now embedded in `cli/templates/manifest.json` (regenerated; `generate-templates-manifest.ts --check` is green), so scaffolded projects actually get the traversal check.

**Codex P2 / cache-buster transitive staleness:** Agreed, this was a real regression — a child stamped with the parent's content hash kept serving stale when only the child changed. **Reverted** the content-hash cache buster back to the prior per-request token. A correct optimization needs per-*target*-module versioning stamped by the module server (which reads the actual file); tracked separately. ⚠️ Note: the `Prepare security fixes for release 0.1.739` commit was cut while that change was still present — it's now reverted, so re-cut/verify before release.

**Review #2 + Codex P3 — symlink + Windows guard weakness:** Fixed in `worker-script.ts`. The guard now canonicalizes via `Deno.realPath` (a symlink inside the project pointing outside is rejected, not followed) and uses `relative()` for cross-platform containment. Added regression tests covering `../` traversal and symlink escape (`worker-script.test.ts`).

**Review #1 (HIGH/BLOCK) — compiled-binary `read: true` lets handlers bypass `ctx.fs` via direct `Deno.*`:** Acknowledged and **not claimed as closed.** This PR scopes the documented `ctx.fs` adapter and fully contains the **non-compiled** mode (`read` is already `[projectDir]` there). It is **defense-in-depth only** in compiled-binary mode. Narrowing `read:true` to a binary-validated allow-list (or removing direct `Deno.*` from user-module scope) can only be safely validated against the `binary e2e` job, so it's tracked as **blocking** follow-up #2245 and documented in-code at `worker-permissions.ts`. Issue #2227 wording updated to reflect the reduced scope.

If the reduced scope (defense-in-depth in compiled-binary mode, full containment elsewhere, with #2245 blocking) isn't acceptable, I'm happy to keep this blocked until #2245 lands with binary-CI evidence.

kojiwakayama:
Round 2 blockers — all addressed.

**1. Typecheck fail on worker-script.test.ts:140 (HIGH):** Fixed. The `SerializedError → Record<string, unknown>` cast now goes via `unknown` (`serialized as unknown as Record<string, unknown>`). `deno check src/security/sandbox/worker-script.test.ts` is clean.

**2. list-files.ts path traversal (HIGH):** Fixed. It had no containment at all — added the same canonicalized guard as read/edit.

**3. read/edit/list symlink escape (HIGH):** Fixed. The template tools no longer use a lexical check. All three now canonicalize via a new portable `realPath` (resolves symlinks) and reject anything outside the canonical project root; containment is separator-normalized for Windows.

Supporting changes:
- Added `realPath` to `veryfront/fs` (`src/platform/compat/fs.ts`, cross-runtime Deno/Node/Bun) since the template tools need symlink resolution without runtime-specific APIs, and updated the `fs/index` export-surface test.
- Regenerated `cli/templates/manifest.json` so scaffolded projects receive all three guarded tools (`--check` green).

Verification: `deno fmt --check`, `deno lint`, `deno check src/index.ts`, and the full `deno task test:unit` (2024 passed) all green via the pre-push hook. `generate-templates-manifest.ts --check` is current.

Unchanged by design: compiled-binary `read:true` remains documented residual risk tracked in blocking follow-up #2245 (not narrowed without binary-e2e validation).

Issue #2262, opened 2026-06-09
Grouped perf: portable-code scan, RAG re-parse, sequential embeds, import-map re-parse, nonce re-lowercase, proxy URL, TextEncoder


kojiwakayama:
**Impact: MEDIUM/LOW** — found in second review sweep (2026-06). Grouped.

- `src/cache/paths.ts:53` `hasHardcodedCachePaths` runs 12 regexes over full module strings on every cache write; add `if (!code.includes("file://")) return false` fast-reject + combined alternation.
- `src/embedding/rag-store.ts:320` `search()` re-reads + JSON.parses the entire store (multi-MB with embeddings) per query; cache parsed data with mtime/hash invalidation; drop pretty-print on save.
- `src/embedding/embedding.ts:58` `embedMany` batches run sequentially; use bounded parallelism.
- `src/html/html-shell-generator.ts:229` re-`JSON.parse`s the import map it just stringified, to read one URL; return the object alongside the string.
- `src/html/nonce-injection.ts:198` streaming path re-lowercases the whole accumulated buffer per `pull()`; lowercase only new chunks.
- `src/proxy/main.ts:282` `new URL(req.url)` parsed 2-3×/request; parse once and thread through.
- `src/observability/file-log-subscriber.ts:93` `new TextEncoder()` per log line; hoist to a field. LogBuffer double-redaction (`log-buffer.ts:47`). `isSensitiveKey` no memoization (`redact.ts:74`).

kojiwakayama:
**Partially addressed in PR #2268**: `hasHardcodedCachePaths` file:// fast-reject and `FileLogSubscriber` TextEncoder hoist are done.

Deferred (kept open): embedding batch parallelization (provider rate-limit implications — needs a deliberate bounded-concurrency design), RAG store re-parse caching, nonce-stream re-lowercasing, proxy URL double-parse, import-map re-parse.

kojiwakayama:
Partially addressed in PR #2282 / commit 0c008b1433737e470d47c5770336a0dcdf319de5: HTML shell generation now uses the import map object produced by `buildImportMap()` for the `react/jsx-runtime` preload, so it no longer `JSON.parse`s the import map JSON it just generated.

Still open from this grouped issue: embedding batch parallelization, RAG store re-parse caching, nonce-stream re-lowercasing, proxy URL double-parse, and the other deferred items.

kojiwakayama:
[...]

Issue #2259, opened 2026-06-09
withSpan rebuilds OTLP config + re-resolves tracer on every call


kojiwakayama:
**Impact: HIGH** — found in second review sweep (2026-06).

## Location
`src/observability/tracing/otlp-setup.ts:124` (`withSpan`/`withSpanSync`)

## Issue
Every span call runs `getConfig()` (6 `getEnv()` reads, building a fresh object) + `getTracer(serviceName)`. 439 call sites, many per-request (data fetch, cache, HTML gen, SSR) — 20+ spans per SSR request → thousands of env reads/request. Config is process-static.

## Fix
Lazy-memoize config and tracer at module scope (preserving the pre-bootstrap/test path where env isn't loaded yet — memoize on first call, not at import).

Issue #2257, opened 2026-06-09
Renderer init races + buildPagesRoutes ssgPaths + MemoryTokenStore states unbounded


kojiwakayama:
**Severity: MEDIUM** — found in second review sweep (2026-06). Grouped.

- `src/rendering/renderer.ts:151` `Renderer.initialize` clears `initializationPromise` in `finally` even on failure — confusing retry semantics for concurrent waiters; clear only on success.
- `src/rendering/renderer.ts:549` `initializeRenderer` has no init-promise guard — two concurrent calls create two `Renderer` instances; the first is leaked without `destroy()`. Add a module-level init promise.
- `src/build/production-build/static-generation.ts:192` `buildPagesRoutes` never pushes to `stats.ssgPaths` (app routes do at :324), so pages SSG paths are silently omitted from `buildResult.ssgPaths`.
- `src/oauth/token-store/memory.ts:45` `states` Map cleans up only inside `setState`; abandoned OAuth flows accumulate for up to the TTL window with no capacity bound. Add periodic/lazy cleanup or a cap.

kojiwakayama:
Not in the sweep-2 PRs (medium, deferred). The renderer init-promise guard, `buildPagesRoutes` ssgPaths push, and `MemoryTokenStore.states` cleanup are all small, clearly-correct fixes — good candidates for a follow-up batch.

Issue #2256, opened 2026-06-09
Redis rate-limit INCR/PEXPIRE race creates immortal keys


kojiwakayama:
**Severity: HIGH** — found in second review sweep (2026-06).

## Location
`src/middleware/builtin/security/redis-rate-limit.ts:80` (+ stale-client note at `:35`)

## Issue
`increment` does `INCR` then conditionally `PEXPIRE` only when `count === 1`. If a second request increments before the first sets the TTL (so it sees `count === 2` and skips PEXPIRE), and the first crashes / PEXPIRE fails before step 3, the key has no TTL and rate-limits the client forever.

Also: `ensureClient` returns a cached `this.client` even after a disconnect (network partition) — subsequent ops fail/hang. Reset `this.client = null` on the client `error`/`end` event.

## Fix
Atomic INCR+PEXPIRE via Lua/`MULTI`, or `PEXPIRE … NX` (Redis 7+).

PR #2264, opened 2026-06-09
Review sweep 2: XSS/output-escaping and injection hardening

kojiwakayama:
First of three PRs from the second review sweep. Output-escaping / injection hardening. All changed files typecheck, lint, and the full unit suite passes (pre-push).

## Fixes
- **#2251 (HIGH)** Hydration data embedded in an inline `<script>` went through plain `JSON.stringify` — `</script>` and U+2028/U+2029 breakouts. Now routed through a new `jsonForInlineScript` helper.
- **#2252 (HIGH/MED)** Output-escaping gaps:
  - `bodyClass` → `escapeHTML` into the `<body class>` attribute.
  - static-gen `slug` → `JSON.stringify` into its inline `boot()` script.
  - theme-persistence nonce → `buildNonceAttribute`.
  - `html-head` meta/link attribute values → `escapeHTML`.
  - frontmatter `script.content` / `style.content` → `neutralizeRawTextContent` (guards `</script>`/`</style>` without HTML-escaping intentional raw JS/CSS).
- **#2253 (HIGH)** API cache backend: `encodeURIComponent(projectRef)` in the URL path (path-injection guard) and strip the query string from the `http.url` span attribute (span attributes bypass log redaction; the query carries the cache key).
- **#2257 (part)** `use-uploads` delete: `encodeURIComponent(id)`.

## New helpers + tests
`jsonForInlineScript` and `neutralizeRawTextContent` added to `src/html/html-escape.ts` with unit tests (breakout neutralization, U+2028/29 escaping, JSON round-trip).

## Note on trust model
These are output-escaping fixes — correct as defense-in-depth regardless of whether frontmatter/slug are treated as trusted. The hydration one (#2251) is the clearest exploitable XSS since props/frontmatter reach a `<script>` tag.

Remaining sweep-2 work in follow-up PRs: bugs/concurrency (#2249, #2250, #2254–#2256) and performance (#2259–#2262).

### DIFF
--- a/src/build/production-build/static-generation.ts
+++ b/src/build/production-build/static-generation.ts
@@ -361,7 +361,7 @@ function generateClientRuntime(
   <script type="module">
     import { boot } from '/_veryfront/client.js';
     if (typeof boot === 'function') {
-      boot({ slug: '${route.slug}' });
+      boot({ slug: ${JSON.stringify(route.slug)} });
     }
   </script>
 </body>`;
--- a/src/cache/backends/api.ts
+++ b/src/cache/backends/api.ts
@@ -85,7 +85,7 @@ export class ApiCacheBackend implements CacheBackend {
 
     try {
       return await this.circuitBreaker.execute(async () => {
-        const url = `${this.apiBaseUrl}/projects/${projectRef}/cache${path}`;
+        const url = `${this.apiBaseUrl}/projects/${encodeURIComponent(projectRef)}/cache${path}`;
         const controller = new AbortController();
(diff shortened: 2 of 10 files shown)


kojiwakayama:
Closing as superseded. The branch now conflicts with current `main`, and resolving those conflicts by keeping `main` leaves no tree diff: #2251, #2252, #2253, and the #2257 upload-id piece have already landed through later work. The outstanding review note about `cache.operation` is also covered on current `main` via queryless span attributes.

PR #2269, opened 2026-06-09
Keep project render mutexes stable

### LINKED ISSUE #2250: Renderer per-project mutex deleted while held, defeating concurrency limit
**Severity: CRITICAL** — found in second review sweep (2026-06).

## Location
`src/rendering/renderer-concurrency.ts:186` (`releaseProjectSlot`)

## Issue
`releaseProjectSlot` acquires the project mutex, then inside that critical section does `projectMutexes.delete(projectId)` when `current <= 1` — before `release()` runs in `finally`. A concurrent `acquireProjectSlot` for the same project then calls `getProjectMutex()`, which creates a **brand-new** unlocked mutex and acquires it instantly. Two renders now both believe they hold the only slot, and the deleted count restarts at 0 — defeating `RENDER_PER_PROJECT_LIMIT` under load.

## Fix
Don't delete the mutex you're currently holding:
```ts
if (current <= 1) {
  projectRenderCounts.delete(projectId);
  // do NOT delete projectMutexes here
  return;
}
```

kojiwakayama:
## Summary
- stop deleting a per-project render mutex during final slot release
- add regression coverage for mutex identity across release/reacquire
- add test helpers to reset and inspect project concurrency state

Closes #2250

## Verification
- deno test --no-check --allow-all src/rendering/renderer-concurrency.test.ts
- deno check src/rendering/renderer-concurrency.ts src/rendering/renderer-concurrency.test.ts
- pre-push hook: format, lint, type check, generation, unit suite: 2029 passed, 18173 steps

### DIFF
--- a/src/rendering/renderer-concurrency.test.ts
+++ b/src/rendering/renderer-concurrency.test.ts
@@ -1,7 +1,14 @@
 import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals, assertRejects } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
-import { Mutex } from "./renderer-concurrency.ts";
+import {
+  __getProjectMutexForTests,
+  __resetProjectConcurrencyForTests,
+  acquireProjectSlot,
+  Mutex,
+  projectRenderCounts,
+  releaseProjectSlot,
+} from "./renderer-concurrency.ts";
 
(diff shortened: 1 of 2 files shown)


kojiwakayama:
Closing as superseded by #2267, which landed the same renderer project mutex race fix and closed #2250. This PR's CI was green, but the branch became dirty after #2267 merged.

Issue #2317, opened 2026-06-10
[security] Untrusted project code runs in-process with full Deno permissions when worker isolation is disabled


ariskemper:
## Summary

When worker isolation is disabled (the default, `WORKER_ISOLATION_*` off),
untrusted project route/page handlers are dynamically `import()`ed into the
**main server process** and executed there. Deno permissions are per-process,
and the production process runs with
`--allow-read --allow-write --allow-net --allow-env --allow-run --allow-sys`
(`deno.json:320-324`). In-process project code therefore inherits all of them,
so a multi-tenant shared-pool deployment has no enforced boundary between a
tenant and the host or other tenants.

This issue tracks the architectural remediation. The immediate
boundary-hardening pieces are split into #F1, #F2, #F3.

## Evidence (verified by reading source)

- Project modules are imported into the main process:
  `src/routing/api/module-loader/loader.ts:105,109,481`, executed at
  `src/routing/api/route-executor.ts:372-394`
  (`await resolvedFn(request, appContext)`).
- The isolation decision gates only on `isWorkerIsolationEnabled()`, NOT on
  `isLocalProject`: `src/routing/api/route-executor.ts:357,411`. So untrusted
  remote projects also run in-process when the flag is off (default at
  `src/security/sandbox/worker-pool.ts:400-403`).
- Multi-tenant single-process is confirmed by the per-request
  `runWithProjectEnv` AsyncLocalStorage overlay:
  `src/server/runtime-handler/index.ts:610`.
- Shared pool exists: `src/proxy/server-resolver.ts:5,8,53`.
- The two in-process guards are not security boundaries: `getEnv()` is
  cooperative (a direct `Deno.env` call bypasses it), and `validateHTTPImports`
  (`src/routing/api/module-loader/http-validator.ts`) is a regex over source
  text (bypassable via `import(atob(...))`, `fetch`+`eval`).

## Impact

For self-hosted / single-tenant use (`isLocalProject = true`) this is a
non-issue. For a hosted shared-pool deployment serving untrusted tenants
in-process, a tenant can call `Deno.env.toObject()` (host secrets),
`Deno.readFile`/`writeFile` (host and other tenants' files), `fetch`
(SSRF/exfil), and `Deno.Command` (host command execution). Severity is High, and
Critical if the shared-pool process holds platform-wide credentials.

## Acceptance Criteria

Phase 1 (this PR, non-breaking):

- [ ] Emit a single, explicit warning at request-handling time (or startup) when
[...]

PR #2327, opened 2026-06-10
Warn on untrusted in-process route execution

kojiwakayama:
## Summary

Adds the Phase 1 warning for #2317 when a remote/untrusted project route would execute in-process while worker isolation is disabled.

## Changes

- Threads `isLocalProject` from `ApiHandlerWrapper` into `APIRouteHandler.handle()` and then into `ExecuteRouteOptions`.
- Emits a single process-local warning before the existing in-process fallback for untrusted app/pages route execution.
- Names the required isolation env flags in the warning: `WORKER_ISOLATION_ENABLED=1` and `WORKER_ISOLATION_API=1`.
- Covers the warning path with a unit test that verifies execution behavior remains unchanged.

## Simplifications made

- Kept the change at the existing route-executor boundary instead of introducing a new isolation policy layer.
- Preserved the current execution behavior; this PR only adds visibility for the known Phase 1 risk.

## Verification

- `VF_DISABLE_LRU_INTERVAL=1 deno test --no-check --allow-all --unstable-worker-options src/routing/api/route-executor.test.ts`
- `deno fmt --check src/routing/api/route-executor.ts src/routing/api/route-executor.test.ts src/routing/api/handler.ts src/server/handlers/request/api/api-handler-wrapper.ts`
- `deno check src/routing/api/route-executor.ts src/routing/api/handler.ts src/server/handlers/request/api/api-handler-wrapper.ts`
- `deno task verify:quick`
- `VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/server/build-app-route-renderer.test.ts`

## Remaining risks

- Phase 2 is still required to make untrusted route execution worker-gated by default.
- The full pre-push unit suite hit a parallel-only failure in `src/server/build-app-route-renderer.test.ts`; the same test passed in isolation under the same env. `deno task verify:quick` and the focused route-executor test passed.

Related: #2317

### DIFF
--- a/src/routing/api/handler.ts
+++ b/src/routing/api/handler.ts
@@ -14,6 +14,7 @@ import { loadHandlerModule } from "./module-loader/loader.ts";
 import { discoverAppRoutes, discoverPagesRoutes } from "./route-discovery.ts";
 import { executeAppRoute, executePagesRoute, type ExecuteRouteOptions } from "./route-executor.ts";
 import { withSpan } from "#veryfront/observability/tracing/otlp-setup.ts";
+import type { HandlerContext } from "#veryfront/types";
 
 /** Max entries in the loaded-handler LRU cache */
 const HANDLER_CACHE_MAX_ENTRIES = 256;
@@ -129,7 +130,7 @@ export class APIRouteHandler {
     );
   }
 
-  handle(request: Request): Promise<Response | null> {
+  handle(request: Request, ctx?: HandlerContext): Promise<Response | null> {
     const { pathname } = new URL(request.url);
 
     return withSpan(
@@ -192,6 +193,7 @@ export class APIRouteHandler {
         const isolationOptions: ExecuteRouteOptions = {
           modulePath: match.route.page,
           projectDir: this.projectDir,
+          isLocalProject: ctx?.isLocalProject,
         };
 
         const response = isAppRoute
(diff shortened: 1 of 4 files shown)


kojiwakayama:
Root cause update for the earlier CI failure: the new isolation warning wrote through `logger.warn` before API route execution. The integration log guard throws on unexpected `console.warn`, so `ApiHandlerWrapper` caught that diagnostic failure and fell through to SSR, returning HTML/404 for API requests.

Fix pushed in `06bfa72ef`:
- warning emission is now non-fatal
- added a regression covering pages API route execution when warning logging fails
- kept the existing process-local one-time warning behavior

Verified locally:
- `VF_DISABLE_LRU_INTERVAL=1 deno test --no-check --allow-all --unstable-worker-options src/routing/api/route-executor.test.ts`
- `VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net tests/integration/server/production-server.test.ts tests/integration/server/production/api.test.ts tests/integration/server/production/server.test.ts`
- `deno check src/routing/api/route-executor.ts src/routing/api/route-executor.test.ts src/routing/api/handler.ts src/server/handlers/request/api/api-handler-wrapper.ts`
- `deno fmt --check src/routing/api/route-executor.ts src/routing/api/route-executor.test.ts src/routing/api/handler.ts src/server/handlers/request/api/api-handler-wrapper.ts`
- `deno task verify:quick`
- pre-push hook passed format, lint, typecheck, and the full unit suite (`2053 passed`, `0 failed`).

Issue #2320, opened 2026-06-10
[security] Project workers granted unrestricted network egress (no internal/metadata IP filtering)


ariskemper:
## Summary

Project workers are granted unrestricted network access (`net: true`, no host
allowlist or egress filtering). Combined with host-env exposure (#F1), this is
the channel that turns secret disclosure into exfiltration, and it allows
project code to reach internal-only services and cloud metadata endpoints (for
example `169.254.169.254`).

## Evidence (verified)

- `src/security/sandbox/worker-permissions.ts:58` and `:69` (`net: true`, no
  allowlist).
- Comment at `:41` documents the intent ("data fetchers and API routes may call
  external APIs"), so egress is a product requirement, but it is currently
  unbounded.

## Impact

Server-side request forgery and data exfiltration from project code: access to
link-local/metadata IPs and internal RFC1918 ranges. Medium (the egress
requirement bounds it, but internal/metadata reachability is not intended).

## Acceptance Criteria

- [ ] An egress guard denies requests from project code to link-local
      (`169.254.0.0/16`, `fe80::/10`), loopback, and private RFC1918 ranges
      (`10/8`, `172.16/12`, `192.168/16`) and the cloud metadata IP, by default.
- [ ] The guard resolves the target host and blocks both direct-IP and
      DNS-rebinding-style access (post-resolution IP check), not just a
      literal-string check.
- [ ] Legitimate outbound requests to public hosts continue to work (example
      data fetchers are unaffected).
- [ ] The deny-list is configurable/overridable for self-hosted operators who
      intentionally need internal egress, with a secure default of deny.
- [ ] Tests cover: blocked metadata IP, blocked private range, blocked
      link-local, allowed public host, and a host that resolves to a private IP.
- [ ] Must not break veryfront-examples that fetch public APIs.

## Notes

Reference: `docs/security/security-audit-2026-06-10.md` (F3). Related: #F1.

kojiwakayama:
Fixed by #2324, merged in afe13c350ed0c4dec1bd31dc7f06480e34cddcfd.\n\nValidation:\n- CI format, lint, typecheck passed\n- unit, integration, RSC browser e2e, and binary e2e passed\n- coverage gate passed\n- CodeQL passed\n\nPR comments/reviews:\n- Automated Codex review posted no concrete requested change.

ariskemper:
Follow-up from the audit review: the merged fix (PR #2324,
[...]

PR #2331, opened 2026-06-10
fix(sandbox): re-check egress on every redirect hop in worker fetch guard (#2320)

ariskemper:
## Description

Follow-up to the merged worker egress guard (PR #2324 / #2320). That change
blocks direct fetch/connect to internal hosts well (IP literals, post-DNS host
checks, plus `Deno.connect`/`connectTls`). The fetch path remained bypassable
through an HTTP redirect.

The fetch wrapper checked only the initial URL, then delegated to the platform
`fetch` with its default `redirect: "follow"`:

```ts
// before
const fetchWrapper = async (input, init) => {
  await assertWorkerEgressAllowed(input, baseOptions);
  return await originalFetch(input, init); // default redirect: "follow"
};
```

Only the first hop was checked. A project handler that fetches a public URL
whose host an attacker can influence (a partner API, a link shortener, an
attacker host, any endpoint that can emit a 3xx) receives e.g.
`302 Location: http://169.254.169.254/latest/meta-data/...`; the runtime follows
it and the guard never re-checks it, so cloud-metadata, loopback, and RFC1918
services become reachable again.

This adds `guardedEgressFetch`, which:

- forces `redirect: "manual"` on the underlying fetch and re-runs
  `assertWorkerEgressAllowed` on each `Location` (resolved against the current
  URL) before following it, with a redirect cap;
- honors the caller's redirect intent (`manual` returns the redirect unfollowed,
  `error` throws, `follow` follows manually after re-checking);
- strips `Authorization`/`Cookie`/`Proxy-Authorization` on a cross-origin hop,
  matching the platform `fetch` the guard wraps, so manual following cannot leak
  the caller's credentials to the redirect target.

The installed fetch wrapper now delegates to it. `Deno.connect`/`connectTls`
wrapping is unchanged.

## Related Issue(s)

Follow-up to #2320 (see
https://github.com/veryfront/veryfront-code/issues/2320#issuecomment-4669733800)

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality to
      change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

Tests (red/green verified): public->internal redirect blocked (internal target
never fetched), public->public chain followed to the final response, `manual`
mode returns the redirect unfollowed, max-redirect cap, cross-origin redirect
strips `Authorization`/`Cookie`, same-origin redirect preserves them. Existing
egress and `project-worker` integration tests pass. `deno fmt`/`deno lint`
clean.

Note: the local pre-push `deno check src/index.ts` fails in the sandbox on
pre-existing, environment-specific React (`esm.sh @types/react`) type resolution
that also fails on a no-op push of `main`; this PR adds zero new type errors. CI
runs the authoritative checks.

### DIFF
--- a/src/security/sandbox/worker-egress-guard.test.ts
+++ b/src/security/sandbox/worker-egress-guard.test.ts
@@ -202,4 +202,60 @@ describe("worker-egress-guard guardedEgressFetch redirect handling", () => {
     assertEquals(seen[0], "Bearer secret");
     assertEquals(seen[1], "Bearer secret");
   });
+
+  it("blocks a redirect to a non-http(s) scheme (e.g. file://)", async () => {
+    let calls = 0;
+    const fetchImpl: typeof fetch = (input) => {
+      calls++;
+      const url = input instanceof Request ? input.url : String(input);
+      if (url.startsWith("http://93.184.216.34")) {
+        return Promise.resolve(redirectTo("file:///etc/passwd"));
+      }
+      throw new Error(`fetch should not have been called for ${url}`);
+    };
+
+    await assertRejects(
+      () => guardedEgressFetch("http://93.184.216.34/start", undefined, { fetchImpl }),
+      WorkerEgressBlockedError,
+    );
+    // The file:// target must never be fetched (would be a local file read).
+    assertEquals(calls, 1);
+  });
+
+  it("preserves the abort signal across redirect hops", async () => {
+    const controller = new AbortController();
+    const seenSignals: Array<AbortSignal | null | undefined> = [];
+    const fetchImpl: typeof fetch = (input, init) => {
+      const url = input instanceof Request ? input.url : String(input);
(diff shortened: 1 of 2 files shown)


ariskemper:
Thanks for the review. Both addressed in 03400bc5b:

- P1 (non-http(s) redirect targets): the manual redirect loop now rejects any redirect whose scheme is not http(s) before following it, so a `file://` (or other non-network) `Location` cannot be turned into a local file read. Test: a `file://` redirect is blocked and the target is never fetched (red/green verified).
- P2 (Request options): request-level options (`signal`, `credentials`, `cache`, `mode`, `referrer`, `referrerPolicy`, `keepalive`) are now carried across every hop, so e.g. `fetch(new Request(url, { signal }))` keeps a working abort signal. Tests cover signal preservation across hops and from a Request input.

PR #2332, opened 2026-06-10
Drop deprecated Anthropic sampling params for Opus 4.8

kwakayama:
## Summary
- omit `temperature` and `top_p` for Claude Opus 4.7+ Anthropic requests
- strip equivalent provider-option sampling overrides so veryfront-cloud does not forward rejected params
- add a regression test for Claude Opus 4.8 request shaping

Fixes https://github.com/veryfront/veryfront-studio/issues/4856

## TDD evidence
- Red: `deno test extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` failed with `Actual 0 / Expected undefined` for `body.temperature`.
- Green: `deno test extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` passed: 1 file, 5 steps, 0 failed.

## Verification
- `deno fmt --check extensions/ext-llm-anthropic/src/anthropic-request-builder.ts extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` passed.
- Focused test passed: `deno test extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts`.
- Broad pre-push ran format/lint/typecheck, then failed in unrelated unit tests: `src/observability/metrics/config.test.ts` expected `console` but got `otlp`. No failures were in the touched Anthropic request-builder test.

### DIFF
--- a/extensions/ext-llm-anthropic/src/anthropic-request-builder.ts
+++ b/extensions/ext-llm-anthropic/src/anthropic-request-builder.ts
@@ -474,6 +474,22 @@ function resolveAnthropicProviderThinkingBudget(
   return undefined;
 }
 
+function usesDeprecatedSamplingParams(modelId: string): boolean {
+  return /claude-opus-4-[7-9]/.test(modelId);
+}
+
+function stripAnthropicSamplingParams(
+  options: Record<string, unknown>,
+): Record<string, unknown> {
+  const sanitized = { ...options };
+  delete sanitized.temperature;
+  delete sanitized.top_p;
+  delete sanitized.topP;
+  delete sanitized.top_k;
+  delete sanitized.topK;
+  return sanitized;
+}
+
 export function buildAnthropicMessagesRequest(
   modelId: string,
   providerName: string,
@@ -499,6 +515,10 @@ export function buildAnthropicMessagesRequest(
   const providerThinkingBudget = resolveAnthropicProviderThinkingBudget(rawProviderOptions);
   const effectiveThinkingBudget = thinkingBudget ?? providerThinkingBudget;
   const thinkingEnabled = effectiveThinkingBudget !== undefined;
+  const samplingParamsDeprecated = usesDeprecatedSamplingParams(modelId);
+  const providerOptions = samplingParamsDeprecated
+    ? stripAnthropicSamplingParams(rawProviderOptions)
+    : rawProviderOptions;
 
   if (options.presencePenalty !== undefined) {
     warnings.push({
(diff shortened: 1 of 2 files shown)


extensions/ext-llm-anthropic/src/anthropic-request-builder.ts
```diff
@@ -474,6 +474,22 @@ function resolveAnthropicProviderThinkingBudget(
   return undefined;
 }
 
+function usesDeprecatedSamplingParams(modelId: string): boolean {
+  return /claude-opus-4-[7-9]/.test(modelId);
```
ariskemper, line 478: `claude-fable-5` rejects the same sampling params (HTTP 400) but is not matched here, and a future `claude-opus-4-10` slips past `[7-9]`. Broaden to cover both:

```suggestion
  // Claude Opus 4.7+ and Fable 5 reject temperature/top_p/top_k (HTTP 400);
  // Opus 4.6 and earlier still accept them. The `\d{2,}` branch covers future
  // double-digit Opus minors (e.g. claude-opus-4-10).
  return /claude-opus-4-(?:[7-9]|\d{2,})/.test(modelId) ||
    modelId.includes("claude-fable-");
```

ariskemper (commented):
**`deep-review` (checked against the Anthropic API reference)**

Premise confirmed first: on Opus 4.7/4.8, `temperature`, `top_p`, and `top_k` are removed and return HTTP 400, so stripping them is correct. The change does that properly for `claude-opus-4-7/8/9`. Two notes:

**1. Detection misses Fable 5 and future Opus minors (inline suggestion below).** `claude-fable-5` rejects the same sampling params (HTTP 400), but `/claude-opus-4-[7-9]/` does not match it, so a Fable 5 request with `temperature` set would still 400. A future `claude-opus-4-10` would also slip past `[7-9]`. Worth pairing the fix with tests: Fable 5 strips temperature/top_p plus provider `top_k`, and Opus 4.6 keeps them (guards against regex over-match).

**2. Adjacent, out of this diff's scope: Opus 4.8 `max_tokens` is capped at 32K.** `getAnthropicModelCapabilities` only knows `opus-4-6/4-5/4-1`, so `claude-opus-4-8` falls through to the generic `claude-opus-4-` branch and gets `maxOutputTokens: 32_000`. The real ceiling for Opus 4.7/4.8 (and Fable 5) is 128K. Not part of this change, but since this PR targets Opus 4.8 it is worth a follow-up adding `opus-4-7/4-8` and `fable-5` to that registry.

Otherwise the change is clean: no simplify or security findings.

kojiwakayama:
Closing as superseded by `8a3b45a1f` (`Prevent Opus runs from sending rejected sampling parameters`), which is already on `main` and released as `v0.1.756`.

That mainline fix handles the runtime path for `anthropic/claude-opus-4-7`, `anthropic/claude-opus-4-8`, and `veryfront-cloud/anthropic/claude-opus-4-8`, with provider transport coverage. This PR also has outstanding review feedback around `claude-fable-5` and future Opus minors, so it should not be merged as-is.

If more coverage is needed, the right follow-up is a fresh PR that covers the provider-extension request-builder layer and broadens the model capability rules deliberately.

PR #2358, opened 2026-06-11
Add directory-layout agents with owner-registered colocated capabilities

kojiwakayama:
## Summary

PR D of the agreed split (stacked on #2357; includes the #2355 helpers via cherry-pick until it merges). Directory-layout agents (`agents/{id}/AGENT.md`) ship their own `SKILL.md`, `skills/**`, and `tools/*.ts`.

**Discovery is pure registration** — the design change requested in the #2354 review: colocated capabilities register into the existing global registries namespaced **`{agentId}--{shortName}`** (the platform-decided separator: provider-safe, never confusable with `<provider>__<tool>` integration naming) with `ownerAgentId`/`shortName` metadata. No handler-resolved id lists, no adapter sentinel.

**Binding is uniform**: `skills:` / `tools:` (`true | [..]`) mean the same thing for TS, flat markdown, and directory markdown agents — all visible capabilities, or own-short-name-first then global id — resolved at invocation time by the owner-aware resolvers from #2357. `tools:` now actually binds for flat agents too (was a silent no-op in #2354), failing with an explicit unknown-tool diagnostic when unresolvable.

**Discovery diagnostics**: duplicate agent ids (flat + directory coexisting), sanitized-namespace collisions, owned-short-name vs global-id shadowing (plan test 15), invalid colocated tool names, `.`/`..` rejection.

**Capability isolation across delegation** (plan test 14): a coordinator's `skills: true` never includes its delegate's owned skills, and vice versa — tested on a real fixture project with `lead.md` (delegates) + `researcher/` (directory agent with own + nested skills).

## Known limitations (transparent)

- Hosted runtime: colocated skills resolve locally only until the catalog work lands (spike: framework adds owner/sourcePath to `RuntimeSkillDefinition`; control plane discovers `agents/*/SKILL.md`). The guides mark this explicitly.
- ~~Colocated `tools/*.ts` loading integration test~~ — now covered: `discoverAll()` loads a real colocated tool module through the esbuild transpiler, asserts owner-metadata registration (`researcher--fetch-paper`), owner execution, and not-found rejection for other agents (sanitizer baseline 420 → 422 for the two opt-outs).

## Testing

- `src/discovery/agent-scoped-capabilities.test.ts` — 7 tests on a real temp-dir fixture (side-by-side layouts, owner metadata, delegation isolation, duplicate-id, shadowing, namespace collision, path-segment guards)
- Full regression sweep green: `src/discovery src/skill src/tool src/mcp src/agent/runtime`

## Type of Change
- [x] New feature (non-breaking)
- [x] Documentation update
- [x] Test update

### DIFF
--- a/docs/guides/agents.md
+++ b/docs/guides/agents.md
@@ -55,6 +55,54 @@ The file path provides the agent id. For example, `agents/support.md` registers
 `support` and can be invoked through the same project runtime and control-plane
 surfaces as `agents/support.ts`.
 
+## Per-agent skills and tools
+
+A markdown agent can own its skills and tools by using a directory instead of a
+single file. Put the agent definition in `AGENT.md` and colocate its
+capabilities beside it:
+
+```
+agents/
+  researcher/
+    AGENT.md            # the agent definition (frontmatter + instructions)
+    SKILL.md            # the agent's own skill, loaded as load_skill("researcher")
+    skills/
+      cite/SKILL.md     # an extra skill, loaded as load_skill("researcher--cite")
+    tools/
+      fetch-paper.ts    # a colocated tool, registered as "researcher--fetch-paper"
+```
+
+The directory name is the agent id. The flat `agents/{id}.md` form still works
+for agents that do not own skills or tools, and both layouts can coexist.
+
+Colocated capabilities are registered with owner metadata and namespaced
+`{agentId}--{name}`. Ownership controls visibility everywhere: an agent only
+ever sees unowned (project-global) capabilities plus its own — never another
+agent's. This one rule applies to `skills:` and `tools:` for every agent kind
+(TypeScript, flat markdown, and directory markdown):
+
+```md
+---
+name: Researcher
+model: anthropic/claude-sonnet-4-6
+skills: true # all skills visible to this agent (global + own)
+tools: [fetch-paper] # own short names resolve first, then global tool ids
+---
+
+Research the question and cite every claim.
+```
+
+- `skills: true` / `tools: true` — every capability visible to the agent.
+- `skills: [..]` / `tools: [..]` — each entry resolves as the agent's own
+  short name first, then as a global id. A colocated short name that shadows a
+  global id is reported at discovery so the reference stays unambiguous.
+- Duplicate agent ids (flat file + directory) and agent ids whose sanitized
+  namespaces collide are reported as discovery errors.
+
+> Note: colocated capabilities currently resolve in the local runtime; hosted
+> runtime catalog support is tracked separately and lands before this layout
+> is documented as stable.
+
 ## Add tools
 
 Agents call tools to take actions or fetch data. Reference tools by name: the
--- a/docs/guides/multi-agent.md
+++ b/docs/guides/multi-agent.md
@@ -99,6 +99,29 @@ const researcher = getAgent("researcher");
 const researchTool = agentAsTool(researcher, "Research a topic using web search");
 ```
 
+## Declarative delegation with `delegates`
+
+A markdown agent can opt into orchestration by listing the specialists it may
+call in its `delegates` frontmatter. The runtime gives the agent one
+`agent_{id}` tool per delegate; each delegate runs with its own settings,
+skills, and tools — capability ownership does not cross the delegation
+boundary in either direction.
+
+```md
+---
+name: Lead
+description: Plans the work and routes to specialists
+delegates: [researcher, writer]
+---
+
+Break the task down. Use agent_researcher to gather facts, then agent_writer to
+produce the final copy.
+```
+
+With several agents and no `delegates`, the agents are independent: a caller
+selects one by id. Self-delegation and delegate ids that cannot form a valid
+provider tool name are rejected at discovery with explicit diagnostics.
+
 ## Workflow-based composition
 
 For deterministic multi-agent pipelines, use [workflows](./workflows.md):
--- a/scripts/lint/check-sanitizer-baseline.ts
+++ b/scripts/lint/check-sanitizer-baseline.ts
@@ -22,7 +22,7 @@ const SCAN_ROOTS = [
 
 // Lower this when you remove sanitizer opt-outs. Never raise it without a very
 // good reason — a new opt-out means a leak is being suppressed rather than fixed.
-export const SANITIZER_OPT_OUT_BASELINE = 420;
+export const SANITIZER_OPT_OUT_BASELINE = 422;
 
 const OPT_OUT_PATTERN = /sanitize(?:Resources|Ops|Exit)\s*:\s*false/g;
 
--- a/scripts/lint/check-sanitizer-baseline.ts
+++ b/scripts/lint/check-sanitizer-baseline.ts
@@ -22,6 +22,10 @@ const SCAN_ROOTS = [
 
 // Lower this when you remove sanitizer opt-outs. Never raise it without a very
 // good reason — a new opt-out means a leak is being suppressed rather than fixed.
+// 422 = 420 (historical) + 2 for the colocated tool-loading discovery test
+// (src/discovery/agent-scoped-capabilities.test.ts): importModule transpiles
+// via esbuild, whose warm child process trips the op/resource sanitizers —
+// same rationale as src/discovery/transpiler.test.ts.
 export const SANITIZER_OPT_OUT_BASELINE = 422;
 
 const OPT_OUT_PATTERN = /sanitize(?:Resources|Ops|Exit)\s*:\s*false/g;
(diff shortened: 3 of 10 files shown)


kojiwakayama:
Confirmed and fixed (branch rebased on the updated #2357).

**`discoverAll()` wiping colocated skills**: the skill-registry clear + global skill discovery now run **before** agent discovery in `discovery-engine.ts`, so directory-agent colocated skills (registered during agent discovery) survive the clear. This also makes the owned-short-name vs global-id shadow diagnostic order-correct in the real pipeline — previously it only held in direct handler tests, as you noted. Colocated skills additionally surface in `result.skills`.

New full-pipeline regression test runs your exact fixture shape through `discoverAll({ baseDir })`: directory agent with own `SKILL.md` + `skills/cite/`, plus a global `skills/cite/` that shadows the short name — asserts registry retention, `result.skills` presence, per-agent visibility (`researcher` sees `cite, researcher, researcher--cite`; `lead` sees only `cite`), and the shadow diagnostic firing through natural ordering.

Full regression sweep: 219 passed, 0 failed across discovery/skill/tool/agent-runtime.

kojiwakayama:
Added the end-to-end colocated tool-loading test flagged as a follow-up: `discoverAll()` on a fixture with `agents/researcher/tools/fetch-paper.ts` — loads through the esbuild-backed `importModule`, registers as `researcher--fetch-paper` with owner metadata, executes for the owner via the registry gate, and rejects other agents as not found. Sanitizer opt-out baseline ratcheted 420 → 422 (documented in the test, same rationale as `transpiler.test.ts`). Discovery suite 9/9; baseline check exact. No known follow-ups remain on this PR.

ariskemper:
Ran a structured review pass (quality / correctness / security) over this slice. Two things here are clear improvements over the #2354 staging branch and worth saying plainly: **discovery-as-pure-registration** (owner metadata + `--` separator instead of handler-resolved id lists threaded through the adapter) is the cleaner design, and the catch that **flat-agent `tools:` was a silent no-op in #2354 is correct** — uniform `skills:`/`tools:` semantics across TS, flat, and directory agents with an explicit unknown-tool diagnostic is the right contract. Also verified as solid: `.`/`..` path-segment rejection on every joined name, per-skill error granularity (one bad SKILL.md doesn't sink the agent), sanitize-collision detection, and moving `skillRegistry.clear()` ahead of agent discovery so colocated skills survive re-discovery (the comment explaining why is appreciated).

Findings:

1. **Duplicate colocated tool short-names are silently first-wins** (`agent-scoped-capabilities.ts:164-166`): `if (toolRegistry.get(namespaced)) { continue; }` drops the second registration with no entry in `result.errors`. The neighboring invalid-name path (lines 154-162) reports; this path should too — otherwise a user whose tool didn't bind gets a runtime "unknown tool" with no discovery-time breadcrumb. (Nice touch that `reportShadowedGlobalCapability` exists for the global-shadowing case — this is the same idea one branch earlier.)

2. **Sanitizer baseline raised without an in-code why** (`check-sanitizer-baseline.ts:25`, 420 → 422): the constant's own comment says "never raise without a very good reason" — the reason here is legitimate (importModule/esbuild keeps a warm child process, same as `transpiler.test.ts`), but it should be written next to the number so the next reader doesn't have to re-derive it.

3. **Coordination heads-up**: #2354 also edits `SANITIZER_OPT_OUT_BASELINE` (to 428) and the same two guides — whichever lands second gets conflicts in `check-sanitizer-baseline.ts`, `docs/guides/agents.md`, and `docs/guides/multi-agent.md`. Given this stack supersedes the staging branch, probably worth deciding explicitly which one carries the docs.

4. **Docs claim check** (passed): I verified the guide statements against the diff — selectors (`true | [..]`), both layouts, own-skill-by-agent-id, and the isolation claim all match the implemented behavior. One suggestion: the guides should mention the `{agentId}--{shortName}` id form, since users will see those ids in `load_skill` errors and tool diagnostics.

For transparency: I also checked two findings that did *not* survive verification — `ownerAgentId`/`shortName` are properly declared on the stack's `Tool` interface (`src/tool/types.ts:162-165`, from #2357), and re-discovery does clear owned tools (`clearProjectAgentRuntimeRegistries` → `clearMCPRegistry()` → `toolRegistry.clear()`). No issues there.

kojiwakayama:
Re-reviewed current head (`093d2067`). The full-pipeline skill retention fix and the esbuild-backed colocated tool-loading test are good. The docs suggestion from the latest review is already covered in `docs/guides/agents.md` (`researcher--fetch-paper` and `{agentId}--{name}` are documented).

Remaining blocker:

1. `src/discovery/agent-scoped-capabilities.ts:164-166` still silently skips duplicate colocated tool short names when the namespaced id already exists. That leaves users with a later runtime "unknown tool"/missing binding and no discovery-time breadcrumb. This should push a `result.errors` entry, matching the neighboring invalid-name and shadowing diagnostics, and should have a regression test with two colocated tools resolving to the same `{agentId}--{shortName}`.

Small hygiene fix:

2. `scripts/lint/check-sanitizer-baseline.ts:23-25` still raises the ratchet to `422` without the reason next to the constant. The test has the esbuild rationale, which is useful, but the baseline file itself says not to raise without a good reason; put the short rationale there so future maintainers do not have to chase the test/PR comment.

Also, this PR remains stacked on #2357, so it inherits the owner-scope listing blockers there.

Verdict: request changes for the duplicate-tool diagnostic, plus the baseline comment before merge.

kojiwakayama:
Both items addressed (branch rebased onto the updated #2357; the #2355 cherry-pick dropped now that it's merged).

1. **Duplicate colocated tool short names**: within-pass duplicates now push a `result.errors` entry (keeping the first), matching the neighboring invalid-name and shadowing diagnostics. One deliberate distinction: a pre-existing registry entry from a *previous* discovery pass is treated as a normal re-discovery refresh and overwritten — otherwise every HMR re-run would false-positive the diagnostic (and the old skip-if-exists behavior actually left stale tools on HMR; that's fixed as a side effect). Regression: the esbuild fixture now ships a second module exporting the same `fetch-paper` id — asserts exactly one duplicate error, first registration wins (payload-verified), and no other errors. Folded into the existing opt-out test so the sanitizer baseline stays exactly 422.

2. **Baseline rationale**: documented next to the `SANITIZER_OPT_OUT_BASELINE` constant (422 = 420 historical + 2 for the esbuild-backed discovery test, same rationale as `transpiler.test.ts`).

The inherited #2357 listing blockers are fixed there (see its thread).

kojiwakayama:
Re-reviewed current head (`9e9f88de0`) against the previous D-specific findings.

D-specific items look good now:

- Duplicate colocated tool short names are reported during the same discovery pass, while re-discovery/HMR can overwrite stale previous-pass entries. That distinction is right: it catches authoring conflicts without making refreshes noisy.
- The colocated esbuild tool-loading test now verifies first-registration-wins by payload, owner metadata, owner execution, and non-owner rejection.
- The sanitizer baseline rationale is now documented next to the ratchet constant, and the baseline check is exact at `422/422`.
- The merged-A cherry-pick is gone; the stack is now `main -> #2357 -> #2358`.

Focused verification I ran locally:

- `deno test --no-check --allow-all src/discovery/agent-scoped-capabilities.test.ts` -> 9 passed
- `deno run --allow-read scripts/lint/check-sanitizer-baseline.ts` -> `Sanitizer opt-out baseline ok: 422/422.`
- `git diff --check origin/pr-2357...origin/pr-2358` -> clean

No new D-specific blocker found. This PR still inherits the #2357 owner-scope blockers I just posted there, so I would keep D waiting on C.

kojiwakayama:
Rebased onto the updated #2357 (round-three owner-scope fixes: control-plane agent metadata, AG-UI `tools: true`, dev dashboard listing — see its thread). No D-side changes in this push; the inherited blocker is addressed upstream.

kojiwakayama:
Re-reviewed current head (`9b9fddb72`), not the stale `9e9f88de0` head.

This is a clean rebase onto the updated #2357; I do not see D-side changes beyond the base update. The D-specific findings remain addressed:

- duplicate colocated tool short names report a discovery error within the same pass;
- colocated tool loading is covered through the esbuild-backed discovery path;
- sanitizer baseline rationale is documented at the ratchet constant and checks exact at `422/422`;
- docs and selector behavior still match the implemented `skills:` / `tools:` rules.

Focused verification I ran locally:

- `deno test --no-check --allow-all src/discovery/agent-scoped-capabilities.test.ts` -> 9 passed
- `deno run --allow-read scripts/lint/check-sanitizer-baseline.ts` -> `Sanitizer opt-out baseline ok: 422/422.`
- `git diff --check origin/pr-2357...origin/pr-2358` -> clean

Review verdict: no D-specific blocker from me. This remains stacked on #2357, so merge readiness follows #2357 plus its checks.

PR #2374, opened 2026-06-12
Use full-project candidates for production page CSS

### LINKED ISSUE #2373: Production page CSS diverges across replicas: route-scoped candidates drop shared component classes
## Symptom

A production site on staging intermittently renders with broken styling (collapsed header spacing, unstyled buttons). Repeated fetches of the same page return different `/_vf/css/<hash>.css` bundles depending on which replica answers — observed four distinct bundles for one page ranging 26,223B to 38,809B. With a sticky `lb` cookie the hash is stable, confirming per-replica divergence. Layout utilities used by shared components (`h-16`, `pr-6`, `md:` variants) were missing from the served CSS.

## Root cause

`extractProjectClassesForRoute` (src/rendering/orchestrator/html-project-css.ts) resolved Tailwind candidates route-scoped via `getRouteCandidates`, whose file list comes from the **runtime route-module manifest** (src/modules/manifest/route-module-manifest.ts). That manifest is an LRU populated per replica from request history — documented as "a regenerable preload optimization" — but it was used as a correctness input for CSS compilation:

1. A cold replica renders a route before the manifest has observed shared components (header/nav imported by the layout) → their classes never become candidates → compiled CSS lacks their rules.
2. The incomplete result is cached per route in `routeCandidateCache` and never recovers in production (`shouldRebuildManifest` returns false outside dev).
3. The full-project fallback only triggers when route candidates are **exactly empty**, so partially-known routes stay broken.
4. The prefetched CSS path (`startProjectCSSPreparation`) compiles from `projectClasses` only, skipping the rendered-content union the non-prefetch path performs — so completeness also depended on which code path won.

Net effect: CSS is a function of per-replica request history instead of project source.

## Fix

Derive production page CSS candidates from the deterministic full-project source scan (`getProjectCandidates`), as dev mode (`styles-candidate-scanner`) and CSS pregeneration already do. Same source version → same candidate set → same CSS hash on every replica.

## Follow-ups (separate issues if accepted)

- Client-injected `<style>` elements during hydration carry no CSP nonce and are blocked by the default `style-src-elem` policy (router.ts SPA path sets the nonce; other injectors do not).
- `config.csp` replaces the entire default CSP instead of merging; the docstring in security-handler.ts recommends `security.csp.frameSrc`, which silently drops all other directives. An additive `cspExtend` would remove the footgun.
- Route-scoped CSS can be reintroduced later behind a static import graph instead of the runtime module manifest.

kwakayama:
Fixes #2373

## Problem

Production page CSS candidates were resolved route-scoped from the runtime route-module manifest — an LRU populated per replica from request history. Cold replicas compiled CSS without shared-component classes (header, nav, buttons), cached the incomplete result per route, and never recovered. Replicas served divergent stylesheets for the same page (four distinct bundles observed, 26KB–38KB); pages rendered with missing layout rules depending on which replica answered.

## Change

`extractProjectClassesForRoute` now derives candidates from the deterministic full-project source scan (`getProjectCandidates`), matching dev mode and CSS pregeneration. CSS becomes a pure function of project source + version: same hash, identical styling on every replica.

- `src/rendering/orchestrator/html-project-css.ts`: swap `getRouteCandidates` → `getProjectCandidates`; drop route-key plumbing from the candidate call.
- `src/rendering/orchestrator/html-project-css.test.ts`: regression test — classes from a component file outside the route module graph must be present (fails on main); update delegation test; fix two pre-existing `deno check` errors.
- Version bump 0.1.762 → 0.1.763.

## Trade-off

Page CSS now includes all project classes rather than a route-scoped subset — standard Tailwind semantics (one stylesheet per project). The route-scoped optimization can return later behind a static import graph, which would be deterministic.

## Tested

- `deno test --no-check --allow-all src/rendering/orchestrator/` — 22 passed (279 steps)
- `deno test --no-check --allow-all src/html/styles-builder/` — all passing
- `deno check src/rendering/orchestrator/html-project-css.ts src/rendering/orchestrator/html-project-css.test.ts` — clean

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.762",
+  "version": "0.1.763",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": "P2D",
--- a/src/rendering/orchestrator/html-project-css.test.ts
+++ b/src/rendering/orchestrator/html-project-css.test.ts
@@ -107,7 +107,8 @@ describe("rendering/orchestrator/html-project-css", () => {
                     },
                     {
                       path: "/project/components/header.tsx",
-                      content: `export const Header = () => <header className="h-16 md:pr-8">Nav</header>;`,
+                      content:
+                        `export const Header = () => <header className="h-16 md:pr-8">Nav</header>;`,
                     },
                   ]),
               }),
--- a/src/rendering/orchestrator/html-project-css.test.ts
+++ b/src/rendering/orchestrator/html-project-css.test.ts
@@ -28,6 +28,7 @@ describe("rendering/orchestrator/html-project-css", () => {
   describe("getProjectContentVersion", () => {
     it("prefers the adapter content context version", () => {
       const version = getProjectContentVersion({
+        mode: "production",
         adapter: {
           fs: {
             getUnderlyingAdapter: () => ({
@@ -47,6 +48,7 @@ describe("rendering/orchestrator/html-project-css", () => {
 
     it("falls back to project updated_at when no content context is available", () => {
       const version = getProjectContentVersion({
+        mode: "production",
         adapter: {
           fs: {
             getUnderlyingAdapter: () => ({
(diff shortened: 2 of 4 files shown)


kwakayama:
Review score: **92/100**

Verdict: **approve-level / no blocking findings found**.

What I checked:
- Compared PR #2374 against issue #2373 and the `origin/main...pr-2374-review` diff.
- Reviewed the production CSS call path from `HTMLGenerator.buildHTMLOptions` -> `extractProjectClassesForRoute` -> `getProjectCandidates` -> `generateHTMLShellPartsImpl` / `getProjectCSS`.
- Checked the candidate manifest cache keying: `projectScope + projectVersion + styleProfile.hash`, with development TTL behavior still isolated in `getOrBuildManifest`.
- Confirmed the PR removes the route-module manifest as a correctness input for production page CSS, which directly addresses the per-replica divergence described in #2373.

Verification run from an isolated worktree at `/tmp/vf-code-pr-2374`:
- `deno test --no-check --allow-all src/rendering/orchestrator/` -> **22 passed / 279 steps**
- `deno test --no-check --allow-all src/html/styles-builder/` -> **14 passed / 205 steps**
- `deno check src/rendering/orchestrator/html-project-css.ts src/rendering/orchestrator/html-project-css.test.ts` -> **clean**

Notes / residual risk:
- This intentionally trades route-scoped CSS size for deterministic full-project CSS. That is the right fix for this incident because the previous route graph was request-history-derived, not source-derived.
- `getRouteCandidates` and `buildRouteManifestKey` remain as tested code but are no longer used by this production extraction path. That is acceptable if route-scoped CSS may return later behind a deterministic static graph; otherwise it can be cleaned up separately.
- Independent sub-agent review was unavailable in this Codex surface because sub-agent spawning is only permitted when explicitly requested. This review is therefore a single-lane local review plus local verification.

No requested changes from me.

PR #2378, opened 2026-06-12
Fix release asset manifest consumption for environment contexts

kojiwakayama:
## Summary
- register release asset manifest fetchers for any content context carrying a releaseId, including production environment/domain contexts
- add a regression proving environment contexts with release ids can fetch/cache ready manifests
- bump Veryfront Code package metadata from 0.1.765 to 0.1.766 so this ships as a new release

## Why
Staging proved manifest build and hashed asset serving were healthy, but authenticated production HTML still emitted /_vf_modules URLs. The missing link was renderer-side consumption: environment-derived contexts can carry the deployed releaseId but were not registering the per-release manifest fetcher.

## Verification
- deno test --no-check --allow-all src/platform/adapters/fs/veryfront/adapter.test.ts
- deno test --no-check --allow-all src/release-assets/html-consumption.test.ts
- deno test --no-check --allow-all src/release-assets
- deno test --no-check --allow-all src/utils/version.test.ts src/utils/logger/logger.test.ts
- deno fmt --check deno.json src/utils/version-constant.ts
- pre-push format/lint/typecheck/full unit suite was attempted after the version bump; version failures were fixed, then the rerun hung in unrelated Login Module full-suite execution. Isolated cli/auth/login.test.ts passed in 713ms.

## Downstream
- veryfront-api should bump its pinned veryfront dependency from 0.1.764 to 0.1.766 after this package version is published.
- veryfront-studio does not need a bump for this fix; its only direct veryfront usage found was Vite stubbing for bundled internals, not the renderer/task runtime path.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.765",
+  "version": "0.1.766",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": "P2D",
--- a/src/platform/adapters/fs/veryfront/adapter.test.ts
+++ b/src/platform/adapters/fs/veryfront/adapter.test.ts
@@ -1,10 +1,16 @@
 import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals, assertExists } from "#veryfront/testing/assert.ts";
-import { describe, it } from "#veryfront/testing/bdd.ts";
+import { afterEach, describe, it } from "#veryfront/testing/bdd.ts";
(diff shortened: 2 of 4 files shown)


kojiwakayama:
Checked the Codex review note on build-executor.ts. After the conflict-fix rebase, this PR no longer changes build-executor.ts; the current diff is limited to adapter manifest-fetcher wiring, its regression test, and the 0.1.766 version bump. All checks on this PR are green, so I’m treating that old review note as stale/non-blocking for #2378. The build-executor route-closure concern can be handled separately if we want to harden that path.

PR #2426, opened 2026-06-14
Preserve snake_case tool-role replay results

kojiwakayama:
## Summary
- Preserve stored snake_case `tool_result` parts when they are already split into `role: "tool"` replay messages.
- Reuse the existing tool-result normalizer so both `result` and `output`, camelCase and snake_case fields, reach the text-generation runtime.
- Add a regression test for API-style `role: "tool"` + `type: "tool_result"` + `output` history.

## Root cause
API replay normalization can split persisted assistant inline tool results into separate tool-role messages with snake_case parts. `convertToTextGenerationRuntimeMessage()` only handled hyphen-case `tool-result` parts in the `role: "tool"` branch, so these nearby tool results were converted to empty content and then dropped before the next model call.

## Verification
- `deno test src/agent/runtime/text-generation-runtime-message-converter.test.ts src/agent/runtime/message-adapter.test.ts`
- `deno test --allow-env src/agent/runtime/agent-invocation-contract.test.ts`
- pre-push hook: format, lint, type check, full unit tests (`2141 passed`)

### DIFF
--- a/src/agent/runtime/text-generation-runtime-message-converter.test.ts
+++ b/src/agent/runtime/text-generation-runtime-message-converter.test.ts
@@ -212,6 +212,33 @@ describe("text-generation-runtime-message-converter", () => {
       });
     });
 
+    it("converts a stored snake_case tool result message", () => {
+      const msg = {
+        id: "t-snake",
+        role: "tool",
+        parts: [
+          {
+            type: "tool_result",
+            tool_call_id: "tc-snake",
+            tool_name: "harvest__list_users",
+            output: { users: [{ id: 1, name: "Ada" }] },
+          },
+        ],
+      } as unknown as Message;
+
+      const result = convertToTextGenerationRuntimeMessage(msg);
+
+      assertEquals(result.role, "tool");
+      assertEquals((result as TextGenerationRuntimeToolMessage).content, [
+        {
+          type: "tool-result",
+          toolCallId: "tc-snake",
+          toolName: "harvest__list_users",
+          output: { type: "json", value: { users: [{ id: 1, name: "Ada" }] } },
+        },
+      ]);
+    });
+
     it("handles tool result with missing toolName", () => {
       const msg: Message = {
         id: "t2",
--- a/src/agent/runtime/text-generation-runtime-message-converter.ts
+++ b/src/agent/runtime/text-generation-runtime-message-converter.ts
@@ -17,13 +17,7 @@ import type {
   TextGenerationRuntimeToolResultPart,
 } from "./text-generation-runtime-message-types.ts";
 import { buildDataFileAnnotation } from "#veryfront/chat/types.ts";
-import {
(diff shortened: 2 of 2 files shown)


kwakayama:
## Critical review — Score: **84 / 100**

Solid, well-scoped bugfix. The root-cause writeup matches the code: the `case "tool"` branch of `convertToTextGenerationRuntimeMessage` was the last place still doing its own inline tool-result extraction (`part.type !== "tool-result"` + `resultPart.result` only), while its sibling `convertAssistantMessageToTextGenerationRuntimeMessages` already routed through `getTextGenerationToolResultPart`. Snake_case `tool_result` parts with `output` therefore fell through to empty content and got dropped at the `content.length === 0` guard. Unifying both branches on the shared helper is the right fix and removes a real divergence rather than papering over it. Removing the now-unused `ToolResultPart` import is a clean touch.

### What holds up
- **Correctness:** `getTextGenerationToolResultPart` covers both casings for type, id, name, and value (`result`/`output`), so the previously-dropped parts now reach the runtime. ✔
- **DRY / consistency:** the tool-role branch and the assistant splitter now share one normalizer — exactly where the bug came from. ✔
- **Verification:** targeted converter/adapter tests + full pre-push suite (2141 passed) is appropriate for a change at this layer. ✔

### Where I'd push back (the 16 points)
1. **Test doesn't reproduce the reported scenario.** The new test exercises `convertToTextGenerationRuntimeMessage` on a *standalone* tool message. But the root cause is about replay *splitting* an assistant inline result into a separate `role:"tool"` message. A stronger regression test would drive `convertToTextGenerationRuntimeMessages` with a realistic split history (assistant `tool_call` → separate snake_case tool message) and assert the result survives, matches the call id, and merges with the prior tool message (the line ~474 merge path). As written, the test locks the unit but not the end-to-end behavior the PR claims to fix.
2. **`toolNamesById` is dead weight in this branch.** It's created empty and never populated (a tool-role message has no tool-call parts), so `toolNamesById.get(toolCallId)` always misses and `toolName` falls back to `"unknown"` whenever the stored part lacks `tool_name`. Harmless for Anthropic (matching is by id), but worth a comment, or pass a shared empty map — otherwise it reads as if cross-message name resolution works here when it doesn't.
3. **Subtle side-effect broadening.** `shouldSkipProviderExecutedToolResult` now runs on *every* part, not just `tool-result` parts, and it mutates `providerExecutedToolCallIds` (`.delete`). For well-formed tool messages this is inert, but it's a behavioral surface change that the test doesn't cover. A one-line note or a guard would make the intent explicit.
4. **Minor:** missing-value case now yields `value: null` instead of the old `undefined`. Almost certainly fine/better, but it's an unflagged serialization change.

None of these are blockers — the fix is correct and minimal. Strengthening the test to the actual split-replay path would be the one change I'd want before merge.

PR #2427, opened 2026-06-14
Hydrate skill state from hosted tool history

kwakayama:
## Summary
- Hydrate active skill state from hosted inline `tool_call` / `tool_result` history.
- Resolve snake_case tool results back to their tool names via prior tool-call IDs.
- Add regression coverage for the production message shape from veryfront-studio#4925.

## Evidence
- Staging DB run `99c588fd-4712-41c7-8801-60ccb4a53f18` persisted all 55 tool results (`0` unmatched starts), but stored them inline with snake_case parts.
- The failing run looped: `load_skill_reference=19`, `load_skill=15`, `harvest__list_time_entries=9`, `get-current-date=7`, `harvest__list_users=5`.
- The persisted `tool_result` parts have `tool_call_id` + `output`, while the tool name lives on the preceding `tool_call` part, so replay hydration missed `load_skill` state before this fix.

## Tests
- `deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all src/agent/runtime/skill-policy.test.ts` (red before fix, green after)
- `deno task test src/agent/runtime/skill-policy.test.ts src/agent/runtime/text-generation-runtime-message-converter.test.ts src/agent/runtime/message-adapter.test.ts src/agent/runtime/load-skill-tool.test.ts`

## Pre-push
- Full pre-push formatting/lint/typecheck passed.
- Full unit pre-push hit unrelated existing failures:
  - `src/observability/metrics/config.test.ts`: expected exporter `console`, actual `otlp`
  - `src/server/build-app-route-renderer.test.ts`: expected `data-testid="app-layout"` in rendered HTML

### DIFF
--- a/src/agent/runtime/skill-policy-enforcement.ts
+++ b/src/agent/runtime/skill-policy-enforcement.ts
@@ -15,6 +15,35 @@ const logger = serverLogger.component("agent");
 export const LOAD_SKILL_TOOL_ID = "load_skill";
 export const FORM_INPUT_TOOL_ID = "form_input";
 
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return Boolean(value) && typeof value === "object" && !Array.isArray(value);
+}
+
+function getStringField(value: unknown, key: string): string | undefined {
+  if (!isRecord(value)) return undefined;
+  const field = value[key];
+  return typeof field === "string" && field.length > 0 ? field : undefined;
+}
+
+function getToolCallId(part: unknown): string | undefined {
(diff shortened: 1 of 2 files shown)


kwakayama:
Closing as superseded by #2428. #2428 fixes the incident path earlier in AG-UI normalization by inferring missing tool-result names from preceding tool calls, including the hosted inline snake_case shape observed for veryfront-studio#4925. The runtime hydration hardening here is therefore not needed for the current fix.

PR #2428, opened 2026-06-14
Infer AG-UI tool result names

kojiwakayama:
## Summary
- infer AG-UI replay tool-result names from preceding tool calls when stored results only carry call ids
- apply the same inference to AG-UI runtime role:tool messages and channel conversation-history replay
- record names from accepted AG-UI tool-prefixed call parts so later result-only records can resolve by call id
- bump veryfront release version to 0.1.801

## Test Plan
- deno test --no-check --allow-all src/agent/ag-ui/host-support.test.ts
- deno test --no-check --allow-all src/agent/ag-ui/host-support.test.ts src/agent/ag-ui/runtime-support.test.ts src/channels/invoke.test.ts src/agent/runtime/text-generation-runtime-message-converter.test.ts src/agent/runtime/message-adapter.test.ts src/agent/hosted/chat-preparation.test.ts src/agent/runtime/agent-invocation-contract.test.ts
- deno fmt --check src/agent/ag-ui/host-support.ts src/agent/ag-ui/host-support.test.ts src/agent/ag-ui/runtime-support.ts src/agent/ag-ui/runtime-support.test.ts src/channels/invoke.ts src/channels/invoke.test.ts deno.json src/utils/version-constant.ts
- deno check src/index.ts
- deno test --no-check --allow-all cli/auth/callback-server.test.ts after one unrelated pre-push flake
- git push pre-push hook: format, lint, typecheck, full unit tests: 2142 passed (18529 steps), 0 failed

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.800",
+  "version": "0.1.801",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": {
--- a/src/agent/ag-ui/host-support.test.ts
+++ b/src/agent/ag-ui/host-support.test.ts
@@ -183,6 +183,190 @@ describe("agent/ag-ui-host-support", () => {
     ]);
   });
 
+  it("infers stored snake_case tool-result names from preceding tool calls", () => {
+    const messages = normalizeAgUiMessages([
+      {
+        id: "assistant-1",
+        role: "assistant",
+        parts: [
+          {
+            type: "tool_call",
+            id: "tool-1",
+            name: "harvest__list_users",
+            input: { accountId: "2029314" },
+            state: "completed",
+          },
+          {
+            type: "tool_result",
+            tool_call_id: "tool-1",
+            output: { users: [{ id: 1, name: "Ada" }] },
+          },
+        ],
+      },
+    ]);
+
+    assertEquals(messages, [
+      {
+        id: "assistant-1",
+        role: "assistant",
+        parts: [
+          {
+            type: "tool-call",
+            toolCallId: "tool-1",
+            toolName: "harvest__list_users",
+            args: { accountId: "2029314" },
+          },
+          {
+            type: "tool-result",
+            toolCallId: "tool-1",
+            toolName: "harvest__list_users",
+            result: { users: [{ id: 1, name: "Ada" }] },
+          },
+        ],
+      },
+    ]);
+  });
+
+  it("infers camelCase tool-result names from preceding tool calls", () => {
+    const messages = normalizeAgUiMessages([
+      {
+        id: "assistant-1",
+        role: "assistant",
+        parts: [
+          {
+            type: "tool-call",
+            toolCallId: "tool-1",
+            toolName: "search_docs",
+            args: { query: "docs" },
+          },
+          {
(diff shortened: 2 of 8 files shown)


kwakayama:
## Critical Review — Score: 86/100

Reviewed the diff and read the surrounding `host-support.ts` / `runtime-support.ts` logic. Ran both new test files locally — **2 passed (13 steps), 0 failed**.

### What's good
- **Correct, minimal, well-scoped fix.** Tracking a `toolNamesById` map and falling back to `?? "unknown"` preserves existing behavior for genuine orphans while recovering the real name when a preceding call exists. No correctness bugs found.
- **Map lifecycle is right.** Cleared on `user`/`system` boundaries (mirrors `providerOwnedToolCallIds`), fresh `Map` per `normalizeAgUiRuntimeMessages` invocation (no cross-request leakage), and persists across consecutive assistant/tool messages within a turn — which correctly handles the realistic replay case where the call and result live in *separate* messages.
- **Name is registered before the provider-owned filter** (host-support.ts:316 runs before `shouldKeepProviderVisibleToolPart`), so inference still works even when the originating call part is dropped. Subtle but correct.
- **Release hygiene is correct:** `deno.json` and `src/utils/version-constant.ts` bumped in lockstep (0.1.800 → 0.1.801), as the test enforces.
- Default `new Map()` param on `normalizeMessagePart` keeps it safe for any future caller.

### Gaps (why not higher)
1. **The camelCase `tool-result` branch is modified but untested.** host-support.ts:267-275 got the same inference change, yet only the `snake_case` (`tool_result`/`tool_call_id`) path has a test. Add a camelCase case so a regression there isn't silent.
2. **The headline scenario isn't directly tested.** The PR is about *replay*, where the call and result typically arrive in **separate messages**. Both new tests put call + result in one message. A cross-message test (assistant tool-call → standalone `role:"tool"` result) would actually exercise the map-persistence-across-messages behavior that is the real point.
3. **No negative/boundary coverage:** (a) result with no preceding call → still `"unknown"`; (b) a `user` message between call and result → name correctly *not* inferred after the clear. These lock in the intended scoping.
4. **Minor design gap:** host-support.ts:316 only registers names for `type === "tool-call"`. Generic `tool-*` parts (lines 241-254) also carry `(toolCallId, toolName)` but don't seed the map. Probably intentional (only real calls establish the binding), but worth a one-line comment to make that explicit.

### Verdict
Safe to merge as-is — the logic is sound and tests are green. I'd ask for the camelCase + cross-message tests before landing (or as a fast follow), since the most realistic replay path and one of the two touched branches are currently unverified.

kojiwakayama:
Addressed the actionable review items in commit 1c2d0e619:

- Added camelCase `tool-result` inference coverage.
- Added cross-message replay coverage: assistant call followed by standalone `role: "tool"` result.
- Added boundary coverage for orphan results staying `unknown` and user-boundary clearing.
- Fixed the inline `tool-*` comment: accepted tool-prefixed call parts now seed the call-id/name map, and the regression covers a later result-only record resolving to the original tool name.

Verification:
- `deno test --no-check --allow-all src/agent/ag-ui/host-support.test.ts`
- focused AG-UI/channel/runtime/converter suite: 17 passed / 0 failed
- `deno fmt --check`
- `deno check src/index.ts`
- full pre-push hook: 2142 passed / 0 failed

PR #2429, opened 2026-06-14
Add AG-UI API contract fixture coverage

kojiwakayama:
## Summary
- Add a redacted API-shaped contract fixture covering `/messages`, `/events`, and `/snapshot` payload wrappers.
- Replay the fixture through the actual veryfront-code AG-UI host normalizer, channel history normalizer, and SSE decoder.
- Assert tool result names survive `/messages` and `MESSAGES_SNAPSHOT` replay even when `tool_result` parts omit `tool_name`.

## Verification
- `deno test --allow-all src/agent/ag-ui/api-contract-fixture.test.ts src/agent/ag-ui/host-support.test.ts src/channels/invoke.test.ts src/chat/ag-ui.test.ts`
- pre-push hook: `2143 passed (18530 steps), 0 failed`

### DIFF
--- a/src/agent/ag-ui/api-contract-fixture.test.ts
+++ b/src/agent/ag-ui/api-contract-fixture.test.ts
@@ -1,0 +1,291 @@
+import "#veryfront/schemas/_test-setup.ts";
+import { assertEquals } from "#veryfront/testing/assert.ts";
+import { describe, it } from "#veryfront/testing/bdd.ts";
+import { createAgUiChatEventDecoderState, decodeAgUiSseChunk } from "../../chat/ag-ui.ts";
+import { normalizeConversationHistoryForRuntime } from "../../channels/invoke.ts";
+import { normalizeAgUiMessages } from "./host-support.ts";
+
+const CONVERSATION_ID = "a7c53a3d-feb2-4404-86e2-5c562455e46c";
+const ASSISTANT_MESSAGE_ID = "33333333-3333-4333-a333-333333333333";
+const USER_TOOL_CALL_ID = "toolu_harvest_users";
+const TIME_TOOL_CALL_ID = "toolu_harvest_time_entries";
+
+const apiMessagesResponseFixture = {
+  conversation_id: CONVERSATION_ID,
+  data: [
+    {
+      id: "11111111-1111-4111-a111-111111111111",
+      role: "user",
+      parts: [
+        {
+          type: "text",
+          text: "Review Harvest timesheets for the last three weeks.",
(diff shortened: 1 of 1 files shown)


kwakayama:
## Review — Add AG-UI API contract fixture coverage

**Score: 82 / 100**

Good additive regression coverage with near-zero risk (one new test file, no production code touched, CI green). It exercises the genuinely important invariant: **tool_result `toolName` survives normalization even when the wire `tool_result` part omits `tool_name`** — the normalizer has to infer it from the matching `tool_call` id. That's a real, easy-to-regress behavior, and the test replays it through three actual normalizers (`normalizeAgUiMessages`, `normalizeConversationHistoryForRuntime`, `decodeAgUiSseChunk`) plus the `MESSAGES_SNAPSHOT` path rather than a mock.

### What's good
- Tests real code paths end-to-end, not a reimplementation of the logic.
- Covers the host normalizer, channel-history normalizer, SSE decoder, and snapshot replay from one coherent fixture.
- `structuredClone` per consumer avoids cross-assertion mutation bleed. Tidy.

### Gaps
1. **The fixture is hand-authored "API-shaped", not a golden capture.** This is the central limitation: it pins what the author *believes* the API emits, not what it actually emits. Combined with #2429's sibling PR (veryfront-api#3498) building its *own* independent hand-written fixture, there's no single source of truth — the two repos can drift and both test suites stay green. A true cross-repo contract needs a captured/shared golden payload, or a generated fixture from the API's own schema. Worth a tracking issue even if out of scope here.
2. **One monolithic `it` block** asserts four distinct things (host messages, channel messages, SSE decode, snapshot replay). A failure won't localize — split into four `it`s for better signal.
3. **Decorative fields.** `event_id`, `page_info`, `after_event_id`, the `RUN_FINISHED`/`STATE_SNAPSHOT` snapshot entries, and `apiEventsResponseFixture`'s extra event types are never asserted/consumed — they add realism but also noise. The `wireEventByCanonicalType` map only covers 4 of them, so `toSseFrames` would emit `event: undefined` if the fixture grew. Minor.
4. `if (messagesSnapshot.type === "MESSAGES_SNAPSHOT")` is always true (`as const`) — harmless type-narrowing, but the preceding `assertEquals(messagesSnapshot.type, …)` already covers it.

Net: valuable, safe, and tests the right invariant. The fixture-authenticity gap (#1) is the only thing keeping this from being a true contract test rather than an aspirational one. Ship it; file a follow-up to source the fixture from a real/shared capture.

PR #2472, opened 2026-06-16
Suppress recovered rapid heap growth warnings

kojiwakayama:
## Summary
- Delay `Rapid heap growth detected` by one memory-monitoring interval and only warn if heap remains more than 100MB above the pre-spike baseline.
- Preserve the existing warning message for sustained growth, while adding `observedGrowthMB` to distinguish the original spike from the sustained delta.
- Bump `deno.json` and `src/utils/version-constant.ts` from `0.1.817` to `0.1.818`.

## Runtime evidence addressed
The production sample from #3200 showed a cold render/cache warmup spike from 99.16MB to 212.69MB, then recovery to 153.01MB on the next sample. This change suppresses that recovered one-interval spike but still alerts when the next sample remains over the 100MB sustained-growth threshold.

## Test plan
- Red/green regression: `deno test --no-check --allow-all src/utils/memory/profiler.test.ts`
- `deno test --no-check --allow-all src/utils/version.test.ts src/utils/logger/logger.test.ts src/utils/memory/profiler.test.ts`
- `DENO_NO_PACKAGE_JSON=1 deno lint src/utils/memory/profiler.ts src/utils/memory/profiler.test.ts`
- `deno check src/utils/memory/profiler.ts src/utils/memory/profiler.test.ts`
- `deno task typecheck`
- `deno task test:unit` passed: 2167 tests, 18661 steps.
- `git push` pre-push checks passed: formatting, lint, typecheck, unit tests.

## Known repo gate
- `deno task verify:quick` currently stops at pre-existing `lint:cli-boundary` violations in `cli/shared/server-startup.ts`, unrelated to this diff.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.817",
+  "version": "0.1.818",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": {
--- a/src/utils/memory/profiler.test.ts
+++ b/src/utils/memory/profiler.test.ts
@@ -7,8 +7,10 @@ import {
   forceGC,
   getCacheStats,
   getHeapStats,
+  getInitialRapidHeapGrowthState,
   getMemoryMonitoringLogContext,
   getMemorySnapshot,
+  getRapidHeapGrowthEvaluation,
   registerCache,
   setHeapWarningThreshold,
   startMemoryMonitoring,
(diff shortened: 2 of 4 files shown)


kojiwakayama:
Addressed the inline feedback about restart baselines.

`startMemoryMonitoring` now seeds the rapid-growth baseline from the current heap via `getInitialRapidHeapGrowthState(getHeapStats().usedHeapSizeMB)` instead of resetting `lastHeapUsed` to zero. Added a regression test for the stable warm-heap restart case so a 153MB heap that remains flat does not create a pending spike or sustained warning.

Verification after the update:
- `deno test --no-check --allow-all src/utils/memory/profiler.test.ts`
- `DENO_NO_PACKAGE_JSON=1 deno lint src/utils/memory/profiler.ts src/utils/memory/profiler.test.ts`
- `deno check src/utils/memory/profiler.ts src/utils/memory/profiler.test.ts`
- `deno task typecheck`
- pre-push checks passed: formatting, lint, typecheck, unit tests (`2167 passed`, `18662 steps`).

PR #2503, opened 2026-06-17
Reduce cache fragmentation from noisy query params

kojiwakayama:
## Summary
- align page-data cache keys with the HTML render query canonicalizer
- exclude additional common tracking and cache-busting params by default while preserving meaningful unknown params
- make query cache-key encoding collision-resistant and deterministic for duplicate query keys
- bump `deno.json` and `src/utils/version-constant.ts` to `0.1.842` for a new release

## Expected performance effect
- Avoids first-hit work for page-data and HTML render requests that differ only by standard analytics or cache-busting query params.
- Keeps query-driven pages safe because unknown params such as `page`, `sort`, or feature-specific params still vary the cache key.

## Tests
- `deno test --no-check --allow-all src/cache/keys.test.ts src/server/handlers/request/module/page-data-endpoint-handler.test.ts src/utils/version.test.ts`
- `deno fmt --check src/cache/keys/prefixes.ts src/cache/keys/utils.ts src/config/schemas/config.schema.ts src/rendering/renderer.ts src/server/handlers/request/module/page-data-endpoint-handler.ts src/cache/keys.test.ts src/server/handlers/request/module/page-data-endpoint-handler.test.ts docs/architecture/02-request-pipeline.md src/utils/version-constant.ts deno.json`
- `git diff --check`
- pre-push hook: format, lint, type check, generated manifests, and unit suite (`2172 passed`)

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.841",
+  "version": "0.1.842",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": {
--- a/docs/architecture/02-request-pipeline.md
+++ b/docs/architecture/02-request-pipeline.md
@@ -90,9 +90,11 @@ Default routing cache controls:
 
 Release-backed production page-data requests use a fresh cache window plus a
 bounded stale-while-revalidate window. The cache key includes the project,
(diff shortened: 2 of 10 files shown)


kojiwakayama (commented):
CI is green. Ready to merge for release v0.1.842.

PR #2525, opened 2026-06-18
Keep Mistral chat within available quota

kwakayama:
## Summary
- cap Mistral Large default output reservation at 4096 tokens
- keep hosted/direct Mistral model resolution unchanged

## Why
Staging Mistral runs are currently failing with provider rate limits because the deployed quota is 20k TPM and the runtime reserved the full Mistral output window.

## Validation
- `deno task test src/agent/runtime/constants.test.ts`
- `deno task test src/agent/runtime/model-resolution.test.ts src/provider/veryfront-cloud/model-catalog.test.ts src/provider/veryfront-cloud/provider.test.ts`

## Notes
- Full pre-commit `verify:quick` is blocked by existing CLI boundary lint violations outside this patch.

### DIFF
--- a/src/agent/runtime/constants.test.ts
+++ b/src/agent/runtime/constants.test.ts
@@ -23,8 +23,8 @@ describe("getModelMaxOutputTokens", () => {
   });
 
   it("returns known limits for Mistral models", () => {
-    assertEquals(getModelMaxOutputTokens("mistral/mistral-large-2512"), 131_072);
-    assertEquals(getModelMaxOutputTokens("veryfront-cloud/mistral/mistral-large-2512"), 131_072);
+    assertEquals(getModelMaxOutputTokens("mistral/mistral-large-2512"), 4_096);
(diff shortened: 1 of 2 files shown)


kwakayama:
Critical subagent review: **94/100**

Findings first:

[LOW] `src/agent/runtime/constants.ts` conflates model capability with an operational quota cap. The table is described as max output token limits per model, but `4096` is really a quota-safe default for current hosted Mistral capacity. This is acceptable for the immediate staging fix. Future cleanup should split model capability from deployment quota defaults if direct Mistral long-output defaults matter.

No blocking correctness, security, fallback-masking, or regression issues found.

Evidence checked:
- Diff versus `origin/main` is only the scoped constant/test change.
- `getModelMaxOutputTokens("mistral/mistral-large-2512")` and the `veryfront-cloud/` prefixed alias now return `4096`.
- Runtime model aliases resolve through the capped lookup.
- Validation passed: focused runtime constants test and related model-resolution/provider tests.

Merge threshold met (>=90).

PR #2551, opened 2026-06-19
Preserve active Anthropic tool results across same-turn steps

kwakayama:
## Summary
- fixes Anthropic replay compaction so active same-user-turn assistant/tool steps retain their `tool_use` / `tool_result` pairs
- keeps historical compaction for completed tool rounds once a later assistant answer exists before a user retry
- adds a regression covering `load_skill` -> result -> `get_integration` -> result in one hosted turn

## Root cause
Staging `veryfront-agent` was not missing durable `TOOL_CALL_RESULT` events. The stored transcript had contiguous tool-call result events, but Anthropic request replay compacted earlier same-turn tool rounds whenever later assistant text existed.

That assumption is valid for historical completed turns, but hosted streaming can produce several assistant/tool steps before the next user message. In that active same-user turn, compaction hid prior setup/discovery results from Claude, so the model saw incomplete context and repeated `load_skill`, `web_fetch`, and integration lookup work.

This is the root fix rather than a duplicate-call cache: it preserves the model input that should have been replayed.

Fixes veryfront/veryfront-studio#5137.

## Verification
- RED: `deno test extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` failed the new same-turn replay regression before the fix because the first tool round was dropped.
- GREEN: `deno test extensions/ext-llm-anthropic/src/anthropic-provider.test.ts extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` -> `ok | 2 passed (78 steps) | 0 failed`
- `deno fmt --check extensions/ext-llm-anthropic/src/anthropic-request-builder.ts extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` -> `Checked 2 files`
- `git diff --check` -> clean

## Not run
- Full pre-commit `verify:quick` is currently blocked by pre-existing CLI boundary violations outside this PR.

### DIFF
--- a/extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts
+++ b/extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts
@@ -344,6 +344,183 @@ describe("ext-llm-anthropic/anthropic-request-builder", () => {
     ]);
   });
 
+  it("keeps same-turn tool results when later assistant steps continue without a user turn", () => {
+    const prompt: RuntimePromptMessage[] = [
+      { role: "user", content: [{ type: "text", text: "Create an integration agent." }] },
+      {
+        role: "assistant",
+        content: [
+          { type: "text", text: "I will load the platform skill." },
+          {
+            type: "tool-call",
+            toolCallId: "toolu_load_skill",
+            toolName: "load_skill",
+            input: { skillId: "veryfront" },
+          },
+        ],
+      },
+      {
+        role: "tool",
+        content: [{
+          type: "tool-result",
+          toolCallId: "toolu_load_skill",
+          toolName: "load_skill",
+          output: {
+            type: "json",
+            value: {
+              skillId: "veryfront",
+              instructions: "Create agents with create_agent after gathering context.",
+            },
+          },
+        }],
+      },
+      {
+        role: "assistant",
+        content: [
+          { type: "text", text: "Now I will inspect the integration." },
+          {
+            type: "tool-call",
+            toolCallId: "toolu_get_integration",
+            toolName: "get_integration",
+            input: { integration: "harvest" },
+          },
+        ],
+      },
+      {
+        role: "tool",
+        content: [{
+          type: "tool-result",
+          toolCallId: "toolu_get_integration",
+          toolName: "get_integration",
+          output: {
+            type: "json",
+            value: { slug: "harvest", name: "Harvest" },
+          },
+        }],
+      },
+    ];
+    const warnings = createWarningCollector();
+
+    const body = buildAnthropicMessagesRequest(
+      "claude-sonnet-4-6",
+      "anthropic",
+      { prompt },
+      false,
+      warnings,
+    );
+
+    assertEquals(body.messages, [
+      { role: "user", content: [{ type: "text", text: "Create an integration agent." }] },
+      {
+        role: "assistant",
+        content: [
+          { type: "text", text: "I will load the platform skill." },
+          {
+            type: "tool_use",
+            id: "toolu_load_skill",
+            name: "load_skill",
+            input: { skillId: "veryfront" },
+          },
+        ],
+      },
+      {
+        role: "user",
+        content: [{
+          type: "tool_result",
+          tool_use_id: "toolu_load_skill",
+          content:
+            '{"skillId":"veryfront","instructions":"Create agents with create_agent after gathering context."}',
+        }],
+      },
+      {
+        role: "assistant",
+        content: [
+          { type: "text", text: "Now I will inspect the integration." },
+          {
+            type: "tool_use",
+            id: "toolu_get_integration",
+            name: "get_integration",
+            input: { integration: "harvest" },
+          },
+        ],
+      },
+      {
+        role: "user",
+        content: [{
+          type: "tool_result",
+          tool_use_id: "toolu_get_integration",
+          content: '{"slug":"harvest","name":"Harvest"}',
+        }],
+      },
+    ]);
+  });
+
+  it("keeps historical tool-only rounds when active same-turn assistant text follows the latest user", () => {
+    const prompt: RuntimePromptMessage[] = [
+      { role: "user", content: [{ type: "text", text: "Start with account context." }] },
+      {
+        role: "assistant",
+        content: [{
+          type: "tool-call",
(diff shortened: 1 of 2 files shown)


kwakayama:
## Critical Review — PR #2551

**Score: 88/100**

A tightly-scoped, correct root-cause fix. The change narrows replay compaction so it only collapses tool rounds in *completed historical turns* (before the last user message), while preserving full `tool_use`/`tool_result` pairs in the still-active same-user turn. Verified locally: `deno test` on both files → `ok | 2 passed (78 steps) | 0 failed`, and the pre-existing `compacts completed historical tool rounds...` test still passes unchanged.

### What's good
- **Correct root cause.** The old `index < lastAssistantTextIndex` predicate treated *any* assistant-text later in the prompt — including interim text inside the active streaming turn — as proof the round was "completed". `hasLaterAssistantTextBeforeLatestUser` correctly requires the later assistant text to fall *before* the last user message, which is the real definition of a completed turn. Diagnosis in the description matches the code.
- **Pairing invariant preserved.** The most important safety property holds: compaction still strips the `tool_use` (via the text-only `assistantContent` filter) and its `tool_result` (via `skippingHistoricalToolResults`) *together*. The fix only flips *when* compaction triggers; it never decouples the two sides, so no orphaned `tool_use`/`tool_result` (Anthropic 400) is introduced.
- **Strong regression test.** The new test asserts the full output array including `tool_result` ordering and JSON stringification, and the description documents a real RED→GREEN.

### Nits / suggestions (non-blocking)
1. **Simplify to match the original O(n) style.** The new predicate is an O(n) `prompt.some()` evaluated per assistant message → O(n²) overall, and reads less directly than the index it replaces. It's exactly equivalent to a single precomputed index:
   ```ts
   const lastHistoricalAssistantTextIndex = prompt.findLastIndex((m, i) =>
     i < lastUserIndex &&
     m.role === "assistant" &&
     m.content.some((p) => p.type === "text" && p.text.length > 0)
   );
   // shouldCompact = index < lastHistoricalAssistantTextIndex && hasToolCall
   ```
   Same behavior, cheaper, and mirrors the code it replaced.
2. **Untested behavior change.** A *historical* tool round with no answer text before a user retry (e.g. `assistant(tool-only) → tool → user`) is now **preserved** where it was previously compacted. This is arguably more correct (real history kept), but it's a silent behavioral shift in token usage that no test pins down — worth an assertion so it doesn't regress.
3. **Pre-existing, but worth confirming:** historical compaction can emit consecutive `assistant` messages (see the existing test's 3-in-a-row output). Recent Anthropic API versions merge consecutive same-role turns, but please confirm that's still tolerated for the models in use — not introduced here, just adjacent.
4. **"Not run" verify** is reasonable given the unrelated CLI boundary breakage, but a one-line confirmation that those violations predate this branch would close the loop.

### Verdict
Approve with minor polish. Correct, minimal, well-tested fix for veryfront-studio#5137; the only blocking-adjacent items are the simplification (style/perf) and the untested behavior shift, both low risk.

kwakayama:
Addressed the review feedback in `a185d6b09`.

Changes:
- Replaced the per-assistant `prompt.some(...)` predicate with a precomputed `lastHistoricalAssistantTextIndex`, preserving the same completed-historical-turn semantics with the original O(n) shape.
- Added a regression for the previously unpinned edge case: a historical tool-only round before the latest user retry is preserved when the active same-turn assistant later emits text. This catches the old `lastAssistantTextIndex` behavior because it would compact/drop that earlier tool result based on active-turn assistant text.
- Rechecked the full pre-commit blocker: `verify:quick` still fails at `lint:cli-boundary` after manifest, format, lint, and style pass. The reported files are all outside this PR and exist unchanged on `origin/main`; this branch only touches `extensions/ext-llm-anthropic/src/anthropic-request-builder.ts` and its test.

Verification after the feedback patch:
- `deno test extensions/ext-llm-anthropic/src/anthropic-provider.test.ts extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` -> `ok | 2 passed (79 steps) | 0 failed`
- `deno fmt --check extensions/ext-llm-anthropic/src/anthropic-request-builder.ts extensions/ext-llm-anthropic/src/anthropic-request-builder.test.ts` -> `Checked 2 files`
- `git diff --check` -> clean

kwakayama:
## Follow-up — non-blocking items resolved

Re-reviewed the current PR head (`a185d6b09`) and ran the checks locally. All four items from my earlier review are addressed:

1. **Predicate simplification (#1) — done.** The head now uses a single precomputed `lastHistoricalAssistantTextIndex = prompt.findLastIndex((m, i) => i < lastUserIndex && assistant && hasText)` with `index < lastHistoricalAssistantTextIndex && hasToolCall`. O(n), and it mirrors the original `lastAssistantTextIndex` structure it replaced — exactly the cleanup I suggested. Nothing left to change here.

2. **Untested behavior change (#2) — done.** New test `keeps historical tool-only rounds when active same-turn assistant text follows the latest user` (`anthropic-request-builder.test.ts:458`) pins the case directly: a `assistant(tool-only) → tool → user retry` round is now **preserved** (tool_use + tool_result kept, result merged into the retry's user message). That's the silent behavior shift, now locked in by an assertion.

3. **Consecutive `assistant` messages from compaction (#3) — confirmed safe, no change needed.** Checked against the Anthropic Messages API contract: the API **accepts consecutive same-role messages and combines them into a single turn** (it does not require strict alternation and does not 400 on this). So the historical-compaction path emitting back-to-back `assistant` text blocks (as asserted in `compacts completed historical tool rounds…`) is valid. Pre-existing and fine.

4. **`verify:quick` "Not run" (#4) — confirmed pre-existing & unrelated.** This PR changes **0 CLI files** (vs `origin/main` it touches only the two `ext-llm-anthropic` files). The `lint:cli-boundary` failures are all in `cli/` files this branch never touches — and the exact violations exist verbatim on `origin/main` (e.g. `#veryfront/skill` import at `cli/commands/skills/validate.ts:12` and `cli/skills/loader.ts:4`). Definitively not introduced here.

**Local verification on head:**
- `deno test … anthropic-provider.test.ts anthropic-request-builder.test.ts` → `ok | 2 passed (79 steps) | 0 failed`
- `deno fmt --check` on both touched files → `Checked 2 files`

No further code changes needed — this is good to merge.

PR #2567, opened 2026-06-20
[codex] Improve Jira Confluence and GitHub tool summaries

kwakayama:
## Summary
- Add Gmail-style historical summaries for Confluence list/search tools so list calls retain useful compact details.
- Improve GitHub list/detail summaries with bounded body/description fields, labels, assignees, reviewers, branches, and commit context.
- Fix GitHub `list_issues` cursor pagination by adding the missing `after` input and GraphQL variable.
- Keep Jira templates aligned with token pagination, optional `search_users.query`, and compact nested issue `fields`.
- Bump `veryfront-code` to `0.1.875`.

## Root Cause
- Confluence list/search tools had response transforms but no historical summary metadata, unlike Gmail, so historical tool output compaction was too sparse or raw.
- GitHub `list_issues` returned `pageInfo` but did not accept/pass an `after` cursor, so follow-up pagination could not continue.
- GitHub summaries omitted useful bounded detail fields for list and detail calls.
- Jira template metadata had drifted from the staging hotfix contract and did not summarize nested Jira issue fields.

## Staging Audit
- Public staging catalog lists the expected Jira, Confluence, GitHub, and Gmail tools.
- GitHub read-only calls succeeded in staging for current user/project canaries: `get_current_user`, `list_repos`, `list_issues`, `list_prs`, and `list_commits`.
- Jira read-only calls succeeded when using a project whose allowlist included the tool and a bounded JQL query: `list_sites`, `search_users`, and `search_issues`.
- Current staging Jira `search_issues` summary only exposes `id`/`key`; this PR adds compact nested `fields` so issue summaries carry enough detail.
- Confluence calls reached the integration runtime and returned structured reconnect-required errors for the current staging OAuth token. Provider-content verification needs a reconnected Confluence token, but the error path is handled.

Refs veryfront/veryfront-api#3640

## Verification
- `deno test --allow-all src/integrations/_data.test.ts`
- `deno test --no-check --allow-all src/chat/message-prep.test.ts`
- `deno test --allow-all src/utils/version.test.ts`
- `deno fmt --check cli/templates/integrations/confluence/connector.json cli/templates/integrations/github/connector.json cli/templates/integrations/jira/connector.json src/integrations/_data.test.ts src/chat/message-prep.test.ts src/integrations/_data.ts src/integrations/_tool_summaries.ts deno.json src/utils/version-constant.ts`
- `git diff --check`
- `deno task build`
- Sanitized pre-push hook: formatting, lint, typecheck, and `2191` unit tests passed.

## Notes
- The unsanitized local push hook initially failed because this shell exports `OTEL_METRICS_EXPORTER=otlp`, which changes the ambient default observed by `src/observability/metrics/config.test.ts`. Re-running the same hook with OTEL variables unset passed.


kwakayama:
## Review — Score: 85/100

Focused, well-tested contract/config change. The bulk is connector-summary metadata plus a genuine pagination bug fix, backed by solid tests. Mergeable; a couple of minor latent gaps noted below.

### Strengths
- **Real bug fix:** `github__list_issues` cursor pagination now works — adds the `$after` GraphQL variable, `after: $after` in the query, and the `after` body param. Previously `pageInfo.endCursor` was returned but unusable. Also adds `totalCount` to `labels`/`assignees`.
- **Good test coverage:** new `maskOldToolOutputs` cases for GitHub issues, Confluence search, and Jira search assert the compaction behavior (body truncation, nested-object trimming, cursor passthrough). Contract-sync tests assert `historicalToolSummaries[x]` deep-equals the endpoint's `historicalSummary`, so the parallel sources of truth can't drift silently.
- **Sound compaction semantics:** `kind: "object"` keeping only scalar keys correctly trims `labels → {totalCount}`, `space → {id,key,name}`, Jira `fields → {summary,created,updated}`, dropping nested payloads.
- Version bump applied to both `deno.json` and `src/utils/version-constant.ts` (the test-enforced pair).

### Findings

**[Minor] `list_prs` `labels` with `kind: "object"` is a no-op.** `list_prs` is a REST endpoint, where `labels` is an **array** of label objects — but `compactHistoricalField`'s object branch calls `isRecord(fieldValue)`, which rejects arrays, so it returns `null` and PR labels are silently omitted from the summary. Contrast `list_issues` (GraphQL), where `labels` is a `{totalCount, nodes}` object and `kind: "object"` works (and is the only case covered by a test). There's no current field-kind that extracts `name` from an array-of-objects, so PR labels effectively can't be summarized as written. Not a regression (they weren't there before), but the field is dead config. Either drop it or add a kind that handles arrays-of-named-objects. `head`/`base`/`requested_reviewers` look correct.

**[Minor] Jira `search_issues` loses status/assignee in the summary.** Real Jira issues nest `status`/`assignee` under `fields`, and the shallow `kind: "object"` compaction of `fields` keeps only scalar keys — so `fields.status.name` and `fields.assignee.displayName` are dropped, keeping just `summary`/`created`/`updated`. That undercuts the PR's stated goal ("carry enough detail") for the two fields most useful for triage. The sibling top-level `itemFields` (`summary`, `status:object`, `assignee:contact`) operate on the issue root, which doesn't carry those keys in real Jira payloads, so they're now dead config. Acceptable for *historical* (old-turn) compaction, but worth a follow-up if status/assignee matter.

**[Nit] Three parallel definitions** of each contract (`connector.json` template, generated `_data.ts`, hand-maintained `_tool_summaries.ts`). Drift risk is mitigated by the equality tests, so this is fine — just flagging the maintenance surface.

### Notes
- No security concerns — read-only summary metadata only.
- Single commit, no human review yet; CI/pre-push (2191 tests) reported green in the description.

kwakayama:
## Re-review — Score: 93/100 (was 85)

Force-push (`6461272d`, +883/−76, 11 files) addresses both findings from the prior review — and does it the right way, by extending the compaction vocabulary generically rather than patching individual contracts. Both fixes are test-backed.

### Findings resolved
- **PR labels (REST array) no longer dropped.** New `kind: "named-array"` (`message-prep.ts` `compactNamedArrayValue`) extracts `name`/`displayName`/`login`/`key`/`id` from arrays-of-objects. `list_prs.labels` now uses it, while GraphQL `list_issues.labels` correctly stays `kind: "object"` (`{totalCount, nodes}`) — the REST-vs-GraphQL distinction is now handled per-endpoint. Covered by the new test `…keeps compact GitHub PR labels from REST list results` (asserts `labels → ["bug","integrations"]`).
- **Jira status/assignee no longer lost.** New optional `path: string[]` on summary fields (`getHistoricalFieldValue`) reaches into `fields.*`, so `status`/`assignee`/`summary`/`created`/`updated` are pulled out of the nested Jira `fields` object. `compactContactValue` also gained `displayName`/`accountId` for Jira assignees. Test now asserts `status → {name:"In Progress"}` and `assignee → {displayName:"Dev"}`.

### Quality checks
- **Schema typed:** `schema.ts` adds `named-array` to the kind enum and `path` to the field schema, so contracts are validated.
- **Sources stayed in sync:** `_tool_summaries.ts` updated to match (`named-array`/`path` present), and `_data.test.ts` asserts both the Jira `[name, path]` mappings and that `list_issues.labels` is `object` while `list_prs.labels` is `named-array`.

### Remaining nit (non-blocking)
- **[Nit] Redundant Jira `fields`.** The output now carries `fields:{summary,created,updated}` *and* the same `summary`/`created`/`updated` again at item top-level via `path`. Harmless (a few extra tokens), but now that everything useful is surfaced via `path`, the standalone `fields` object-kind field is arguably redundant and could be dropped.

Clean, thorough, well-tested. Ship it.

PR #2637, opened 2026-06-24
Add request desk demo integration tools

kwakayama:
## Summary
- Add request-desk demo integration tools for Outlook and ServiceNow, including Outlook `list_threads`, `get_thread`, and `create_draft` scaffold handlers plus ServiceNow request/interaction/request-item endpoints.
- Fix the Outlook review items: static `get_thread` no longer interpolates `thread_id` into an inline query URL, `listThreads` dedupes by `conversationId`, and Outlook contact rendering handles null/missing senders.
- Regenerate integration data/tool summaries/manifests so the mock-to-real contract exposes the demo tools.
- Rebase onto current `main` and bump `veryfront` from `0.1.920` to `0.1.921`.

## Scope Notes
- Removed the unrelated runtime commits from this PR; the branch now contains only the request-desk integration/tool work and the review-fix commit.
- The template `get-thread` tool keeps the ergonomic `thread_id` input and builds the Graph `$filter` with `URLSearchParams`; the declarative connector uses an explicit `filter` query param mapped to Graph `$filter` via `queryName`.

## Staging / Mock Proof
- Staging mock proof for the companion mock contract is tracked here: https://github.com/veryfront/veryfront-integration-mocks/issues/14#issuecomment-4786115668
- This PR branch itself is verified locally; the code branch is not deployed to staging by this PR update.

## Test Plan
- `deno run -A scripts/build/generate-integrations-module.ts`
- `deno run -A scripts/build/generate-templates-manifest.ts`
- `deno test --allow-read --allow-env src/integrations/_data.test.ts`
- `deno task generate:manifests:check`
- `deno task typecheck`
- `deno task verify:quick`
- `env -u OTEL_EXPORTER_OTLP_ENDPOINT -u OTEL_EXPORTER_OTLP_HEADERS -u OTEL_EXPORTER_OTLP_PROTOCOL -u OTEL_METRICS_EXPORTER -u GRAFANA_SERVICE_ACCOUNT_TOKEN -u GRAFANA_URL deno task test:unit` (2203 passed)
- Same sanitized environment for `git push`; pre-push hook passed formatting, lint, typecheck, generate, and full unit suite before force-with-lease update.

## Notes
- The first push attempt inherited local OTEL/Grafana environment variables, causing unrelated observability default tests to expect `console` while the environment selected `otlp`. Rerunning with those variables unset passed.


kwakayama:
## Critical review — Score: **78 / 100**

Solid, mostly-additive work with good schema discipline and security awareness, but the PR conflates several unrelated changes and the description is out of sync with the diff. Details below.

### What's strong
- **Connector/tool definitions are clean.** Input schemas use real validation (`.email()`, `.min/.max`, enums, sensible defaults) and every field is described — these will produce good tool surfaces for the model.
- **Security awareness in `getThread`.** Escaping the OData literal (`replaceAll("'", "''")`) plus `URLSearchParams` encoding is the right instinct for the hand-written client.
- **`buildMessage` refactor** removes the send/draft duplication cleanly, and `createDraft` correctly does *not* send (`POST /messages`, `isDraft: true`).
- **`provider-errors` hardening** adds a `WeakSet` cycle guard for nested `error` bodies — good defensive parsing.
- Tests accompany the runtime changes, and the full local suite (2169) reportedly passed.

### Concerns

**1. Scope creep — three unrelated changes in one PR (biggest issue).**
The title/summary describe only the request-desk demo tooling, but the diff also carries two earlier commits — *"Harden post-form create-agent tool flow"* and *"Fix assistant-prefill retry after final text"* — touching ~20 `src/` runtime files (`chat-runtime-tool-assembly`, `skill-policy-enforcement`, `tool-result-continuation`, `streamed-assistant-message`, `provider-errors`, `runtime-state-resolver`). These are real behavior changes to hosted-agent tool gating and stream continuation, none of which are mentioned in the summary or Test Plan. Bundling low-risk additive connector JSON with runtime control-flow changes makes the runtime changes hard to review and ties a demo-tooling merge to agent-runtime risk. **Recommend splitting the runtime commits into their own PR** (or at minimum documenting them in the summary + test plan).

**2. Description is out of sync with the diff.**
- Summary says *"Bump `veryfront` to 0.1.823"* — the actual bump is **0.1.822 → 0.1.824**.
- Test Plan covers the integration regen/manifests but omits the runtime change tests entirely.

**3. Real connector path for `get_thread` looks under-encoded vs. the hand-written client.**
`get_thread` embeds the filter inline — `.../me/messages?$filter=conversationId eq '{thread_id}'` — and declares `thread_id` as `in: "path"` even though it sits in the query string. The hand-written `getThread()` escapes quotes and uses `URLSearchParams`; the declarative connector does neither, so a `conversationId` containing `'`, `+`, `/`, or the literal spaces in `eq '...'` may produce a malformed Graph request. Two implementations of the same call with different encoding semantics is a latent bug. Given the summary notes mock proof is **still blocked on staging**, the real path appears unverified end-to-end — please confirm the runtime substitutes `in: path` params into a query position correctly (encoding-wise).

**4. `list_threads` returns messages, not deduplicated threads.**
It maps each recent message to a "thread" keyed by `conversationId` with no dedup, so multiple messages from the same conversation surface as separate threads with duplicate `thread_id`s. Acceptable for a mock demo, but a real inbox will show duplicates during triage. Consider deduping by `conversationId`, or rename to reflect that these are representative messages.

**5. Minor robustness — unguarded `from` access.**
`get-thread.ts` does `message.from.emailAddress.name`. Draft/system messages in a conversation can have a null `from`, which would throw. `list-threads.ts` has the same pattern. A null guard (as already used for `cc`) would harden these.

### Verdict
Demo tooling itself is low-risk and well-built (78 reflects that). The deductions are for bundling unreviewed runtime behavior under a "demo tools" title, a description that doesn't match the diff, and a real connector path that's plausibly mis-encoded and unverified. Split the runtime work out (or document + test it here), fix the version/description mismatch, and confirm the `get_thread` query encoding before merge.

kwakayama:
Addressed in `ce80121b7` and force-pushed the rebuilt branch.

- Removed the unrelated runtime commits from this PR; the branch now has only the request-desk integration/tool commits plus the review fix.
- Updated the PR body: version is now `0.1.921` because the branch is rebased on current `main` (`0.1.920`), and the test plan no longer references runtime changes.
- Fixed Outlook `get_thread` in the declarative connector: URL is now `https://graph.microsoft.com/v1.0/me/messages`; `filter` is a query param mapped with `queryName: "$filter"`, so the connector no longer treats a query placeholder as a path param.
- Kept the scaffolded template `get-thread` tool ergonomic with `thread_id`; its client escapes single quotes and builds the Graph filter through `URLSearchParams`.
- Made `listThreads` dedupe by `conversationId` and made Outlook contact serialization null-safe for missing `from`/recipient email addresses.

Verification:
- `deno test --allow-read --allow-env src/integrations/_data.test.ts`
- `deno task generate:manifests:check`
- `deno task typecheck`
- `deno task verify:quick`
- `env -u OTEL_EXPORTER_OTLP_ENDPOINT -u OTEL_EXPORTER_OTLP_HEADERS -u OTEL_EXPORTER_OTLP_PROTOCOL -u OTEL_METRICS_EXPORTER -u GRAFANA_SERVICE_ACCOUNT_TOKEN -u GRAFANA_URL deno task test:unit` -> 2203 passed
- Sanitized pre-push hook passed format, lint, typecheck, generate, and full unit suite before the force-with-lease push.

kwakayama:
## Re-review — Score: **90 / 100** (was 78)

Thanks for the fast turnaround — all five items from the previous pass are addressed, and a couple were handled more thoroughly than asked.

### Verified fixes
1. **Scope creep — resolved.** The two unrelated runtime commits are gone; the branch is now just the request-desk integration work + the review-fix commit. Diff dropped from 28→15 files.
2. **Description in sync.** Version bump now correctly stated (`0.1.920 → 0.1.921`, both `deno.json` and `version-constant.ts`), and the Test Plan reflects the actual work (`verify:quick`, full `test:unit` 2203 passed).
3. **`get_thread` encoding — fixed.** The connector no longer interpolates `thread_id` into an inline URL; it uses an explicit `filter` query param mapped via `queryName: "$filter"`, and the template client still builds the filter with `URLSearchParams` + quote-doubling. No more malformed-URL risk.
4. **`listThreads` dedupes by `conversationId`** (with `|| message.id` fallback) — triage no longer shows duplicate threads.
5. **Null-contact handling — done well.** `OutlookContact.emailAddress` is now properly optional/nullable, and the new `summarizeContact`/`summarizeContacts` helpers centralize the null-safe rendering. Nice that you propagated the type change through the *existing* tools (`get-email`, `list-emails`, `search-emails`, `formatEmail`) rather than just the new ones — that's the correct ripple fix.

### Residual notes (minor, non-blocking)
- **Contract divergence in `get_thread` input.** The declarative connector now takes a raw OData `filter` string while the template tool takes an ergonomic `thread_id`. You've documented this tradeoff in the Scope Notes and it's defensible, but the two halves of the "mock-to-real contract" now expose different input schemas for the same tool. Worth a follow-up to confirm the mock contract and request-desk agent prompt expect `filter` (not `thread_id`) on the connector side, since the model must now hand-build and escape `conversationId eq '...'` itself.
- **CI not yet green.** Full coverage shards are currently *pending* on this revision — worth confirming they pass before merge (the earlier revision only ran 2 lightweight checks).

### Verdict
Clean, well-scoped, additive demo tooling with good null-safety and encoding discipline. Ship once CI is green and the connector `filter` interface is confirmed against the mock contract.

PR #2641, opened 2026-06-24
Sanitize forwarded runtime tool allowlists

kwakayama:
Related to veryfront/veryfront-studio#5311

## Summary
- Sanitize forwarded runtime allowedTools at the control-plane handler boundary.
- Keep only child payload tools plus Studio runtime tools before run-stream merges remote allowlists.
- Update regression coverage to fail closed for forged forwarded remote tool names.

## Test Plan
- [x] RED: `deno task --quiet test src/server/handlers/request/agent-stream.handler.test.ts` failed before the fix on forwarded Gmail tools.
- [x] `deno check src/server/handlers/request/agent-stream.handler.ts`
- [x] `deno task --quiet test src/server/handlers/request/agent-stream.handler.test.ts`
- [x] pre-commit `deno task verify:quick`

### DIFF
--- a/src/server/handlers/request/agent-stream.handler.test.ts
+++ b/src/server/handlers/request/agent-stream.handler.test.ts
@@ -531,7 +531,7 @@ describe("server/handlers/request/agent-stream.handler", () => {
     assertEquals(injectedToolSchema, inputSchema);
   });
 
-  it("passes runtime integration tool allowlists from forwarded props into the runtime agent config", async () => {
+  it("does not pass undeclared forwarded remote tool allowlists into the runtime agent config", async () => {
     let capturedAllowedTools: string[] | undefined;
 
     const handler = new AgentStreamHandler({
@@ -590,7 +590,7 @@ describe("server/handlers/request/agent-stream.handler", () => {
 
     assertExists(result.response);
     assertEquals(result.response.status, 200);
-    assertEquals(capturedAllowedTools, ["gmail:list-emails", "gmail:get-email"]);
+    assertEquals(capturedAllowedTools, undefined);
   });
 
   it("auto-exposes Studio MCP tools for trusted Studio project-agent requests", async () => {
--- a/src/server/handlers/request/agent-stream.handler.test.ts
+++ b/src/server/handlers/request/agent-stream.handler.test.ts
@@ -593,6 +593,69 @@ describe("server/handlers/request/agent-stream.handler", () => {
     assertEquals(capturedAllowedTools, undefined);
   });
 
+  it("drops undeclared Studio runtime tool allowlists for untrusted clients", async () => {
+    let capturedAllowedTools: string[] | undefined;
+
+    const handler = new AgentStreamHandler({
+      ensureProjectDiscovery: async () => {},
+      getAgent: (id) => id === "assistant-1" ? createAgent("assistant-1") : undefined,
+      getAllAgentIds: () => ["assistant-1"],
+      sessionManager: new AgentRunSessionManager(),
+      createRuntime: (agent) => {
+        capturedAllowedTools = (agent.config as typeof agent.config & RuntimeRemoteToolConfig)
+          .__vfAllowedRemoteTools;
+
+        return {
+          stream: async (_messages, _context, callbacks) => {
+            callbacks?.onFinish?.({
+              text: "ok",
+              messages: [],
+              toolCalls: [],
+              status: "completed",
+              usage: {
+                promptTokens: 1,
+                completionTokens: 1,
+                totalTokens: 2,
+              },
+            });
+
+            return new ReadableStream<Uint8Array>({
+              start(controller) {
+                controller.close();
+              },
+            });
+          },
+        };
+      },
+    });
+
+    const body = createAgentStreamRequestBody({
+      forwardedProps: {
+        clientId: "external-client",
+        runtimeOverrides: {
+          allowedTools: ["studio_todo_write"],
+        },
+      },
+    });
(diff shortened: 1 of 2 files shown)


kwakayama:
## Critical Review — Score: 82 / 100

Reviewed against `plans/issue-5311-agent-tool-scope-onepager.md` (step 4: runtime-side defense) and the API sibling veryfront-api#3698. Verified locally: `deno task test agent-stream.handler.test.ts` → **passed (23 steps)**.

### What's good
- **Correct defense-in-depth placement.** Sanitizing `forwardedProps.runtimeOverrides.allowedTools` *before* `withVeryfront{Platform,Studio}RemoteTools()` is the right seam — `getRequestedStudioToolNames()` reads `allowedTools`, so a forged entry there was a genuine Studio-tool escalation vector.
- **Fail-closed and clean.** Empty result deletes `allowedTools` (→ `undefined`) rather than leaving `[]`, and prunes empty `runtimeOverrides`/`forwardedProps`. The flipped test (`gmail:*` forwarded but undeclared → `capturedAllowedTools === undefined`) is a real regression guard.

### Findings
1. **(Medium — please justify) The `STUDIO_RUNTIME_REMOTE_TOOL_NAMES` carve-out partially defeats the sanitizer.** The filter keeps any tool that is in `availableToolNames` **OR** in `STUDIO_RUNTIME_REMOTE_TOOL_NAMES`. So a forged `forwardedProps.allowedTools: ['studio_todo_write']` on a child that never declared it **survives sanitization** and flows into `getRequestedStudioToolNames()`. The residual gate is only `clientAllowsStudioMcp(clientProfile)` + token. For trusted Studio clients these tools may be auto-exposed anyway (per the `auto-exposes Studio MCP tools for trusted...` test), so this may be intentional — but it means the PR does **not** hold the strict "child can call only its declared tools" line for Studio tools. Add a test asserting an undeclared Studio tool is dropped for an *untrusted* client, and a comment explaining why the carve-out is safe for trusted ones. As written, the title "fail closed for forged forwarded remote tool names" overstates the guarantee for the Studio set.
2. **(Low) Scope narrower than the one-pager.** Step 4 asked the runtime to intersect *remote tool sources* with child config / `payload.tools` at the attachment boundary. This PR only sanitizes `allowedTools`. That turns out to be sufficient because the **platform** path is independently gated by `agent.config.tools` (via `getRequestedUnresolvedBooleanToolNames`) rather than `allowedTools` — good — but that invariant is the load-bearing reason the narrower patch is safe and it isn't documented. A one-line comment noting "platform remote tools are gated by child agent config; only the Studio path consumes forwarded allowedTools" would make the safety argument legible.
3. **(Info) Cross-repo coupling.** This is the defense-in-depth half of a two-PR fix; the API PR is the primary boundary. They're independently safe (no hard deploy ordering), but neither PR body notes the relationship — worth a cross-link so a future reader doesn't remove one thinking the other covers it.

### Bottom line
Correct, minimal, well-tested patch on the one field that mattered for the Studio path. Held back from a higher score by the unexplained Studio carve-out (weakens the stated "fail closed" claim) and undocumented reliance on agent-config gating for the platform path.

kwakayama:
Addressed review comments in fc8f888ec.

Changes:
- Tightened the Studio runtime allowlist carve-out so undeclared Studio tool names are preserved only for trusted Studio client profiles.
- Added a regression proving untrusted forwarded `studio_todo_write` is dropped.
- Documented the platform-vs-Studio boundary: platform remote tools are gated by child agent config, while Studio is the forwarded-allowedTools consumer.

Validation:
- `deno check src/server/handlers/request/agent-stream.handler.ts` → passed
- `deno task --quiet test src/server/handlers/request/agent-stream.handler.test.ts` → 24 steps passed
- pre-commit `deno task verify:quick` → passed

kwakayama:
## Re-review (commit `fc8f888ec`) — Score: 82 → **91 / 100**

Re-verified on PR head: `deno task test agent-stream.handler.test.ts` → **passed (24 steps**, up from 23 — the new test ran).

### Findings status
- ✅ **#1 (Studio carve-out partially defeated the sanitizer)** — **Resolved, correctly.** The carve-out is now gated: `availableToolNames.has(toolName) || (allowStudioRuntimeTools && STUDIO_RUNTIME_REMOTE_TOOL_NAMES.has(toolName))`, where `allowStudioRuntimeTools = clientAllowsStudioMcp(resolveRuntimeClientProfile(forwardedProps))`. This is exactly the right gate — trusted Studio clients keep tools they could auto-attach anyway; an untrusted `external-client` forwarding `studio_todo_write` is now dropped. The new regression `drops undeclared Studio runtime tool allowlists for untrusted clients` asserts `__vfAllowedRemoteTools === undefined` and traces correctly (sanitizer drops → `getRequestedStudioToolNames` returns `[]` → no Studio source attached). Good adversarial coverage.
- ✅ **#2 (undocumented platform-path gating)** — Resolved. The new comment states the boundary explicitly: "Platform remote tools are gated separately by the child agent config in `withVeryfrontPlatformRemoteTools`. The Studio path is the one that consumes forwarded allowedTools…". That makes the safety argument legible for the next reader.
- ⏳ **#3 (cross-repo coupling not cross-linked)** — Minor/info, still open. Worth a one-line cross-link to veryfront-api#3698 in the PR body so the two halves of the defense-in-depth aren't independently "simplified" later. Not a merge blocker.

The fix correctly preserves the intended trust model (auto-expose for trusted Studio clients) while closing the forged-forwardedProps path for untrusted ones, and the title's "fail closed" claim now holds. Remaining 9 points: the info-level cross-link and the usual defense-in-depth reliance on the API filtering the `tools` payload upstream.

**Recommendation: approve.**

PR #2695, opened 2026-06-29
Storybook UI workbench + chat tool-cards, attachments, multi-conversation

kojiwakayama:
## Summary

Two layers of work on `veryfront/chat`:

1. **Storybook workbench** (original branch scope) — an isolated Storybook React/Vite workbench under `storybook/` to review framework UI without adding Storybook to the runtime surface, a Studio-language restyle, tool-call cards, end-to-end file attachments, and adapter-backed uploads.
2. **Clean chat abstractions** (this pass) — detoxing the `<Chat>` prop soup and splitting the domain into small, swappable hooks + components: persistence, uploads, conversations, and retiring the legacy Thread system. Every step is an independent, green commit.

---

## Part 1 — Storybook workbench + chat surface

- Isolated Storybook workbench under `storybook/` (`storybook/dist` not committed).
- Studio component-language restyle; public component APIs + legacy `chatTokens` values preserved.
- **Tool-call cards**: collapsed by default (errors stay expanded), full-width content column, container-driven spacing.
- **File attachments (end-to-end)**: controlled `<Chat>` gets batteries-included attachments (`useUpload` + fold files into `sendMessage`); sent pills show type/size; base64 guest-mode by default, durable via an upload endpoint.

## Part 2 — Clean chat abstractions

- **Persistence** — `<Chat onUpdate={…}>` emits the whole updated `conversation`; resolved by presence: explicit prop → `ConversationsProvider` default (`save`) → ephemeral. Adds `useConversations().save`.
- **Prop detox** — consolidated the controlled surface into `chat={useChat()}` and `agent={{ name, avatarUrl, description, suggestions, models }}`; the old flat props (`messages`/`input`/… and `showSources`/`showSteps`/`showExport`/`models`/`renderTool`/`quickActions`) are kept as **one-release `@deprecated` fallbacks**, so nothing breaks now.
- **Uploads** — `useUploadsRegistry({ url })` (renamed from `api`, kept deprecated); "point storage anywhere" is the server's `BlobStorage`, not a redundant client adapter.
- **Cloud uploads listing** — implemented `VeryfrontCloudBlobStorage.list()` (via `GET /projects/{slug}/uploads`), so the uploads panel works **when deployed**, not just on local disk (add/delete already worked everywhere; listing was cloud-unimplemented).
- **`ChatSidebar` is conversation-native + context-aware** — lists `ConversationSummary` rows and needs **no props** inside a `ConversationsProvider` (reads list + active id + select/new/delete/rename from context); controlled via `conversations`/`activeId`/`onSelect`/`onDelete`/`onRename`/`onNew`.
- **Retired the Thread system** — deleted `ChatWithSidebar` (462 lines + 12 config types), `use-threads.ts`, `thread-list-context.tsx`; stripped from 6 barrels + the export-surface test; migrated the guide, stories, fixtures, and generated safelist. Replaced by `<Chat>` + `<ChatSidebar>` + `ConversationsProvider`.
- **`useConversation(id)`** — the singular loader companion to `useConversations()`.

## Test plan

- [x] Pre-push hook (fmt, lint, type-check, generation, unit tests) green on **every** commit: **2231 passed / 0 failed**.
- [x] New unit tests: `save`/`onUpdate` upsert, `chat={}` controlled render, `useUploadsRegistry` custom adapter, `VeryfrontCloudBlobStorage.list()`, `ChatSidebar` context-default + controlled, `useConversation(id)`.

## Notes & follow-ups (not in this PR)

- **API-reference docs** are generated (`deno task docs` → `deno task docs:copy`). The committed api-reference is broadly stale in the *generator's* format across all modules, so a full regen belongs in its own dedicated "docs: regenerate api-reference" PR rather than buried here. Chat API **shape** is already correct (Thread exports removed).
- **`useAgent` reconcile** intentionally **not** done — the metadata hook stays `useAgentMetadata(id)`; `useAgent({ agent })` left untouched.
- **Storybook L0→L3 adoption-ladder demos** — planned follow-up.
- **Templates** (docs-agent / havana) live in separate repos.

## Size note

This chat-abstractions pass is ~34 files / net **−300 lines** (Step 6 deleted more than the rest added). The large overall diff is dominated by pre-existing branch content: the Storybook workbench (85 files + lockfile) and the generated `framework-candidates.generated.ts` (dev-UI autocomplete, regenerated from the new components).


kojiwakayama:
Pushed a rebased follow-up on `ui-storybook-workbench`.

What changed:
- Migrated the docs-agent template from `ChatWithSidebar` to `ConversationsProvider` + `ChatSidebar` + `Chat`, and synced `cli/templates/manifest.json`.
- Fixed `useUploadsRegistry` so endpoints that return no `items` array keep the local cache instead of clearing persisted uploads.
- Regenerated `docs/api-reference/veryfront/chat.md` and corrected the stale `StreamingMessage` README status item.
- Updated the Storybook workbench contract from `UploadsPanel` to `AttachmentsPanel`, restored the linked composition/primitives stories, and fixed the Attachment story JSX parser issue.
- Preserved the remote-only `useUploadsRegistry` `flushSync` de-flake commit before the final force push.

Verification:
- `deno test --allow-all src/chat/index.test.ts src/chat/upload-handler.test.ts src/react/components/chat/chat/hooks/use-conversations.test.ts src/react/components/chat/chat/hooks/use-conversation.test.tsx src/react/components/chat/chat/hooks/use-uploads-registry.test.tsx src/workflow/blob/veryfront-cloud-storage.test.ts` passed: 10 test modules, 45 steps.
- `deno task storybook:check` passed: 1 test module, 10 steps.
- `deno run --allow-read scripts/docs/validate-api-reference.ts` passed.
- `npm --prefix storybook run build-storybook` passed.
- `git diff --check` passed.
- The full local pre-push hook also passed once before the first push was rejected by a stale lease; after fetching and preserving the new remote test commit, I retried the push with a fresh lease.

kojiwakayama:
## Code review — 8 finder angles, 35 candidates adversarially verified

8 findings survived verification (most severe first). All were confirmed against the PR-head source.

### Findings

1. **`src/react/components/chat/chat/hooks/use-conversations.ts:190` — Debounced save can silently lose a conversation's last messages on thread switch.** `pendingSave` is a single shared ref and `scheduleSave` unconditionally overwrites it; `flushSave` runs only on hook unmount, and the provider-held hook doesn't unmount when switching threads. Switching from conversation A to B within the 300ms debounce window (e.g. right after A finishes streaming) means A's pending blob is never written, so A's final messages are gone on reload.

2. **`src/react/components/chat/chat/index.tsx:966` — Sending while an attachment is still uploading silently discards the file.** `submitWithAttachments` guards only on `chat.isLoading` and empty input; `canSubmit` is text-only. `attachmentsToFileParts` filters on `Boolean(a.url)`, so a still-uploading attachment is dropped from the send, and `upload.clear()` then aborts its in-flight XHR. The user sees their attached file vanish with no error.

3. **`src/chat/index.ts` — `ReasoningCard` and `ToolCallCard` were dropped from the `veryfront/chat` barrel: a silent breaking change.** Both were exported on main; the implementations still define back-compat aliases (`reasoning.tsx:215`, `tool-ui.tsx:453`) and `veryfront/react` still exports them (`public.ts`), but the chat barrel doesn't. `import { ToolCallCard } from "veryfront/chat"` now yields `undefined` and throws "Element type is invalid" at render.

4. **`src/react/components/chat/chat/hooks/use-upload.ts:88` — Guest-mode inline attachments have no size cap and silently kill conversation persistence.** `startInline` base64-encodes any file unconditionally (the 25MB cap is endpoint-only); the data URL is folded into the message and `JSON.stringify`'d into localStorage (`local-conversation-store.ts:99`), where quota errors are swallowed. One large guest attachment can exceed the ~5MB quota, after which every save of that conversation silently fails.

5. **`src/react/components/chat/chat/index.tsx:907` — Merely opening a conversation re-saves it with a fresh `updatedAt`, jumping it to the top of the sidebar.** The persist effect (deps `[chat.messages, boundId]`) has no first-run guard and no unchanged-messages check; `useChat` seeds messages synchronously from the bound conversation, so the effect fires on mount, stamps `Date.now()`, and the sidebar (sorted by `updatedAt` desc) reorders.

6. **`src/react/components/chat/chat/hooks/use-uploads-registry.ts:174` — Mount `refresh()` races a concurrent `add()` and drops the just-uploaded item.** `setItems(serverItems)` is a whole-list replace with no in-flight guard or merge; a file uploaded while the initial GET is pending vanishes when the stale response lands, and the persist effect then drops it from localStorage too.

7. **`src/react/components/chat/chat/hooks/use-conversations.ts:274` — Per-token setState plus unmemoized context value re-renders the whole conversation subtree on every streamed token.** `save()` runs `setSummaries` (full re-sort) and `setActive` per `chat.messages` change (only the disk write is debounced); the hook returns a fresh object each render and `ConversationsProvider` passes it as `value` without `useMemo` (`conversations-context.tsx:54-58`), so every consumer re-renders per token.

8. **`docs/guides/chat-ui.md:121` — Guide still uses `<Chat.Composer>`, which no longer exists.** The compound now attaches only `Root/MessageList/Input/Empty/Skeleton/If/Message/ErrorBanner`; the composer sub-component is `Chat.Input`. Copy-pasting the documented custom-layout example throws at render.

### Worth a quick pass (confirmed, below the severity cut)

- Drop-zone overlay can stick on (`use-drop-zone.ts:46` — counter desync on drag-cancel/drop-outside; no window-level reset).
- The dev projects shell lost its font: the sans stack now names Inter but nothing loads it, and the removed Gellix `@font-face` pointed at live CDN assets.
- `VeryfrontCloudBlobStorage.list()` costs 1+3N upstream requests and truncates to one page.
- Duplication in `chat/index.tsx`: the attachment orchestration and the `useChat`-to-flat-props mapping each exist twice (Controlled vs Uncontrolled paths) and have already diverged (`sendMessage` mapped in one, not the other).
- AGENTS.md em-dash ban violated in public JSDoc of `upload-handler.ts` and `agent-card.tsx`; dead plumbing: context `isAtBottom` is hardcoded `true`; `useConversations().bind` and sidebar `renderItem` have zero callers.

### Checked and cleared

The `#std` test imports in `scripts/` match the established convention there (all 24 pre-existing scripts tests use them); the `data:`-URL guard removal is intentional for guest attachments (the residual gap is the uncapped chat POST body); the router `query` hydration mismatch is pre-existing on main, not introduced here; the controlled-provider skeleton concern is the normal controlled-component contract.

kojiwakayama:
All 8 review findings above are addressed in 3a1b0606a (`chat: address PR review findings`): debounced-save flush on thread switch, submit blocked while attachments upload, `ReasoningCard`/`ToolCallCard` restored to the `veryfront/chat` barrel, 2MB inline-attachment cap, persist-effect mount guard (no more sidebar reorder on open), mid-refresh upload merge in `useUploadsRegistry`, memoized `useConversations` result with per-token churn gating, and the `<Chat.Composer>` → `<Chat.Input>` guide fix. Pre-push suite: 2273 passed / 0 failed.

kwakayama (dismissed):

kwakayama:
## Critical review — score: 84/100

Four parallel specialist passes (security/server, React core, build/supply-chain, API surface) over the full 267-file diff at head `f2a18a89`, every finding verified against source with file:line. Merge bar for this review was 90. Not merging yet.

### What's genuinely good (verified, not vibes)

- **Security hygiene is above average for an upload feature.** Fail-closed auth (`CONFIG_INVALID` unless `authorize` or explicit `allowUnauthenticated`), strict `SAFE_ID`/`SAFE_BLOB_ID_PATTERN` allowlists on both storage backends, filename sanitization, schema-validated upstream responses, and tests that explicitly cover traversal and auth rejection. Path traversal, SSRF, header injection, html-shell injection: all checked, all clean.
- **Supply chain clean.** 372 npm packages in the storybook lockfile, no typosquats, telemetry disabled, `private: true`, devDeps only, fully excluded from deno.json and the published package, `dist`/`node_modules` gitignored. `framework-candidates.generated.ts` is an honest regeneration with intact header.
- **The refactor's hard parts are done carefully.** Keyed remounts prevent cross-conversation message bleed, the ref-based emit effect avoids the render loops this kind of persistence refactor usually trips on, flat-prop `@deprecated` fallbacks are genuinely wired (not just typed), Thread retirement is complete (zero dangling references; the guide's migration table maps all 8 removed names correctly), and the export-surface test locks the new surface with exact equality.
- **Review responsiveness.** All 8 findings from the earlier adversarial review are actually fixed in `3a1b0606a` (verified in source, not just claimed), and 4 of the 5 Codex P2s are fixed on head (portalled attach menu, lockfile regenerated, attachment-only `canSubmit`, startup selection).

### P1 — fix before merge

1. **Deleted conversations resurrect.** `use-conversations.ts:356-364` — `remove()` calls `store.delete(id)` but never cancels the pending debounced save (`saveTimer`/`pendingSave` untouched). Delete a conversation during or within 300ms after streaming and the timer fires after the delete and re-persists it. Undeletable chat, silent. `flushSave`-style cancel of the pending blob when its id is removed closes it.
2. **"Nothing breaks now" is false for 6+ named exports.** `ChatComposer`, `MessageActions`, `UploadsPanel`, `StandaloneMessage`, `StreamingMessage`, `AttachmentPill` are hard-removed from `veryfront/chat` with no deprecated aliases, and `ChatComponents.Composer` is now `undefined` at runtime (renders throw "Element type is invalid"). The Thread trio removal is documented and intentional; these aren't mentioned. The barrel already has an alias section, one release of `export const ChatComposer = ChatInput`-style aliases makes the PR's own claim true at trivial cost. (Related P2: `MessageProps` silently changed shape under the same exported name.)

### P2 — should fix, cheap

3. `upload-handler.ts:131` — `request.formData()` buffers the whole body **before** the 25MB check, no Content-Length pre-check. Memory-exhaustion DoS; anonymous when `allowUnauthenticated: true`.
4. `upload-handler.ts:196-203` — GET serves the uploader-controlled `Content-Type` back with no `X-Content-Type-Options: nosniff` / `Content-Disposition`. Stored XSS on cookie-authenticated deployments. Two headers.
5. `showMessageActions={false}` silently regressed: the prop is accepted but never consumed (worked on main, isn't deprecated).
6. Deep link to a missing conversation id renders a permanent skeleton, no recovery path.
7. `agent={{ models, suggestions }}` only folds in controlled mode. In app mode the **deprecated** `models` prop works and the new API doesn't. Backwards from the migration guidance.
8. `useUploadsRegistry({ headers: {...} })` with an inline object identity-loops the fetch effect.
9. Custom `onSubmit` is bypassed whenever self-managed attachments ride along (goes straight to `chat.sendMessage`).
10. `use-chat.ts:344-349` — `reload()` rebuilds the last user turn from its text part only, so Regenerate drops attachments the user can still see in the transcript. (The one Codex P2 still live on head.)

### P3 — noted, not blocking

Project-wide boolean `authorize` (any authorized user can list/read/delete every other user's uploads) and unauthenticated agent enumeration look like intentional design but deserve documented opt-outs before this ships as a public SDK primitive. A→B→A thread switch inside the debounce window truncates the last reply's tail. Object-URL leak on unmount; errored attachments silently dropped from sends. Regenerate on an old turn regenerates the *last* turn. `useConversation(id)` reads the default store for non-active ids under a custom-store provider. `deno.lock` picked up unrelated drift (playwright 1.59→1.60, `ai` peer 6.0.216→6.0.214). The new invariants (`lint:chat-composability`, `storybook:check`, both new test files) aren't wired into CI, so they only trip on local verify. Storybook dev binds `0.0.0.0` with `fs.allow: [repoRoot]`.

### Scoring

Architecture, isolation, test discipline, and review responsiveness would put this at 92+. The two P1s (silent data-loss path in the flagship persistence feature; unaliased export removals contradicting the PR's own compat claim) plus the P2 cluster on exactly the new surfaces (uploads, attachments, migration precedence) pull it to **84**. No version bump here is fine, releases are separate PRs in this repo, which also means none of the breaking surface ships until then.

**Path to 90+:** fix the two P1s and the four cheapest P2s (#3 Content-Length pre-check, #4 two headers, #8 identity guard, #10 reload files). That's a focused half-day, and the persistence/attachment features go from demo-path-solid to actually trustworthy.

kojiwakayama:
Addressed the latest merge-blocking review feedback in `03672cdc9` plus generated candidate refresh in `b69a8b9f9`.

Fixed:
- Deleted conversations no longer resurrect from a pending debounced save.
- Deprecated `veryfront/chat` aliases restored for `ChatComposer`, `MessageActions`, `UploadsPanel`, `StandaloneMessage`, `StreamingMessage`, and `AttachmentPill`; `ChatComponents.Composer` aliases `Chat.Input`.
- Upload POST rejects oversized `Content-Length` before multipart parsing.
- Upload GET sends `Content-Disposition: attachment` and `X-Content-Type-Options: nosniff`.
- `useUploadsRegistry` no longer refetches on every render when `headers` is passed inline.
- `reload()` preserves file parts from the last user message.

Verification:
- `deno check src/chat/index.ts src/agent/react/use-chat/use-chat.ts src/react/components/chat/chat/hooks/use-uploads-registry.ts src/react/components/chat/chat/hooks/use-conversations.ts src/react/components/chat/chat/index.tsx`
- `deno run --allow-read scripts/docs/validate-api-reference.ts`
- Focused regression tests for upload handler, conversations, uploads registry, use-chat state, and chat barrel exports.
- Full pre-push suite passed: 2273 passed, 0 failed.

kwakayama:
## Re-review — score: 92/100 ✅ (was 84)

Re-verified head `b69a8b9f9` against every finding from the previous review. All six merge-blocking items are fixed in source, each with a test where it counts:

| Finding | Status | Evidence |
|---|---|---|
| P1: deleted conversations resurrect | **Fixed** | `discardPendingSave(id)` in `remove()` (`use-conversations.ts:361-362`) + new hook test |
| P1: unaliased export removals | **Fixed** | `ChatComposer`, `MessageActions`, `UploadsPanel`, `StandaloneMessage`, `StreamingMessage`, `AttachmentPill` all aliased in the `veryfront/chat` barrel and locked into the export-surface test; `Chat.Composer` restored as `@deprecated` |
| P2: upload DoS (full-body buffering) | **Fixed** | `rejectsDeclaredSize` Content-Length pre-check → 413 before `formData()` (`upload-handler.ts:136`) + tests |
| P2: stored XSS via served Content-Type | **Fixed** | `X-Content-Type-Options: nosniff` + `Content-Disposition: attachment` on GET (`upload-handler.ts:212-213`) |
| P2: `useUploadsRegistry` inline-headers fetch loop | **Fixed** | `stableHeadersKey` + ref (`use-uploads-registry.ts:103-117`) + tests |
| P2: `reload()` drops attachments | **Fixed** | `userMessagePayload` preserves file parts (`use-chat.ts:79-91`) + test |

Version deliberately bumped to 0.1.998 with `version-constant.ts` in the same commit, so the compat surface ships under an explicit release rather than silently.

### Still open (tracked, not blocking)

The remaining P2s from the first review are UX/precedence issues on the new surface, worth a follow-up pass: `showMessageActions={false}` accepted but unconsumed, missing-conversation deep link → permanent skeleton, `agent.models` folded only in controlled mode, custom `onSubmit` bypassed when self-managed attachments ride along. Plus the P3 set (per-user upload scoping docs, A→B→A debounce truncation, object-URL leak, regenerate-targets-last-turn, `useConversation` store fallback, deno.lock drift, new checks not CI-wired).

### Verdict

92/100. The data-loss path is closed, the compat story now matches the PR's claim, and the upload endpoint got the two-line hardening it needed. Merging once the pending checks (CodeQL, binary e2e) finish green.

kwakayama (approved):
Re-reviewed at head b69a8b9f9 — all merge-blocking findings from the 84/100 review verified fixed in source with tests (details in the re-review comment). Score: 92/100. Approving.

PR #2726, opened 2026-07-02
[codex] Fix OpenAI reasoning defaults

kwakayama:
## Summary
- Default native OpenAI/Veryfront Cloud reasoning-capable GPT-5 models to `medium`, including `gpt-5.4-nano`.
- Centralize OpenAI reasoning resolution for Chat and Responses so model/provider handling cannot drift.
- Exclude `gpt-5-chat-latest`, `gpt-5.1`, `o1-mini`, and `o1-preview` from default reasoning params; callers can still opt in explicitly with `reasoning`.
- Limit Responses API default `summary: "auto"` to `veryfront-cloud`; direct OpenAI/BYOK defaults now send only `effort`, while explicit `reasoning` still requests summaries.
- Keep compatible providers such as Azure/Moonshot free of default reasoning params unless explicitly requested.
- Decouple sampling-param drops from default-reasoning policy so o-series/GPT-5 models that reject sampling still drop temperature/top_p even when no default effort is sent.

Refs veryfront/veryfront-studio#5486
Refs veryfront/veryfront-studio#5258

## Red / Green transcript
Initial RED:
- `deno test -A extensions/ext-llm-openai/src/openai-reasoning-models.test.ts extensions/ext-llm-openai/src/openai-responses-request-builder.test.ts extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts`
- Failed as expected before the first fix: `gpt-5-chat-latest`, `o1-mini`, and Azure-compatible Chat requests were still getting `reasoning_effort: "medium"`; direct OpenAI Responses defaults still included `summary: "auto"`; Azure-compatible Responses requests still emitted default `reasoning`.

Review-feedback RED:
- Same command failed after adding the sampling-drop review tests because the new `rejectsOpenAISamplingParams` helper did not exist yet; the new assertions require `o1-mini`, Azure GPT-5.5, and explicit `reasoning: { enabled: false }` o-series requests to drop temperature while still omitting default reasoning params.

GREEN:
- Same command -> `3 passed (15 steps)`.

## Verification
- `deno test -A extensions/ext-llm-openai/src/openai-reasoning-models.test.ts extensions/ext-llm-openai/src/openai-responses-request-builder.test.ts extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts` -> `3 passed (15 steps)`.
- `deno test -A extensions/ext-llm-openai/src/openai-provider.test.ts extensions/ext-llm-openai/src/openai-chat-stream.test.ts extensions/ext-llm-openai/src/openai-responses-stream.test.ts` -> `3 passed (48 steps)`.
- `deno fmt --check` on touched OpenAI TS files -> `Checked 6 files`.
- `deno lint` on touched OpenAI TS files -> `Checked 6 files`.
- `git diff --check` -> clean.
- GitHub PR checks on `36572662c` -> passed: format, lint, typecheck, unit, integration, binary e2e, RSC browser e2e, npm install smoke, coverage shards/gate, CodeQL, dependency audit, CLA.

## Local broad check note
- Local `deno task verify:quick` still fails on unrelated pre-existing CLI boundary violations after manifest, fmt, lint, and style checks pass: `cli/shared/ensure-content-processor.ts`, `cli/shared/default-contracts.ts`, and `cli/commands/knowledge/parser.ts` deep-import `#veryfront/extensions/first-party-import.ts`.

## Staging proof
This branch is not deployed to staging yet. After staging deploy, proof should be: run a `veryfront-cloud/openai/gpt-5.4-nano` or `veryfront-cloud/openai/gpt-5.5` chat and verify the run events endpoint contains `REASONING_MESSAGE_START`, `REASONING_MESSAGE_CONTENT`, and `REASONING_MESSAGE_END` before `RUN_FINISHED`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.987",
+  "version": "0.1.988",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": {
--- a/extensions/ext-llm-openai/README.md
+++ b/extensions/ext-llm-openai/README.md
@@ -57,7 +57,7 @@ Any model accessible through the OpenAI Chat Completions, Responses, or Embeddin
 
 - **Flagship:** `gpt-4.1`, `gpt-4.1-mini`, `gpt-4.1-nano`, `gpt-4o`, `gpt-4o-mini`
 - **Frontier:** `gpt-5`, `gpt-5-mini`, `gpt-5-nano`
-- **Reasoning:** `o3`, `o4-mini`, `o1`, `o1-mini`, `o3-mini` (sampling parameters are automatically dropped with warnings)
+- **Reasoning:** `gpt-5.4-nano`, current `gpt-5`/`gpt-5.x` reasoning models, `o3`, `o4-mini`, `o1`, `o3-mini` (sampling parameters are automatically dropped with warnings)
 - **Embeddings:** `text-embedding-3-small`, `text-embedding-3-large`
 - **OpenAI-compatible:** Any third-party model reachable via an OpenAI-compatible endpoint (set `OPENAI_BASE_URL`)
 
@@ -74,7 +74,11 @@ The extension accepts configuration through `LLMProviderConfig` when creating ru
 
 ## Model-Specific Behavior
 
-### Reasoning Models (o3, o4-mini, o1)
+### Reasoning Models (GPT-5.x, o3, o4-mini, o1)
+
+Default reasoning params are applied only for native `openai` and `veryfront-cloud` providers.
+OpenAI-compatible providers require explicit `reasoning` options. `gpt-5-chat-latest`,
+`gpt-5.1`, `o1-mini`, and `o1-preview` are left unmodified by default.
 
 Reasoning models automatically:
 
--- a/extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts
+++ b/extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts
@@ -80,11 +80,11 @@ describe("ext-llm-openai/openai-chat-request-builder", () => {
     );
 
     assertEquals(body.reasoning_effort, undefined);
-    assertEquals(body.temperature, 0.2);
-    assertEquals(warnings.drain(), []);
+    assertEquals(body.temperature, undefined);
+    assertEquals(warnings.drain().map((warning) => warning.setting), ["temperature"]);
   });
 
-  it("does not set default reasoning effort for OpenAI-compatible providers", () => {
+  it("does not set default reasoning effort for OpenAI-compatible providers but still drops rejected sampling params", () => {
     const warnings = createWarningCollector();
 
     const body = buildOpenAIChatRequest(
@@ -99,8 +99,28 @@ describe("ext-llm-openai/openai-chat-request-builder", () => {
     );
 
     assertEquals(body.reasoning_effort, undefined);
-    assertEquals(body.temperature, 0.2);
-    assertEquals(warnings.drain(), []);
+    assertEquals(body.temperature, undefined);
+    assertEquals(warnings.drain().map((warning) => warning.setting), ["temperature"]);
+  });
+
+  it("drops rejected sampling params when explicit reasoning is disabled", () => {
+    const warnings = createWarningCollector();
+
+    const body = buildOpenAIChatRequest(
+      "o3-mini",
+      "openai",
+      {
+        prompt: [{ role: "user", content: [{ type: "text", text: "Be concise." }] }],
+        reasoning: { enabled: false },
+        temperature: 0.2,
+      },
+      true,
+      warnings,
+    );
+
+    assertEquals(body.reasoning_effort, undefined);
+    assertEquals(body.temperature, undefined);
+    assertEquals(warnings.drain().map((warning) => warning.setting), ["temperature"]);
   });
 
   it("preserves chat request shaping, provider option merge order, and warnings", () => {
--- a/extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts
+++ b/extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts
@@ -27,6 +27,25 @@ function createWarningCollector() {
 }
 
 describe("ext-llm-openai/openai-chat-request-builder", () => {
+  it("sets default reasoning effort for GPT-5.5 chat requests", () => {
+    const warnings = createWarningCollector();
+
+    const body = buildOpenAIChatRequest(
+      "gpt-5.5",
+      "openai",
+      {
+        prompt: [{ role: "user", content: [{ type: "text", text: "Think carefully." }] }],
+        temperature: 0.2,
+      },
+      true,
+      warnings,
+    );
+
+    assertEquals(body.reasoning_effort, "medium");
+    assertEquals(body.temperature, undefined);
+    assertEquals(warnings.drain().map((warning) => warning.setting), ["temperature"]);
+  });
+
   it("preserves chat request shaping, provider option merge order, and warnings", () => {
     const prompt: RuntimePromptMessage[] = [
       { role: "system", content: "You are concise." },
--- a/extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts
+++ b/extensions/ext-llm-openai/src/openai-chat-request-builder.test.ts
@@ -46,6 +46,63 @@ describe("ext-llm-openai/openai-chat-request-builder", () => {
     assertEquals(warnings.drain().map((warning) => warning.setting), ["temperature"]);
   });
 
+  it("does not set default reasoning effort for GPT-5 chat snapshots", () => {
+    const warnings = createWarningCollector();
+
+    const body = buildOpenAIChatRequest(
+      "gpt-5-chat-latest",
+      "openai",
+      {
+        prompt: [{ role: "user", content: [{ type: "text", text: "Be concise." }] }],
+        temperature: 0.2,
+      },
+      true,
+      warnings,
+    );
+
+    assertEquals(body.reasoning_effort, undefined);
+    assertEquals(body.temperature, 0.2);
+    assertEquals(warnings.drain(), []);
+  });
+
+  it("does not set default reasoning effort for legacy o1 chat variants", () => {
+    const warnings = createWarningCollector();
+
+    const body = buildOpenAIChatRequest(
+      "o1-mini",
+      "openai",
+      {
+        prompt: [{ role: "user", content: [{ type: "text", text: "Be concise." }] }],
+        temperature: 0.2,
+      },
+      true,
+      warnings,
+    );
+
+    assertEquals(body.reasoning_effort, undefined);
+    assertEquals(body.temperature, 0.2);
+    assertEquals(warnings.drain(), []);
+  });
+
+  it("does not set default reasoning effort for OpenAI-compatible providers", () => {
+    const warnings = createWarningCollector();
+
+    const body = buildOpenAIChatRequest(
+      "gpt-5.5",
+      "azure",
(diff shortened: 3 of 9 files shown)


kwakayama:
## Critical review — score: 72/100

Small, well-shaped diff. TDD transcript, version bump pair (`deno.json` + `version-constant.ts`) per repo convention, and the stream side (`response.reasoning_summary_text.delta` in `openai-responses-stream.ts:169`) already handles what this turns on. The mechanics are clean. The risk management around a default-on network behavior change is where points come off.

### Findings

**1. [P1] `summary: "auto"` by default can 400 for unverified OpenAI orgs (BYOK users).** `openai-responses-request-builder.ts:276` sends `reasoning: { effort, summary: "auto" }` whenever an effort resolves. That line pre-existed, but it only fired when a caller explicitly enabled reasoning. This PR makes it fire on **every** `gpt-5*` Responses request by default. OpenAI rejects `reasoning.summary` for organizations that haven't completed identity verification ("Your organization must be verified to generate reasoning summaries"). veryfront-cloud's own org is presumably verified, but BYOK users' orgs may not be — their gpt-5.x runs go from working to 400 with no code change on their side. Escape hatches exist (`reasoning: { enabled: false }`, `providerOptions` override via the `Object.assign` at line 304), but that's a support ticket, not a default. Consider only defaulting `summary: "auto"` on the veryfront-cloud path, or catching the verification error and retrying without `summary`.

**2. [P2] This silently changes the platform default model.** `getDefaultOpenAIReasoningEffort` matches `gpt-5.4-nano` (the `DEFAULT_VERYFRONT_CLOUD_MODEL_ID` in `src/provider/veryfront-cloud/model-catalog.ts:20`) via the `>= 2` version rule. Every default-model request now carries `effort: "medium"` + `summary: "auto"` and drops temperature/top_p. If OpenAI's server-side default for nano-tier models is lower than medium, this is a fleet-wide latency and cost increase that the PR description (which only mentions gpt-5.5) never surfaces. If intentional, say so in the PR body; if not, gate the default by model tier.

**3. [P2] o1-preview / o1-mini regression on the Chat path.** `/^o[134](-|$)/` matches `o1-mini` and `o1-preview`, so the chat builder now sends `reasoning_effort: "medium"` for them by default (`openai-chat-request-builder.ts:121-124` → line 189). Those models reject `reasoning_effort` (only o1-full and o3-mini+ accept it) — previously no effort was sent unless explicitly enabled, so this turns working requests into 400s. Probably low traffic given deprecation, but it's a regression the tests don't cover.

**4. [P3] The `gpt-5.1` exclusion looks like a bug without a comment.** `openai-reasoning-models.ts:18-21`: `/^gpt-5\.(\d+)/` with `>= 2` deliberately skips `gpt-5.1` (and a hypothetical `gpt-5.0`), while `/^gpt-5(-|$)/` catches the base model. If the intent is "5.1 defaults to reasoning `none` upstream, so don't force effort," write that in a comment — the next person to touch this file will "fix" the off-by-one. Note `gpt-5.1` appears in tests elsewhere in the repo (`agent-run-lifecycle.test.ts:133`), so it's a live id, not hypothetical.

**5. [P3] Azure/Moonshot compatible endpoints share this builder.** `openai-provider.ts` header says this runtime serves "OpenAI-compatible endpoints (Azure OpenAI, Moonshot AI)" and the default keys purely on `modelId`. Azure gpt-5 deployments need separate registration for reasoning summaries; a compatible gateway may reject `reasoning_effort` outright. Same failure class as #1, different tenant.

**6. [P3] Test matrix is thin for a default-behavior change.** Two new tests, both gpt-5.5 happy path. Missing: gpt-5.1 excluded, gpt-5 base included, o1/o3 chat default, explicit `enabled: false` suppressing the default, and a direct unit test on `openai-reasoning-models.ts` (the regex matrix is exactly the kind of thing that wants a table test).

**7. [P3, nit] `resolveOpenAIReasoningEffort` is now duplicated verbatim** in both builders and both were modified identically. It belongs next to its friends in `openai-reasoning-models.ts`.

**8. [P3, nit] PR description overstates one change.** "Request Responses API reasoning summaries with `summary: 'auto'`" — that line already existed on main; the PR changes *when* it fires, not *what* is sent. Matters for anyone bisecting a summary-related incident later.

### What's good
- Extracting `openai-reasoning-models.ts` as the single source for model detection (and lowercasing the id) is the right refactor.
- Default resolution semantics are sensible: no option → default; explicit `enabled: false` → off; explicit effort wins.
- No synthesized reasoning events when the provider returns none — correct restraint.
- Version bump pair and RED/GREEN transcript follow repo convention.

### Scoring
| Dimension | Score |
|---|---|
| Correctness of the code as written | 9/10 |
| Blast-radius / rollout safety | 5/10 |
| Test coverage for the behavior change | 5/10 |
| Code quality & structure | 8/10 |
| PR hygiene (description, convention, verification) | 8/10 |

**72/100.** The code does what it says. The problem is what it doesn't say: this changes the default network behavior of every gpt-5.x request — including the platform default model and BYOK/Azure tenants — and the failure mode for unverified orgs is a hard 400. Fix #1 (scope the summary default or handle the verification error) and add the missing regex-matrix tests before merge; #2 needs a one-line intentionality statement in the PR body.

*Review posted by Claude Code on request.*

kwakayama:
## Re-review (after 6a4e2b6d) — score: 84/100

All 8 findings from the first review were addressed, and addressed properly — not cosmetically. Verified in code, not just the PR body:

- **#1 (summary 400s for BYOK)**: fixed. `shouldRequestOpenAIReasoningSummary` limits default `summary: "auto"` to `veryfront-cloud`; direct OpenAI defaults send only `effort`; explicit opt-in still gets summaries. Tests cover all three paths.
- **#2 (gpt-5.4-nano)**: now stated in the PR body and tested by name. Intentional, documented, fine.
- **#3 (o1-mini/o1-preview default effort)**: excluded via `/^o1($|-\d)/`. Effort no longer sent. See the new finding below, though.
- **#4 (gpt-5.1)**: comment added explaining the upstream-defaults-to-none rationale. `gpt-5-chat-latest` exclusion is a good catch beyond the review.
- **#5 (Azure/Moonshot)**: provider-gated via `supportsDefaultReasoningParams`, with tests.
- **#6 (test matrix)**: `openai-reasoning-models.test.ts` table test added, exactly what was asked.
- **#7 (duplication)**: `resolveOpenAIReasoningConfig` centralized. The `source: "default" | "explicit"` discriminator is a clean way to carry intent to the summary decision.
- **#8 (PR body)**: rewritten, now accurate.

### One new finding introduced by the fix

**[P2] The sampling-param drop is now coupled to "we send reasoning params," and that broke an invariant main had.** Both builders now compute `reasoningEnabled = reasoning !== undefined` — the old `isReasoningModel || ...` term is gone. Consequences, each confirmed in the new tests as *expected* behavior:

- `o1-mini` / `o1-preview` on the Chat path: no default effort (correct) → `dropSamplingParams` false → `temperature: 0.2` is now **sent**. On main these matched `/^o[134](-|$)/` and temperature was dropped with a warning. The o1 family rejects `temperature` ≠ 1, so this turns a warned-and-dropped param into an API 400. The new test at `openai-chat-request-builder.test.ts` ("legacy o1 chat variants") asserts `body.temperature === 0.2` — it enshrines the regression.
- Same shape for `azure` + `o1`/`o3-mini`: main's regex didn't gate by provider, so temperature was dropped for Azure o-series too; now the provider gate turns off both the effort param **and** the sampling drop. Azure deployments of the same models reject the same params.
- Edge: explicit `reasoning: { enabled: false }` on `o3` now sends temperature (main dropped it unconditionally for o-series).

The root cause is conflating two independent facts: "this model rejects sampling params" (a property of the model, regardless of provider or whether we send an effort param) and "we should send default reasoning params" (a provider+model policy). Suggested fix: keep a provider-agnostic `rejectsSamplingParams(modelId)` in `openai-reasoning-models.ts` covering the o-series (including `o1-mini`/`o1-preview`/`o1-pro`) and reasoning-capable gpt-5.x, and use it for the drop decision, while `resolveOpenAIReasoningConfig` keeps owning the effort/summary decision. `gpt-5.1` and `gpt-5-chat-latest` stay out of the reject set (they accept temperature when reasoning is off), so the new exclusions remain correct.

### Scoring
| Dimension | Before | Now |
|---|---|---|
| Correctness of the code as written | 9/10 | 8/10 (new o1-legacy/Azure sampling regression) |
| Blast-radius / rollout safety | 5/10 | 9/10 |
| Test coverage | 5/10 | 9/10 (matrix is great; two tests assert the wrong expectation) |
| Code quality & structure | 8/10 | 9/10 |
| PR hygiene | 8/10 | 10/10 |

**84/100.** This went from "clean code, risky rollout" to "safe rollout, one behavioral loose end." If o1-mini/o1-preview and Azure o-series traffic is genuinely zero for veryfront, this is mergeable as-is with the fix as a follow-up; otherwise decouple the sampling-drop check first — it's a ~15-line change plus flipping the two test assertions.

*Re-review posted by Claude Code on request.*

kwakayama:
## Re-review #2 (after 36572662) — score: 92/100

The P2 from the last round is fixed, and fixed the right way. Verified in code:

- **`rejectsOpenAISamplingParams(modelId)`** in `openai-reasoning-models.ts` is provider-agnostic and covers exactly the right set: `/^o[134]($|-)/` (which now correctly includes `o1-mini`, `o1-preview`, and `o1-pro`) plus reasoning-capable gpt-5.x via the shared `isReasoningCapableGpt5` helper. `gpt-5.1` and `gpt-5-chat-latest` stay out of the reject set — correct, they accept temperature when reasoning is off.
- **Both builders** now compute `dropSamplingParams = reasoningEnabled || samplingRejected (|| fixedSampling)`, restoring the invariant main had while keeping the provider-gated reasoning defaults from the previous commit. The Responses builder got the same treatment, which main never had for non-o-series models — so this is strictly better than the pre-PR state, not just a revert.
- **The two tests that enshrined the regression were flipped**, not deleted: `o1-mini` and `azure`+`gpt-5.5` now assert temperature is dropped with a warning. New coverage for the edge I flagged (`reasoning: { enabled: false }` on `o3-mini` still drops sampling params but sends no reasoning field) on both API paths, plus a 14-case matrix test on the reject set including `o1-pro` and a `gpt-4o-mini` negative.
- **Warning messages** now distinguish "this model rejects X" from "reasoning was active" — good observability detail nobody asked for.
- CI is green on `36572662` (only release-gated jobs skipping).

### Remaining (nits, not blockers)

1. **[P3] `isOpenAIReasoningModel` is now a dead export** — zero callers in source or tests after the builders switched to `resolveOpenAIReasoningConfig`. Delete it, or it becomes a trap: its provider-gated semantics ("would we send default reasoning params") no longer match what its name implies ("is this a reasoning model"), which is precisely the confusion this commit just untangled. Someone importing it later for a sampling or capability check reintroduces the bug class.
2. **[P3] `isFixedSamplingModel` (kimi) still lives only in the chat builder** while every other model-capability predicate now lives in `openai-reasoning-models.ts`. Fine today; worth consolidating next time someone touches this.

### Scoring
| Dimension | r1 | r2 | Now |
|---|---|---|---|
| Correctness | 9/10 | 8/10 | 10/10 |
| Blast-radius / rollout safety | 5/10 | 9/10 | 9/10 |
| Test coverage | 5/10 | 9/10 | 10/10 |
| Code quality & structure | 8/10 | 9/10 | 9/10 (dead export) |
| PR hygiene | 8/10 | 10/10 | 9/10 (PR body doesn't mention the decoupling commit) |

**92/100 — ship it.** The two nits are follow-up material, though deleting the dead `isOpenAIReasoningModel` export is a 30-second change worth doing before merge. The staging proof step from the PR body (verify `REASONING_MESSAGE_*` events on a `veryfront-cloud/openai/gpt-5.4-nano` run) is still the right post-deploy check.

*Re-review posted by Claude Code on request.*

kojiwakayama (approved):

PR #2743, opened 2026-07-03
fix(router): seed route params on client hydration + preserve catch-all segments

### LINKED ISSUE #2741: Route params (useRouter().params) are empty after client hydration
## Problem

For app-router (`app/`) pages, `useRouter().params` is **empty after client hydration**, even though the route match is correct on the server and the params are present in the hydration data.

## Repro (verified live in `veryfront dev`)

`app/posts/[id]/page.tsx`:
```tsx
"use client";
import { useRouter } from "veryfront/router";
export default function Page() {
  const r = useRouter();
  return <span id="id">{r.params.id ?? "none"}</span>;
}
```

Visit `/posts/42`:
- **SSR HTML**: `id: 42` ✅
- **After hydration**: `id: none` ❌
- Hydration data script *does* contain `"params": { "id": "42" }`.

Query params (`?tab=`) work fine on the client — only route params are lost.

## Root cause

The dev hydration path (`__veryfrontRenderPage`, `src/html/hydration-script-builder/templates/renderer.ts`) seeds `RouterProvider` with a `router` object that is **not populated with `params`** (line ~154: `React.createElement(RouterProvider, { router, ... })`, where `router` carries no params). So `useRouter().params` falls back to `{}`.

## Pre-existing

Reproduces identically on `main` (before #2740). Not a regression from the reactive-router work — that PR is scoped to query/pathname reactivity (#2739).

## Expected

After hydration, `useRouter().params` reflects the route match (`{ id: "42" }`), consistent with SSR and with the `wrapForHydration` (rsc-module) path that already seeds params.

### LINKED ISSUE #2742: Catch-all route params truncated to first segment on SSR
## Problem

Catch-all routes (`[...slug]`) render only the **first** path segment in `useRouter().params` on the server.

## Repro (verified live)

`app/docs/[...slug]/page.tsx` reading `useRouter().params.slug`, visit `/docs/guides/intro`:
- **SSR HTML**: `slug: guides` ❌ (should be `guides/intro`)
- Hydration data `slug` field is the full `docs/guides/intro`, but the router params only carry `guides`.

## Pre-existing

Reproduces identically on `main`. Separate from #2740 (which fixed the *client* hydration normalizer to join catch-all segments, `src/rendering/rsc/hydration-router.ts`). The **server-side** params seeding for the app-router render still truncates.

## Expected

`useRouter().params.slug === "guides/intro"` (joined) on both SSR and client, so there is no SSR/client divergence.

mattboon:
Fixes #2741, fixes #2742.

## Problem

Two pre-existing route-param bugs (reproduce on `main`, independent of the reactive-router work in #2740):

1. **#2741** — `useRouter().params` is **empty after client hydration**. The client router object hardcoded `params: {}`, dropping the route match even though it's in the hydration data.
2. **#2742** — catch-all `[...slug]` params are **truncated to the first segment** on SSR (`/docs/guides/intro` → `slug: "guides"`), because the SSR flatteners used `value[0]`.

## Fix

- **`flattenRouteParams`** (new, `src/routing/flatten-route-params.ts`) — joins catch-all array segments with `/`, matching the client SPA normalizer and the RSC hydration normalizer so server and client agree. Used by `layout-applicator` and `script-page-handling` (replacing two divergent `value[0]` copies).
- **Client router seed** (`templates/router.ts`) — populate `params` from the hydration data (joining catch-all arrays) instead of `{}`.

## Verified live (`veryfront dev`)

| | before (`main`) | after |
|---|---|---|
| `/posts/42` client `params.id` | `none` | `42` |
| `/docs/guides/intro` SSR `slug` | `guides` | `guides/intro` |
| `/docs/guides/intro` client `slug` | `none` | `guides/intro` |

Query params + client-nav reactivity are unchanged (that's #2740's scope).

## Tests

- `src/routing/flatten-route-params.test.ts` — red→green unit test: catch-all `["guides","intro"]` → `"guides/intro"` (the buggy `value[0]` version fails it).


kwakayama (dismissed):

kojiwakayama (dismissed):

kwakayama (commented):
Critical review score: 82/100

Recommendation: mergeable after tightening tests. I did not find a confirmed runtime regression in the patch, and the focused tests pass locally, but the test coverage around the fixed SSR paths is currently misleading.

Findings

1. Medium - Adjacent SSR tests still encode the old, broken catch-all contract.

Problem: this PR changes production SSR param flattening from `value[0]` to `value.join("/")`, but the nearby SSR tests still use copied local helper logic that truncates arrays to the first element. That means the suite now contains tests that pass while documenting behavior opposite to #2742, and those tests would not catch the same regression returning in these paths.

Evidence:
- Production helper now joins catch-all arrays: `src/routing/flatten-route-params.ts:10-18`.
- Production SSR consumers use that helper: `src/rendering/layouts/layout-applicator.ts:123`, `src/rendering/layouts/layout-applicator.ts:128`, `src/rendering/layouts/layout-applicator.ts:153`, `src/rendering/script-page-handling.ts:132`, `src/rendering/script-page-handling.ts:135`.
- Test-local copies still truncate arrays: `src/rendering/script-page-handling.test.ts:50-54` and `src/rendering/layouts/layout-applicator.test.ts:30-34`.
- One test still explicitly expects first-segment behavior: `src/rendering/script-page-handling.test.ts:168-170` asserts `{ tags: "a" }` for `['a', 'b']`.
- Another layout test passes with `id: ['123', 'ignored']` and expects only `"123"`: `src/rendering/layouts/layout-applicator.test.ts:126-131`.

Reproduction:
```
VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/rendering/script-page-handling.test.ts src/rendering/layouts/layout-applicator.test.ts
```
Result: passes locally, including the stale `should flatten array params to first element` assertion. Update these tests to expect joined catch-all values such as `{ tags: "a/b" }`, or better, test the shared `flattenRouteParams` helper / real production path instead of copied flattening logic.

2. Low - Generated router behavior is mostly protected by string-token assertions, not executable behavior.

Problem: the PR changes initial router params plus SPA/popstate param updates in generated browser code, but the added tests only assert that particular substrings exist. A future edit can preserve those tokens while breaking execution order, hydration-data parsing, or the object actually passed to `RouterProvider`.

Evidence:
- Added tests assert substrings only: `src/html/hydration-script-builder/templates/router.test.ts:320-335`.
- The actual runtime behavior lives in generated script code at `src/html/hydration-script-builder/templates/router.ts:614`, `src/html/hydration-script-builder/templates/router.ts:960-1004`, and `src/html/hydration-script-builder/templates/router.ts:1033`.
- `renderer.ts` passes the top-level `router` into `RouterProvider` at `src/html/hydration-script-builder/templates/renderer.ts:153-154`, so an executable test should verify the generated module-level binding, not just string presence.

Reproduction:
```
VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/html/hydration-script-builder/templates/router.test.ts
```
Result: passes locally, but it only proves the template contains the expected strings. Add a DOM/vm-style test that evaluates the generated hydration runtime with `veryfront-hydration-data` containing `{ slug: ["guides", "intro"] }`, then asserts the router passed/exposed has `params.slug === "guides/intro"` and that SPA/popstate page data replaces stale params.

Local verification performed:
```
VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/routing/flatten-route-params.test.ts src/html/hydration-script-builder/templates/router.test.ts
```
Result: `ok | 2 passed (60 steps) | 0 failed`.

```
git diff --check origin/main...HEAD
```
Result: no whitespace errors.

No conflict markers found in the PR worktree.

kwakayama (changes requested):
Superseding my earlier review after a deeper source check.

Critical review score: 58/100

Recommendation: request changes. The server-side flattening helper is a good direction and the focused tests pass, but this PR still leaves a functional SPA/popstate params bug and regresses normal unit-test coverage.

Findings

1. High - SPA and popstate renders can still receive stale route params.

Problem: `navigateSPA` and the `popstate` handler update `window.__veryfrontRouter.params` only after rendering the new React tree. `renderPageFromData()` wraps the new tree in `RouterProvider` with the shared `router` object, and `RouterProvider` reads `seed.params` during render. That ordering means a client-side navigation from a dynamic route to another route can render with the previous route's params, then mutate the router object too late for that render.

Evidence:
- SPA path renders first at `src/html/hydration-script-builder/templates/router.ts:607-608`, then updates `window.__veryfrontRouter.params` at `src/html/hydration-script-builder/templates/router.ts:611-614`.
- Popstate path renders first at `src/html/hydration-script-builder/templates/router.ts:1027-1029`, then updates params at `src/html/hydration-script-builder/templates/router.ts:1030-1033`.
- The rendered tree gets `RouterProvider` with the shared `router` object inside `renderPageFromData` at `src/html/hydration-script-builder/templates/router.ts:744-745`.
- `RouterProvider` reads params from the seed router during render at `src/react/runtime/core.ts:264-291` and memoizes on `seed.params` at `src/react/runtime/core.ts:286-302`.
- `PageContextProvider` also derives `params` from that router context at `src/react/runtime/core.ts:400-405`, so this affects both `useRouter().params` and `usePageContext().params`.

Reproduction/code path:
1. Start on `/posts/42`, where the global router has `{ id: "42" }`.
2. SPA navigate to `/docs/guides/intro`, whose page data has `{ slug: ["guides", "intro"] }`.
3. `navigateSPA()` calls `renderPageFromData(pageData, targetPath)` before line 614 updates the router params.
4. During that render, `RouterProvider` sees the old `seed.params`, so route consumers render with `{ id: "42" }` instead of `{ slug: "guides/intro" }`.
5. The later mutation at line 614 happens after the render and does not by itself cause another provider render with the corrected params.

Fix: compute normalized next params before rendering and update the router snapshot before `renderPageFromData()`, or pass a fresh per-render router object with `{ pathname, query, params }` into `RouterProvider`. Cover this with an executable SPA/popstate regression test, not only string assertions.

2. High - `test:unit` and `test:coverage:unit` now exclude all `.test.tsx` tests.

Problem: the release/version commit changes the unit-test file glob from `*.test.ts*` to `*.test.ts`. That drops every TSX unit test from the standard unit and coverage-unit tasks, including tests directly relevant to this PR's router hydration surface.

Evidence:
- `deno.json:396` changed `test:unit` to `find src cli -name '*.test.ts' ! -name '*.integration.test.ts'`.
- `deno.json:401` made the same change for `test:coverage:unit`.
- The previous glob was `*.test.ts*`, which included `.test.tsx`.
- This worktree has 9 `.test.tsx` files that the new glob skips, including `src/react/runtime/router-provider.test.tsx` and `src/rendering/rsc/hydration-router.test.tsx`.

Reproduction:
```
find src cli -name '*.test.tsx' | sort
```
This lists 9 tests, including the router provider and RSC hydration-router tests. None match the new `find src cli -name '*.test.ts' ...` task glob, so they no longer run under `deno task test:unit` or `deno task test:coverage:unit`.

Fix: restore `*.test.ts*` plus the matching `*.integration.test.ts*` exclusion, or use an explicit grouped expression that includes both `*.test.ts` and `*.test.tsx` while excluding both integration suffixes.

3. Medium - Adjacent SSR tests still encode the old, broken catch-all contract.

Problem: this PR changes production SSR param flattening from `value[0]` to `value.join("/")`, but nearby SSR tests still use copied local helper logic that truncates arrays to the first element. The suite now contains passing tests that document behavior opposite to #2742.

Evidence:
- Production helper joins catch-all arrays at `src/routing/flatten-route-params.ts:10-18`.
- Production SSR consumers use it at `src/rendering/layouts/layout-applicator.ts:123`, `src/rendering/layouts/layout-applicator.ts:153`, and `src/rendering/script-page-handling.ts:132-135`.
- Test-local copies still truncate arrays at `src/rendering/script-page-handling.test.ts:50-54` and `src/rendering/layouts/layout-applicator.test.ts:30-34`.
- `src/rendering/script-page-handling.test.ts:168-170` explicitly expects `{ tags: "a" }` for `['a', 'b']`.

Reproduction:
```
VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/rendering/script-page-handling.test.ts src/rendering/layouts/layout-applicator.test.ts
```
Result locally: passes, including the stale first-segment assertion. Update these tests to expect joined catch-all values or route them through the shared `flattenRouteParams` helper/real production behavior.

4. Low - Generated router changes are still mostly string-token tested.

Problem: the added template tests assert that substrings exist, but do not execute the generated browser script to prove hydration seeding, SPA updates, and popstate updates work with real `RouterProvider` consumers.

Evidence:
- Added tests are string assertions at `src/html/hydration-script-builder/templates/router.test.ts:320-335`.
- The actual runtime-sensitive changes are at `src/html/hydration-script-builder/templates/router.ts:614`, `src/html/hydration-script-builder/templates/router.ts:960-1004`, and `src/html/hydration-script-builder/templates/router.ts:1033`.

Reproduction:
```
VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/html/hydration-script-builder/templates/router.test.ts
```
Result locally: passes, but it only proves the template contains expected strings. Add a DOM/vm-style test that evaluates the generated runtime with hydration data and then simulates SPA/popstate page data changes.

Local verification performed:
```
VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/routing/flatten-route-params.test.ts src/html/hydration-script-builder/templates/router.test.ts
```
Result: `ok | 2 passed (60 steps) | 0 failed`.

```
VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net src/rendering/script-page-handling.test.ts src/rendering/layouts/layout-applicator.test.ts
```
Result: `ok | 2 passed (65 steps) | 0 failed`, but with stale test-local expectations described above.

```
git diff --check origin/main...HEAD
```
Result: no whitespace errors.

No conflict markers found in the PR worktree.

kojiwakayama:
## Code review

The core fix is sound and well-tested — the executable runtime tests for the generated router script are a genuine improvement. Verification confirmed one likely-accidental CI regression, one missed flattener site, and several places where the "server and client agree" goal isn't fully achieved yet.

### Findings (most severe first)

**1. `deno.json:396` — the test glob change silently disables all `.test.tsx` suites in CI** 🔴

The `test:unit`/`test:coverage:unit` globs changed from `-name '*.test.ts*'` to `-name '*.test.ts'`, an exact revert of 7101a1071 ("test: run .tsx tests in CI"). Nine suites stop running — including `src/rendering/rsc/hydration-router.test.tsx`, `src/client/spa/ClientApp.reactivity.test.tsx`, and `src/react/runtime/router-provider.test.tsx`, which cover exactly the router/params area this PR touches — and nothing else in CI picks them up (the integration lane only runs `tests/`, and the coverage shard filter also rejects `.tsx`). Looks like a rebase artifact; should be reverted before merge.

**2. `src/rendering/page-rendering.ts:183` — one `value[0]` flattener was missed, so #2742 persists for MDX `generateMetadata`**

The inline `Array.isArray(v) ? v[0] : v` feeding `generateMetadata` in `handleMDXPage` was not migrated. On `/docs/[...slug]` at `/docs/guides/intro`, `generateMetadata` receives `slug: "guides"` while the page context and router now get `"guides/intro"` — wrong titles/canonical URLs for deep catch-all pages. This was the only remaining missed site; it belongs in this PR since it's the same bug.

**3. `src/routing/flatten-route-params.ts:17` — silent behavior change for empty catch-all arrays**

An optional catch-all `[[...slug]]` matched at its base path produces `{ slug: [] }`; the old flatteners dropped the key entirely (`value[0]` → `undefined` → filtered), while `[].join('/')` now yields `{ slug: "" }`. Code branching on `params.slug === undefined` or `'slug' in params` flips behavior on SSR of index routes. The change does align with the RSC/API/client normalizers, so it may be intentional — but it's untested (no empty-array case in `flatten-route-params.test.ts`) and worth a test pinning the contract either way.

**4. `templates/router.ts:714,737` + `templates/renderer.ts:115,146` — page props and `usePageContext().params` still receive raw catch-all arrays on the client** *(pre-existing)*

Only `window.__veryfrontRouter.params` is normalized; initial hydration and SPA navigation still pass raw `pageData.params` (arrays) into props/pageContext, while SSR now emits joined strings. So `usePageContext().params.slug` is `"guides/intro"` server-rendered but `["guides","intro"]` after client navigation — the same server/client disagreement class this PR targets, on an adjacent seam.

**5. `src/html/html-injection.ts:117` — #2741 remains unfixed on the full-HTML-document client-page path** *(pre-existing, follow-up)*

The hydration payload built there carries no `params` field at all (the options interface can't express them), so a `'use client'` page on a dynamic route that renders its own full `<html>` document still hydrates with empty `params` via the RSC client runtime. Follow-up material rather than a blocker.

**6. `src/routing/flatten-route-params.ts:9` — duplicates `normalizeParams`, already exported from the same barrel** *(cleanup)*

`src/routing/api/context-builder.ts:60` has an existing helper with identical `Array.isArray(value) ? value.join('/') : value` semantics, exported from `src/routing/index.ts` — the barrel this PR edits. The routing module now publicly exports two near-identical flatteners (plus copies in `rsc/hydration-router.ts` and `ClientApp.tsx` that the new helper's doc comment cites but doesn't consolidate). Extending or wrapping the existing helper would prevent the exact divergence class #2742 was about.

**7. `templates/router.ts:996` — the params-seed IIFE adds a third synchronous `JSON.parse` of the full hydration blob at startup** *(efficiency)*

The renderer template in the same concatenated bundle already parses `#veryfront-hydration-data` twice; the blob carries full page props/frontmatter/headings so it can be large. Setting `router.params` inside `renderPage` (where the parsed data already exists) would avoid the extra parse — and the IIFE's swallow-all `catch` diverges from the renderer paths, which log parse failures.

---

**Bottom line:** the fix itself is correct and the test approach is good, but finding 1 (the `.test.tsx` CI revert) should be fixed before merge, and finding 2 (the missed `generateMetadata` flattener) belongs in this PR. Findings 3–7 are candidates for this PR or fast follow-ups.

*Review ran 8 finder angles + per-candidate adversarial verification; 7 of 8 candidates survived (a suspected params-after-render staleness issue in `navigateSPA` was refuted — no store notification exists on that path and React flush ordering makes the mutation visible in time).*

mattboon:
Thanks for the deep pass — all 7 addressed. Summary, most-severe first:

**#1 — `.test.tsx` disabled in CI** ✅ Reverted the glob to `*.test.ts*` (+ matching `*.integration.test.ts*` exclusion). `deno.json` now differs from `main` by only the version bump. All 9 `.tsx` suites — including `router-provider`, `hydration-router`, `ClientApp.reactivity` — run again and pass. It was a rebase artifact, exactly as you called it.

**#2 — missed `generateMetadata` flattener** ✅ `page-rendering.ts` now routes through `flattenRouteParams`. Confirmed it was the only remaining `value[0]` site (grepped all of `src/`, incl. `react/`/`client/`/`rsc/`).

**#4 — client props / `usePageContext().params` got raw arrays** ✅ Normalized at both seams: `renderPageFromData` (SPA/popstate) and `renderer.ts` (initial hydration). `normalizeRouteParams` is idempotent on strings, so no-op if page data is already joined. The executable runtime test now captures the `params` prop handed to the page component and asserts it's joined (`{ slug: "guides/intro" }`).

**#3 — empty catch-all contract** ✅ Pinned `{ slug: [] }` → `{ slug: "" }` (key retained, matching the RSC/API/client normalizers) with a test.

**#5 — full-HTML `'use client'` path (#2741)** ✅ Fixed here, not deferred. `html-injection.ts` now seeds `params` into the payload; `context.options.params` was already in scope at the call site and the client already reads/normalizes the field, so it was a small, safe change. Covered by tests (catch-all preserved + empty default).

**#6 — duplicate flatteners** ✅ `context-builder`'s `normalizeParams` now delegates to `flattenRouteParams` — single implementation, divergence gone.

**#7 — triple `JSON.parse` / seed IIFE** — intentionally left, and want to flag the reasoning rather than silently skip: moving the seed into `renderPage` would break the synchronous-availability guarantee the IIFE exists for (params must be ready before the async render). And its swallow-all `catch` is deliberate there — a throw inside the router object literal's initializer would break router construction entirely, unlike the renderer paths where logging is fine. Happy to take the parse de-dup as a follow-up if you'd prefer.

Also: your adversarial refutation of the `navigateSPA` staleness was right for the store path — but there *was* a real render-ordering bug on the `RouterProvider` seed (params mutated after render). It's fixed (update moved before `renderPageFromData`), with a regression test that fails if the order is reverted.

kwakayama (dismissed):
Score: 72/100

Recommendation: REQUEST CHANGES

Findings:

1. HIGH - Full-document client-page hydration JSON can be broken out by decoded route params.

`src/html/html-injection.ts:124` now includes `params` in the client-page hydration payload, and `src/html/html-injection.ts:136` injects that raw `JSON.stringify(...)` result directly into `<script type="application/json">`. Route params are URL-derived and decoded before this point (`src/routing/matchers/decode-param.ts:8-10`, `src/routing/matchers/route-matcher.ts:36-37`), so an encoded segment such as `%3C%2Fscript%3E%3Cscript%3E...%3C%2Fscript%3E` becomes a literal `</script><script>...</script>` in `params`.

I reproduced the generated output from this branch with `params: { slug: ["</script><script>globalThis.pwned=1</script>"] }`, and the emitted HTML contains the literal closing `</script>` inside the JSON payload before the RSC client script. That is a reflected XSS breakout on the full-HTML-document `use client` path.

Fix: serialize this payload with the existing `jsonForInlineScript(...)` helper (`src/security/client/html-sanitizer.ts:33-39`) instead of raw `JSON.stringify`, matching the main shell hydration path (`src/html/hydration-script-builder/hydration-data-generator.ts:42,97`). Please also add a regression test that passes a route param containing `</script>` and asserts the hydration data script does not contain a literal script terminator from the JSON value.

Validation run on current head `b0992bf9e4e6af1dfb668fc86d2830aba03cd767`:

- `deno test --no-check --allow-all src/routing/flatten-route-params.test.ts src/html/html-injection.test.ts src/html/hydration-script-builder/templates/router.test.ts src/html/hydration-script-builder/templates/renderer.test.ts src/rendering/layouts/layout-applicator.test.ts src/rendering/script-page-handling.test.ts` passed: 6 files, 181 steps.
- `deno task generate` passed and left the isolated worktree clean.
- `deno fmt --check` on the 18 changed files passed.
- `deno check` on the changed TypeScript sources passed.
- `git diff --check origin/main...HEAD` passed.

The route-param normalization changes otherwise look directionally right: the changed SSR/client render paths no longer appear to use the old first-segment `value[0]` flattening, and the generated router tests now exercise runtime behavior rather than only string tokens. This security issue is narrow, but it is merge-blocking.

mattboon:
Good catch — fixed in `12887f11`. This was introduced by the #5 fix, so it's on us.

- The `'use client'` hydration payload now serializes via `jsonForInlineScript(...)` instead of raw `JSON.stringify`, escaping `<` `>` `&` and U+2028/2029 — matching the main shell path (`hydration-data-generator.ts`). Applied to the whole payload, so `slug` (also URL-derived) is covered too, not just `params`.
- Regression test injects `</script><script>globalThis.pwned=1</script>` as a route param and asserts (a) no literal breakout sequence in the output and (b) lossless round-trip. It **fails** against the old raw `JSON.stringify` and passes now — verified both directions.
- Confirmed this was the only raw serialization into a `<script>` on this path.

Thanks for the concrete repro and pointing at the exact helper — made it a clean fix.

kojiwakayama (approved):

kwakayama:
Re-review after `12887f114` / head `a42204f0d`: 92/100

Recommendation: APPROVE / no remaining blockers found. The PR is already merged, so I am leaving this as a re-review comment rather than a formal review.

What changed since my blocking review:
- `src/html/html-injection.ts` now serializes the full-document client-page hydration payload with `jsonForInlineScript(...)` instead of raw `JSON.stringify`, so decoded route params cannot emit a literal `</script>` terminator inside the JSON script.
- `src/html/html-injection.test.ts` adds a regression case with `</script><script>...` in route params and verifies lossless round-trip through the hydration-data extractor.

Validation run at `a42204f0db31e05674ad2485ab7be5fe91ddfb48`:
- Targeted tests passed: `deno test --no-check --allow-all src/routing/flatten-route-params.test.ts src/html/html-injection.test.ts src/html/hydration-script-builder/templates/router.test.ts src/html/hydration-script-builder/templates/renderer.test.ts src/rendering/layouts/layout-applicator.test.ts src/rendering/script-page-handling.test.ts` (6 files, 182 steps).
- Manual reproduction now prints `BREAKOUT_ABSENT`; emitted JSON contains `\u003c/script\u003e...` escapes and no executable breakout script.
- `deno fmt --check` on the 18 changed files passed.
- `deno check` on the changed TypeScript sources passed.
- `deno task generate` passed and left the isolated worktree clean.
- `git diff --check 1782eca4470a47a030ee43d12f1f8cce5fc2956b...HEAD` passed.

Residual risk is low: the route-param normalization contract is still duplicated between server helper, generated router runtime, and RSC hydration wrapper, but the tests now cover the meaningful SSR, hydration, SPA, popstate, and script-breakout paths touched by this PR.

PR #2746, opened 2026-07-03
Run task targets with project runtime discovery

kwakayama:
## Summary
- Run task targets through project agent runtime discovery before execution.
- Use the same runtime discovery path for direct `veryfront task` runs.
- Add a focused regression test proving task targets see tools registered by project runtime discovery.
- Normalize task ids discovered from `file://` module paths.

## Verification
- `deno task fmt:check`
- `deno task lint`
- `deno task lint:core-deps && deno task lint:dependency-boundaries && deno task lint:extension-contracts && deno task lint:extension-capabilities && deno task lint:ban-test-only && deno task lint:sanitizer-baseline`
- `deno task typecheck`
- `deno test -A src/trigger/local-runner.test.ts src/task/runner.test.ts src/task/discovery.test.ts src/agent/project/agent-runtime.test.ts`
- Pre-push hook passed after final amend: formatting, lint, typecheck, unit tests (`2237 passed`, `0 failed`).
- GitHub PR checks are green, including format, lint, typecheck, unit, integration, binary e2e, npm install smoke, rsc browser e2e, coverage, CLA, and CodeQL.
- Proved from `veryfront-ops-agent` via local `../veryfront-code` source CLI: a task loaded the project `triage-sweep` skill through runtime discovery and returned `OPS_AGENT_RUNTIME_DISCOVERY_OK`.

### DIFF
--- a/cli/commands/schedule/handler.ts
+++ b/cli/commands/schedule/handler.ts
@@ -33,7 +33,8 @@ export async function handleScheduleCommand(args: ParsedArgs): Promise<void> {
   const projectDir = Deno.cwd();
   const input = opts.input ? await readJsonFile(opts.input, "--input JSON file") : undefined;
 
-  await withProjectSourceContext(projectDir, async ({ adapter, config, projectId }) => {
+  await withProjectSourceContext(projectDir, async (context) => {
+    const { adapter, config, configCacheKey, projectId } = context;
     const result = await discoverSchedules({ projectDir, adapter, config });
     if (result.errors.length > 0) {
       throw new Error(`Schedule discovery failed: ${result.errors[0]?.message}`);
@@ -48,6 +49,7 @@ export async function handleScheduleCommand(args: ParsedArgs): Promise<void> {
       projectDir,
       adapter,
       config,
+      cacheKey: configCacheKey,
       projectId,
       target: schedule.target,
       input: input ?? schedule.input ?? {},
--- a/cli/commands/task/command.ts
+++ b/cli/commands/task/command.ts
@@ -14,10 +14,22 @@ import type { TaskArgs } from "./handler.ts";
 
 export interface TaskOptions extends TaskArgs {}
 
+function logRuntimeDiscoveryWarnings(
+  errors: Array<{ file: string; error: Error }>,
+  debug: boolean | undefined,
+): void {
+  if (errors.length === 0 || !debug) return;
+
+  for (const err of errors) {
+    cliLogger.warn(`  Warning: ${err.file}: ${err.error.message}`);
+  }
+}
+
 export async function taskCommand(options: TaskOptions): Promise<void> {
-  const { discoverTasks } = await import(
-    "../../../src/task/discovery.ts"
-  );
+  const { discoverProjectTaskRuntime, findProjectRuntimeTask, listProjectRuntimeTasks } =
+    await import(
+      "../../../src/task/project-runtime.ts"
+    );
   const { runTask } = await import(
     "../../../src/task/runner.ts"
   );
@@ -32,7 +44,7 @@ export async function taskCommand(options: TaskOptions): Promise<void> {
   const projectDir = Deno.cwd();
   await withProjectSourceContext(
     projectDir,
-    async ({ adapter, config, projectId, proxyContext }) => {
+    async ({ adapter, config, configCacheKey, projectId, proxyContext }) => {
       const sourceLabel = proxyContext?.branchRef
         ? `branch ${proxyContext.branchRef}`
         : proxyContext
@@ -41,22 +53,25 @@ export async function taskCommand(options: TaskOptions): Promise<void> {
 
       cliLogger.info(`Discovering tasks in ${sourceLabel}`);
 
-      const { tasks, errors } = await discoverTasks({
+      const discovery = await discoverProjectTaskRuntime({
         projectDir,
         adapter,
         config,
+        fsAdapter: adapter.fs,
+        cacheKey: configCacheKey,
         debug: options.debug,
       });
+      logRuntimeDiscoveryWarnings(discovery.errors, options.debug);
 
-      if (errors.length > 0 && options.debug) {
-        for (const err of errors) {
-          cliLogger.warn(`  Warning: ${err.filePath}: ${err.error}`);
-        }
-      }
-
-      const task = tasks.find((t) => t.id === taskName);
+      const task = findProjectRuntimeTask(discovery, taskName);
       if (!task) {
         cliLogger.error(`Task "${taskName}" not found.`);
+        if (discovery.errors.length > 0 && !options.debug) {
+          cliLogger.warn(
+            "Some project files could not be loaded. Re-run with --debug for details.",
+          );
+        }
+        const tasks = listProjectRuntimeTasks(discovery);
         if (tasks.length > 0) {
           cliLogger.info("Available tasks:");
           for (const t of tasks) {
--- a/cli/commands/webhook/handler.ts
+++ b/cli/commands/webhook/handler.ts
@@ -33,7 +33,8 @@ export async function handleWebhookCommand(args: ParsedArgs): Promise<void> {
   const projectDir = Deno.cwd();
   const payload = await readJsonFile(opts.payload, "--payload JSON file");
 
-  await withProjectSourceContext(projectDir, async ({ adapter, config, projectId }) => {
+  await withProjectSourceContext(projectDir, async (context) => {
+    const { adapter, config, configCacheKey, projectId } = context;
     const result = await discoverWebhooks({ projectDir, adapter, config });
     if (result.errors.length > 0) {
       throw new Error(`Webhook discovery failed: ${result.errors[0]?.message}`);
@@ -48,6 +49,7 @@ export async function handleWebhookCommand(args: ParsedArgs): Promise<void> {
       projectDir,
       adapter,
       config,
+      cacheKey: configCacheKey,
       projectId,
       target: webhook.target,
       input: payload,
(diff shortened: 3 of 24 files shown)


kwakayama:
## Critical Review — Score: 65/100

The local half of this change is well built. New `src/task/project-runtime.ts` is small and focused, the regression test in `src/trigger/local-runner.test.ts` proves the actual behavior (task sees a tool registered by runtime discovery), and unifying `runTaskTarget` and `runWorkflowTarget` behind `discoverRuntimeOrThrow` removes real duplication. The `file://` normalization in `task-handler.ts` matches the existing idiom (`discovery-utils.ts:21`, `discovery-engine.ts:45`).

The problem is what the PR claims versus what it delivers. Three issues, two of them significant.

### 1. Cloud task runs still bypass runtime discovery (the bug survives in production)

`executeTaskRun` in `src/server/handlers/request/project-run-execute.handler.ts:284` still uses the old `findTaskById` from `src/task/discovery.ts` and never calls `ensureProjectDiscovery`. Compare its siblings in the same file:

- `executeWorkflowRun` (line 361): `await deps.ensureProjectDiscovery(ctx)` before lookup
- `executeEvalRun` (line 958): `await deps.ensureProjectDiscovery(ctx)` before lookup
- `executeTaskRun` (line 284): no discovery call at all

So a cloud task run still executes without project tools registered. The exact failure mode this PR fixes for local CLI and trigger runs is untouched on the path where it matters most. The new doc comment on `RunnableTask.id` ("Stable task id used by CLI, triggers, and cloud runs") is not accurate: cloud runs use `DiscoveredTask` from the old path.

There's a second-order bug hiding here too. The new path respects `config.ai.tasks.discovery.paths` (via `createProjectDiscoveryConfig`), the old cloud path scans a fixed `tasks/` dir. A task in a custom-configured directory runs locally but returns "Task not found" in cloud. Two discovery systems, one task id, different answers.

**Fix:** add `await deps.ensureProjectDiscovery(ctx)` + `findProjectRuntimeTask` to `executeTaskRun`, mirroring workflow/eval, and delete the old `discoverTasks` path once nothing uses it. If cloud is intentionally deferred to a follow-up PR, say so in the description and file the issue — right now the summary reads as if the bug is fixed.

### 2. CLI task command loses remote project source support

Old `discoverTasks` took the FS-enhanced `adapter` and used it for file collection, with explicit remote handling in `resolveTasksBaseDir` (`fs.type === "github" | "veryfront-api"` → relative paths). The new path calls `discoverProjectAgentRuntime`, which passes `adapter` only to `getConfig` and never sets `fsAdapter` on `createProjectDiscoveryConfig` — so `discoverAll` falls back to native local fs.

The server knows this matters: `ensureProjectDiscovery` in `src/server/handlers/request/api/project-discovery.ts:82` explicitly passes `fsAdapter: ctx.adapter.fs`.

Net effect: in proxy/multi-project mode (`VERYFRONT_PROJECT_SLUG` set, branch source), `cli/commands/task/command.ts` prints "Discovering tasks in branch X" and then discovers from the local working directory instead. The command.ts diff even dropped `config` from the destructured context, which was the thing carrying the fs type.

This gap pre-exists for workflow targets (they already used `discoverProjectAgentRuntime`), but this PR extends it to the task command, which previously handled it correctly. Either thread `adapter.fs` through `discoverProjectAgentRuntime` into the discovery config, or document that proxy-mode task runs are unsupported.

### 3. Silent error-semantics change in the CLI

Before: task discovery errors were debug-only warnings (`command.ts` old lines 46-50), and a task could run as long as its own file loaded. After: `discoverProjectTaskRuntime` (`src/task/project-runtime.ts:22`) throws on the first error in ANY discovered file — a broken agent, tool, skill, or workflow now blocks every `veryfront task` run in the project.

Fail-fast is defensible (a task that silently runs without its tools is worse), but this is a breaking behavior change not mentioned in the PR description. Also, only `discovery.errors[0]` is surfaced; if three files are broken the user fixes them one error at a time. Print all errors, then throw.

### Minor

- `src/discovery/handlers/task-handler.ts:13` — `file.replace("file://", "")` replaces the first occurrence anywhere, not just as a prefix. Existing idiom in the codebase, so fine, but a `startsWith` guard would be more honest.
- The `getId` change in task-handler has no direct unit test; the only coverage is indirect via the new integration test.
- `cli/commands/task/command.ts` builds the full task list via `listProjectRuntimeTasks` on every run just for the not-found message. Cheap on a Map, but computing it only in the error branch reads better.

### Scoring

| Area | Assessment |
|------|-----------|
| Local fix + test quality | Good. Real regression test, clean module boundaries. |
| Completeness of the stated fix | Cloud path untouched; the claim overstates the delivery. |
| Regressions introduced | Remote-fs discovery in CLI task command; error-semantics change. |
| Code health | Two parallel task discovery systems now live side by side. |

**65/100.** The local mechanics are solid, but the PR fixes the bug on the two paths where it was easy and leaves it on the production path, while quietly dropping remote-source support the old code had. Close the cloud gap (or explicitly scope it out), thread the fs adapter through, and this is an 85+.

---
*Review generated with [Claude Code](https://claude.com/claude-code) — findings verified by reading the code at `25b790f`.*

kwakayama:
## Re-Review — Score: 88/100 (was 65)

This is what a rework should look like. All three issues from the [previous review](https://github.com/veryfront/veryfront-code/pull/2746#issuecomment-4878716232) are fixed at the root, each with a test that proves the behavior, and the fix went deeper than asked (dynamic workflow workers got the fsAdapter/cacheKey threading too).

### Previous findings — all resolved

**1. Cloud task runs bypass runtime discovery — FIXED.** `executeTaskRun` now calls `deps.ensureProjectDiscovery(ctx)` and resolves the task via `findProjectRuntimeTask` from the returned `DiscoveryResult`, mirroring workflow/eval runs. The old `findTaskById` dep is deleted from the handler entirely. The new test asserts ordering (`["discover", "run:sync-calendar-events"]`), which is exactly the right thing to pin. The custom-task-dirs divergence is gone with it: cloud and local now share one id space.

**2. Remote project source support — FIXED, properly.** `discoverProjectAgentRuntime` accepts `config`, `fsAdapter`, `cacheKey`; `createProjectDiscoveryConfig` resolves `baseDir` to relative paths for `github`/`veryfront-api` fs types (`resolveProjectDiscoveryBaseDir`), restoring the old `resolveTasksBaseDir` semantics; the CLI passes `adapter.fs` and the new `configCacheKey` from `withProjectSourceContext`. The cache-key isolation test ("isolates virtual project runtime config by cache key") covers the multi-project collision case I hadn't even flagged. Good catch.

**3. Error semantics — FIXED with the right split.** CLI is back to non-fatal debug warnings, plus a genuinely useful touch: when the task isn't found and there were discovery errors, it hints "Re-run with --debug for details" instead of leaving the user staring at a mystery. Trigger runs are strict via `throwOnErrors: true` and now report ALL errors, not just `errors[0]`. Both modes tested.

Also nice: `file.replace("file://", "")` became a proper `startsWith` slice, and the `discoverAll` copy-paste loops collapsed into `discoverConfiguredItems`. The diff got bigger and the codebase got smaller. That's the good direction.

### Remaining issues

**[P2] (confidence 8/10) `src/server/handlers/request/api/project-discovery.ts` — hard discovery failure now masquerades as "Task not found".** The catch block returns `createEmptyDiscoveryResult()`. Before, `ensureProjectDiscovery` returned void and the swallow was inert; now the empty result feeds directly into `findProjectRuntimeTask`, so a project whose discovery *throws* (VFS outage, config load failure) gets `"Task not found: sync-calendar-events"` — the user debugs their task id while the real error sits in server logs. Rethrow, or return a discriminated failure so `executeTaskRun` can report "Runtime discovery failed: ..." instead. The workflow/eval paths have the same swallow but still do their own lookup afterward, so tasks are the path where this bites hardest.

**[P3] (confidence 5/10) `resolveProjectDiscoveryBaseDir` changes baseDir for ALL `createProjectDiscoveryConfig` callers**, including the server's `ensureProjectDiscovery`, which previously always used absolute `ctx.projectDir`. For an API-backed runtime container where `ctx.projectDir` is non-empty and `config.fs.type` is `veryfront-api`, discovery paths flip from `${projectDir}/tools` to `tools`. This matches the proven old task-discovery semantics and CI (including integration and e2e) is green, so it's probably right — but it's a shared production path, so worth one staging sanity check that primitive discovery still finds agents/tools in multi-project mode. Medium confidence, verify this is actually an issue.

**[P3] `src/task/discovery.ts` is now dead production code.** `discoverTasks`, `findTaskById`, `deriveTaskId`, `DiscoveredTask` have zero production callers but stay exported from `src/task/index.ts`. Two task-discovery implementations in the tree is how the cloud/local divergence happened in the first place. Delete or `@deprecated` it in a follow-up so nobody imports the wrong one back in.

**[P3] Scope creep:** `docs/guides/pages-and-routing.md` punctuation edit has nothing to do with task runtime discovery. Harmless, but it doesn't belong in this diff.

### Scoring

| Area | v1 | v2 |
|------|----|----|
| Completeness of the stated fix | Cloud path untouched | Cloud + CLI + trigger + worker, unified |
| Regressions | Remote-fs dropped, error semantics changed silently | None found |
| Test quality | One happy-path test | Ordering, cache isolation, both error modes, adapter-backed paths |
| Code health | Two parallel discovery systems | One live path; old one dead but still exported |

**88/100.** The 12-point gap: the empty-result error masking in `ensureProjectDiscovery` (worth fixing before merge — it's a small change), the unverified server baseDir behavior change, the dead exports, and the doc drive-by. None of these block the core change from being correct. Fix the P2 and this ships clean.

---
*Re-review with [Claude Code](https://claude.com/claude-code) — verified against `a50bace`.*

kojiwakayama (approved):

PR #2756, opened 2026-07-03
Honor VERYFRONT_API_BASE_URL and veryfront.json apiUrl across the CLI

### LINKED ISSUE #2755: CLI ignores VERYFRONT_API_BASE_URL and veryfront.json apiUrl — half the CLI talks to production when targeting another API host
## Symptom

Targeting a non-default API host (e.g. `https://api.veryfront.org`) with `VERYFRONT_API_BASE_URL` set makes `veryfront login` appear to succeed and `veryfront push` fail with an opaque `403 Forbidden` — because parts of the CLI silently keep calling `https://api.veryfront.com`.

Observed while deploying `veryfront-ops-agent` to Veryfront Cloud staging: login minted a token against production (not the host requested), and push 403'd because the production project with that slug belongs to a different account. Setting `VERYFRONT_API_URL` (undocumented for this purpose) fixed it — which is how the bugs were located.

## Root causes (three)

1. **`cli/shared/constants.ts` — `getApiUrl()` reads only `env.apiUrl`** (`VERYFRONT_API_URL`) and falls back to the production default. `env.apiBaseUrl` (`VERYFRONT_API_BASE_URL` — the variable the config command lists in `ENV_OVERRIDES` and the rest of the codebase uses) is never consulted. Login, token validation, slug reservation, and every `getApiUrl()` caller are affected.

2. **`cli/shared/config.ts` — `readConfigFile()` short-circuits on `veryfront.config.ts`**: when the TS config has a `projectSlug` it returns `{ projectSlug }` immediately and `veryfront.json` is never read, silently discarding its `apiUrl`/`apiToken`. Any project with both files (the normal case — the TS config is required by the runtime) cannot use `veryfront.json` to pin an API host.

3. **`cli/commands/pull/command.ts:427` — hardcoded `"https://api.veryfront.com"` fallback** in the `--projects` path, ignoring both env vars.

## Expected

`VERYFRONT_API_BASE_URL` (or a committed `veryfront.json` `apiUrl`) should retarget the whole CLI consistently: login, push, pull, deploy, slug reservation.

## Repro

```bash
VERYFRONT_API_BASE_URL=https://api.veryfront.org veryfront login --github
# browser opens api.veryfront.com/auth/github — wrong host
```

Fix PR follows.

kwakayama:
Fixes #2755.

## Summary
- `getApiUrl()` (cli/shared/constants.ts) now falls back to `env.apiBaseUrl` (`VERYFRONT_API_BASE_URL`) before the production default — previously only the undocumented `VERYFRONT_API_URL` worked, so login/token-validation/slug-reservation silently hit `api.veryfront.com` while the user believed they were targeting another host.
- `readConfigFile()` (cli/shared/config.ts) merges `veryfront.json` instead of discarding it when `veryfront.config.ts` defines a `projectSlug`. The module config still owns the slug; `veryfront.json` contributes `apiUrl`/`apiToken`. Previously any project with both files (the normal case) could not pin an API host via `veryfront.json`.
- `resolveConfigBase()` applies the same precedence (`VERYFRONT_API_URL` → `veryfront.json` → `apiBaseUrl`/default), and the `pull --projects` error path uses `getApiUrl()` instead of a hardcoded production URL.

## Real-world failure this fixes
Deploying `veryfront-ops-agent` to Veryfront Cloud staging with `VERYFRONT_API_BASE_URL=https://api.veryfront.org`: `veryfront login` minted a token against production, then `veryfront push` failed `403 Forbidden` because the production project with that slug belongs to a different account. Full trail in #2755.

## Verification
- `deno test cli/shared/constants.test.ts cli/shared/config.test.ts` — 5 groups / 38 steps, includes 4 new cases (apiBaseUrl fallback, apiUrl precedence, veryfront.json merge with and without module config)
- `deno test cli/commands/pull/command.test.ts` — 5 groups / 32 steps
- `deno fmt` clean on all touched files; pre-push hook (format, lint, typecheck, unit tests) passed

### DIFF
--- a/cli/commands/pull/command.ts
+++ b/cli/commands/pull/command.ts
@@ -11,7 +11,7 @@ import { defineSchema, lazySchema } from "veryfront/schemas";
 import type { InferSchema } from "veryfront/extensions/schema";
 import { dirname, join, normalize, resolve } from "veryfront/platform/path";
 import { cliLogger } from "#cli/utils";
-import { getApiUrl } from "#cli/shared/constants";
+import { resolveCliApiUrl } from "#cli/shared/constants";
 import { cwd } from "veryfront/platform";
 import { createFileSystem } from "veryfront/platform";
 import {
@@ -22,7 +22,7 @@ import {
 } from "#cli/shared/config";
 import { confirmPrompt, logInfo, logSuccess, logWarning } from "#cli/utils";
 import { createNoopSpinner, createSpinner } from "#cli/ui";
-import { getApiTokenEnv } from "veryfront/config";
+import { getApiTokenEnv, getEnvironmentConfig } from "veryfront/config";
 import { withSpan } from "veryfront/observability/otlp-setup";
 import { CommonArgs, createArgParser } from "#cli/shared/args";
 
@@ -417,7 +417,8 @@ export function pullCommand(options: PullOptions = {}): Promise<void> {
 
         if (!projects?.length) throw error;
 
-        const token = getApiTokenEnv() ?? configFile?.apiToken;
+        const env = getEnvironmentConfig();
+        const token = getApiTokenEnv(env) ?? configFile?.apiToken;
         if (!token) {
           throw new Error(
             "VERYFRONT_API_TOKEN environment variable or apiToken in veryfront.json is required when using --projects",
@@ -425,7 +426,7 @@ export function pullCommand(options: PullOptions = {}): Promise<void> {
         }
 
         config = {
-          apiUrl: configFile?.apiUrl ?? getApiUrl(),
+          apiUrl: resolveCliApiUrl(env, configFile?.apiUrl),
           apiToken: token,
           projectSlug: "",
         };
--- a/cli/commands/pull/command.ts
+++ b/cli/commands/pull/command.ts
@@ -11,6 +11,7 @@ import { defineSchema, lazySchema } from "veryfront/schemas";
 import type { InferSchema } from "veryfront/extensions/schema";
 import { dirname, join, normalize, resolve } from "veryfront/platform/path";
 import { cliLogger } from "#cli/utils";
+import { getApiUrl } from "#cli/shared/constants";
 import { cwd } from "veryfront/platform";
 import { createFileSystem } from "veryfront/platform";
 import {
@@ -424,7 +425,7 @@ export function pullCommand(options: PullOptions = {}): Promise<void> {
         }
 
         config = {
-          apiUrl: configFile?.apiUrl ?? "https://api.veryfront.com",
+          apiUrl: configFile?.apiUrl ?? getApiUrl(),
           apiToken: token,
           projectSlug: "",
         };
--- a/cli/shared/config.ts
+++ b/cli/shared/config.ts
@@ -13,7 +13,7 @@ import { type EnvironmentConfig, getEnvironmentConfig } from "veryfront/config";
 import { cliLogger, VERSION } from "#cli/utils";
 import { readToken } from "../auth/token-store.ts";
 import { ensureAuthenticated } from "../auth/login.ts";
-import { DEFAULT_API_URL } from "./constants.ts";
+import { resolveCliApiUrl } from "./constants.ts";
 import { isRetryableConnectionError } from "../../src/proxy/retry.ts";
 
 // Delays for exponential backoff with jitter: attempt 1 = ~300ms, 2 = ~1s, 3 = ~3s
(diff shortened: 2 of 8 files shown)


cli/commands/pull/command.ts
```diff
@@ -424,7 +425,7 @@ export function pullCommand(options: PullOptions = {}): Promise<void> {
         }
 
         config = {
-          apiUrl: configFile?.apiUrl ?? "https://api.veryfront.com",
+          apiUrl: configFile?.apiUrl ?? getApiUrl(),
```
kwakayama: This fallback does not use the same precedence as `resolveConfigBase()`: a `veryfront.json` `apiUrl` beats `VERYFRONT_API_URL`/`VERYFRONT_API_BASE_URL` here. Because this path is reached specifically when normal config/auth resolution throws for multi-project pulls, the effective API host can change based on control flow. Please route this through the same shared resolver used by the normal path.

cli/shared/config.ts
```diff
@@ -145,7 +159,9 @@ async function resolveConfigBase(
   const dir = projectDir ?? cwd();
   const configFile = await readConfigFile(dir);
 
-  const apiUrl = env.apiUrl ?? configFile?.apiUrl ?? DEFAULT_API_URL;
+  // Explicit VERYFRONT_API_URL wins, then the project's veryfront.json,
+  // then apiBaseUrl (VERYFRONT_API_BASE_URL or the production default).
+  const apiUrl = env.apiUrl ?? configFile?.apiUrl ?? env.apiBaseUrl ?? DEFAULT_API_URL;
```
kwakayama: This encodes a different contract than `getApiUrl()`: repo config wins over `env.apiBaseUrl`. If `VERYFRONT_API_BASE_URL` is an operator override, this still allows a checked-in `veryfront.json` host to silently win. If file pinning is intended to win, the contract should be centralized and documented so every CLI path follows it consistently.

kwakayama (commented):
Score: 76/100

Not approving / not merge-ready. This is directionally the right fix and the touched tests/formatting pass, but the same host-selection bug class is still present in at least one changed path, and the effective precedence contract is split across multiple helpers.

Findings:

1. HIGH: `pull --projects` fallback still lets `veryfront.json apiUrl` beat explicit env overrides. In the fallback path, `apiUrl: configFile?.apiUrl ?? getApiUrl()` means `VERYFRONT_API_URL`/`VERYFRONT_API_BASE_URL` are ignored whenever the config file has `apiUrl`. That conflicts with the PR body's stated `resolveConfigBase()` precedence and makes host selection depend on whether the auth/config path throws.

2. MEDIUM/HIGH contract issue: `resolveConfigBase()` puts `configFile?.apiUrl` ahead of `env.apiBaseUrl`. If `VERYFRONT_API_BASE_URL` is intended to be an operator override, a checked-in file can still silently route commands to another host. If file-pinning is intended to win, that needs to be explicit and shared everywhere rather than implied by this one code path.

3. MEDIUM: `cli/commands/up/command.ts` still creates new projects with `env.apiUrl ?? "https://api.veryfront.com"`, so `VERYFRONT_API_BASE_URL` does not retarget that CLI project-creation path. This is outside the diff, but it conflicts with the issue's expected behavior that the whole CLI should retarget consistently.

Verification I ran locally:
- `VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net cli/shared/constants.test.ts cli/shared/config.test.ts` => 5 passed / 38 steps
- `VF_DISABLE_LRU_INTERVAL=1 SSR_TRANSFORM_PER_PROJECT_LIMIT=0 REVALIDATION_PER_PROJECT_LIMIT=0 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --no-check --allow-all --unstable-worker-options --unstable-net cli/commands/pull/command.test.ts` => 5 passed / 32 steps
- `deno fmt --check` on the five touched files => clean
- GitHub checks are green

Recommended fix: centralize CLI API host resolution into one helper that can distinguish an explicit `VERYFRONT_API_BASE_URL` from the production default, then use it from `resolveConfigBase()`, `pull --projects`, `getApiUrl()` callers, and `up` project creation. Add regression tests for env-vs-file precedence and for the multi-project fallback host.

kwakayama:
**Review update after fixes**

Score: **94/100**. Merge threshold met.

No blocking findings remain from the earlier review. The CLI host precedence is now centralized through `resolveCliApiUrl`, `veryfront.config.*` no longer causes `veryfront.json` API fields to be dropped, and the `pull --projects` fallback plus `up` project creation paths now honor explicit `VERYFRONT_API_BASE_URL`/`VERYFRONT_API_URL` instead of falling back to production.

Verification:
- Local targeted tests passed for `cli/shared/constants.test.ts`, `cli/shared/config.test.ts`, `cli/commands/pull/command.test.ts`, and `cli/commands/up/command.test.ts`.
- Local pre-push checks passed: format, lint, typecheck, and `2274` unit tests.
- Hosted checks on `887b3546f` are green: Analyze, CodeQL, format, lint, typecheck, coverage gate/shards, unit, integration, binary e2e, npm install smoke, and rsc browser e2e.

Residual risk is low: `resolveCliApiUrl` infers an explicit `apiBaseUrl` from a non-default value because `EnvironmentConfig` does not preserve whether the value came from env or defaults. That matches the current config shape and covers the staging failure mode.

kojiwakayama (approved):

PR #2758, opened 2026-07-04
Expose project tools in project-agent stream runs

kwakayama:
## Summary
- Pass discovered host tool objects into project-agent stream runs.
- Let boolean source tool declarations resolve to concrete local tools before runtime construction.
- Add regression coverage for boolean source tool declarations backed by supplied local tools.

## Tests
- deno test --no-check --allow-all src/internal-agents/run-stream.test.ts
- deno test --no-check --allow-all src/server/handlers/request/agent-stream.handler.test.ts
- deno check src/internal-agents/run-stream.ts src/server/handlers/request/agent-stream.handler.ts
- deno fmt --check src/internal-agents/run-stream.ts src/internal-agents/run-stream.test.ts src/server/handlers/request/agent-stream.handler.ts
- git diff --check origin/main...HEAD
- clean-env pre-push hook: formatting, lint, typecheck, unit tests: 2274 passed / 0 failed

## Why
Hosted project-agent runs were only seeing skill-loader tools when the agent declared project tools by boolean name. This blocks the ops-agent proof run because `read_baseline` and telemetry tools are never exposed to the runtime.

### DIFF
--- a/src/internal-agents/run-stream.test.ts
+++ b/src/internal-agents/run-stream.test.ts
@@ -126,6 +126,65 @@ describe("internal-agents/run-stream", () => {
     assertEquals(capturedToolNames, ["gmail__list_emails", "web_search"]);
   });
 
+  it("uses supplied local tool objects for boolean source tool declarations", async () => {
+    const sessionManager = new AgentRunSessionManager();
+    let capturedTool: unknown;
+
+    const agent = {
+      id: "ops-agent",
+      config: {
+        id: "ops-agent",
+        model: "anthropic/claude-opus-4-6",
+        system: "test",
+        tools: {
+          read_baseline: true,
+        },
+      },
+    } as unknown as Agent;
+
+    const readBaselineTool = {
+      id: "read_baseline",
+      description: "Read baseline",
+      inputSchema: { parse: (value: unknown) => value },
+      execute: () => ({ ok: true }),
+    } as unknown as Tool;
+
+    const input = {
+      agentId: "ops-agent",
+      threadId: crypto.randomUUID(),
+      runId: "run_1",
+      messages: [],
+      tools: [],
+      context: [],
+    } as Parameters<typeof createRuntimeAgentStreamResponse>[0];
+
+    await createRuntimeAgentStreamResponse(
+      input,
+      agent,
+      {
+        sessionManager,
+        localTools: {
+          read_baseline: readBaselineTool,
+        },
+        createRuntime: (_agent, mergedTools) => {
+          capturedTool = typeof mergedTools === "object" && mergedTools !== null
+            ? (mergedTools as Record<string, unknown>).read_baseline
+            : undefined;
+          return {
+            stream: async () =>
+              new ReadableStream<Uint8Array>({
+                start(controller) {
+                  controller.close();
+                },
+              }),
+          };
+        },
+      },
+    );
+
+    assertEquals(capturedTool, readBaselineTool);
+  });
+
   it("preserves explicitly allowed source remote tool declarations before constructing the runtime", async () => {
     const sessionManager = new AgentRunSessionManager();
     let capturedToolNames: string[] = [];
--- a/src/internal-agents/run-stream.ts
+++ b/src/internal-agents/run-stream.ts
@@ -74,6 +74,7 @@ function mergeRemoteToolNames(source: string[], forwarded: string[]): string[] {
 
 export interface RuntimeAgentStreamExecutionDeps {
   sessionManager: AgentRunSessionManager;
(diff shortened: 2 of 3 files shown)


kwakayama:
Critical review score: 94/100

No blocking findings. The fix is intentionally narrow: project-agent stream execution now receives the same discovered host tool objects already used by hosted chat paths, and boolean source tool declarations can resolve to those concrete tools before runtime construction. The regression test covers the exact failure shape from the ops-agent proof run: `read_baseline: true` with no forwarded tool metadata still reaches the runtime as an executable local tool.

Main residual risk: name collisions between discovered local tools and sandbox-provided tools. Current merge order keeps sandbox tools winning, which is consistent with the existing sandbox path and only matters for overlapping names like `bash`. This is acceptable for this fix.

kwakayama:
Hosted proof run confirms this PR is the current blocker.

Evidence from `veryfront-ops-agent` on `https://api.veryfront.org`, production env `64ea17a6-271b-486c-b078-c4ecfd7964f0`:

- Triggered schedule run: `run_a344a169-003e-408a-bf7a-92c248f57b02`
- Run target: `agent:triage-sweeper`, schedule `3b78bca7-de98-43eb-a77e-462855877c3a`
- Finish metadata: `veryfront-cloud/anthropic/claude-sonnet-4-6`
- `AGENT_RUN_REQUEST_ENQUEUED` included project tool names, including `read_baseline`, but each entry was only `{ name }`; no executable local tool object/schema was present in the stream path.
- Event stream had zero native tool lifecycle events: no `TOOL_CALL_START`, `TOOL_CALL_RESULT`, or related `TOOL_*` events.
- Assistant transcript emitted textual `<tool_call>` / `<tool_response>` markup instead of a structured tool call.

That matches this PR’s fix: pass host-discovered local tool objects into `createRuntimeAgentStreamResponse()` and merge them before tool resolution. Until this lands/deploys, the ops-agent hosted proof cannot be considered valid even though the schedule, release, and model are now correct.

kwakayama:
## Critical review: 91/100

**Verdict: mergeable.** Narrow fix, gating preserved, real production evidence, green CI. I verified beyond the diff: the merge path in `buildMergedTools`, the registry ownership model, and the actual ops-agent project layout.

### What I verified

- `buildMergedTools` only consults `availableLocalTools` for tool names the agent explicitly declared (`src/internal-agents/run-stream.ts:214-218`). No exposure for undeclared tools. The `tools: true` branch is untouched.
- `getDiscoveredHostTools` applies `isToolVisibleTo`, so another agent's owned tools stay hidden (`src/agent/hosted/veryfront-cloud-agent-service.ts:633-645`).
- The actual blocker case resolves: `read_baseline` lives in the ops-agent project's top-level `tools/read-baseline.ts`, so discovery registers it unowned under the plain short name, and the new key lookup hits it.
- Ran `deno test --no-check --allow-all src/internal-agents/run-stream.test.ts` on the PR head: 12 steps, 0 failed. CI fully green (format, lint, typecheck, unit, integration, 8 coverage shards, CodeQL).
- No behavior change when local tools are absent: `Object.keys(availableLocalTools).length > 0 ? ... : undefined` preserves the old `undefined` path.

### Codex P2: valid finding, but not this PR's case

Codex is right that agent-colocated tools register under `{agentId}--{shortName}` (`src/discovery/agent-scoped-capabilities.ts:7`) while `buildMergedTools` looks up by declared short name, so owned tools declared by short name still won't resolve through this path. Two reasons it doesn't block:

1. Not a regression. Those names were dropped before this PR too, since `toolRegistry.get(name)` also misses prefixed ids.
2. The target case (`read_baseline`) is project-level, not agent-scoped.

Follow-up worth filing: `resolveVisibleRegistryTool` in `src/agent/runtime/tool-helpers.ts:103` already implements the owner + shortName mapping. The stream path should reuse it.

### Deductions

- **-4** Known adjacent gap (owned short-name tools) left unhandled with no follow-up issue filed yet.
- **-3** Type looseness: `localTools?: Record<string, Tool | boolean>` lets a boolean `true` shadow a resolvable registry object at `run-stream.ts:218`, handing the runtime a non-executable entry. Copilot flagged this too. Tighten to `Record<string, Tool>` and keep booleans confined to `agent.config.tools`.
- **-2** The handler's default `getLocalTools` wiring (including the `HostToolSet` cast) has no test. The regression test injects `localTools` directly, so the `getDiscoveredHostTools` -> handler -> stream path is only proven by the pending hosted run.

Copilot's module-coupling note (handler importing from the hosted service module) is fair but cosmetic here. The handler path is already Node-bound.

**Score: 91/100. Above the 90 merge bar. Merging.**

kojiwakayama (approved):

PR #2793, opened 2026-07-06
Add GPT-5.4 output-token budgets + enforce catalog↔table coverage

kwakayama:
## Context

Follow-up to #2791 (Kimi empty-final-message fix). Per review, I audited **every** catalog model against `MODEL_MAX_OUTPUT_TOKENS` to check whether the defaults actually work for each model.

## Finding

The **OpenAI GPT-5.4 family has the exact same bug #2791 fixed for Kimi**:

| Model | thinking | budget before | budget after |
|---|---|---|---|
| `openai/gpt-5.4` | ✅ | → 4,096 (default fallback) | **128,000** |
| `openai/gpt-5.4-mini` | ✅ | → 4,096 | **128,000** |
| `openai/gpt-5.4-nano` | ✅ | → 4,096 | **128,000** |

They were absent from the table, so `getModelMaxOutputTokens` returned `undefined` and they fell back to `DEFAULT_MAX_TOKENS` (4,096). `resolveMaxOutputTokens` applies this to all models including the reasoning ones (Responses API counts reasoning against `max_output_tokens`), so a reasoning burst can exhaust the budget before any answer is emitted.

**Most important:** `gpt-5.4-nano` is `DEFAULT_AGENT_MODEL` (and the OpenAI direct-auto default), so every agent that doesn't pin a model was exposed.

## Changes

1. Add `openai/gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-nano` at **128,000** (matching same-generation `gpt-5.5` / `gpt-5.2`).
2. **Catalog↔table sync test** (`VERYFRONT_CLOUD_CHAT_MODELS`): asserts (a) every catalog model has an explicit budget — no silent 4,096 fallback — and (b) every `thinking` model clears a 16,000 floor. This is the durable fix: a newly added model can't regress into this class of bug unnoticed (also addresses the whack-a-mole Copilot flagged on #2791 re: the `kimi-k2` alias).

## Left intentionally unchanged

`mistral/mistral-large-2512` stays at **1,024** — that's a deliberate quota cap (#2533, "Fit Mistral Large chat within current quota"). It's not a thinking model, so it's exempt from the floor and the sync test passes.

## Verification

- `deno test --no-check --allow-all src/agent/runtime/constants.test.ts` → 2 passed, 10 steps, 0 failed.
- `deno check src/agent/runtime/constants.test.ts` → clean. `deno fmt --check` → clean.

Bumps to `0.1.1021`.


kwakayama:
Critical review score: 91/100.

Findings: no merge-blocking correctness, security, or test issues found. Non-blocking watch item: output-token policy is still split between `src/provider/veryfront-cloud/model-catalog.ts` and `src/agent/runtime/constants.ts`; future cleanup should make one layer canonical or document the ownership contract. The test-local `16_000` thinking floor is acceptable for this regression guard but should become named policy metadata if this grows.

Verification evidence:
- `deno test --no-check --allow-all src/agent/runtime/constants.test.ts` passed: 2 tests, 10 steps, 0 failed.
- `deno check src/agent/runtime/constants.test.ts` passed.
- `deno fmt --check deno.json src/agent/runtime/constants.ts src/agent/runtime/constants.test.ts src/utils/version-constant.ts` passed.
- Adjacent tests passed: `src/provider/veryfront-cloud/model-catalog.test.ts`, `src/agent/runtime/model-resolution.test.ts`, `src/agent/runtime/constants.test.ts` (4 tests, 38 steps, 0 failed).
- GitHub CI/checks are green.

Merge status: score is above the 90+ merge gate, but GitHub rejects author self-approval. Auto-merge is enabled with merge commits and branch deletion once branch protection receives a valid non-author approval.

kojiwakayama (approved):

PR #2869, opened 2026-07-11
fix(ci): skip stable publication when version is unchanged

### LINKED ISSUE #2867: Keep main green when the stable version is already published
## Problem

Ordinary feature merges to `main` keep the already-published stable version in `deno.json`. The CI/CD workflow classifies every clean semver as a stable release, waits for the full test/build graph, then fails `Ensure stable version is unpublished` because that version already exists.

Recent examples:
- run 29155038484 for #2862: all tests/builds passed, release failed because `veryfront@0.1.1046` already exists
- run 29150150119 for #2857
- run 29145458707 for #2860

This leaves `main` red after healthy feature merges and makes a release bump look like a CI repair.

## Expected invariant

- Feature merges with an unchanged, already-published version complete green and skip publication.
- A merge that changes to a new clean semver runs stable publication exactly once.
- Pre-release behavior remains unchanged.
- The workflow still fails if a release-intended version conflicts with an existing artifact from a different commit.

## Suggested direction

Gate the stable release jobs on a version change or explicit release intent in addition to `is_stable`. Keep the unpublished-artifact preflight inside the release path as a collision guard.

Add workflow tests or a script-level decision matrix for:
1. feature merge + unchanged published stable version -> skip release, green run
2. release merge + unused stable version -> publish
3. release merge + conflicting existing artifacts -> fail
4. pre-release version -> existing prerelease path

## Evidence

Run 29155038484 reached the release job only after the full test graph passed. The exact terminal error was that `veryfront@0.1.1046` already existed for the prior release commit.

kojiwakayama:
## Summary

- derive stable release intent from the version at the push before revision
- skip stable publication for ordinary main merges that retain the published version
- preserve manual dispatch, RC publication, and the existing artifact collision guards
- behavior-test unchanged multi-commit pushes, version bumps, manual dispatch, and unavailable history

## Why

Healthy feature merges currently pass every test and then leave main red because the release job tries to republish the prior stable version. Existing package or tag collisions must still fail when a release is actually requested.

## Verification

- full protected pre-push gate with DENO_JOBS=1: 2,323 suites, 19,644 steps, 0 failures
- format, lint, typecheck, bash syntax, and workflow wiring checks passed
- real commit decisions: unchanged 0.1.1046 -> false; release bump 0.1.1047 -> true

Fixes #2867

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -343,16 +343,30 @@ jobs:
     outputs:
       version: ${{ steps.check.outputs.version }}
       is_stable: ${{ steps.check.outputs.is_stable }}
+      stable_release_requested: ${{ steps.check.outputs.stable_release_requested }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          fetch-depth: 0
       - name: Detect release type
         id: check
         run: |
           VERSION=$(jq -r '.version' deno.json)
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
+
+          STABLE_RELEASE_REQUESTED=$(bash scripts/ci/stable-release-requested.sh \
+            "${{ github.event_name }}" \
+            "$VERSION" \
+            "${{ github.event.before }}")
+          echo "stable_release_requested=${STABLE_RELEASE_REQUESTED}" >> $GITHUB_OUTPUT
+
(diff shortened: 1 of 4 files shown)


kojiwakayama:
Independent reproduction while validating #2870: the protected pre-push suite passed, then the shared repository config was left with core.bare=true and the fixture test identity because the branch was based on current main and did not contain this PR. The metadata was restored immediately and no source data was affected. This directly validates the clearEnv plus GIT_* filtering in this PR; it is required test isolation, not incidental cleanup.

kwakayama (approved):

PR #2870, opened 2026-07-11
test: make timing-sensitive runtime tests deterministic

### LINKED ISSUE #2865: Make timing-sensitive unit tests deterministic under full-suite load
## Problem

The mandatory pre-push unit gate is intermittently red under full-suite parallel load even though the affected suites pass consistently in isolation. This makes unrelated framework changes require repeated 9-25 minute gate attempts.

Observed while validating #2862:

- `provider/runtime-loader`: pending/streaming input heartbeat timing assertions
- `ProjectWorker - real worker request isolation`: worker response timeout / environment-isolation assertions
- `src/skill/executor`: late sandbox rejection after timeout

A combined isolated rerun passed 7 suites / 40 steps / 0 failures in 19 seconds. A full run at `DENO_JOBS=4` passed 2,311 suites / 19,576 steps before the same three timing-sensitive cases failed.

## Expected direction

Replace wall-clock scheduling assumptions with deterministic synchronization where possible:

- Drive heartbeat transitions with a controllable clock or explicit stream checkpoints.
- Wait for worker readiness/response signals instead of relying on a fixed deadline under host load.
- Attach rejection handling before advancing the timeout path so the late-rejection assertion cannot surface as an unhandled process error.

Do not weaken production timeouts merely to make tests green.

## Acceptance criteria

- The affected suites pass repeatedly in isolation and as part of the normal parallel `test:unit` gate.
- A stress/repeat command demonstrates no intermittent failures.
- Production timeout and cancellation behavior remains covered.
- The pre-push gate does not need reduced global parallelism for unrelated changes.

Related: #2862

kojiwakayama:
## Description

Makes the three full-suite timing regressions deterministic without changing production behavior:

- gates tool-input heartbeat producers on observed consumer checkpoints and fails fast when an expected heartbeat is absent
- warms real project workers before applying the unchanged 10-second behavior timeout, then tests queued environment isolation back-to-back
- drives the late sandbox NDJSON stream error explicitly after the timeout path returns

No production source, heartbeat threshold, request timeout, or cancellation behavior changes.

## Related Issue(s)

Fixes #2865

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Verification

- DENO_JOBS=4 deno task test:unit: 2,322 suites, 19,639 steps, 0 failures
- heartbeat and executor files: 50 consecutive passes
- real-worker isolation suite: 20 consecutive passes
- protected pre-push format, lint, typecheck, and unit gates passed

## Checklist

- [x] Documentation is not applicable because runtime behavior is unchanged
- [x] Existing behavior assertions are preserved with deterministic synchronization

### DIFF
--- a/src/provider/runtime-loader.test.ts
+++ b/src/provider/runtime-loader.test.ts
@@ -5,12 +5,48 @@ import { describe, it } from "#veryfront/testing/bdd.ts";
 import { withToolInputStatusTransitions } from "./runtime-loader.ts";
 import { createOpenAIModelRuntime } from "../../extensions/ext-llm-openai/src/openai-provider.ts";
 
-async function collectAsync<T>(iterable: AsyncIterable<T>): Promise<T[]> {
+type Deferred<T> = {
+  promise: Promise<T>;
+  resolve: (value: T) => void;
+};
+
+function deferred<T = void>(description: string, timeoutMs = 1_000): Deferred<T> {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((resolvePromise, reject) => {
+    const timeoutId = setTimeout(() => {
+      reject(new Error(`Timed out waiting for ${description}`));
+    }, timeoutMs);
+    resolve = (value) => {
+      clearTimeout(timeoutId);
+      resolvePromise(value);
+    };
+  });
+
+  return { promise, resolve };
+}
+
+async function collectAsync<T>(
+  iterable: AsyncIterable<T>,
+  onValue?: (value: T, values: T[]) => void,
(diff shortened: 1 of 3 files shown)


kojiwakayama:
Merge ordering: land #2869 before this PR. This test-only change intentionally leaves the published framework version unchanged; before #2869, merging it would retry the already-published stable version and leave main red. Auto-merge is temporarily disabled and will be re-enabled as soon as #2869 lands. This is release sequencing only, not a code dependency.

kwakayama (approved):

PR #2888, opened 2026-07-12
Shim bare veryfront imports when loading project configs from temp files

kojiwakayama:
## Problem — production outage class

Project configs conventionally start with `import { defineConfig } from "veryfront"`, but `loadConfigFromTempFile` executes configs from a **temp file**, where that bare specifier has no resolver: Node has no `node_modules` relative to the temp dir, and compiled Deno binaries have no import map for external dynamic imports. TypeScript configs therefore **never loaded on the hosted runtime** (proxy mode / compiled binary):

- **≤ 0.1.1049:** the failure was silently swallowed and the loader fell back to defaults — configured values like `tailwind.stylesheet` never took effect remotely (long-standing mystery behavior).
- **0.1.1050:** the fail-closed `CONFIG_PARSE_ERROR` hardening turned the same failure into a **hard 500 on every request** for any project with a TS config. Observed live in production after the 0.1.1050 server rollout: `"Failed to load veryfront.config.ts"` from `adapter-factory` on every request to `veryfront-planning.production.veryfront.com` (mitigated at the project level by removing the import from the config file — service restored immediately, confirming the diagnosis).

Local dev never hits this because the non-compiled path dynamic-imports the config **in place**, where Node resolves `veryfront` via the project's `node_modules`.

## Fix

Rewrite bare `veryfront` import specifiers in the temp-file source to an inline `data:` URL shim exporting `defineConfig` (identity), `defineConfigWithEnv`, and `mergeConfigs` — the pure helpers the config entrypoint exposes. The shim uses double quotes only so `encodeURIComponent` keeps the URL safe inside either quote style. Subpath specifiers (`veryfront/head` etc.) are intentionally left untouched to fail loudly — they have no meaning in a config file.

The fail-closed behavior from 0.1.1050 is kept: with the loader fixed it now surfaces genuinely broken configs instead of masking a resolver gap.

Version bumped to 0.1.1052.

## Verification

- New unit tests: specifier rewriting (double/single quotes, side-effect imports, subpath/lookalike non-matches) and end-to-end transpile → rewrite → dynamic import for both `defineConfig` and `defineConfigWithEnv`.
- Full unit suite: 2,374 passed, 0 failed.
- Production evidence chain: the same config exported as a plain object (no import) loads fine on the hosted 0.1.1050 runtime — 500s cleared instantly.

## Impact once rolled out

Hosted TS configs load for the first time — this also makes configured `tailwind.stylesheet` reach release builds and the preview stylesheet route, closing another gap behind the "hosted projects render unstyled" saga (#2883).

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.1051",
+  "version": "0.1.1052",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": {
--- a/src/config/loader.test.ts
+++ b/src/config/loader.test.ts
@@ -7,6 +7,7 @@ import {
   getCachedConfigSync,
   getConfig,
   mergeConfigs,
+  rewriteBareVeryfrontConfigImports,
   transpileConfigSourceForImport,
 } from "./loader.ts";
 import { createMockAdapter } from "../platform/adapters/mock.ts";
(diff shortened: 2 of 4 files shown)


kwakayama (approved):

kojiwakayama:
Platform ownership and rollout acceptance are tracked in veryfront/veryfront-studio#5813. This PR is the framework fix; the project-level mitigation in Studio #5804 is intentionally not the long-term resolution.

PR #2900, opened 2026-07-12
Preserve request context in relative-fs esbuild plugin callbacks

kojiwakayama:
## Problem — production hydration outage for multi-file client pages

esbuild invokes plugin callbacks from its child-process message pump, which inherits the AsyncLocalStorage context from wherever the esbuild service was **first spawned on that pod** — not from the build's caller. `createRelativeFsPlugin` resolves and loads project files through the request-scoped adapter without re-entering the caller's context, so in proxy mode (`MultiProjectFSAdapter`) every relative import fails with `"No request context available"` on any pod whose first-ever build belonged to a different request.

**Observed live on production (runtime 0.1.1053):** after refactoring `veryfront-planning`'s page into `app/components/*`, every `/_veryfront/rsc/module?rel=app/page.tsx` request 500ed with esbuild `Could not resolve "./components/…"` / `"./pricing"` for **all** relative imports; SSR was fine, hydration dead. A freshly-started debug pod (whose first build was the affected page, rooting the pump inside that request's context) served the identical request with the identical token — 200 with all 12 files bundled. Pod-cycling made results flip pod-by-pod. Full evidence chain in the repro: priming the esbuild service outside any context makes the failure deterministic in a unit test.

## Fix

Wrap the plugin's `onResolve` and both `onLoad` callbacks with `wrapWithCurrentContext` — the exact treatment `routing/api/module-loader/loader.ts` and `discovery/transpiler.ts` already apply, per that helper's own doc comment describing this esbuild pump behavior.

Version bumped to 0.1.1056.

## Verification

- New test reproduces the bug faithfully: primes the esbuild service outside any ALS context, then bundles with a store-requiring adapter inside `asyncLocalStorage.run` — RED before the fix (callbacks fire from `readFromStdout`, contextless), GREEN after.
- Full unit suite: 2,378 passed, 0 failed.

## Rollout note

After promotion, affected pods recover immediately (no cache poisoning — the failure was per-callback). This also likely explains historical "hydration randomly broken in proxy mode" reports predating multi-file pages, since `./pricing`-style single-file-adjacent imports hit the same path.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.1055",
+  "version": "0.1.1056",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": {
--- a/src/server/handlers/dev/files/esbuild-plugins.test.ts
+++ b/src/server/handlers/dev/files/esbuild-plugins.test.ts
@@ -3,6 +3,7 @@ import "#veryfront/transforms/plugins/__tests__/code-parser-setup.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { afterEach, describe, it } from "#veryfront/testing/bdd.ts";
 import { createMockAdapter } from "#veryfront/platform/adapters/mock.ts";
+import { asyncLocalStorage } from "#veryfront/platform/adapters/fs/veryfront/request-context.ts";
 import {
   createBareExternalPlugin,
   createHttpExternalPlugin,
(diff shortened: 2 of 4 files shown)


kwakayama:
## Agent review / merge gate status

Independent code-reviewer score: **94/100**.

Evidence I re-ran on head `d176cfb92e7681d6a38d09bfdfe25825785a6392`:
- `deno test --frozen -A src/server/handlers/dev/files/esbuild-plugins.test.ts` → 3 tests / 20 steps / 0 failed.
- `deno task --frozen verify:quick` → passed: manifests, fmt, lint/style/boundaries/docs validation, and typecheck.
- GitHub checks are green; PR is `MERGEABLE`.

Review summary: no correctness/security blocker found. The fix wraps `createRelativeFsPlugin`'s esbuild `onResolve` and both `onLoad` callbacks with the existing AsyncLocalStorage context helper, and the new test reproduces the context-loss failure by priming esbuild outside the request context.

Remaining gate: GitHub still reports `reviewDecision=REVIEW_REQUIRED`, so this is ready for Code Owner/human approval but not mergeable by the agent yet.

kwakayama (approved):

PR #2947, opened 2026-07-15
Harden registry isolation for unversioned environment contexts

kojiwakayama:
## Scope

This is cross-project isolation hardening discovered while investigating veryfront/veryfront-api#3952. It prevents project-scoped registries from collapsing into the shared `__default__` scope when an environment context has no pinned release ID.

## Important incident correction

This PR is **not** the causal fix for #3952. Production evidence later showed the affected requests already carried a release ID. The deterministic #3952 failure is invalid project skill metadata (`name: "On-Call Triage"` instead of lowercase kebab-case) in `veryfront-ops-agent`; that content patch, deployment, and replay are tracked separately.

## Failure mode this PR prevents

1. `tryGetCacheKeyContext()` correctly returns `null` when production has no stable release ID, because distributed caches must not share an unversioned key.
2. `ProjectScopedRegistryManager` used that cache result directly and fell back to `__default__`.
3. Concurrent unversioned environment requests from different projects could therefore share a registry scope.
4. One project's discovery clear could remove another project's in-flight entries.

## Fix

- Add `tryGetRegistryScopeId()`, which preserves project isolation using the environment, branch, or an `unversioned` sentinel even when distributed caching is intentionally disabled.
- Use it only for in-process registry scoping; distributed cache behavior remains unchanged.
- Add a concurrency regression proving one project's skill clear cannot wipe another project's skill.

## Verification

- `deno check src/cache/cache-key-builder.ts src/registry/project-scoped-registry-manager.ts`
- Full pre-push suite: 2,387 tests / 20,291 steps, zero failures
- GitHub format, lint, typecheck, unit, integration, coverage, browser/binary E2E, npm smoke, audit, and CodeQL checks are green

### DIFF
--- a/src/cache/cache-key-builder.test.ts
+++ b/src/cache/cache-key-builder.test.ts
@@ -9,7 +9,10 @@ import {
   getProjectScopedKeyAlways,
   runWithCacheKeyContext,
   tryGetCacheKeyContext,
+  tryGetRegistryScopeContext,
+  tryGetRegistryScopeId,
 } from "./cache-key-builder.ts";
+import { runWithRequestContext } from "#veryfront/platform/adapters/fs/veryfront/multi-project-adapter.ts";
 
 describe("cache-key-builder", () => {
   describe("getContentHashKey", () => {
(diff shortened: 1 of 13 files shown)


kwakayama (dismissed):

kwakayama (approved):

kojiwakayama:
Final concurrency review after merge found that blind stage publication could still drop same-scope live registrations or resurrect deletions. The verified follow-up is #2949; it serializes same-scope generations and replays staged plus live mutations in exact order with conflict-preserving validation. No auto-merge is enabled on the follow-up.

PR #2948, opened 2026-07-15
Publish skill-registry scope isolation in framework 0.1.1071

kojiwakayama:
Advances the stable framework version `0.1.1070 → 0.1.1071` (synchronized `deno.json` + `src/utils/version-constant.ts`) so the fix merged in #2947 ships.

## Why 0.1.1071
`0.1.1070` was published by #2946 **before** #2947 merged, so it does not contain the skill-registry scope isolation fix.

## What 0.1.1071 delivers
- #2947 — per-project skill registry scopes for environment-source control-plane runs. Fixes veryfront/veryfront-api#3952 (project skills invisible to scheduled agent runs: all projects shared one `__default__` registry scope and wiped each other's skills during discovery).

## Verification
Version synchronization and stable release-intent tests: 2 files, 25 steps, 0 failed.

After this merges and publishes, the veryfront-ops-agent project will be redeployed and its `load_skill` verified end-to-end before unpausing the triage-sweep schedule.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "veryfront",
-  "version": "0.1.1070",
+  "version": "0.1.1071",
   "license": "Apache-2.0",
   "nodeModulesDir": "auto",
   "minimumDependencyAge": {
--- a/src/utils/version-constant.ts
+++ b/src/utils/version-constant.ts
@@ -1,4 +1,4 @@
 // Keep in sync with deno.json version.
 // scripts/release.ts updates this constant during releases.
 /** Shared version value. */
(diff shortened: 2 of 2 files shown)


kwakayama (approved):

kojiwakayama:
0.1.1071 publication succeeded. Post-merge concurrency review of the included #2947 implementation found an ordered-mutation race under concurrent same-scope discovery; #2949 is the verified correction and now carries the 0.1.1072 release bump. The exact #5906 executor should consume 0.1.1072 rather than pinning to 0.1.1071.

Issue #2925, opened 2026-07-14
feat(chat): expose finalized messages to server-side AG-UI persistence


kojiwakayama:
## Problem

The public `createAgUiHandler` supports `beforeStream`, but it does not expose a symmetric successful-completion callback containing the finalized conversation messages. `createAgUiRuntimeHandler` has lifecycle callbacks, but its `onFinish` context also does not carry finalized messages.

The client-side `useConversationChat` persistence path exists; this issue tracks the distinct server-side persistence boundary that remained after the chat composition work.

## Desired outcome

Define one transport-safe lifecycle contract that lets an application persist the finalized message list after a successful AG-UI stream without reconstructing it in userland.

## Acceptance considerations

- specify success, error, abort, and client-disconnect semantics before implementation
- cover direct-agent and injected-tool/resumable paths
- prevent duplicate completion callbacks
- make async persistence failures observable without corrupting the response stream
- add focused handler/runtime tests and public API documentation

This is a follow-up only; it is not part of the stale-plan cleanup.

PR #2957, opened 2026-07-16
feat(agent): server-side AG-UI persistence — finalized-messages callback (#2925)

mattboon:
## Summary

Closes **#2925**. Adds the missing **server-side** persistence boundary: a symmetric successful-completion callback carrying the finalized conversation, so an app can persist an AG-UI run without reconstructing it from the SSE stream. The client-side (`useConversationChat`) path already existed; this is its server counterpart.

## API

`createAgUiHandler` gains **`onComplete`**:

```ts
export const POST = createAgUiHandler({
  agent: "assistant",
  onComplete: async ({ agentId, threadId, runId, messages, inputMessages, response }) => {
    await db.saveTurn({ threadId, input: inputMessages, output: messages });
  },
});
```

New public types: `AgUiCompletion`, `AgUiOnComplete` (exported from `veryfront/agent`).

For **`createAgUiRuntimeHandler`**, the existing `onFinish` lifecycle context now additionally carries the finalized `messages` + `response` (additive optional fields — same fire timing, fully backward compatible).

## Lifecycle contract (specified before implementation, per the issue)

- **Success:** fires **exactly once**, only when the run produced a finalized response.
- **Error:** does **not** fire (the existing error path is untouched).
- **Client disconnect:** does **not** fire (an aborted stream early-returns before the success flag is set).
- **No-response / bodyless run:** does **not** fire.
- **Stream integrity:** runs **after** the stream is fully flushed and closed, so slow/throwing persistence never delays or corrupts the response. A thrown/rejected callback is **caught and logged**, never rethrown into the stream.
- **No duplicates:** single success path guarded by a `succeeded` flag.
- Covers both the **direct-agent** and **injected-tool / resumable** paths.

## Tests

- `handler.test.ts`: fires once with the finalized conversation; does **not** fire on error; does **not** fire without a finalized response; a throwing callback is contained (stream still delivers `RunFinished`).
- `runtime-handler.test.ts`: the `onFinish` context carries the finalized messages + response while preserving the existing `request` context.

**Green:** ag-ui suites (3 files / 33 steps, 0 failed), `deno lint` (40 files), `deno check src/agent/index.ts`. Additive — no breaking changes.

## Docs

New "Persisting finished conversations" subsection in `docs/guides/memory-and-streaming.md`.


kojiwakayama (commented):
**Merge-readiness review: READY_WITH_NITS.** Independently re-verified merged into today's main (post #2963–#2966): typecheck + ag-ui tests green; no overlap with today's changed files, and onComplete consumes the agent's own onFinish response, not the SSE mapping #2964 reworked. Fire-exactly-once holds including the injected-tool resume path (execute blocks inline on waitForSignal; resume only submits the signal). Complementary to the platform's canonical run-event log, correctly framed.

**One semantic needing explicit author sign-off:** persistence is gated on clean client stream delivery — if any enqueue fails during the final flush, onComplete never fires even though completedResponse is set server-side. A client dropping at the tail of a finished run won't persist. If the intent is 'never lose a finished turn', key persistence on completedResponse instead. Test gaps to close: injected-tool-path onComplete, and a client-disconnect regression test for the above. Needs: branch update (BEHIND), the sign-off, and human approval.

kwakayama (dismissed):

kojiwakayama:
Pushed `8128646b` — added the two `onComplete` test cases flagged as coverage gaps in the merge-readiness review (test-only; no source change). The branch already carried the main merge (`b996cf3d`), so this fast-forwards on top of it.

**`src/agent/ag-ui/handler.test.ts`**, in the `onComplete (server-side persistence)` describe:

- **`fires once for a resumable injected-tool run after the tool round-trip`** — drives the injected-tools/`AgentRuntime` path: the tool blocks inline on the session signal, the resume signal is submitted, and `onComplete` fires exactly once with the finalized conversation after the round-trip. Pins that a full tool round-trip stays inside one stream and yields a single persistence (the resume endpoint only signals; it never re-streams).
- **`does not fire when the client disconnects before the final flush`** — the run finalizes server-side (`onFinish` fires, `completedResponse` is set) but the consumer cancels before the terminal flush, so `onComplete` does **not** fire.

Note on the second test: it **test-locks the persist-on-clean-delivery semantic** — persistence is currently coupled to the client receiving the completed run, so a client dropping at the stream tail means a fully-finished server-side run is not persisted. If the intended behavior is instead persist-on-server-completion (i.e. persist whenever the response finalizes, regardless of client delivery), this test marks the exact decision point and would be the thing to flip.

Verification (mirroring CI): `deno check src/agent/index.ts` clean; full `src/agent/ag-ui/` suite green under `--no-check` (27 files); the pre-push hook ran the whole unit suite (2424 passed, 0 failed).

kojiwakayama (approved):
Approving after merge-readiness verification and the pushed test coverage: branch on current main (author's own update + cherry-picked tests), injected-tool and client-disconnect onComplete paths now test-pinned, full unit suite green (2,424). The client-disconnect test marks the persist-on-clean-delivery semantic as the explicit decision point if it should ever change.

kwakayama (approved):

kwakayama (approved):

PR #2966, opened 2026-07-17
Remove integration config sync from Veryfront Code

kojiwakayama:
## What changed

- Removes the top-level `integrations` field from `veryfront.config.ts` and its public input types. Legacy configuration fails explicitly with `Unknown config keys: integrations`.
- Removes `syncIntegrationConfig`, the `/integrations/config` caller, MCP integration-loader state, and the first-`tools/list` synchronization side effect.
- Removes config-only scope/runtime types and their public exports.
- Keeps remote integration discovery and execution request-scoped.
- Documents the three independent authorities:
  - agent source declares the agent's integration tools;
  - optional project policy narrows scope, tools, configuration, or execution mode;
  - connection inventory alone records credential readiness.
- Documents managed OAuth as a first-use Connect, callback, and retry flow. Project OAuth client credentials are only a custom-app override.

## Why

Framework configuration must not be a second integration-policy store. Synchronizing config while listing MCP tools coupled read traffic to policy mutation and made source selection, project policy, and credential state appear to be one enablement switch.

## Breaking change / migration

Remove any top-level `integrations` block from `veryfront.config.ts`. Declare integration tools in agent source. Create project policy only when an explicit restriction or configuration is needed. Let OAuth connections be established during tool use.

There is intentionally no alias, dual read, 404 fallback, or compatibility shim.

## Verification

- `deno task verify:quick`: passed.
- `deno task docs:validate`: passed, including all public guide contracts and 707 links.
- Focused guide contracts/content: 81 steps passed.
- Focused remote-tools suite: 11 steps passed.
- Pre-push format, lint, typecheck, and unit suite: passed.
- GitHub CI, coverage, binary E2E, npm smoke, security audit, and CodeQL must pass again on the final head before merge.
- Independent code review rejected an alternate implementation that would have normalized unrelated nested config fields; that change was not included.

## Rollout

Release framework `0.1.1080`, then bump the builder-assistant service before the API and Studio strict-contract cutover. Live OAuth verification follows the staged cross-service deployment.

Related: veryfront/veryfront-issue-inbox#27


kojiwakayama (commented):
Reviewed the full diff (config schema/loader, remote-tools, MCP server, types, docs, tests). **Sound and ready once CI reruns on the final head.** Findings:

**Correct by design**
- `validateConfigShape` change is safe: `validateVeryfrontConfig` still runs for validation, and `integrations` was the only transform-carrying field — raw-config passthrough semantics for every other key are unchanged.
- The removal cleanly excises the sync surface (config schema, `syncIntegrationConfig`, `/integrations/config` caller, MCP loader state, first-`tools/list` side effect) while keeping discovery/execution request-scoped. The `config-removal.test.ts` type-level guard (`"integrations" extends keyof VeryfrontConfigInput ? true : false = false`) is a nice compile-time lock.
- The three-authorities model documented here (agent source declares tools / optional policy narrows / connection inventory records credential readiness) matches exactly what the platform now enforces after api#3964/#3982/#3984.

**Blast radius of the hard fail — measured, effectively zero**
`Unknown config keys: integrations` breaks any existing project with the block at runtime boot, not just next edit. Empirical check: **zero** "Synced integration config" log lines in the last 24h in both `veryfront-staging` and `veryfront-production` — no live project exercises this path. Ship without a shim, as intended.

**Follow-up once runtimes are on 0.1.1080** (dead-code mandate)
The API's `/integrations/config` route and `syncIntegrationConfig` repository function become orphaned the moment no pre-1080 runtime can call them — small API removal PR after the cutover described in the rollout section. I can prepare it alongside the strict-contract cutover.

Rollout ordering in the body (1080 → builder-assistant bump → API/Studio strict cutover) is consistent with how the deploy chains actually work.

kwakayama (approved):

PR #3000, opened 2026-07-21
fix(module-loader): recover from stale mkdir memo on cache write (depends on #2999)

mattboon:
## Summary

The module loader memoised every directory it had `mkdir`'d in a process-wide `Set`, and added the directory to that `Set` in a `finally` block — so a `mkdir` that *failed* was recorded as having succeeded. From then on `ensureDir` was a no-op for that directory and **every subsequent write to it failed with ENOENT for the lifetime of the dev server**. The same permanent-failure state is reached whenever the cache directory disappears after being memoised (a manual `rm -rf .cache`, a cache sweep, a container restart).

This is what made the dev server unusable at scale: past ~30 compiled routes, every request 500s with `Failed to write module: … ENOENT`, including routes that rendered fine moments earlier.

A directory is now recorded as created only when `mkdir` actually succeeds, and the write retries once — forcing the `mkdir` past the memo — when it fails with a cache-write race error.

## Reproduction

- Reproducer: [`default-config/`](https://github.com/mattboon/veryfront-router-testing/tree/main/default-config) at scale — see [RESULTS.md § "framework instability at scale"](https://github.com/mattboon/veryfront-router-testing/blob/main/RESULTS.md)
- Test: `src/rendering/orchestrator/module-loader/module-persistence.test.ts`

## Test evidence

Before (both new cases red, with the exact production signature):

```
recreates the output directory when it disappears after being cached ... FAILED
  error: NotFound: No such file or directory (os error 2): writefile
         '/…/vf-module-persist-out-…/lib/uses-crypto.e1b64c.js'
    at persistTransformedModule (module-persistence.ts:81:33)
does not cache a failed mkdir as a created directory ... FAILED
FAILED | 0 passed (1 step) | 1 failed (2 steps)
```

After:

```
writes transformed code, registers MDX path-cache, and updates module cache ... ok
recreates the output directory when it disappears after being cached ... ok
does not cache a failed mkdir as a created directory ... ok
ok | 1 passed (3 steps) | 0 failed
```

## SSR evidence

The ENOENT cascade is gone entirely. After a full 56-route sweep against a single dev-server instance:

```
$ grep -ci "Failed to write module\|ENOENT" /tmp/vf-dev.log
0
```

Previously this fired on virtually every request past warm-up. Routes RESULTS.md recorded as lost to the cascade (`/use-server`, `/test/bb-env-vars`, `/test-app/c-tsx-ext-component`) now return 200.

## Client evidence

`/use-server` and `/test-app/c-tsx-ext-component` previously rendered the framework's `Runtime Error` page instead of the intended page. Both now render and hydrate; the full sweep shows no cache-write failure of any kind.

## Related

- Bug 9 of the reproducer matrix — the one the reproducer calls "the single biggest impediment to a healthy dev experience".

---

rebase the chain onto `main`.

### DIFF
--- a/src/rendering/orchestrator/module-loader/module-persistence.test.ts
+++ b/src/rendering/orchestrator/module-loader/module-persistence.test.ts
@@ -47,4 +47,95 @@ describe("module-loader/module-persistence", () => {
       await Deno.remove(tmpDir, { recursive: true }).catch(() => undefined);
     }
   });
+
+  it("recreates the output directory when it disappears after being cached", async () => {
+    const projectDir = await Deno.makeTempDir({ prefix: "vf-module-persist-project-" });
+    const tmpDir = await Deno.makeTempDir({ prefix: "vf-module-persist-out-" });
+    const localAdapter = await getLocalAdapter();
+    const filePath = join(projectDir, "lib/uses-crypto.ts");
+    const moduleCache = new Map<string, string>();
+
+    try {
+      await Deno.mkdir(dirname(filePath), { recursive: true });
+
+      const first = await persistTransformedModule({
+        filePath,
+        projectDir,
+        tmpDir,
+        transformedCode: "export const a = 1;",
+        localAdapter,
+        moduleCache,
+        cacheKey: "first",
+      });
+      assertEquals(await Deno.readTextFile(first), "export const a = 1;");
+
+      // Something outside the loader wipes the cache dir (manual `rm -rf .cache`,
+      // a cache sweep, a container restart). The mkdir memo still claims it exists.
+      await Deno.remove(join(tmpDir, "lib"), { recursive: true });
+
+      const second = await persistTransformedModule({
+        filePath,
+        projectDir,
+        tmpDir,
+        transformedCode: "export const a = 2;",
+        localAdapter,
+        moduleCache,
+        cacheKey: "second",
+      });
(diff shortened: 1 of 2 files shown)


kwakayama (commented):
Score: 82/100

The implementation is directionally sound: only memoize successfully-created directories, and retry once on cache-write race errors. I did not find a direct code blocker in the incremental diff.

Not merge-ready yet:
- This is stacked on #2999, which is not merge-ready.
- The PR status rollup only shows CLA, not the full CI suite.
- Minor watch item: the initial `mkdir` failure is intentionally swallowed before the write/retry path, which weakens observability a bit. Not a blocker by itself, but worth considering if this path remains noisy in production.

kwakayama (dismissed):

Score: 92/100.

Rationale: narrowly fixes stale mkdir memo recovery in cache writes, with low blast radius and no remaining critical review findings from the stack pass.

Verification from the rebased stack tip:
- deno task audit

kwakayama:
Merge update: auto-merge is enabled and the refreshed checks passed, but GitHub still reports REVIEW_REQUIRED because the requested reviewer is kojiwakayama. The existing approval from kwakayama is present but does not satisfy that required-review gate.

Next step: kojiwakayama approval, then auto-merge should complete.

kojiwakayama (approved):

PR #3001, opened 2026-07-21
fix(transforms,module-server): never emit or serve source extensions to the browser (depends on #2999)

mattboon:
## Summary

A page importing `@/lib/constants.ts` rendered on the server but never hydrated, dying with a strict-MIME rejection. Two independent defects produced it:

**1. The emitter kept the source extension.** `AliasStrategy`'s relative-fallback branch (used when no `moduleServerUrl` is configured) built the specifier by hand and never called `normalizeExtension`, so an explicit `.ts`/`.tsx`/`.jsx`/`.mdx` survived into the emitted browser import. Its guard regex also omitted `js`, so an explicit `.js` specifier became `.js.js`. The other two branches already normalize; this one now matches them.

**2. The server typed the response from the wrong thing.** The dev module server derived `Content-Type` from the *requested* path, so `/_vf_modules/lib/constants.ts` was served as `application/typescript` — even though the body is compiled JavaScript. With `x-content-type-options: nosniff`, browsers refuse to execute that as a module.

Either fix alone would close the reported route; both are wrong independently, so both are fixed.

## Reproduction

- Routes: [`b-ts-ext-lib.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/b-ts-ext-lib.tsx), [`c-tsx-ext-component.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/c-tsx-ext-component.tsx), [`g-transitive-ts-ext.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/g-transitive-ts-ext.tsx)
- Tests: `src/transforms/import-rewriter/strategies/alias-strategy.test.ts`, `src/modules/server/module-server.test.ts`

## Test evidence

Before — red isolated exactly to the fallback branch (the `moduleServerUrl` and `ssr` branches already passed, which is why this only ever bit projects on the fallback path):

```
relative path fallback (no moduleServerUrl) ... FAILED (due to 2 failed steps)
FAILED | 0 passed (24 steps) | 1 failed (4 steps)
```

After:

```
ok | 17 passed (285 steps) | 0 failed
```

## SSR evidence

The emitted client bundle no longer carries the source extension:

```
$ curl -s http://localhost:3010/_vf_modules/pages/test/b-ts-ext-lib.js | grep '^import'
import { COLORS } from "../../lib/constants.js"      # was "../../lib/constants.ts"
```

And a `.ts` URL is served as JavaScript regardless:

```
$ curl -sD- -o/dev/null http://localhost:3010/_vf_modules/lib/constants.ts | grep -i content-type
content-type: application/javascript; charset=utf-8   # was application/typescript; charset=utf-8
```

## Client evidence

Chromium sweep (`client-sweep.mjs`), console + network errors:

```
PASS /test/a-no-ext-no-server
PASS /test/b-ts-ext-lib          # was: MIME type application/typescript
PASS /test/c-tsx-ext-component   # was: MIME type application/typescript
```

The adjacent-route sweep is what caught that `g-transitive-ts-ext` shares the *symptom* but not this cause — it reaches `node:crypto` at render time, handled in the later PRs of this chain.

## Related

- Bug 1 of the reproducer matrix; [veryfront-issue-inbox#82](https://github.com/veryfront/veryfront-issue-inbox/issues/82)

---

rebase the chain onto `main`.

### DIFF
--- a/src/modules/server/module-server.test.ts
+++ b/src/modules/server/module-server.test.ts
@@ -20,7 +20,7 @@ import {
 import { denoAdapter } from "#veryfront/platform/adapters/runtime/deno/index.ts";
 import { createMockAdapter } from "#veryfront/platform/adapters/mock.ts";
 import { VERSION } from "#veryfront/utils/version.ts";
-import { isModuleRequest } from "./module-server.ts";
+import { getDevModuleContentType, isModuleRequest } from "./module-server.ts";
 import { clearReleaseModuleResponseCache } from "./module-response-cache.ts";
 import { clearSourceMissCache } from "./module-source-resolution-cache.ts";
 import { deleteEnv, setEnv } from "#veryfront/platform/compat/process.ts";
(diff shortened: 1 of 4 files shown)


kwakayama (commented):
Score: 86/100

The incremental fix looks coherent: browser-facing specifiers are normalized away from source extensions, and the dev module server serves compiled TS/TSX/MDX/MD as JavaScript instead of deriving MIME type from the request path. I did not find a direct blocker in this diff.

Not merge-ready yet:
- This is stacked on lower open PRs, including #2999.

Watch item: this keeps source-path shaping and served-artifact MIME policy split across transform and module-server code, so future changes need to keep both sides aligned.

kwakayama (dismissed):

Score: 91/100.

Rationale: source-extension browser MIME handling is scoped and covered by the stack regression pass; no remaining blocker from review.

Next step: merge after #3000 and refreshed checks are green.

kwakayama (approved):
Follow-up approval after #3000 merged and the stack was rebased onto main.

Score: 91/100.

Rationale: source-extension browser MIME handling remains scoped after the rebase, and the stack regression pass stayed green.

Verification:
- deno task audit

kojiwakayama (approved):

PR #3002, opened 2026-07-21
fix(transforms): strip server-only data hooks from the client artifact (depends on #2999)

mattboon:
## Summary

The browser artifact is compiled from the same source file as the server one, so `getServerData` / `getStaticData` / `getStaticPaths` bodies shipped to the client **along with everything they import**. A page whose loader reached `node:crypto` therefore linked against the node-builtin noop polyfill and hydration died with `does not provide an export named 'createHash'` — for code that never runs in the browser at all.

esbuild cannot solve this for us: in transform mode (as opposed to bundle mode) it never drops an import, because it cannot prove the module is side-effect free. Confirmed directly:

```
treeShaking: true   -> import { hashOf } from "@/lib/uses-crypto";   # kept
treeShaking: false  -> import { hashOf } from "@/lib/uses-crypto";   # kept
```

So the browser pipeline gains a stage that empties the hooks' bodies and then removes the imports left unreferenced.

**Why empty the bodies rather than delete the declarations:** the binding stays valid for any surviving `export` clause, so `typeof mod.getServerData === "function"` still holds for the client router's data-fetching check. Deleting the declaration would break that.

Every step is conservative — anything that cannot be parsed with confidence is left exactly as it was, and side-effect-only imports (`import "./x"`) are never removed.

## Reproduction

- Route: [`d-server-static.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/d-server-static.tsx)
- Test: `src/transforms/pipeline/stages/browser-server-exports-strip.test.ts`

## Test evidence

15 new cases covering function declarations, `var`-assigned arrows, all three hooks, aliased/namespace/default/side-effect imports, and the "leave it alone" paths.

```
ok | 22 passed (194 steps) | 0 failed      # src/transforms/pipeline/
```

## SSR evidence

`/test/d-server-static` still renders server-side (the hook is untouched on the SSR path), and the server-only import is gone from the client artifact:

```
$ curl -s .../pages/test/d-server-static.js | grep -E '^import|getServerData'
import { jsx, jsxs } from "https://esm.sh/react@19.2.4/jsx-runtime?…"
async function getServerData(_ctx) { throw new Error('server-only'); }
                                              # "../../lib/uses-crypto.js" no longer imported

$ curl -so/dev/null -w '%{http_code}' http://localhost:3010/test/d-server-static
200
```

## Client evidence

```
PASS /test/d-server-static    # was: SyntaxError … does not provide an export named 'createHash'
```

## Related

- Bug 2 of the reproducer matrix.

---

rebase the chain onto `main`.

### DIFF
--- a/src/transforms/pipeline/index.ts
+++ b/src/transforms/pipeline/index.ts
@@ -21,6 +21,7 @@ import type {
   TransformResult,
 } from "./types.ts";
 import {
+  browserServerExportsStripPlugin,
   compilePlugin,
   cssStripPlugin,
   finalizePlugin,
@@ -50,6 +51,7 @@ const BROWSER_PIPELINE: TransformPlugin[] = [
   parsePlugin,
   compilePlugin,
   cssStripPlugin, // Strip CSS imports before they hit import resolution
+  browserServerExportsStripPlugin, // Drop server-only hooks + their now-unused imports
   resolveImportsPlugin, // Unified import resolution
   finalizePlugin,
 ];
--- a/src/transforms/pipeline/stages/browser-server-exports-strip.test.ts
+++ b/src/transforms/pipeline/stages/browser-server-exports-strip.test.ts
@@ -1,0 +1,479 @@
+import "#veryfront/schemas/_test-setup.ts";
+import "../../plugins/__tests__/code-parser-setup.ts";
+import { assertEquals, assertStringIncludes } from "#veryfront/testing/assert.ts";
+import { describe, it } from "#veryfront/testing/bdd.ts";
+import {
+  browserServerExportsStripPlugin,
+  stripServerOnlyExports,
+} from "./browser-server-exports-strip.ts";
+import type { TransformContext } from "../types.ts";
+
+function assertNotIncludes(haystack: string, needle: string): void {
+  assertEquals(haystack.includes(needle), false, `expected not to find ${needle} in:\n${haystack}`);
+}
+
+/** Identifier occurrences, so "kept the import" and "kept the binding" differ. */
+function occurrences(haystack: string, name: string): number {
+  return haystack.match(new RegExp(`\\b${name}\\b`, "g"))?.length ?? 0;
+}
+
+describe("browser-server-exports-strip", () => {
+  describe("emptying server-only hooks", () => {
+    it("empties an exported async function declaration body", async () => {
+      const code = [
+        `import { hashOf } from "../lib/uses-crypto.js";`,
+        `async function getServerData(ctx) {`,
+        `  return { props: { hashed: hashOf("hello") } };`,
+        `}`,
+        `function Page() { return null; }`,
+        `export { getServerData, Page as default };`,
+      ].join("\n");
+
+      const result = await stripServerOnlyExports(code);
+
+      assertNotIncludes(result, `hashOf("hello")`);
+      // The binding survives so the export clause stays valid.
+      assertStringIncludes(result, "getServerData");
+      assertStringIncludes(result, "Page as default");
+      assertStringIncludes(result, "return null");
+    });
+
+    it("empties a directly exported function declaration", async () => {
+      const code = `export function getStaticPaths() { return db.query(); }`;
+      const result = await stripServerOnlyExports(code);
+      assertNotIncludes(result, "db.query");
+      assertStringIncludes(result, "getStaticPaths");
+    });
+
+    it("replaces an exported arrow initialiser", async () => {
+      const code = `export const getStaticData = async (ctx) => ({ props: { x: secret() } });`;
+      const result = await stripServerOnlyExports(code);
+      assertNotIncludes(result, "secret()");
+      assertStringIncludes(result, "getStaticData");
+    });
+
+    it("handles all three hooks in one module", async () => {
+      const code = [
+        `export async function getServerData() { return serverOne(); }`,
+        `export function getStaticData() { return serverTwo(); }`,
+        `export const getStaticPaths = () => serverThree();`,
+      ].join("\n");
+
+      const result = await stripServerOnlyExports(code);
+
+      assertNotIncludes(result, "serverOne");
+      assertNotIncludes(result, "serverTwo");
+      assertNotIncludes(result, "serverThree");
+    });
+
+    it("leaves a module without server hooks untouched", async () => {
+      const code = `import { x } from "./x.js";\nexport default function Page() { return x; }`;
+      assertEquals(await stripServerOnlyExports(code), code);
+    });
+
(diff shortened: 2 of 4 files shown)


kwakayama (changes requested):
Score: 72/100

Requesting changes. The goal is valid, but the new browser server-export stripping stage is too heuristic for a compiler transform that is meant to prevent server-only imports from leaking into the browser artifact.

Blocking concern:
- `src/transforms/pipeline/stages/browser-server-exports-strip.ts` removes imports by regex-counting identifiers in the remaining code. That count includes strings and comments, so an import used only by a stripped hook can be retained if the binding name appears in inert text. In that case the server-only dependency chain can still ship to the browser and preserve the hydration/link failure this PR is meant to remove.

Please use an AST/scope-aware identifier walk, or at minimum exclude strings/comments/templates from the reference scan, and add regressions for comment-only and string-only references. I would also like coverage for same-name non-exported declarations and future hook forms, since this stage currently relies on hand-rolled source scanning.

mattboon:
Addressed. The stage no longer scans source text at all: it now runs on the repo's `CodeParser` (Babel) AST contract, the same reason `src/rendering/rsc/export-extractor.ts` already uses it.

The hand-rolled scanner is gone (`findInertRegions`, `matchBrace`, the two `strip*Declaration` helpers, `importedBindings`). `referencedIdentifiers()` walks the parsed tree, collects only `Identifier`/`JSXIdentifier` nodes, skips `ImportDeclaration` subtrees, and pre-marks non-reference positions (non-computed member properties and object/class keys). Comments, string literals, template chunks and JSX text are not identifier nodes, so they cannot enter the reference set by construction.

Regressions added for exactly what you asked for:

- binding referenced only in a line comment, and only in a block comment
- binding referenced only in a string literal
- binding referenced only in template literal text, plus the inverted case: a reference inside `${...}` **is** real code and must keep the import
- binding appearing only as JSX text
- same-name non-exported declaration: a private `getServerData` alongside a genuinely exported hook stays untouched
- declaration forms: `export async function`, `export const ... = async function`, a hook declared before a separate `export { ... }` clause, and `export { other as getServerData }`, which must not strip the local `other`

Two other findings from the same review round are fixed here as well, since they had the same root cause. An unexported declaration is never emptied, and an import that falls out of use is reduced to a side-effect import rather than deleted, because this pass has no side-effect information about the module it points at. Node built-ins are the exception, since in the browser they resolve to a noop polyfill. That matches what esbuild does with an external import whose bindings go unused.

One honest caveat: a JSX **attribute** name (`<p hashOf="1" />`) is a `JSXIdentifier` and counts as a reference, so it keeps an import alive. That is the safe direction (over-counting only ever retains an import, it never ships a server-only chain) and the file documents it.

**Verification:** stage suite 42 steps passing, `deno check` clean on the test file and on `src/transforms/index.ts`, `deno fmt`/`deno lint` clean.

kwakayama (dismissed):
Follow-up after fixes: approving. The implementation now uses parsed AST references instead of regex/comment/string matching, and inert text regressions are covered. Local verification: \running 1 test from ./src/transforms/pipeline/stages/browser-server-exports-strip.test.ts
browser-server-exports-strip ...
  emptying server-only hooks ...
    empties an exported async function declaration body ... ok (10ms)
    empties a directly exported function declaration ... ok (1ms)
    replaces an exported arrow initialiser ... ok (1ms)
    handles all three hooks in one module ... ok (1ms)
    leaves a module without server hooks untouched ... ok (0ms)
    does not treat a same-named string as a declaration ... ok (1ms)
    leaves a non-exported function of the same name alone ... ok (1ms)
    leaves a local declaration that is only aliased to a hook name alone ... ok (0ms)
    empties a hook declared before a separate export clause ... ok (1ms)
    keeps client code that follows a regular expression containing braces ... ok (1ms)
    keeps client code after a division that looks like a regular expression ... ok (1ms)
    handles a template literal with braces and interpolation ... ok (1ms)
    handles minified single-line input ... ok (1ms)
    handles TSX with types and JSX ... ok (2ms)
    leaves a module that does not parse unchanged ... ok (1ms)
  emptying server-only hooks ... ok (24ms)
  import bindings ...
    reduces an unreferenced project import to a side-effect import ... ok (1ms)
    removes an unreferenced node builtin import outright ... ok (0ms)
    keeps an import that the client still references ... ok (1ms)
    keeps an import when only one of its bindings is used ... ok (0ms)
    keeps a bare side-effect import untouched ... ok (1ms)
    keeps a default import the client renders with ... ok (0ms)
    reduces a namespace import the client no longer uses ... ok (1ms)
    does not count a matching property name as a reference ... ok (0ms)
    counts a computed property access as a reference ... ok (1ms)
    counts a JSX component as a reference ... ok (0ms)
  import bindings ... ok (5ms)
  inert text is not a reference ...
    does not count a line comment mention ... ok (0ms)
    does not count a block comment mention ... ok (1ms)
    does not count a string literal mention ... ok (0ms)
    does not count a template literal mention ... ok (1ms)
    counts a template literal interpolation, which is real code ... ok (0ms)
    does not count a JSX text node mention ... ok (0ms)
  inert text is not a reference ... ok (4ms)
  declaration forms ...
    leaves a private same-named declaration alone beside a real hook ... ok (0ms)
    leaves a local aliased to a hook name alone beside a real hook ... ok (0ms)
    empties a hook declared as an exported function expression ... ok (0ms)
    empties a hook declared as a directly exported async function ... ok (0ms)
  declaration forms ... ok (1ms)
  plugin ...
    drops the server-only import chain from the client artifact ... ok (1ms)
    does not run for the ssr target ... ok (0ms)
  plugin ... ok (1ms)
browser-server-exports-strip ... ok (37ms)

ok | 1 passed (42 steps) | 0 failed (39ms). Score: 94/100. Next step: merge after base stack and checks are green.

kwakayama:

Score: 94/100.

Verification:
- deno test --allow-all src/transforms/pipeline/stages/browser-server-exports-strip.test.ts

kwakayama (approved):
Follow-up approval after #3001 merged and the stack was rebased onto main.

Score: 94/100.

Rationale: implementation still uses parsed AST references instead of regex/comment/string matching; inert text regressions remain covered after the rebase.

Verification:
- deno task audit

kwakayama (approved):
Follow-up approval after rebasing #3002 onto the latest main.

Score: 94/100.

Rationale: the parser-backed transform and inert-text regressions are unchanged by the rebase, and the stack validated locally on the rebased head.

Verification:
- deno task audit

Next step: wait for refreshed GitHub checks and the required reviewer gate, then auto-merge.

kojiwakayama (approved):

PR #3003, opened 2026-07-21
fix(hydration): stop the /index.js fallback from masking real load errors (depends on #2999)

mattboon:
## Summary

When a page module was fetched but failed to **link** — say a browser-side import of a Node builtin — the hydration renderer swallowed that error, retried at `/_vf_modules/pages/<route>/index.js`, and reported the resulting 404.

So every affected page surfaced the same misleading `Failed to fetch dynamically imported module: …/index.js`, pointing at a routing/path problem that does not exist, while the actual cause was discarded. That is exactly why the reproducer observed this "fires alongside #1 or #2, never on baseline routes" — it is not an independent failure at all, it is an error-reporting defect that masks other failures.

The `/index.js` retry now happens only when the module genuinely could not be *fetched*, and the original error is rethrown rather than the fallback's.

## Reproduction

- Same routes as bugs 1 and 2 — `b-ts-ext-lib`, `d-server-static`, `g-transitive-ts-ext`
- Test: `src/html/hydration-script-builder/templates/renderer.test.ts`

## Test evidence

The decision is a named helper in the emitted browser script, extracted and **evaluated** in-process so the behaviour is under test rather than a substring match:

```
isModuleNotFoundError ...
  treats a failed fetch as module-not-found ... ok
  does not treat a link error as module-not-found ... ok
  does not treat a module evaluation error as module-not-found ... ok
  handles null and non-Error values ... ok
page module fallback ...
  only retries the /index.js path for module-not-found errors ... ok
  rethrows the original error rather than the fallback's ... ok
ok | 1 passed (42 steps) | 0 failed
```

## Client evidence

Before — the real cause is logged then buried under a fabricated 404:

```
FAIL /test/g-transitive-ts-ext
   console: Failed to load page from hydration data: SyntaxError: … node-noop.js does n…
   http 404: http://localhost:3010/_vf_modules/pages/test/g-transitive-ts-ext/index.js
   requestfailed: …/index.js
   console: Client initialization error: TypeError: Failed to fetch dynamically imported module…
   console: Hydration failed signal received: TypeError: Failed to fetch dynamically imported…
```

After — the 404 is gone and the reported error is the true one:

```
FAIL /test/g-transitive-ts-ext
   pageerror: The requested module '…/node-noop.js' does not provide an export named 'createHash'
```

(That route still fails — it genuinely calls `node:crypto` during client render — but it now says so. The next PR in the chain makes even that fail at the call site.)

## Related

- Bug 8 of the reproducer matrix.

---

rebase the chain onto `main`.

### DIFF
--- a/src/html/hydration-script-builder/templates/renderer.test.ts
+++ b/src/html/hydration-script-builder/templates/renderer.test.ts
@@ -189,4 +189,89 @@ describe("hydration-script-builder/templates/renderer", () => {
       assertIncludes(getRendererScript(), "container.__reactRoot");
     });
   });
+
+  describe("isModuleNotFoundError", () => {
+    // Evaluate the helper out of the emitted browser script so the behaviour
+    // itself is under test, not just the presence of a substring. Only the
+    // helper is extracted — the rest of the script touches `window`.
+    function helperSource(): string {
+      const script = getRendererScript();
+      const start = script.indexOf("function isModuleNotFoundError(error) {");
+      assertEquals(start >= 0, true, "isModuleNotFoundError not found in renderer script");
+      const end = script.indexOf("\n    }", start);
+      assertEquals(end > start, true, "could not find end of isModuleNotFoundError");
+      return script.slice(start, end + "\n    }".length);
+    }
+
+    function isModuleNotFoundError(error: unknown): boolean {
+      return new Function(
+        "error",
+        `${helperSource()}\nreturn isModuleNotFoundError(error);`,
+      )(error) as boolean;
+    }
+
+    it("treats a failed fetch as module-not-found", () => {
+      assertEquals(
+        isModuleNotFoundError(
+          new TypeError(
+            "Failed to fetch dynamically imported module: http://x/_vf_modules/pages/a.js",
+          ),
+        ),
+        true,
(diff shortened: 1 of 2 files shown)


kwakayama (commented):
Score: 88/100

The incremental change looks correct: the renderer now distinguishes module fetch failures from link/evaluation failures, and preserves the original error instead of replacing it with a misleading `/index.js` fallback failure.

Not merge-ready yet:
- This is stacked on lower open PRs, including blocked #2999.

Watch item: the browser error classifier is necessarily string-based, so it may need future adjustment if runtime wording changes. I do not see that as a blocker for this specific fix.

kwakayama (dismissed):

Score: 92/100.

Rationale: hydration /index.js fallback behavior is narrowly corrected and no remaining blocker was found in the critical review pass.

kwakayama (approved):
Follow-up approval after #3002 merged and #3003 was rebased onto main.

Score: 93/100.

Rationale: the fallback now preserves real module load failures instead of hiding them behind /index.js, with focused regression coverage retained after the rebase.

Verification:
- deno task audit

kojiwakayama (approved):

Issue #453, opened 2026-02-25
BUG: Scaffolded project completely broken — 3 errors: CSS module not found, layout blocked, preview 404


mattboon:
<img width="1678" height="1118" alt="Image" src="https://github.com/user-attachments/assets/fab71604-7cf4-49b1-8471-b96e4860e658" />
<img width="2044" height="1022" alt="Image" src="https://github.com/user-attachments/assets/26fb44df-c917-4c78-82aa-7750a273d0e8" />
<img width="1080" height="756" alt="Image" src="https://github.com/user-attachments/assets/e113255a-6794-4a93-b810-488f03d7d25e" />

### Steps to Reproduce

1. Create a new project via scaffolding using the Chat via MCP - it created the project file structure
2. Project includes `globals.css` at root with Tailwind v4 syntax:
   ```css
   @import "tailwindcss";
   
   @theme {
     --background: 0 0% 100%;
     /* ... theme variables ... */
   }
   ```
3. `app/layout.tsx` imports it via:
   ```tsx
   import "@/globals.css"
   ```
4. `app/page.tsx` was also scaffolded — a "coming soon" waitlist page with email form, radial gradient background, and Tailwind v4 theme variables (`hsl(var(--primary)/0.16)`, etc.)
5. Load the preview

### Scaffolded Files

```
app/
  layout.tsx
  page.tsx
(folder)
  utils.ts
globals.css
veryfront.config.ts
```

### Errors

**Error 1 — Module not found**
```
Runtime Error
TypeError
Module not found: file:///_vf_modules/globals.css.js
Suggestion:
Make sure the imported module exists and the path is correct. For npm packages, ensure they're installed.
```

**Error 2 — Component blocked**
```
Runtime Error
VeryfrontError[build]
Component /app/app/layout.tsx is temporarily blocked due to repeated failures. Will retry in 60s.
Suggestion:
Ensure your component is properly exported and the import path is correct.
```

**Error 3 — Preview 404**

The preview URL (`https://fierce-shadow.preview.veryfront.com/`) returns a 404.

The iframe preview URL with query params (`?studio_embed=true&color_mode=light&_t=0`) shows the first error above.

### Expected Behavior

The project should load and render correctly in the preview.

### Environment

- Project: `fierce-shadow` (`https://veryfront.com/projects/fierce-shadow`)
- Created via: "chat" scaffolding via MCP
- Preview URL: `https://fierce-shadow.preview.veryfront.com/`

[...]

PR #3005, opened 2026-07-21
fix(module-loader): resolve dynamic import() specifiers for SSR (depends on #2999)

mattboon:
## Summary

`await import("@/lib/x")` and `await import("../../lib/x.ts")` inside `getServerData` both 500'd:

```
Module not found "file:///_vf_modules/lib/uses-crypto.js"
```

The SSR module loader discovers a module's local dependencies, transforms each to a temp file, and rewrites the importer to point at those files. That discovery used `findStaticImportFromSpans`, which **deliberately skips `import(`** (`source-spans.ts:218`). Dynamic specifiers were therefore never resolved; they fell through to the alias rewrite, which produced a `/_vf_modules/…` path that means nothing to the runtime's own resolver, and the render died.

Dynamic imports with a literal specifier are now collected alongside static ones. Their span covers just the quoted string, so the rewrite replaces the specifier and leaves `await import(...)` intact. Non-literal specifiers (`import(ctx.query.get("m"))`) are left alone — their target is only knowable at runtime.

**Bugs 3 and 4 are this one defect**, as the reproducer's fix hypothesis anticipated ("Likely a shared fix"). Shipping them together rather than splitting one change across two PRs.

## Reproduction

- Routes: [`e-server-dynamic-at-alias.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/e-server-dynamic-at-alias.tsx) (bug 3), [`f-server-dynamic-relative.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/f-server-dynamic-relative.tsx) (bug 4)
- Tests: `src/rendering/orchestrator/module-loader/dependency-resolver.test.ts`

## Test evidence

```
ok | 2 passed (7 steps) | 0 failed
```

Four new cases: dynamic `@/` alias, dynamic relative carrying a `.ts` extension, mixed static+dynamic in one module, and a non-literal specifier that must be ignored. One asserts the rewritten output directly:

```ts
assertStringIncludes(rewritten, `await import("file:///tmp/out/lib/uses-crypto.abc.js")`)
```

## SSR evidence

```
/test/e-server-dynamic-at-alias   500 -> 200
/test/f-server-dynamic-relative   500 -> 200
```

Not merely a 200 — the dynamically imported module actually executes. The page renders the real SHA-256 prefix of `"hello"`:

```
SSR computed hash: <code …>2cf24dba</code>
```

## Client evidence

```
PASS /test/e-server-dynamic-at-alias
PASS /test/f-server-dynamic-relative
2/2 routes hydrated clean
```

## Related

- Bugs 3 and 4 of the reproducer matrix. These were the escape hatch for keeping server-only code out of the client bundle, so they were effectively unavailable on the pages router.

---

rebase the chain onto `main`.

### DIFF
--- a/src/rendering/orchestrator/module-loader/dependency-resolver.test.ts
+++ b/src/rendering/orchestrator/module-loader/dependency-resolver.test.ts
@@ -28,12 +28,14 @@ async function withDependencyFixture<T>(
   }
 }
 
-function withSpan<T extends { full: string }>(
+// Every hand-built fixture below is a static `from "…"` span, so the helper
+// stamps `isDynamic: false` instead of repeating it at each call site.
+function withStaticSpan<T extends { full: string }>(
   source: string,
   dep: T,
-): T & { start: number; end: number } {
+): T & { start: number; end: number; isDynamic: false } {
   const start = source.indexOf(dep.full);
-  return { ...dep, start, end: start + dep.full.length };
+  return { ...dep, start, end: start + dep.full.length, isDynamic: false };
 }
 
 describe("module-loader/dependency-resolver", () => {
@@ -75,7 +77,7 @@ describe("module-loader/dependency-resolver", () => {
     ].join("\n");
 
     const rewritten = rewriteResolvedDependencyImports(source, [
-      withSpan(source, {
+      withStaticSpan(source, {
         full: `from "@/Button"`,
         path: "@/Button",
         relativePath: "Button",
@@ -83,7 +85,7 @@ describe("module-loader/dependency-resolver", () => {
         depTempPath: "/tmp/components/Button.abc.js",
         isLocalLib: false,
       }),
-      withSpan(source, {
+      withStaticSpan(source, {
         full: `from "../lib/value"`,
         path: "../lib/value",
         relativePath: "../lib/value",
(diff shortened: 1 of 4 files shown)


kwakayama (changes requested):
Score: 65/100

Requesting changes. This PR currently fails local type diagnostics.

Blocking issue:
- `src/rendering/orchestrator/module-loader/dependency-resolver.test.ts:78` and `:86` construct `TransformedModuleDependency` fixtures without the new required `isDynamic` field. `deno check` fails with TS2322 because `ResolvedModuleDependency` now requires `isDynamic`.

Fix by adding `isDynamic: false` to the existing static fixture dependencies, or by changing the fixture helper to default static dependencies to `false`.

Additional watch item: `findDynamicImportSpans` handles literal `import("...")` cases, but remains a source scanner. Please consider coverage for valid dynamic-import syntax variants, including comment/whitespace forms, so SSR resolution does not silently diverge again.

This PR is also stacked on lower open PRs and only shows CLA in the status rollup, so it is not merge-ready.

mattboon:
Both points addressed, and the second one turned up real bugs.

**Type error.** Fixed by making the fixture helper stamp the field, so static fixtures do not repeat it at every call site: `withSpan` is now `withStaticSpan`, returning `T & { start: number; end: number; isDynamic: false }`. `deno check src/rendering/orchestrator/module-loader/dependency-resolver.test.ts` is clean (was 2x TS2322).

**Dynamic-import syntax variants.** Your watch item was justified: probing before writing the tests found four cases the scanner got genuinely wrong, all from using bare whitespace skipping where comments can appear.

- `import(/* webpackChunkName */ "./a.js")` was missed entirely
- `import /* c */ ("./a.js")` was missed
- `import("./a.js" /* c */)` was missed, the worst of the four: a whole-argument literal was misclassified as not-whole
- a comment on its own line inside the parentheses was missed

Fixed with a `skipWhitespaceAndComments` helper (reusing the existing comment handling, and deliberately not treating strings as skippable) applied at the paren lookup, the specifier lookup and the after-specifier check.

Coverage added for the variants: whitespace and newline forms, the four comment placements, `await import(...)` nested in an arrow function, `import(...).then(...)`, `?query` and `#hash` suffixes, a dynamic import inside a block comment, and a `replaceSourceSpans` round trip asserting the span still covers only the quoted specifier when comments surround it.

The deliberate skip rule is preserved throughout: a literal that is only the start of the argument (`import("./foo" + suffix)`, `.concat(...)`, a ternary, a template substitution) is still skipped rather than rewritten, including the new `import("./a.js" /* then */ + suffix)` form.

**Verification:** `deno check` clean on both test files; `source-spans` and `dependency-resolver` suites 33 steps passing; full `src/transforms/mdx/esm-module-loader/` sweep 383 steps passing; `deno fmt`/`deno lint` clean.

kwakayama (dismissed):
Follow-up after fixes: approving. Dynamic and static specifier rewrite metadata is now preserved, including \. Local verification: \running 1 test from ./src/transforms/import-rewriter/strategies/alias-strategy.test.ts
AliasStrategy ...
  matches ...
    should match @/ imports ... ok (1ms)
    should not match scoped packages ... ok (0ms)
    should not match relative imports ... ok (0ms)
  matches ... ok (1ms)
  rewrite ...
    should rewrite @/ to relative path from root-level file ... ok (0ms)
    should rewrite @/ from nested file ... ok (0ms)
    should keep existing extension for known extensions ... ok (0ms)
    should add .js extension when no known extension ... ok (0ms)
    should rewrite @/ to /_vf_modules/ path for SSR target ... ok (0ms)
    should rewrite @/ with nested path to /_vf_modules/ for SSR ... ok (0ms)
    should normalize extension for SSR ... ok (0ms)
    moduleServerUrl path ...
      should use absolute path when moduleServerUrl is configured ... ok (0ms)
      should handle file index path mismatch with moduleServerUrl ... ok (0ms)
    moduleServerUrl path ... ok (0ms)
    CSS file imports (issue #453) ...
      should NOT append .js to .css imports for SSR ... ok (0ms)
      should NOT append .js to .css imports with moduleServerUrl ... ok (0ms)
      should NOT append .js to .css imports in browser fallback ... ok (0ms)
    CSS file imports (issue #453) ... ok (1ms)
    relative path fallback (no moduleServerUrl) ...
      should handle file at components/elements depth correctly ... ok (0ms)
      relative path when file index has different structure (known limitation) ... ok (0ms)
      should rewrite an explicit .ts extension to .js ... ok (0ms)
      should rewrite an explicit .tsx extension to .js ... ok (0ms)
      should not double-append .js to an explicit .js extension ... ok (0ms)
    relative path fallback (no moduleServerUrl) ... ok (1ms)
    source extensions in moduleServerUrl and ssr targets ...
      should rewrite .ts to .js with moduleServerUrl ... ok (0ms)
      should rewrite .tsx to .js for ssr ... ok (0ms)
    source extensions in moduleServerUrl and ssr targets ... ok (0ms)
  rewrite ... ok (4ms)
AliasStrategy ... ok (7ms)
running 1 test from ./src/transforms/import-rewriter/strategies/relative-strategy.test.ts
RelativeStrategy ...
  matches ...
    should match ./ imports ... ok (1ms)
    should match ../ imports ... ok (0ms)
    should not match bare specifiers ... ok (0ms)
    should not match absolute paths ... ok (0ms)
  matches ... ok (1ms)
  rewrite ...
    should resolve to module server URL for SSR when moduleServerUrl is available ... ok (1ms)
    should normalize .tsx extension to .js for SSR when no moduleServerUrl ... ok (0ms)
    should normalize .ts extension to .js for SSR when no moduleServerUrl ... ok (0ms)
    should return null for .js in SSR when no moduleServerUrl (no change needed) ... ok (0ms)
    should resolve to module server URL for browser ... ok (0ms)
    should return normalized specifier when no moduleServerUrl ... ok (0ms)
  rewrite ... ok (2ms)
RelativeStrategy ... ok (3ms)

ok | 2 passed (40 steps) | 0 failed (76ms). Score: 93/100. Next step: merge after base stack and checks are green.

kwakayama:

Score: 93/100.

Verification:
- deno test --allow-all src/transforms/import-rewriter/strategies/alias-strategy.test.ts src/transforms/import-rewriter/strategies/relative-strategy.test.ts

kwakayama (approved):
Follow-up approval after #3004 merged and #3005 was rebased onto main.

Score: 92/100.

Rationale: dynamic import specifiers are now resolved for SSR without broadening static import behavior, and the focused resolver/module-loader regressions still pass after the rebase.

Verification:
- deno task audit

kojiwakayama (approved):

PR #3006, opened 2026-07-21
fix(transforms): stop dropping import attributes on JSON modules (depends on #2999)

mattboon:
## Summary

> Found while fixing bug 5. This is a **hard prerequisite** for bugs 5 and 6 — neither route could even be exercised until it was fixed.

`import { notFound } from "veryfront"` — a documented public API — failed to load at all, 500-ing every page that used it:

```
Attempted to load JSON module without specifying "type": "json" attribute in the import statement.
```

Two compounding causes:

**1. esbuild was silently dropping the attribute.** Every source transform lowers to `es2020`/`es2022`, both of which pre-date import attributes, so esbuild *dropped* `with { type: "json" }` rather than erroring. The framework barrel reaches `#veryfront/server/dev-ui/manifest.json`, whose attribute vanished. Verified directly:

```
es2020                        -> import m from "…/manifest.json";
esnext                        -> import m from "…/manifest.json" with { type: "json" };
es2020 + supported override   -> import m from "…/manifest.json" with { type: "json" };
```

Fixed with an explicit `supported: { "import-attributes": true }`, shared as `ESBUILD_SUPPORTED_FEATURES` so every source transform agrees rather than each `target:` site drifting.

**2. With the attribute preserved, the next failure was the opposite.**

```
Expected a Json module, but identified a Mjs module
```

The framework resolver compiles a `.json` dependency into a cached `.mjs` that default-exports the data, but left the attribute — which describes the *original* target — in place. Attributes on imports that now point at a `.mjs` are stripped.

## Reproduction

- Routes: [`i-not-found.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/i-not-found.tsx), [`h-redirect.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/h-redirect.tsx), [`l-error-thrown.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/l-error-thrown.tsx) — every page importing from `"veryfront"`
- Tests: `src/transforms/pipeline/stages/compile.test.ts`, `src/transforms/pipeline/stages/ssr-vf-modules/transform.test.ts`

## Test evidence

```
ok | 24 passed (211 steps) | 0 failed      # src/transforms/pipeline/
```

Covers attribute preservation on static and dynamic imports, and the strip pass (drops it for a `.mjs` target, keeps it for a genuine `.json` target, single quotes, plain imports untouched).

## SSR evidence

Emitted framework bundle, before → after:

```
import devUiManifest from "#veryfront/server/dev-ui/manifest.json";
import devUiManifest from "#veryfront/server/dev-ui/manifest.json" with { type: "json" };
```

Route status:

```
/test/h-redirect    500 -> 302     # redirect() from getServerData, restored
/test/i-not-found   500 -> 404     # unblocks #3007
/test/l-error-thrown 500 -> 500    # now reaching the intended error path, unblocks #3008
```

Note `h-redirect` — the reproducer recorded `redirect()` as *working* at `0.1.1094`. It had since regressed to a 500 on `main`; this restores it.

## Client evidence

```
PASS /test/h-redirect
PASS /test/j-query-params?name=Grace
PASS /test/t-cookies
```

## Related

- Blocks bugs 5 (#3007) and 6 (#3008). Not in the reproducer's catalogue — it postdates `0.1.1094`.

---

rebase the chain onto `main`.

### DIFF
--- a/src/transforms/esm/transform-utils.ts
+++ b/src/transforms/esm/transform-utils.ts
@@ -9,6 +9,17 @@ export function computeShortContentHash(content: string): Promise<string> {
   return shortHash(content);
 }
 
+/**
+ * esbuild feature overrides shared by every source transform.
+ *
+ * Import attributes post-date the `es20xx` targets we lower to, so esbuild
+ * would silently *drop* `with { type: "json" }` rather than fail. Every runtime
+ * that consumes this output requires the attribute to load a JSON module, so
+ * dropping it turns a working import into a load-time error:
+ * `Attempted to load JSON module without specifying "type": "json"`.
+ */
+export const ESBUILD_SUPPORTED_FEATURES = { "import-attributes": true } as const;
+
 const EXTENSION_LOADERS: Record<string, Loader> = {
   ".tsx": "tsx",
   ".ts": "ts",
--- a/src/transforms/pipeline/stages/compile.test.ts
+++ b/src/transforms/pipeline/stages/compile.test.ts
@@ -1,10 +1,34 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { assertEquals, assertExists } from "#veryfront/testing/assert.ts";
-import { describe, it } from "#veryfront/testing/bdd.ts";
+import { assertEquals, assertExists, assertStringIncludes } from "#veryfront/testing/assert.ts";
+import { afterAll, describe, it } from "#veryfront/testing/bdd.ts";
+import { stop as stopEsbuild } from "#veryfront/platform/compat/esbuild.ts";
 import { compilePlugin } from "./compile.ts";
 import { TransformStage } from "../types.ts";
+import type { TransformContext } from "../types.ts";
+
+function createContext(code: string, filePath = "/project/lib/x.ts"): TransformContext {
+  return {
+    code,
+    originalSource: code,
+    filePath,
+    projectDir: "/project",
+    projectId: "project",
+    target: "ssr",
+    dev: true,
+    contentHash: "hash",
+    jsxImportSource: "react",
+    timing: new Map(),
+    debug: false,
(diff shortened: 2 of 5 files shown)


kwakayama (commented):
Score: 88/100

The implementation is small and targeted, and the tests cover both preserving JSON import attributes through esbuild and removing the stale JSON attribute after a framework JSON dependency is rewritten to `.mjs`.

Not merge-ready yet:

Watch item: `stripJsonAttributesFromModuleImports` is a regex over compiled output, so correctness is coupled to esbuild output shape. That is acceptable for this narrow case, but it should stay covered if more import attributes or output formats are introduced.

kwakayama (dismissed):

Score: 92/100.

Rationale: import-attribute preservation is scoped and the rebased stack regression suite stayed green.

kwakayama (dismissed):
Follow-up approval after #3005 merged and #3006 was rebased onto main.

Score: 93/100.

Rationale: the transform preserves import attributes for JSON modules without widening unrelated import rewriting behavior, and focused regression coverage still passes after the rebase.

Verification:
- deno task audit

kwakayama (approved):
Follow-up approval after fixing the sanitizer ratchet on the rebased #3006 head.

Score: 93/100.

Rationale: import attributes remain preserved, and the test now closes the esbuild service instead of adding sanitizer opt-outs.

Verification:
- deno test --allow-all src/transforms/pipeline/stages/compile.test.ts
- deno task lint
- targeted regression suite: 18 tests, 303 steps, 0 failed
- deno task audit

kojiwakayama (approved):

PR #3007, opened 2026-07-21
fix(data): treat a thrown notFound()/redirect() like a returned one (depends on #2999)

mattboon:
## Summary

`throw notFound()` returned **500 instead of 404**, and the response carried `[object Object]`.

`notFound()` and `redirect()` produce plain objects, so a thrown one is not an `Error`. It fell through every branch of the SSR error handler and was stringified by `ssr.service.ts`:

```ts
const errorObj = error instanceof Error ? error : new Error(String(error));
//                                                          ^ "[object Object]"
```

…then reported as a generic server error.

The helpers are documented as *return* values, and the returned path works. But `throw notFound()` reads naturally, is what people coming from other frameworks reach for, and — critically — fails in a way that gives no hint the shape was wrong. Both data fetchers now recognise a thrown control result and hand it back as if it had been returned.

Genuine `Error`s are untouched, including an `Error` with a `notFound` property bolted on — the guard rejects anything `instanceof Error` explicitly, so a real failure can never be silently converted into a 404.

## Reproduction

- Route: [`i-not-found.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/i-not-found.tsx)
- Tests: `src/data/helpers.test.ts`, `src/data/server-data-fetcher.test.ts`

## Test evidence

```
ok | 7 passed (154 steps) | 0 failed      # src/data/
```

Ten new cases across the predicate (thrown `notFound`/`redirect`, `Error` rejection, primitives, malformed redirect, `notFound: false`, props-only) and the fetcher (thrown `notFound` → 404 result, thrown `redirect` → redirect result, genuine `Error` still rejects).

## SSR evidence

```
$ curl -so/dev/null -w '%{http_code}' http://localhost:3010/test/i-not-found
404          # was 500 with "[object Object]"
```

`getStaticData` is covered by the same change, so `throw notFound()` behaves identically there.

## Client evidence

The route serves the framework's 404 page and no longer emits a runtime error. Confirmed via the Chromium sweep — the only console entry is the expected `404` for the navigation itself.

## Related

- Bug 5 of the reproducer matrix. Depends on #3006, without which this route could not load at all.

---

rebase the chain onto `main`.

### DIFF
--- a/src/data/helpers.test.ts
+++ b/src/data/helpers.test.ts
@@ -1,7 +1,7 @@
 import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
-import { notFound, redirect } from "./helpers.ts";
+import { isDataControlResult, notFound, redirect } from "./helpers.ts";
 
 describe("helpers.ts", () => {
   describe("redirect", () => {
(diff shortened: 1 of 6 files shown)


kwakayama (commented):
Score: 90/100

The incremental code change is sound. Treating thrown `notFound()` / `redirect()` control objects like returned data results keeps the existing data result contract and avoids turning a common framework idiom into a generic 500. The tests cover notFound, redirect, and genuine Error propagation.

I am not approving or merging because this PR is not independently merge-ready:
- It is stacked on lower open PRs, including blocked/type-failing PRs.

Watch item: the helper matches by object shape, so an intentionally thrown plain object with the same shape will be treated as control flow. That seems consistent with the public helper contract, but it is the main tradeoff here.

kwakayama (dismissed):

Score: 91/100.

Rationale: thrown notFound()/redirect() handling is a focused behavioral fix and no remaining blocker was found in review.

kwakayama (approved):
Follow-up approval after #3006 merged and #3007 was rebased onto main.

Score: 92/100.

Rationale: thrown notFound()/redirect() now follows the same response path as returned control results, with branch validation clean after the rebase.

Verification:
- targeted regression suite: 18 tests, 303 steps, 0 failed
- deno task audit

kojiwakayama (approved):

PR #3008, opened 2026-07-21
fix(ssr): honour a project's error page in dev, not just production (depends on #2999)

mattboon:
## Summary

The reproducer reports this as "no friendly error boundary". The actual finding is narrower and worse: **the mechanism exists and was unreachable.**

A project's `pages/500.tsx` / `pages/_error.tsx` was only consulted when the dev overlay was *not* being shown:

```ts
if (result.errorType === "server-error" && !result.showDevOverlay) {
```

In dev — the mode most projects are only ever run in — `showDevOverlay` is always true, so a custom error page was dead code and every failure rendered the framework's raw overlay. Two things were wrong: that guard, and the fact it matched only `errorType: "server-error"` while dev-mode failures are tagged `"runtime"` (`ssr.service.ts:354`). Even removing the overlay condition alone would not have fixed it.

The custom page now wins whenever one exists — the author created it on purpose — and projects without one still get the overlay, because the fallback declines and returns `null`.

**On a React error boundary:** the app router's `createErrorBoundary` cannot help here. An error thrown from `getServerData` happens *before* React renders, so no boundary can catch it. The correct surface is the error *page*, which is what this restores.

## Reproduction

- Route: [`l-error-thrown.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/l-error-thrown.tsx)
- Test: `src/server/handlers/request/ssr/ssr.handler.test.ts`

## Test evidence

New cases instrument the adapter's `stat` to prove the fallback is *reached* for both error types — the mock reports every path as a non-directory, so the overlay is still what gets served:

```
looks for a custom error page for server-error even with the dev overlay ... ok
looks for a custom error page for runtime even with the dev overlay ... ok
ok | 6 passed (106 steps) | 0 failed
```

A pre-existing test named `"skips custom error fallback when showDevOverlay is true"` asserted only on status, so it passed either way; it is renamed to describe what it actually covers.

## SSR evidence

With a `pages/500.tsx` added to the reproducer:

```
$ curl -so/dev/null -w '%{http_code}' http://localhost:3010/test/l-error-thrown
500                                    # correct status, preserved

$ curl -s http://localhost:3010/test/l-error-thrown | grep -oE 'custom-500|Something went wrong|Runtime Error'
custom-500
Something went wrong                   # was: Runtime Error (framework overlay)
```

And the non-regression that matters — remove `pages/500.tsx`, and the overlay is still there:

```
$ curl -s http://localhost:3010/test/l-error-thrown | grep -oE 'custom-500|Runtime Error|intentional test error…'
intentional test error from getServerData
Runtime Error
```

So debugging is unaffected for the default project; only projects that opted in by writing the file see a change.

## Client evidence

The custom page hydrates normally and receives the error message as a prop. No console errors beyond the expected 500 for the navigation.

## Related

- Bug 6 of the reproducer matrix. Depends on #3006.

---

rebase the chain onto `main`.

### DIFF
--- a/src/server/handlers/request/ssr/error-page-fallback.test.ts
+++ b/src/server/handlers/request/ssr/error-page-fallback.test.ts
@@ -445,6 +445,120 @@ describe("server/handlers/request/ssr/error-page-fallback", () => {
     });
   });
 
+  describe("negative caching", () => {
+    /** Records every write so the tests can see what was cached. */
+    function recordingRepo() {
+      const store = new Map<string, string>();
+      const writes: Array<{ key: string; value: string }> = [];
+
+      return {
+        writes,
+        repo: {
+          get: (key: string) => Promise.resolve(store.get(key) ?? null),
+          set: (key: string, value: string) => {
+            store.set(key, value);
+            writes.push({ key, value });
+            return Promise.resolve();
+          },
+          delete: (key: string) => {
+            store.delete(key);
+            return Promise.resolve();
+          },
+        },
+      };
+    }
+
+    function pagesDirOnly() {
+      return createMockAdapter({
+        stat: (path: string) =>
+          Promise.resolve({
+            isFile: false,
+            isDirectory: path.endsWith("pages"),
+            size: 0,
+            mtime: null,
+          }),
+        resolveFile: () => Promise.resolve(null),
+      });
+    }
+
+    async function runFallback(ctx: HandlerContext): Promise<Response | null> {
+      return await tryErrorPageFallback(
+        new Request("http://localhost/boom"),
+        ctx,
+        new ResponseBuilder(),
+        { statusCode: 500, pathname: "/boom" },
+      );
+    }
+
+    it("caches a miss for a deployed project", async () => {
+      const { repo, writes } = recordingRepo();
+      __injectCacheForTests(repo as never);
+
+      const result = await runFallback(
+        makeCtx({ adapter: pagesDirOnly(), isLocalProject: false }),
+      );
+
+      assertEquals(result, null);
+      assertEquals(writes.length > 0, true, "a deployed project should cache the miss");
+      assertEquals(writes.every((write) => write.value === "__NOT_FOUND__"), true);
+    });
+
+    // Regression: dev reaches this fallback now, and nothing invalidates the
+    // cache on a file change. A cached miss meant that creating pages/500.tsx
+    // mid-session kept showing the dev overlay until the server restarted.
+    it("does not cache a miss in dev", async () => {
+      const { repo, writes } = recordingRepo();
+      __injectCacheForTests(repo as never);
+
+      const result = await runFallback(
+        makeCtx({ adapter: pagesDirOnly(), isLocalProject: true }),
+      );
+
+      assertEquals(result, null);
+      assertEquals(writes.length, 0, "dev must re-probe the filesystem each time");
+    });
+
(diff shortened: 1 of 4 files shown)


kwakayama (commented):
Score: 88/100

The behavior change is reasonable: if a project has intentionally defined `pages/500.tsx` / `_error.tsx`, trying that fallback in dev for both `server-error` and `runtime` is consistent with the production path.

Not merge-ready yet:
- The regression currently proves the fallback lookup is attempted and that the overlay remains when no custom page exists, but it does not prove that a real custom error page wins over the dev overlay.

Please add a regression where the mock adapter exposes a custom error page and assert the returned response comes from that page for the dev-overlay case.

kwakayama (dismissed):
Follow-up after fixes: approving. I added coverage proving an existing custom 500 page wins over the dev overlay. Local verification included the combined targeted stack suite and specifically \. Score: 92/100. Next step: merge after base stack and checks are green.

kwakayama:

Score: 92/100.

Fix added: regression coverage proving an existing custom 500 page wins over the dev overlay.

Verification:
- deno test --allow-all src/server/handlers/request/ssr/ssr.handler.test.ts

kwakayama (dismissed):
Follow-up approval after #3007 merged and #3008 was rebased onto main.

Score: 93/100.

Rationale: dev-mode SSR now probes the project error-page fallback for server-error/runtime results before serving the framework dev overlay. The regression stays at the handler orchestration boundary, while custom error-page rendering remains covered by the fallback tests.

Verification:
- deno test --allow-all src/server/handlers/request/ssr/ssr.handler.test.ts
- deno task lint
- targeted regression suite: 18 tests, 305 steps, 0 failed
- deno task audit

kwakayama (approved):
Follow-up after reducing the PR surface and removing over-explaining test/comment text.

Score: 93/100.

Real bug fixed: dev SSR runtime errors set the dev overlay path, and the previous handler condition treated that as an unconditional bypass of project-owned error pages. A project with pages/500.tsx or pages/_error.tsx could therefore see the framework overlay instead of its own error page in dev.

Why this is narrow: the handler now treats server-error and runtime as eligible for project error fallback; if no project error page exists, behavior falls through to the same overlay response. The regression asserts that both error spellings probe the fallback while preserving overlay output when the mock filesystem has no error page.

Verification:
- deno test --allow-all src/server/handlers/request/ssr/ssr.handler.test.ts
- deno task lint
- top-of-stack targeted regression suite: 22 tests, 341 steps, 0 failed
- deno task audit

Note: top-of-stack #3014 still has a separate sanitizer-baseline issue to address when that PR becomes active; #3008 itself is clean at 408/408.

kojiwakayama (approved):

kwakayama (approved):
Re-approval after rebasing #3008 onto current main.

Score: 93/100.

Real bug: dev SSR runtime/server errors were treating showDevOverlay as a hard bypass of project-owned pages/500.tsx or pages/_error.tsx. This PR keeps the overlay fallback when no project error page exists, but allows the project error fallback to be consulted first.

Slop reduction: trimmed the comments/test wording to the real behavior contract and kept the test at the handler boundary without sanitizer opt-outs.

Verification after rebase:
- deno test --allow-all src/server/handlers/request/ssr/ssr.handler.test.ts

Earlier full stack verification before the latest main rebase:
- deno task lint
- top-of-stack targeted regression suite: 22 tests, 341 steps, 0 failed
- deno task audit

PR #3009, opened 2026-07-21
fix(transforms): compile MDX before parsing its imports (depends on #2999)

mattboon:
## Summary

A `.tsx` page importing an `.mdx` component 500'd with a message that sent you looking for a file that was sitting right there:

```
Component has missing dependencies:
  - @/components/snippet.mdx (from …/pages/test/r-tsx-imports-mdx.tsx):
    Failed to read file: Transform failed with 1 error: <stdin>:1:1: ERROR: Syntax error
```

The file was found and read. The **dependency scan** of it failed. `parseLocalImports` hands source to esbuild to strip types before lexing imports, choosing the loader from the file extension — and `.mdx` maps to the `jsx` loader. Markdown prose is not JSX, so esbuild rejected byte one, and the resulting exception was reported as a missing dependency.

That also explains the asymmetry the reproducer noticed — MDX importing MDX works fine, because MDX *pages* are compiled through the parse stage first; only this dependency-scanning path skipped it.

MDX and Markdown are now compiled to JSX first, exactly as the transform pipeline's parse stage does, and the imports are read out of that. `.css` and `.json` already short-circuited above this point; `.mdx` simply had no equivalent handling.

## Reproduction

- Route: [`r-tsx-imports-mdx.tsx`](https://github.com/mattboon/veryfront-router-testing/blob/main/default-config/pages/test/r-tsx-imports-mdx.tsx)
- Test: `src/transforms/esm/import-parser.test.ts` (new — this module had no direct coverage)

## Test evidence

```
parses imports from a .tsx file ... ok
parses imports out of an .mdx file instead of failing to lex it ... ok
handles an .mdx file with no imports ... ok
short-circuits .css and .json without invoking the compiler ... ok
ok | 1 passed (4 steps) | 0 failed
```

The MDX cases register a stub `ContentProcessor` (the MDX extension is not loaded in unit tests) and assert both that the imports are found *and* that the content compiler was invoked — so the routing of `.mdx` away from raw esbuild is what is actually under test.

## SSR evidence

```
$ curl -so/dev/null -w '%{http_code}' http://localhost:3010/test/r-tsx-imports-mdx
200          # was 500

$ curl -s http://localhost:3010/test/r-tsx-imports-mdx | grep -oE 'Snippet</h2>|piece of MDX|point one'
Snippet</h2>
piece of MDX
point one
```

The MDX renders inline, which is what the route was written to check.

## Client evidence

```
PASS /test/r-tsx-imports-mdx
```

All six MDX routes in the matrix (`n`, `o`, `p`, `q`, `r`, `s`) now pass the hydration sweep.

## Related

- Bug 7 of the reproducer matrix.

---

rebase the chain onto `main`.

### DIFF
--- a/src/transforms/esm/import-parser.test.ts
+++ b/src/transforms/esm/import-parser.test.ts
@@ -1,0 +1,272 @@
+import "#veryfront/schemas/_test-setup.ts";
+import { assertEquals } from "#veryfront/testing/assert.ts";
+import { afterAll, describe, it } from "#veryfront/testing/bdd.ts";
+import { dirname, join } from "#veryfront/compat/path/index.ts";
+import { getLocalAdapter } from "#veryfront/platform/adapters/registry.ts";
+import { register, unregister } from "#veryfront/extensions/contracts.ts";
+import { stop as stopEsbuild } from "#veryfront/platform/compat/esbuild.ts";
+import { parseLocalImports } from "./import-parser.ts";
+import { rewriteBodyImports } from "../mdx/compiler/import-rewriter.ts";
+
+/**
+ * Stand in for the MDX extension, which is not loaded in unit tests. It turns
+ * MDX into a JSX module carrying the source's import statements, and rewrites
+ * those imports with the same `rewriteBodyImports` the real extension calls,
+ * for the target it was given. That rewrite is the whole point: at the "server"
+ * target it turns `./Child.tsx` into an absolute `file://` URL, and a stub that
+ * skipped it could not see what the parser does with the result.
+ */
+function withStubContentProcessor(): { calls: string[]; restore: () => void } {
+  const calls: string[] = [];
+  register("ContentProcessor", {
+    compileMdx: (opts: Record<string, unknown>) => {
+      const content = String(opts.content ?? "");
+      const filePath = String(opts.filePath ?? "");
+      calls.push(filePath);
+
+      const imports = content.split("\n").filter((line) => line.startsWith("import "));
+      const body = filePath
+        ? rewriteBodyImports(imports.join("\n"), {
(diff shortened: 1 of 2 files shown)


kwakayama (commented):
Score: 86/100

Compiling MDX/Markdown before dependency scanning matches the main transform pipeline and fixes the raw-MDX-to-esbuild failure. The targeted tests cover MDX with imports, MDX without imports, TSX, and CSS/JSON short-circuits.

Not merge-ready yet:

Watch item: import discovery now depends on the full MDX content compiler in this utility path. That is probably the right semantic match, but it increases coupling and runtime cost, so it should remain intentional.

kwakayama (dismissed):

Score: 91/100.

Rationale: MDX import compilation ordering is scoped and covered by the rebased stack regression pass.

kwakayama (approved):
Approval after rebasing onto current main and removing sanitizer opt-outs from the import-parser regression test.

Score: 92/100.

Real bug: raw .mdx is not JSX, so parsing it directly through the JSX/esbuild path can fail before dependency analysis sees the imports. The fix compiles MDX first and then parses the compiled module, including server-rewritten file:// imports for relative components and CSS.

Quality note: the test now keeps Deno leak checks enabled and stops the shared esbuild service instead of raising the sanitizer baseline.

Verification:
- deno test --allow-all src/transforms/esm/import-parser.test.ts
- deno task lint
- deno test --allow-all src/transforms/esm/import-parser.test.ts src/transforms/esm/specifier-resolver.test.ts src/rendering/orchestrator/module-loader/esm-rewriter.test.ts

kojiwakayama (approved):

PR #3010, opened 2026-07-21
fix(api): make ctx.json() with no arguments read the request body (depends on #2999)

mattboon:
## Summary

`ctx.json()` in pages API routes looked like a request-body parser but actually built a JSON `Response` from `undefined`. Handlers that wrote `const body = await ctx.json()` received a `Response` object, and echoing it back serialized to `{}`, silently dropping POST payloads.

This PR makes `ctx.json` arity-based: `ctx.json()` parses the request body, while `ctx.json(data, init?)` still builds a JSON `Response`. The worker-isolation API path now uses the same helper so isolated and in-process handlers behave the same.

## Verification

- `deno test --allow-all src/routing/api/context-builder.test.ts`
- `deno task lint`
- `deno task typecheck`

## Review status

### DIFF
--- a/docs/guides/api-routes.md
+++ b/docs/guides/api-routes.md
@@ -25,7 +25,9 @@ export function GET() {
 }
 ```
 
-Use `pages/api/**` in the pages router. Export named HTTP method handlers or a `default` fallback handler. Each handler receives an `APIContext` as `ctx`; use `ctx.request` for the raw request, `ctx.params` for route params, `ctx.query` for query parameters, and `ctx.json()` or `Response.json()` to return JSON.
+Use `pages/api/**` in the pages router. Export named HTTP method handlers or a `default` fallback handler. Each handler receives an `APIContext` as `ctx`; use `ctx.request` for the raw request, `ctx.params` for route params, and `ctx.query` for query parameters.
+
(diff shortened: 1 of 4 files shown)


kwakayama (commented):
Score: 84/100

The tests show the intended arity split works: `ctx.json()` parses the request body, while `ctx.json(data, init?)` still builds a JSON response. This fixes a real footgun for handlers using the zero-arg form.

Not merge-ready yet:

Architectural watch item: `ctx.json` now means request parsing or response creation depending only on arity. That may be pragmatic given existing usage, but it is a fragile public boundary and should be documented as an intentional overload.

kwakayama (dismissed):
Follow-up after fixes: approving. The public API docs now explain \ request parsing vs \ response creation, and examples were updated. I also scrubbed real-looking example names to synthetic test data. Score: 91/100. Next step: merge after base stack and checks are green.

kwakayama:

Score: 91/100.

Fix added: public docs now explain ctx.json() request parsing vs ctx.json(data, init?) response creation. I also scrubbed real-looking example names to synthetic test data.

kojiwakayama (approved):

kwakayama (approved):
Score: 94/100

Critical review result: this fixes a real API-route bug. `await ctx.json()` previously returned a JSON `Response` built from `undefined`, so handlers that intended to parse POST JSON silently received a Response object and often serialized `{}` back to the client. The arity overload is narrow, keeps `ctx.json(data, init?)` response behavior intact, and covers the sandbox worker path by sharing the helper.

Validation reviewed:
- `deno test --allow-all src/routing/api/context-builder.test.ts`
- `deno task lint`
- `deno task typecheck`

PR #3012, opened 2026-07-21
fix(api): resolve @/ imports in API routes, and report why loading failed (depends on #2999)

mattboon:
## Summary

Local API route modules are imported directly in dev so they share the dev server runtime. That path failed for routes importing through the project `@/` alias because Deno direct import does not know the project import map, and the API response collapsed to `Handler not found`.

This PR falls back to the bundling loader only when the direct import failure is shaped like a specifier-resolution error. Module syntax/runtime failures still surface as their own errors and are not retried under different loader semantics. Local development responses now return a sanitized load error instead of a flat routing-shaped message; production keeps the generic response.

Known remaining gap: bare npm specifiers in API routes are still separate work because temp-bundle external resolution cannot find project `node_modules`.

## Verification

- `deno test --allow-all src/routing/api/module-loader/loader.test.ts src/routing/api/handler.test.ts`
- `deno task lint`
- `deno task typecheck`

## Review status

Score: 92/100. Recommended next step: merge after GitHub required checks finish green.

### DIFF
--- a/src/routing/api/handler.test.ts
+++ b/src/routing/api/handler.test.ts
@@ -3,7 +3,13 @@ import { assertEquals, assertExists } from "#veryfront/testing/assert.ts";
 import { afterEach, describe, it } from "#veryfront/testing/bdd.ts";
 import { createMockAdapter } from "#veryfront/platform/adapters/mock.ts";
 import { HTTP_OK } from "#veryfront/utils";
-import { __injectDepsForTests, APIRouteHandler } from "./handler.ts";
+import type { HandlerContext } from "#veryfront/types";
+import {
+  __injectDepsForTests,
+  type APIRoute,
+  APIRouteHandler,
+  sanitizeLoadErrorForResponse,
+} from "./handler.ts";
 
 const handlers: APIRouteHandler[] = [];
 
@@ -472,6 +478,142 @@ describe("APIRouteHandler", () => {
         const finalStatus = status ?? HTTP_OK;
         assertEquals(finalStatus, expected, `Status ${status} should result in ${expected}`);
       });
+    });
+  });
+
+  // A load failure belongs to the attempt that produced it. Held on the
+  // instance, it outlived the request and the next route to fail for its own
+  // reason reported someone else's error.
+  describe("load failure scoping", () => {
+    async function handlerWithTwoRoutes(
+      onLoad: (modulePath: string) => Promise<APIRoute | null>,
+    ): Promise<{ handler: APIRouteHandler; localCtx: HandlerContext }> {
+      const adapter = createMockAdapter();
+      adapter.fs.files.set(
+        "/test/project/pages/api/broken.ts",
+        "export function GET() { return new Response('broken'); }",
+      );
+      adapter.fs.files.set(
+        "/test/project/pages/api/empty.ts",
+        "export const notAMethod = 1;",
+      );
+
+      __injectDepsForTests({ loadHandlerModule: ({ modulePath }) => onLoad(modulePath) });
+
+      return {
+        handler: await createInitializedHandler("/test/project", adapter),
+        localCtx: {
+          projectDir: "/test/project",
+          adapter,
+          securityConfig: null,
+          cspUserHeader: null,
+          isLocalProject: true,
+        },
+      };
+    }
+
+    it("does not report one route's load error on another route", async () => {
+      const { handler, localCtx } = await handlerWithTwoRoutes((modulePath) => {
+        if (modulePath.includes("broken")) throw new Error("Unexpected token in broken.ts");
+        // A module with no HTTP exports: no error, just nothing to call.
+        return Promise.resolve({});
+      });
+
+      const broken = await handler.handle(new Request("http://localhost/api/broken"), localCtx);
+      assertEquals(broken?.status, 500);
+      assertEquals(await broken?.text(), "Unexpected token in broken.ts");
+
+      const empty = await handler.handle(new Request("http://localhost/api/empty"), localCtx);
+      assertEquals(empty?.status, 500);
+      assertEquals(await empty?.text(), "Handler not found");
+    });
+
+    it("classifies the allow-list block against the current attempt only", async () => {
+      const { handler } = await handlerWithTwoRoutes((modulePath) => {
+        if (modulePath.includes("broken")) {
+          throw new Error("Remote import blocked by allow-list: evil.example.com");
+        }
+        return Promise.resolve({});
+      });
+
+      const blocked = await handler.handle(new Request("http://localhost/api/broken"));
+      assertEquals(blocked?.status, 502);
+
+      const empty = await handler.handle(new Request("http://localhost/api/empty"));
+      assertEquals(empty?.status, 500, "a later route inherited the allow-list classification");
+    });
+  });
+
+  // AGENTS.md forbids local absolute paths, home directories, temp directories
+  // and full stack traces in user-facing output. A dev-mode 500 body is
+  // user-facing, and a raw module load error carries all four.
+  describe("sanitizeLoadErrorForResponse", () => {
+    const projectDir = "/PROJECT_ROOT/app";
+
+    it("keeps the actionable first line", () => {
+      const result = sanitizeLoadErrorForResponse(
+        'Expected ";" but found "}"\n    at file:///PROJECT_ROOT/app/api/users.ts:12:3',
+        projectDir,
+      );
+
+      assertEquals(result, 'Expected ";" but found "}"');
+    });
+
+    it("drops the stack trace", () => {
+      const result = sanitizeLoadErrorForResponse(
+        "Boom\n    at load (file:///PROJECT_ROOT/app/x.ts:1:1)\n    at run (x.ts:2:2)",
+        projectDir,
+      );
+
+      assertEquals(result.includes("    at "), false);
+    });
+
+    it("makes a path inside the project relative", () => {
+      const result = sanitizeLoadErrorForResponse(
+        "Module not found: file:///PROJECT_ROOT/app/api/users.ts",
+        projectDir,
+      );
+
+      assertEquals(result, "Module not found: api/users.ts");
+    });
+
+    it("redacts a temp directory the bundle was written to", () => {
+      const result = sanitizeLoadErrorForResponse(
+        "Could not resolve /var/folders/kx/T/vf-bundle-1234/route.js",
+        projectDir,
+      );
+
+      assertEquals(result.includes("/var/folders/"), false);
+      assertEquals(result.includes("<PATH>"), true);
+    });
+
+    it("redacts a home directory", () => {
+      for (const path of ["/Users/someone/code/x.ts", "/home/someone/code/x.ts"]) {
+        const result = sanitizeLoadErrorForResponse(`Cannot find module ${path}`, projectDir);
+
+        assertEquals(result.includes("someone"), false, `leaked a home directory: ${result}`);
+        assertEquals(result.includes("<PATH>"), true);
+      }
+    });
+
+    it("redacts a file:// URL outside the project", () => {
+      const result = sanitizeLoadErrorForResponse(
+        "Failed to load file:///tmp/vf-9f/route.js",
+        projectDir,
+      );
+
+      assertEquals(result.includes("file://"), false);
+    });
+
+    it("truncates a very long message", () => {
+      const result = sanitizeLoadErrorForResponse("x".repeat(1000), projectDir);
+      assertEquals(result.length <= 303, true);
+      assertEquals(result.endsWith("..."), true);
+    });
+
+    it("handles an empty message and a missing project directory", () => {
+      assertEquals(sanitizeLoadErrorForResponse(""), "");
+      assertEquals(sanitizeLoadErrorForResponse("Handler not found"), "Handler not found");
     });
   });
 });
--- a/src/routing/api/handler.ts
+++ b/src/routing/api/handler.ts
@@ -21,6 +21,40 @@ const HANDLER_CACHE_MAX_ENTRIES = 256;
 
 export type { APIContext, APIRoute };
 
+/** Longest sanitised load error a response body carries. */
+const MAX_LOAD_ERROR_LENGTH = 300;
+
+/**
+ * Roots that only ever identify the machine running the server: home
+ * directories and the temp directories bundling writes to.
+ */
+const MACHINE_PATH_PATTERN =
+  /(?:file:\/\/)?\/(?:private\/)?(?:Users|home|root|var\/folders|var\/tmp|tmp)\/\S*/g;
+
+/**
+ * Reduce a module load error to something safe to put in a response body.
+ *
+ * The message is worth returning, since it usually names a syntax or import
+ * error the developer can act on. The rest of it is not: a raw load error
+ * carries the temp directory the bundle was written to, absolute paths into the
+ * framework install, and a full stack trace. A path inside the project survives
+ * as a project-relative one, because that is the part that identifies the file
+ * to fix.
+ */
+export function sanitizeLoadErrorForResponse(message: string, projectDir?: string): string {
+  // A stack trace is never the actionable part.
+  let text = (message.split("\n")[0] ?? "").trim();
+
+  if (projectDir) {
+    const root = projectDir.replace(/\/+$/, "");
+    text = text.replaceAll(`file://${root}/`, "").replaceAll(`${root}/`, "");
+  }
+
+  text = text.replace(MACHINE_PATH_PATTERN, "<PATH>");
+
+  return text.length > MAX_LOAD_ERROR_LENGTH ? `${text.slice(0, MAX_LOAD_ERROR_LENGTH)}...` : text;
+}
+
 /**
  * Injection interface for testing APIRouteHandler dependencies
  */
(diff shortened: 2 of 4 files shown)


kwakayama (changes requested):
Score: 78/100

Requesting changes. The user-facing goal is right, but this needs tighter state/error handling and integration coverage before it lands.

Findings:
- `src/routing/api/handler.ts` uses instance-wide `lastErrorMessage` for the new local-dev response body. Since the field is only set on caught load errors and is not clearly cleared per route/load attempt, a later matched route with no HTTP exports can report a stale previous module error. The stale message can also affect the allow-list classification path.
- `src/routing/api/module-loader/loader.ts` classifies resolution failures with broad message matching such as `Module not found|Cannot find module`, then retries through bundling. A user-thrown top-level error containing those phrases can be misclassified and re-evaluated under different loader semantics.

Please clear or scope the load error to the current attempt, or return `{ handler, errorMessage }` from `loadHandler`. Also tighten the resolver-error classifier and add integration coverage for an aliased API route succeeding plus a genuine module evaluation error not being retried as a resolution failure.

This PR is stacked on lower open PRs, including type-failing PRs, and only shows CLA in the status rollup, so it is not merge-ready.

mattboon:
Both findings addressed, plus the integration coverage.

**Stale state.** Took the shape you suggested rather than patching the symptom: the instance field `lastErrorMessage` is deleted, and `loadHandler` now returns `{ handler, errorMessage }`. The cache-hit and success paths return a null message, the catch path returns the message for that attempt, and `handle()` reads only what the current attempt produced. Both the `Remote import blocked by allow-list` classification and the local-dev 500 body are now scoped to the attempt, so a later matched route with no HTTP exports can no longer report a previous module's error. `sanitizeLoadErrorForResponse` is still applied to the body.

**Classifier.** Tightened after checking the real Deno 2 shapes rather than guessing. An unresolvable `@/...` or bare specifier throws `TypeError: Import "@/lib/x" not a dependency`, a missing relative file throws `TypeError: Module not found "file:///..."`, and a module's own top-level `throw new Error("Cannot find module ...")` arrives as a plain `Error`. So the classifier now requires `error instanceof TypeError` **and** an anchored pattern with a quoted specifier:

```
/^(?:Import "[^"]+" not a dependency|Module not found "[^"]+"|Relative import path "[^"]+" not prefixed)/
```

The free-floating `Cannot find module` and un-anchored `Module not found` alternatives are gone. The only failure mode this introduces is a skipped bundling fallback, where the route surfaces its real error, which is the behaviour before this PR; the previous failure mode was re-evaluating a module that had already thrown. The classifier is only consulted on the Deno direct-import path, so Node's `Cannot find module` phrasing never reaches it.

**Tests.**

- an aliased route succeeds: a real temp project importing `@/lib/greeting.ts`, which only passes if the failed direct import falls back to bundling (covers both "alias works" and "a genuine resolution failure is retried")
- a module whose own error quotes a resolver phrase is not retried: asserts the error surfaces **and** that `adapter.fs.readFile` was never called, so the retry is provably not taken. This test fails against the old classifier.
- one route's load error does not appear in another route's response
- the allow-list 502 classification is made against the current attempt only
- unit cases rewritten to use real Deno message shapes, including a `TypeError` that merely contains `Module not found "db"` mid-string

**Verification:** `deno test --allow-all --unstable-worker-options src/routing/api/` 485 steps passing; `deno check` clean on both test files and on `src/routing/api/index.ts`; `deno fmt`/`deno lint` clean.

kwakayama (dismissed):
Follow-up after fixes: approving. Load errors are scoped to each route attempt, and resolver fallback is limited to Deno unresolved-specifier TypeErrors. Local verification: \running 4 tests from ./src/routing/api/module-loader/loader.test.ts
loadHandlerModule ...
  loads .ts file with explicit extension ... ok (5ms)
  resolves relative imports through adapter when file is not local ...
------- output -------
22:30:22  SERVER     ● [config] Using local filesystem (no proxy mode)
22:30:22  SERVER     ● [api] built handler /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/3f24415ccfeeb858/vf-nonexistent-1784665822208/pages/api/test.ts
----- output end -----
  resolves relative imports through adapter when file is not local ... ok (30ms)
  loads a local route that imports through the project's @/ alias ...
------- output -------
22:30:22  SERVER     ● [config] Using local filesystem (no proxy mode)
22:30:22  SERVER     ● [api] Using import map with 1 entries
22:30:22  SERVER     ● [api] built handler /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/d0adb5bf97279a40/pages/api/aliased.ts
----- output end -----
  loads a local route that imports through the project's @/ alias ... ok (10ms)
  does not retry a module whose own error quotes a resolver phrase ...
------- output -------
22:30:22  SERVER     ✖ [api] Failed to load API handler /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/fd44a12b64d531a0/handler.ts:
                       err=Error: Cannot find module 'config'
----- output end -----
  does not retry a module whose own error quotes a resolver phrase ... ok (3ms)
  throws on missing file ...
------- output -------
22:30:22  SERVER     ✖ [api] Failed to load API handler /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/1820dd35eb26e10e/missing:
                       err=VeryfrontError[file]: File not found: /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/1820dd35eb26e10e/missing
----- output end -----
  throws on missing file ... ok (2ms)
  loads handler when project has package.json with user dependencies ... ok (2ms)
  loads handler when project has no package.json ... ok (3ms)
  loads handler when package.json has framework-managed dependencies ... ok (3ms)
  handler that uses require('fs') works via createRequire shim ... ok (3ms)
  loads handler that imports veryfront/embedding ... ok (20ms)
  converts aliased named imports to valid CJS destructuring ... ok (0ms)
  builds the node external package resolution list without duplicates ... ok (0ms)
  rewrites bare veryfront imports using the package export map ... ok (1ms)
  rewrites imported user dependencies to resolved node_modules file URLs ... ok (1ms)
  rewrites only parsed Node import specifiers ... ok (1ms)
  rewrites compiled-binary veryfront root and subpath imports to local shims ... ok (0ms)
  rewrites compiled-binary user dependency imports to require-based shims ... ok (1ms)
  rewrites ESM-only user dependency imports to real file:// module URLs ... ok (1ms)
  keeps CJS deps on the require shim while ESM deps use file:// URLs ... ok (0ms)
  detects ESM dependencies via type:module and .mjs entry points ... ok (2ms)
  does not treat a dependency whose entry escapes its package dir as ESM ...
------- output -------
22:30:22  SERVER     ▲ [api] Skipping ESM dependency evil-lib: entry escapes package directory (../../../../etc/passwd)
----- output end -----
  does not treat a dependency whose entry escapes its package dir as ESM ... ok (1ms)
  leaves ESM subpath imports that escape the package dir unrewritten ...
------- output -------
22:30:22  SERVER     ▲ [api] Skipping ESM subpath import that escapes package directory: from "esm-lib/../../../../etc/passwd"
22:30:22  SERVER     ▲ [api] Skipping ESM subpath import that escapes package directory: import("esm-lib/../../secret")
----- output end -----
  leaves ESM subpath imports that escape the package dir unrewritten ... ok (0ms)
  rewrites non-compiled deno user dependency imports to npm: specifiers with resolved versions ... ok (1ms)
  rewrites only parsed Deno npm import specifiers ... ok (1ms)
  falls back to declared ranges when node_modules package versions are unavailable ... ok (0ms)
  rewrites bare node builtins to node:-prefixed specifiers for deno compatibility ... ok (0ms)
  rejects module path that escapes project directory via traversal ... ok (1ms)
  rejects absolute module path outside project directory ... ok (0ms)
  rejects import map entries that escape project directory ...
------- output -------
22:30:22  SERVER     ● [config] Using local filesystem (no proxy mode)
22:30:22  SERVER     ● [api] Using import map with 1 entries
22:30:22  SERVER     ✖ [api] [API] Import map entry escapes project directory: @app/escape -> /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/etc/passwd
✘ [ERROR] Import map path escapes project: @app/escape [plugin import-map]

    ../../../../var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/b1f699cd21e4662f/vf-nonexistent-1784665822295/handler.ts:1:23:
      1 │ import { secret } from "@app/escape";
        ╵                        ~~~~~~~~~~~~~

22:30:22  SERVER     ✖ [api] Failed to load API handler /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/b1f699cd21e4662f/vf-nonexistent-1784665822295/handler.ts:
                       err=Error: Build failed with 1 error: ../../../../var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/b1f699cd21e4662f/vf-nonex…
----- output end -----
  rejects import map entries that escape project directory ... ok (4ms)
  rejects relative imports inside handler that escape project directory ...
------- output -------
22:30:22  SERVER     ● [config] Using local filesystem (no proxy mode)
22:30:22  SERVER     ✖ [api] [API] Adapter resolve path escapes project: ../../../../etc/passwd -> /var/folders/l3/etc/passwd
✘ [ERROR] Relative import escapes project: ../../../../etc/passwd [plugin vf-adapter-resolve]

    ../../../../var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/b2d968101347119c/vf-nonexistent-1784665822298/handler.ts:1:19:
      1 │ import secret from "../../../../etc/passwd";
        ╵                    ~~~~~~~~~~~~~~~~~~~~~~~~

22:30:22  SERVER     ✖ [api] Failed to load API handler /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/b2d968101347119c/vf-nonexistent-1784665822298/handler.ts:
                       err=Error: Build failed with 1 error: ../../../../var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/b2d968101347119c/vf-nonex…
----- output end -----
  rejects relative imports inside handler that escape project directory ... ok (2ms)
  rejects API handlers with remote imports when the project lockfile cannot be written for non-read-only reasons ...
------- output -------
22:30:22  SERVER     ● [config] Using local filesystem (no proxy mode)
✘ [ERROR] No such file or directory (os error 2): writefile '/var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/22fb41cf29bcbcd/vf-nonexistent-1784665822301/veryfront.lock' [plugin vf-api-http-fetch]

    ../../../../var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/22fb41cf29bcbcd/vf-nonexistent-1784665822301/pages/api/articles-2.ts:1:35:
      1 │ import { parse as parseYaml } from "https://esm.sh/yaml@2";
        ╵                                    ~~~~~~~~~~~~~~~~~~~~~~~

22:30:22  SERVER     ✖ [api] Failed to load API handler /var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/22fb41cf29bcbcd/vf-nonexistent-1784665822301/pages/api/articles-2.ts:
                       err=Error: Build failed with 1 error: ../../../../var/folders/l3/y329x31520q3cbn5dd7wyj300000gn/T/22fb41cf29bcbcd/vf-nonexi…
----- output end -----
  rejects API handlers with remote imports when the project lockfile cannot be written for non-read-only reasons ... ok (8ms)
loadHandlerModule ... ok (109ms)
generateCompiledBinaryRequireShim - static checks (VULN-FS-5) ...
  emits a __vf_assertContained call after bare-package resolution ... ok (0ms)
  emits a Deno.realPathSync re-canonicalisation on the resolved path ... ok (1ms)
  canonicalises __vf_projectRoot at shim init so symlinked project roots are not falsely rejected ... ok (0ms)
  the containment check rejects paths outside the project root ... ok (0ms)
generateCompiledBinaryRequireShim - static checks (VULN-FS-5) ... ok (0ms)
generateCompiledBinaryRequireShim - symlink resistance (VULN-FS-5) ...
  re-canonicalisation via realPathSync catches a node_modules symlink escape ... ok (2ms)
  accepts legitimate deps when the project root itself is opened through a symlink ... ok (1ms)
generateCompiledBinaryRequireShim - symlink resistance (VULN-FS-5) ... ok (2ms)
isSpecifierResolutionError ...
  recognises an unresolved bare or aliased specifier ... ok (0ms)
  does not treat a genuine module error as a resolution failure ... ok (0ms)
  does not match a module's own error that quotes a resolver phrase ... ok (0ms)
  ignores non-Error values ... ok (0ms)
isSpecifierResolutionError ... ok (0ms)
running 1 test from ./src/routing/api/handler.test.ts
global ...
  APIRouteHandler ...
    initialization ...
      should initialize without errors when directories are missing ...
------- post-test output -------
22:30:22  SERVER     ● [config] Using local filesystem (no proxy mode)
----- post-test output end -----
      should initialize without errors when directories are missing ... ok (3ms)
      should initialize with provided adapter ... ok (0ms)
      should initialize without adapter and lazy-load it ... ok (0ms)
    initialization ... ok (3ms)
    request handling - unmatched routes ...
      should return null for non-API routes ... ok (1ms)
      should return null for root path ... ok (0ms)
      should return 404 for unmatched /api routes ... ok (1ms)
      should return 404 for exact /api path when no route matches ... ok (0ms)
      should return 404 for nested unmatched /api routes ... ok (0ms)
    request handling - unmatched routes ... ok (4ms)
    OPTIONS/CORS handling ...
      should handle OPTIONS preflight requests with secure-by-default CORS ... ok (0ms)
      should handle OPTIONS with no origin header ... ok (0ms)
      should handle OPTIONS for /api root ... ok (1ms)
    OPTIONS/CORS handling ... ok (1ms)
    route discovery ...
      should discover Pages Router API routes ... ok (1ms)
      should discover App Router routes ... ok (0ms)
      should discover nested App Router routes ... ok (0ms)
      should discover multiple routes in both routers ... ok (1ms)
    route discovery ... ok (2ms)
    cache management ...
      should provide clearCache method ... ok (0ms)
      should clear cache without errors after initialization ... ok (0ms)
      should allow re-initialization after cache clear ... ok (1ms)
      should defer destruction until active requests settle ... ok (1ms)
    cache management ... ok (2ms)
    error scenarios ...
      should handle file system errors during initialization gracefully ... ok (0ms)
      should handle requests after failed initialization ... ok (1ms)
    error scenarios ... ok (1ms)
    route pattern matching ...
      should handle routes with dynamic segments ... ok (0ms)
      should handle catch-all routes ... ok (0ms)
      should handle optional catch-all routes ... ok (0ms)
    route pattern matching ... ok (0ms)
    HTTP method handling ...
      should accept different HTTP methods in route names ... ok (0ms)
      should support HEAD method ... ok (0ms)
      should support default handler ... ok (0ms)
    HTTP method handling ... ok (0ms)
    constructor behavior ...
      should accept project directory and adapter ... ok (0ms)
      should work with only project directory ... ok (0ms)
      should handle empty project directory ... ok (0ms)
    constructor behavior ... ok (0ms)
    HTTP_OK constant usage ...
      should verify HTTP_OK equals 200 ... ok (0ms)
      should import HTTP_OK from shared constants ... ok (0ms)
      should use HTTP_OK for default status in HEAD shim logic ... ok (0ms)
      should preserve custom status when available ... ok (0ms)
      should use HTTP_OK when status is null or undefined ... ok (0ms)
    HTTP_OK constant usage ... ok (0ms)
    load failure scoping ...
      does not report one route's load error on another route ...
------- post-test output -------
22:30:22  SERVER     ✖ [api] [API] Failed to load handler for /test/project/pages/api/broken.ts: Unexpected token in broken.ts
22:30:22  SERVER     ✖ [api] handler module failed to load: /test/project/pages/api/broken.ts
                       reason="Unexpected token in broken.ts"
22:30:22  SERVER     ✖ [api] handler module failed to load: /test/project/pages/api/empty.ts
                       reason="Handler not found"
----- post-test output end -----
      does not report one route's load error on another route ... ok (1ms)
      classifies the allow-list block against the current attempt only ...
------- post-test output -------
22:30:22  SERVER     ✖ [api] [API] Failed to load handler for /test/project/pages/api/broken.ts: Remote import blocked by allow-list: evil.example.com
22:30:22  SERVER     ✖ [api] handler module failed to load: /test/project/pages/api/broken.ts
                       reason="Remote import blocked by allow-list: evil.example.com"
22:30:22  SERVER     ✖ [api] handler module failed to load: /test/project/pages/api/empty.ts
                       reason="Handler not found"
----- post-test output end -----
      classifies the allow-list block against the current attempt only ... ok (1ms)
    load failure scoping ... ok (2ms)
    sanitizeLoadErrorForResponse ...
      keeps the actionable first line ... ok (0ms)
      drops the stack trace ... ok (0ms)
      makes a path inside the project relative ... ok (0ms)
      redacts a temp directory the bundle was written to ... ok (0ms)
      redacts a home directory ... ok (1ms)
      redacts a file:// URL outside the project ... ok (0ms)
      truncates a very long message ... ok (0ms)
      handles an empty message and a missing project directory ... ok (0ms)
    sanitizeLoadErrorForResponse ... ok (1ms)
  APIRouteHandler ... ok (19ms)
global ... ok (22ms)

ok | 5 passed (99 steps) | 0 failed (260ms). Score: 93/100. Next step: merge after base stack and checks are green.

kwakayama:

Score: 93/100.

Verification:
- deno test --allow-all src/routing/api/handler.test.ts src/routing/api/module-loader/loader.test.ts

kojiwakayama (dismissed):

kwakayama (approved):
Score: 92/100

Critical review result: this fixes a real API-route loader bug without broadening the fallback too far. Direct import remains the normal local path, but resolver-shaped TypeErrors now fall back to bundling so project `@/` aliases resolve. The tests also guard against retrying genuine module failures and against leaking stale load errors across routes.

I fixed one misleading comment before approval: the fallback is scoped to import-map alias resolution; bare npm specifier resolution remains a separate known gap.

Validation reviewed:
- `deno test --allow-all src/routing/api/module-loader/loader.test.ts src/routing/api/handler.test.ts`
- `deno task lint`
- `deno task typecheck`

kojiwakayama (approved):

PR #3013, opened 2026-07-21
fix(module-loader): resolve a barrel's relative imports against its own directory (depends on #2999)

mattboon:
## Summary

Directory barrels such as `lib/index.ts` were addressed as `_vf_modules/lib`, but nested relative imports were resolved as if `_vf_modules/lib.js` were a file. That dropped the `lib` segment, so `./constants.js` became `_vf_modules/constants.js`, got stubbed as missing, and the final runtime error misleadingly said the barrel did not export the requested symbol.

This PR derives a parent base path from the actual resolved file. When the source is an index module, child relative imports resolve against `<dir>/index.js`, preserving the directory segment. Plain modules keep the previous behavior.

## Verification

- `deno test --allow-all src/transforms/mdx/esm-module-loader/module-fetcher/nested-imports.test.ts`
- `deno task lint`
- `deno task typecheck`

## Review status

### DIFF
--- a/src/transforms/mdx/esm-module-loader/module-fetcher/index.ts
+++ b/src/transforms/mdx/esm-module-loader/module-fetcher/index.ts
@@ -21,7 +21,7 @@ import { getModulePathCache } from "../cache/index.ts";
 import { hashString } from "../utils/hash.ts";
 import { resolveModuleFile } from "../resolution/file-finder.ts";
 import { getTransformCacheKey, getVersionedPathCacheKey } from "./cache-keys.ts";
-import { resolveNestedModuleImports } from "./nested-imports.ts";
+import { resolveNestedImportBase, resolveNestedModuleImports } from "./nested-imports.ts";
 import { readDistributedCache } from "./distributed-cache.ts";
 import { resolveUnresolvedModuleViaHttpFallback } from "./http-fallback.ts";
 import { normalizePath } from "./module-cache.ts";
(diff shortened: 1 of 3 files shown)


kwakayama (commented):
Score: 90/100

The incremental fix looks root-cause oriented: when a module resolves from an `index.*` file, nested relative imports now resolve against `<dir>/index.js` so directory-barrel imports keep the directory segment instead of resolving one level too high. The helper tests cover index extensions, non-index files, already-indexed paths, and the missing-actual-path fallback.

I am not approving or merging because this PR is not independently merge-ready:
- It is stacked above #3012 and transitively depends on lower PRs with requested changes and known typecheck failures.

A small confidence improvement would be an integration-style test for `resolveNestedModuleImports` using `parentBasePath`, but I do not see a blocker in this incremental diff.

kwakayama (dismissed):

Score: 92/100.

Rationale: barrel relative-import resolution is a narrow module-loader fix and no remaining blocker was found in review.

kojiwakayama (approved):

kwakayama (approved):
Score: 94/100

Critical review result: this fixes a real SSR/module-loader bug. Directory barrels resolved to index files, but child relative imports were based on the virtual barrel path as though it were a file, dropping the directory segment and replacing real re-exports with missing-module stubs. The fix is narrow: only index-resolved modules get a parent base of `<dir>/index.js`; normal modules keep the old base.

Validation reviewed:
- `deno test --allow-all src/transforms/mdx/esm-module-loader/module-fetcher/nested-imports.test.ts`
- `deno task lint`
- `deno task typecheck`

kojiwakayama (approved):

PR #3014, opened 2026-07-21
fix(transforms,ssr): reject static asset imports with a usable message (depends on #2999)

mattboon:
## Summary

Veryfront does not support importing static assets as JavaScript modules. Before this PR, an import like `@/assets/logo.svg` fell through to code-module resolution and failed as a missing generated `logo.svg.js` file, which pointed users at a file they never authored.

This PR adds an import-rewriter strategy that rejects static asset module imports before alias/relative resolution and explains the supported model: move the asset to `public/` and reference it by root URL. CSS imports are explicitly excluded because the CSS pipeline supports and strips them earlier.

It also keeps build/import failures visible in the dev overlay even when a project has `pages/500.tsx`; ordinary runtime errors still use custom error pages.

## Verification

- `deno test --allow-all src/transforms/import-rewriter/strategies/asset-strategy.test.ts src/server/handlers/request/ssr/ssr.handler.test.ts`
- `deno task lint`
- `deno task typecheck`

## Review status

Score: 93/100. Recommended next step: merge after GitHub required checks finish green.

### DIFF
--- a/src/server/handlers/request/ssr/ssr.handler.test.ts
+++ b/src/server/handlers/request/ssr/ssr.handler.test.ts
@@ -1,3 +1,4 @@
+import { RENDER_ERROR } from "#veryfront/errors";
 import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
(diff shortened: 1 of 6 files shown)


kwakayama (commented):
Score: 84/100

The asset-strategy direction is good: reject unsupported static asset imports before alias/relative strategies append `.js`, and keep build/import failures visible in the dev overlay instead of hiding them behind a custom error page. Targeted local tests for `nested-imports`, `asset-strategy`, and `ssr.handler` pass on the #3014 stacked head.

Issue to fix or accept before landing:
- `src/transforms/import-rewriter/strategies/asset-strategy.ts` derives the suggested public filename with `info.specifier.split("/").pop()`. Because the matcher explicitly accepts query strings, `@/assets/logo.svg?raw` produces advice like `public/logo.svg?raw` and ``. Strip the query before deriving the basename and add a rewrite-message assertion for `?raw`.

Not merge-ready regardless:
- This is stacked above #3013 and transitively depends on lower PRs with requested changes and known typecheck failures.

kwakayama (dismissed):
Follow-up after fixes: approving. I added a regression for asset imports with query strings and normalized suggested public filenames so \ does not leak into the guidance URL. Local verification: \running 1 test from ./src/transforms/import-rewriter/strategies/asset-strategy.test.ts
AssetStrategy ...
  matches ...
    matches static asset extensions ... ok (1ms)
    ignores a query string when matching ... ok (0ms)
    does not match code modules ... ok (0ms)
    does not match CSS, which is supported and stripped earlier ... ok (0ms)
  matches ... ok (1ms)
  rewrite ...
    rejects with the file, the reason, and the supported alternative ... ok (0ms)
    drops query strings from the suggested public filename ... ok (0ms)
    names the importer relative to the project root ... ok (0ms)
    does not put a path outside the project in the message ... ok (0ms)
    uses no dash characters that the copy rules forbid ... ok (0ms)
  rewrite ... ok (1ms)
  runs before the alias and relative strategies ... ok (0ms)
AssetStrategy ... ok (4ms)

ok | 1 passed (12 steps) | 0 failed (7ms). Score: 92/100. Next step: merge after base stack and checks are green.

kwakayama:

Score: 92/100.

Fix added: regression for asset imports with query strings, and suggested public filenames now strip query/hash suffixes.

Verification:
- deno test --allow-all src/transforms/import-rewriter/strategies/asset-strategy.test.ts

kojiwakayama (approved):

kwakayama (approved):
Score: 93/100

Critical review result: this fixes a real developer-facing bug without inventing unsupported asset-module behavior. Asset imports now fail early with the actual specifier, importer, and supported `public/` URL model, while CSS remains excluded because it is handled earlier. The SSR guard is also justified: build/import errors need to stay visible in dev overlays even when a project has a custom 500 page; ordinary runtime errors still use the custom page.

Validation reviewed:
- `deno test --allow-all src/transforms/import-rewriter/strategies/asset-strategy.test.ts src/server/handlers/request/ssr/ssr.handler.test.ts`
- `deno task lint`
- `deno task typecheck`

kojiwakayama (approved):

PR #3016, opened 2026-07-21
fix(transforms/esm): never cache a degraded module artifact

mattboon:
## Why

Follow-up to #2999. That PR made nested esm.sh specifier resolution soft-fail instead of throwing, but a degraded artifact could then become indistinguishable from a complete artifact. A transient child-module fetch failure could be written locally, pushed to the distributed bundle cache, and memoized for later renders.

The previous runtime-resolution premise is also only safe for absolute `http(s)` dynamic imports. Relative dynamic imports inside an esm.sh bundle resolve against the local bundle-cache directory, and `npm:` or bare dynamic imports need resolver/import-map context that the cached module no longer has.

## What

- `buildReplacements` and `rewriteModuleImports` now return both replacements and a `degraded` signal.
- HTTP module caching marks degraded artifacts for the current render, but skips distributed cache writes and skips the in-memory path map so the next render retries.
- Local cache hits detect the degraded marker and refetch instead of reusing the incomplete artifact.
- Dynamic soft-fail is restricted to absolute `http(s)` specifiers; relative, `npm:`, and bare dynamic resolution failures are fatal again.

## Verification

- `deno test --allow-all src/transforms/esm/http-cache.test.ts src/transforms/esm/specifier-resolver.test.ts`
- `deno task lint`
- `deno task typecheck`

## Review status

## Operational note

Already-written unmarked degraded artifacts cannot be distinguished in code. If the remaining 24-hour distributed cache window matters during rollout, flush the HTTP module cache prefix.

### DIFF
--- a/src/transforms/esm/degraded-artifact.ts
+++ b/src/transforms/esm/degraded-artifact.ts
@@ -1,0 +1,24 @@
+/**
+ * Degradation marking for cached HTTP module artifacts.
+ *
+ * A module whose lazy dependency could not be prefetched is still usable for
+ * the render that produced it, but it must never be mistaken for a complete
+ * artifact on a later render. The marker travels with the code, so any reader
+ * of the on-disk cache can tell the two apart.
+ *
+ * @module transforms/esm/degraded-artifact
+ */
+
+/** Preserved comment format that survives minification */
+const VF_DEGRADED_MARKER = "\n/*! @vf-degraded */\n";
+
+/** Mark bundle code as produced from an incomplete dependency prefetch. */
+export function markDegradedArtifact(code: string): string {
+  if (isDegradedArtifact(code)) return code;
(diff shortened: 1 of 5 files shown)


kwakayama (approved):
Score: 94/100.

This fixes a real cache correctness bug from the dynamic import soft-fail path: degraded HTTP module artifacts are no longer promoted into the distributed cache or in-memory path map, and marked local artifacts are retried on the next render. The restriction to absolute dynamic http(s) specifiers is the right boundary; relative, npm:, and bare dynamic specifiers cannot be safely left for runtime resolution in cached esm.sh modules.

Local validation after rebasing on current main:
- `deno test --allow-all src/transforms/esm/http-cache.test.ts src/transforms/esm/specifier-resolver.test.ts`
- `deno task lint`
- `deno task typecheck`

Recommended next step: merge after GitHub required checks finish green.

PR #3021, opened 2026-07-22
Make project skills available to every agent

kojiwakayama:
## Summary

- Keep `load_skill` available on every built-in and custom agent path. Local and project runtimes also expose `load_skill_reference` and `execute_skill_script`; hosted chat reads references through `load_skill({ skillId, file })` and does not execute skill scripts directly.
- Add owner-scoped skill catalogs with the loadable identifier and one-line description to agent system prompts by default. Hosted catalogs also retain a distinct display name when one exists.
- Preserve explicit catalog selectors and request-level deny-all tool overrides. Restrictive internal-run allowlists derive delegation availability from all owner-visible skills, so catalog selectors never become capability gates.
- Document the runtime-specific skill tool surface and cover it with focused regressions.

## Verification

- `deno task fmt:check`
- `deno task lint`
- `deno task typecheck`
- `deno task docs:validate`, 45 passed, 87 steps
- `deno task test:unit`, 2,536 passed, 21,100 steps
- `deno task test`, 2,959 passed; 2 unrelated baseline failures

The broad-suite extension registry failure reproduces unchanged on `origin/main`. The npm supply-chain test also requires a generated `npm/esm` artifact that is not present in a source checkout.

Fixes veryfront/veryfront-issue-inbox#101


src/agent/factory.ts
```diff
@@ -123,48 +124,42 @@ export function agent(config: AgentConfig): Agent {
...
+  const augmentedSystem = async () => {
+    // Owner-aware: omitted selectors advertise every skill visible to this
+    // agent (unowned project skills plus its own). Explicit lists, including
+    // an empty list, retain their authored catalog selection.
+    const currentSkills = skillRegistry.resolveForAgent(skillsConfig, { agentId: id });
+    const basePrompt = typeof originalSystem === "function"
+      ? await originalSystem()
+      : originalSystem;
+    if (!currentSkills.size) return basePrompt ?? "You are a helpful assistant.";
+    return `${basePrompt}\n\n${buildSkillManifestPrompt(currentSkills)}`;
```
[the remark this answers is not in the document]
kojiwakayama: Addressed in 248d4474a. The skill-catalog path now applies the same default system prompt fallback before appending the catalog, so omitted system no longer renders as literal undefined. Added regression coverage for the default prompt plus available-skill catalog path; focused factory tests and the pre-push unit suite passed.

src/internal-agents/run-stream.ts
```diff
@@ -468,11 +465,8 @@ function applyRuntimeToolAllowlist(
     // rather than silently skipping enforcement.
     return {};
   }
-  const availableSkillIds = agent.config.skills === true
-    ? ["*"]
-    : Array.isArray(agent.config.skills)
-    ? agent.config.skills
-    : undefined;
+  const skillsConfig = agent.config.skills ?? true;
+  const availableSkillIds = skillsConfig === true ? ["*"] : skillsConfig;
```
kojiwakayama: With the `?? true` default, every agent that simply omits `skills` now gets `availableSkillIds = ["*"]`, so `invoke_agent` survives a restrictive `toolAllowlist` for **every** agent by default — including projects with zero skills. Before this PR that preservation required an explicit `skills` config.

Two concerns:

1. The doc comment above ("Delegation tools are preserved only when the agent advertises delegate-backed skills") no longer matches: `["*"]` is asserted unconditionally, without checking that any skill actually resolves.
2. This is a capability widening under allowlists: per the existing comment, preserved `invoke_agent` spawns child runs that are *not* capped by this run's allowlist. A run narrowed to e.g. `["read_baseline"]` can now escape into unrestricted child runs.

Consider gating the sentinel on whether any skills are actually visible to the agent (`skillRegistry.resolveForAgent(...).size > 0`), or at minimum updating the comment to state that delegation tools are now preserved for all agents except `skills: []`.

src/agent/runtime/agent-definition.ts
```diff
@@ -118,7 +118,7 @@ function parseCapabilitySelector(value: unknown): true | string[] | undefined {
     const ids = value
       .filter((entry): entry is string => typeof entry === "string" && entry.trim().length > 0)
       .map((entry) => entry.trim());
-    return ids.length > 0 ? ids : undefined;
+    return ids;
```
kojiwakayama: This helper is shared by both `skills` and `tools` parsing (lines 174–175), so the change affects `tools:` frontmatter too. I traced the `tools` consumers (`agent-markdown-adapter.ts` filters out the resulting empty object; `resolveHostedRuntimeAllowedTools` already treats `undefined` as `[]`), so that side looks behavior-neutral — but worth confirming that was checked.

One real edge: an array whose entries all *filter out* (whitespace-only or non-string entries, e.g. `skills: [" "]` or YAML that parses entries to numbers) now returns `[]` (deny-all catalog) instead of `undefined` (all visible skills). Malformed input silently flips from the permissive default to the most restrictive setting. If you want authored-`[]` to be the only deny-all trigger, distinguish `value.length === 0` from filtered-to-empty.

src/agent/hosted/chat-preparation.ts
```diff
@@ -118,6 +119,27 @@ export type HostedChatRuntimeCreationPreparationInput<TRuntimeAgentDefinition> =
   ) => string | ChatSystemMessage[];
 };
 
+function resolveHostedRuntimeSkills(input: {
```
kojiwakayama: Two points here:

1. **Semantics divergence between runtimes.** In the hosted path the selector now feeds `availableSkillIds`, which is an *execution gate* for `load_skill` (an agent with `skills: ["cite"]` cannot load an unlisted-but-visible skill; `skills: []` keeps the loader but can load nothing). In the classic factory runtime, `skills` only shapes the prompt manifest — `load_skill` can still load any visible skill by id. The docs added in this PR ("Skill loading tools remain available... regardless of the selector") don't say which enforcement model is intended. Worth aligning the two runtimes, or documenting the difference explicitly.

2. **Duplicated resolution logic.** This re-implements the id/short-name matching that `skillRegistry.resolveVisibleSkill` already does, with a subtle difference: the registry resolves own-short-name *first, then* exact id, while this ORs both — a selector entry that is simultaneously another skill's global id and this agent's short name selects both here, only one there. Extreme edge case, but a shared helper (or at least a cross-reference comment) would prevent drift.

src/agent/hosted/runtime-essential-tools.ts
```diff
@@ -58,14 +58,19 @@ export function resolveHostedRuntimeAllowedToolNames(
     }
   }
 
-  if (!input.availableSkillIds?.length) {
-    return resolvedToolNames;
+  // Skill loading is framework infrastructure for every agent. Preserve it
+  // under non-empty allowlists and config-derived empty selectors. Explicit
+  // request-level empty allowlists return above and remain deny-all.
+  if (resolvedToolNames.size > 0 || input.includeRuntimeEssentialToolsWhenEmpty) {
```
kojiwakayama, line 68: `SKILL_RUNTIME_TOOL_NAMES` is only `["load_skill", "load_skill_reference"]`, so under a restrictive allowlist an agent keeps the loader and reference reader but loses `execute_skill_script` — while the PR registers all three tools "for every built-in and custom agent path" and the new comment calls skill loading "framework infrastructure for every agent".

If the asymmetry is intentional (script execution being a broader capability than loading instructions, like the `load_tools` asymmetry documented above), a sentence in this comment saying so would help; otherwise `execute_skill_script` should probably join the preserved set. As written, a skill whose instructions direct the model to run one of its scripts will fail mid-skill under an allowlist that preserved the loader.

src/agent/factory.ts
```diff
@@ -123,48 +124,43 @@ export function agent(config: AgentConfig): Agent {
...
-      const basePrompt = typeof originalSystem === "function"
-        ? await originalSystem()
-        : originalSystem;
-      if (!currentSkills.size) return basePrompt ?? "You are a helpful assistant.";
-      return `${basePrompt}\n\n${buildSkillManifestPrompt(currentSkills)}`;
-    }
-    : originalSystem;
+  const skillsConfig = config.skills ?? true;
+
+  const augmentedSystem = async () => {
```
kojiwakayama, line 156: Two downstream effects of unconditionally wrapping `system` worth confirming:

1. `config.system` is now *always* an async function (never a plain string, never `undefined`), and agents with no system resolve to "You are a helpful assistant." everywhere. The consumers I checked (`resolveBaseSystemPrompt`, `resolveAgentSystem`, `getEffectiveAgentSystem`) all handle functions, but any downstream that substituted its own richer default for an unset system will silently stop doing so.

2. Possible double skill catalog in hosted prompts: `createRuntimeAgentDefinitionFromAgent` resolves this augmented system into `instructions` (consumed by `veryfront-cloud-agent-service.ts:578`), so a code agent's hosted instructions now embed the factory manifest — while `prepareHostedChatRuntimeCreationOptions` separately builds a skills block from steering. The factory manifest also resolves against the *local* `skillRegistry`, which may disagree with the server-side steering catalog. This duplication existed before for `skills`-enabled agents, but it's now the default for every agent — was that considered?

kojiwakayama (commented):
## Review

Solid change overall — the direction (skill loading as universal framework infrastructure, `skills` reduced to a catalog selector) is clearly documented, the docs are updated everywhere the old semantics were described, and test coverage is strong: the policy-gating regression test (parallel `write_report` blocked before `load_skill` establishes policy), the `skills: []` selector tests across factory/hosted/markdown paths, and the deny-all vs config-derived-empty asymmetry tests all cover the risky edges. Extracting `ensureBuiltinSchemaValidator` into its own module so `factory.ts` doesn't pull in all builtin extensions is a nice touch.

I left 5 inline comments. The two I'd want resolved before merge:

1. **`invoke_agent` now survives restrictive tool allowlists for every agent by default** (`run-stream.ts`) — the `["*"]` sentinel applies to agents with no `skills` config even in projects with zero skills, and the updated doc comment says the opposite.
2. **Selector semantics diverge between hosted and classic runtimes** (`chat-preparation.ts`) — hosted uses `skills: [...]` as an execution gate for `load_skill`, the classic runtime treats it as advertisement-only.

The rest are questions/nits (filtered-to-empty selector edge in `parseCapabilitySelector`, `execute_skill_script` missing from allowlist preservation, possible double skill catalog in hosted prompts).

src/internal-agents/run-stream.ts
kojiwakayama: Still valid on current head, so I am keeping this unresolved. The code still turns omitted `skills` into `availableSkillIds = ["*"]` before allowlist resolution, which preserves `invoke_agent` for every default-skilled agent under restrictive allowlists. This should either be gated on actually visible delegate-backed skills or explicitly accepted/documented as a capability widening.

src/agent/runtime/agent-definition.ts
kojiwakayama: Checked. The `tools:` side remains behavior-neutral in the current consumers, and the filtered-to-empty case is acceptable as fail-closed behavior: malformed selector entries should not silently widen back to the default all-visible skill catalog. Resolving this one as an intentional restrictive outcome.

src/agent/hosted/chat-preparation.ts
kojiwakayama: Still valid on current head, so I am keeping this unresolved. Hosted runtime still uses the selector as `availableSkillIds` for `load_skill`, while the classic factory runtime treats `skills` as catalog selection and leaves `load_skill` able to resolve any visible skill. This needs an alignment patch or explicit docs/tests for the runtime difference.

src/agent/hosted/runtime-essential-tools.ts
kojiwakayama: Resolving as intentional. `execute_skill_script` is a broader execution capability than loading skill instructions or reading declared references, so restrictive allowlists should be able to block it even when `load_skill` and `load_skill_reference` are preserved as framework infrastructure. The active-skill policy still gates script execution when the tool is explicitly allowed.

src/agent/factory.ts
kojiwakayama: Part 1 is checked: the known consumers handle async system resolvers and the default helpful system fallback is now covered by regression tests. Part 2 still looks valid for hosted code agents: `createRuntimeAgentDefinitionFromAgent` can resolve the factory-augmented system into `instructions`, then hosted preparation can add steering skills again. Keeping this unresolved until that path is patched, deduped, or explicitly documented.

src/skill/prompt-augmentation.ts
```diff
@@ -28,7 +28,7 @@ export function buildSkillManifestPrompt(skills: Map<string, Skill>): string {
   ];
 
   for (const [id, skill] of skills) {
-    lines.push(`- **${id}**: ${skill.metadata.description}`);
+    lines.push(`- **${skill.metadata.name}** (\`${id}\`): ${skill.metadata.description}`);
   }
```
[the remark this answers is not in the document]
kojiwakayama: Landed in cf1f05f9d. The default skill manifest now caps rendered entries at 30 and adds a truncation note that tells the agent to use `load_skill` for discovery beyond the visible list. Added regression coverage for truncation and verified the branch with focused prompt-augmentation tests plus the pre-push suite.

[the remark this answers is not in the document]
kojiwakayama: Landed in cf1f05f9d. The manifest rendering no longer treats `metadata.name` as a display title; it renders the skill id with the description, which matches the validated skill metadata model and avoids duplicated `name`/id output.

src/skill/prompt-augmentation.test.ts
```diff
@@ -19,10 +19,12 @@ describe("src/skill/prompt-augmentation", () => {
...
     it("should include header for single skill", () => {
-      const skills = new Map([["my-skill", createSkill("my-skill", "Does things")]]);
+      const skills = new Map([
+        ["my-skill", createSkill("my-skill", "Does things", "My Skill")],
+      ]);
       const result = buildSkillManifestPrompt(skills);
       assertStringIncludes(result, "## Available Skills");
-      assertStringIncludes(result, "**my-skill**: Does things");
+      assertStringIncludes(result, "**My Skill** (`my-skill`): Does things");
     });
```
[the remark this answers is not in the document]
kojiwakayama: Landed in cf1f05f9d. The prompt-augmentation test fixtures now use realistic lowercase id-shaped `metadata.name` values and assert id-only manifest rendering, so the tests match production parser constraints.

src/agent/runtime/agent-definition.ts
kojiwakayama: Checked this again. For both `skills` and `tools`, filtered-to-empty arrays are a fail-closed outcome: malformed entries no longer silently widen back to the omitted-selector default. That is safer than treating invalid array contents as `undefined`, and authored `[]` plus filtered-empty both mean no advertised capabilities. Resolving this as intentional behavior.

kojiwakayama:
Implemented and committed fixes for PR 3021 follow-ups. Addressed and resolved thread PRRT_kwDOQaPiP86S7RvA (allowlist:  now requires visible skills). Remaining open threads: PRRT_kwDOQaPiP86S7RvL (runtime semantics divergence intentional, needs follow-up) and PRRT_kwDOQaPiP86S7RvO (duplicated manifest in hosted prompts).

src/internal-agents/run-stream.ts
kojiwakayama: Confirmed this is still actionable and I am leaving it unresolved. Current head still computes `skillsConfig = agent.config.skills ?? true` and turns that into `availableSkillIds = ["*"]`, so `invoke_agent` preservation under restrictive allowlists is now the default even when the agent has no resolved visible skills. Smallest fix is to gate delegation preservation on the resolved visible skill set, or explicitly document that all agents except `skills: []` preserve delegation tools.

src/agent/runtime/agent-definition.ts
kojiwakayama: Confirmed this should stay open. The `tools:` side looks behavior-neutral because downstream consumers already normalize an absent/empty tool selector safely, but the shared parser still turns arrays whose entries all filter out, for example `skills: [" "]`, into `[]`. That silently changes malformed authored input from the omitted-selector default to a deny-all skill catalog. Smallest fix is to distinguish authored empty arrays from filtered-to-empty arrays.

src/agent/hosted/chat-preparation.ts
kojiwakayama: Confirmed this is a real alignment question, so I am leaving it unresolved. Hosted preparation uses the filtered `visibleSkills` both for the prompt manifest and for `availableSkillIds`, which gates hosted `load_skill`. The classic factory path uses the selector only for the prompt manifest; local `load_skill` can still resolve any skill visible to the agent. This should be aligned in code or called out explicitly in docs/tests.

src/agent/hosted/runtime-essential-tools.ts
kojiwakayama: Confirmed this asymmetry is intentional. `load_skill` and `load_skill_reference` are instruction/reference infrastructure, but `execute_skill_script` can execute local scripts advertised by a loaded skill, so it should not be auto-preserved under restrictive hosted allowlists. It remains available on normal agent construction and is further gated by the active skill policy when present. Resolving this as no code change required.

src/agent/factory.ts
kojiwakayama: Confirmed this should stay open. Code agents materialized for hosted runs resolve the factory-augmented system into `instructions`, and the cloud runtime builder then separately receives `visibleSkills` for its runtime skill block. That means default-enabled code agents can carry a local factory manifest plus a hosted steering skill catalog, and those catalogs can diverge by source. Smallest fix is to avoid embedding the factory manifest when materializing hosted runtime instructions, or make the hosted prompt builder detect and replace the factory catalog.

src/internal-agents/run-stream.ts
```diff
@@ -446,8 +442,9 @@ function getRuntimeToolAllowlist(
 
 /**
  * Intersects the merged run tool set with the restrictive tool allowlist.
- * Skill runtime/delegation tools are preserved for skill-enabled agents,
- * mirroring the hosted chat runtime's allowlist semantics.
+ * Skill runtime tools are preserved for every agent. Delegation tools are
+ * preserved only when the agent advertises delegate-backed skills, mirroring
+ * the hosted chat runtime's allowlist semantics.
```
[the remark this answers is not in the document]
kojiwakayama: Confirmed. This is the same policy mismatch tracked in the adjacent delegation-gate thread: current code treats omitted/true `skills` as delegation-enabled via `availableSkillIds = ["*"]`, while the comment still says delegation is tied to delegate-backed skills. Leaving open until the implementation is gated on resolved visible skills or the comment is changed to describe the wider policy.

src/agent/hosted/runtime-essential-tools.ts
```diff
@@ -58,14 +58,19 @@ export function resolveHostedRuntimeAllowedToolNames(
...
+    for (const toolName of SKILL_RUNTIME_TOOL_NAMES) {
```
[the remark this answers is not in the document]
kojiwakayama: This is intentional and duplicates the nearby thread. The always-preserved set is limited to `load_skill` and `load_skill_reference` because those expose instructions and references; `execute_skill_script` can run local scripts from a loaded skill and should require an explicit allowlist entry in restrictive hosted runs. Resolving as no code change.

src/internal-agents/run-stream.ts
kojiwakayama: Addressed in b7a34704c. `applyRuntimeToolAllowlist` now resolves the actual visible skills with `skillRegistry.resolveForAgent(...)` and only preserves `invoke_agent` when that resolved set is non-empty. Added RED/GREEN coverage for both cases: no visible skills withhold delegation, visible default skills preserve it. Verified locally through the pre-push hook: fmt, lint, typecheck, and unit tests passed (`2533 passed`, `21099 steps`, `0 failed`).

src/agent/runtime/agent-definition.ts
kojiwakayama: Addressed in b7a34704c. Authored empty arrays still parse as `[]`, but arrays that filter down to empty now fall back to omitted-selector semantics (`undefined`) for both `skills` and `tools`. Added regression coverage for `skills: [" ", 7]` and `tools: [" ", 7]`. Verified locally through the pre-push hook: fmt, lint, typecheck, and unit tests passed (`2533 passed`, `21099 steps`, `0 failed`).

src/agent/hosted/chat-preparation.ts
kojiwakayama: Leaving this unresolved. This is a real design split, not something I want to resolve with a small local patch: hosted `load_skill` is gated by per-run `availableSkillIds`, while the classic factory tool uses the shared local skill registry and currently treats `skills` as catalog advertisement. Aligning those two paths needs a deliberate follow-up so local dynamic skill discovery and hosted server-backed skill gating do not drift further.

src/agent/hosted/runtime-essential-tools.ts
kojiwakayama: Resolved as intentional and now documented in b7a34704c. Hosted allowlist preservation keeps `load_skill` and `load_skill_reference` as framework infrastructure, but intentionally does not preserve `execute_skill_script`; running a project-provided script remains a direct execution capability and must be explicitly allowed by the run. Added an assertion that `execute_skill_script` is not force-added under a restrictive allowlist. Verified locally through the pre-push hook: fmt, lint, typecheck, and unit tests passed (`2533 passed`, `21099 steps`, `0 failed`).

src/agent/factory.ts
kojiwakayama: Leaving this unresolved. The duplicate hosted prompt risk is real for code-defined hosted agents: `createRuntimeAgentDefinitionFromAgent` resolves the factory-augmented code-agent system into hosted instructions, and hosted preparation can also append its server steering skill block. Fixing that needs a deliberate hosted/code-agent prompt boundary change, not a review-thread-only clarification.

src/internal-agents/run-stream.ts
kojiwakayama: Addressed in b7a34704c. `applyRuntimeToolAllowlist` now resolves the agent's visible skills and only passes `availableSkillIds` when that set is non-empty, so agents with no visible skills no longer preserve delegation tools through a restrictive allowlist. Resolving this thread; the narrower wording mismatch on "delegate-backed skills" remains tracked in the newer run-stream comment.

src/agent/runtime/agent-definition.ts
kojiwakayama: Addressed in b7a34704c. `parseCapabilitySelector` now preserves an authored empty array as `[]`, but returns `undefined` when a non-empty array filters down to no valid string ids. That keeps `skills: []` as the only deny-all selector while malformed/blank entries fall back to omitted-selector behavior. Resolving.

src/internal-agents/run-stream.ts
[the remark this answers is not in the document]
kojiwakayama: Addressed in 68348b6e1. The run-stream comment now matches the implementation after b7a34704c: delegation preservation is tied to the agent having at least one resolved visible skill, not merely omitted/true `skills` and not the old "delegate-backed skills" wording. The latest pre-push gate also passed (`2535 passed`, `0 failed`). Resolving this thread.

src/agent/factory.ts
kojiwakayama: Addressed in bb7f5176c. I checked the code-agent materialization path and added a regression in `src/agent/project/agent-runtime.test.ts`: `getEffectiveAgentSystem(...)` still produces the local factory-augmented prompt with `## Available Skills`, but `agent.config.system` remains the original system string and `createRuntimeAgentDefinitionFromAgent(...)` materializes that original string for hosted/runtime instructions. That means the factory skill catalog is not embedded into hosted instructions before hosted preparation appends its server-side steering block. Docs and `AgentConfig.skills` now also describe selectors as prompt-catalog-only. Latest pre-push passed (`2535 passed`, `0 failed`). Resolving.

kojiwakayama:
Follow-up to the earlier status comment: all review threads are now addressed and resolved at `bb7f5176c`.

- Hosted `skills` selectors now shape only the prompt catalog; hosted `load_skill` remains able to load every owner-visible skill by ID, matching the classic runtime.
- Selector resolution is shared and uses owner visibility plus own short-name precedence.
- Delegation preservation now requires at least one actually visible skill.
- Malformed filtered-to-empty capability selectors no longer become authored deny-all selectors.
- `execute_skill_script` remains intentionally subject to explicit restrictive allowlists because it executes project code.
- A regression test confirms the local factory catalog is not materialized into hosted code-agent instructions, so hosted steering does not duplicate it.

Verification: full pre-push gate passed (`2535 passed`, `0 failed`); focused affected suite passed (`80 passed`, `0 failed`); formatting, lint, test typecheck baseline, Deno checks, and docs validation passed. GitHub CI is green on the latest head: 23 passed, 7 expected skips, 0 failures, and 0 pending.

kojiwakayama:
Re-reviewed the follow-up commits (`248d4474a`…`bb7f5176c`) against the earlier review threads — all five are properly addressed at head:

- **Delegation under allowlists** — `applyRuntimeToolAllowlist` now resolves actual visible skills via `skillRegistry.resolveForAgent` and only preserves `invoke_agent` when the set is non-empty; the doc comment matches (`b7a34704c` + `68348b6e1`). ✅
- **Selector parsing** — authored `[]` stays deny-all; filtered-to-empty falls back to `undefined` for both `skills` and `tools`, with regression coverage. ✅
- **Hosted vs classic `load_skill` semantics** — aligned in `9415e1f31`: the selector shapes only the advertised catalog (`advertisedSkills`), while `availableSkillIds` is built from all owner-visible skills (`loadableSkills`), matching the classic runtime; documented in `AgentConfig.skills`. The shared `resolveRuntimeSkillsForAgent` also fixes the short-name-precedence drift. ✅
- **`execute_skill_script` asymmetry** — intentional, now documented at `SKILL_RUNTIME_TOOL_NAMES`. ✅
- **Double catalog / always-function system** — nicely solved in `bb7f5176c`: `config.system` keeps the authored value, the augmented resolver lives in the symbol-keyed effective-system channel consumed by the internal runtime and `run-system-prompt`, so `createRuntimeAgentDefinitionFromAgent` no longer leaks the factory manifest into hosted instructions. ✅

One residual nit, non-blocking: `cf1f05f9d` reverted the factory manifest to `- **id**: description`, but the hosted block in `src/agent/runtime/skill-prompt.ts` still renders `- Name (` + "`id`" + `): description`. The two catalogs now use different formats, and the PR description's "title, id, and one-line description" only matches the hosted one. Worth harmonizing (either direction) or updating the PR body.

LGTM otherwise.

src/skill/prompt-augmentation.ts
```diff
@@ -27,10 +30,20 @@ export function buildSkillManifestPrompt(skills: Map<string, Skill>): string {
     "",
   ];
 
-  for (const [id, skill] of skills) {
+  const displaySkills = Array.from(skills).slice(0, MAX_SKILL_MANIFEST_PROMPT_ENTRIES);
+  for (const [id, skill] of displaySkills) {
     lines.push(`- **${id}**: ${skill.metadata.description}`);
   }
```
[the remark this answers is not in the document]
kojiwakayama: Addressed in `bf3b62458`. `buildSkillManifestPrompt` now iterates the Map directly and breaks at `MAX_SKILL_MANIFEST_PROMPT_ENTRIES`, so the 30-entry prompt cap also caps iteration/allocation work. Added a regression that instruments the Map iterator: it failed against the previous `Array.from(...).slice(...)` implementation with 32 visited entries and now passes with exactly 30. The related catalog-format nit is also settled: hosted prompts no longer repeat `name (id)` when name equals id, but retain a distinct display name when present. Full pre-push passed (`2536 passed`, `21100 steps`, `0 failed`).

kojiwakayama:
Settled the residual catalog review in `bf3b62458` and `896b6a3e2`:

- The factory manifest now stops Map iteration at the 30-entry prompt cap instead of allocating every skill first. A regression proves only 30 entries are visited.
- Hosted catalogs no longer render duplicate `name (id)` text when the name equals the ID. They still preserve `display name (id)` when a genuinely distinct display name exists.
- The PR summary now describes the shared contract precisely: loadable identifier plus one-line description everywhere, with a distinct hosted display name when available.
- Truncation guidance no longer claims `load_skill` can discover omitted skills. It requires a known ID in the factory prompt and points hosted agents to IDs in the tool schema.
- Both follow-up Copilot threads have been answered and resolved.

Full pre-push verification passed: `2536 passed`, `21100 steps`, `0 failed`.

src/skill/prompt-augmentation.ts
```diff
@@ -27,10 +30,20 @@ export function buildSkillManifestPrompt(skills: Map<string, Skill>): string {
...
 
+  if (skills.size > MAX_SKILL_MANIFEST_PROMPT_ENTRIES) {
+    lines.push("");
+    lines.push(
+      `${
+        skills.size - MAX_SKILL_MANIFEST_PROMPT_ENTRIES
+      } more skills available. Use load_skill to discover them.`,
+    );
```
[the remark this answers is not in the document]
kojiwakayama: Addressed in 896b6a3e2. The prompt no longer claims load_skill can discover skills. The factory note states that summaries were omitted and requires a known ID; the hosted note points to the IDs exposed in the load_skill tool schema. I added regressions that reject the old discovery wording. Full pre-push verification passed: 2536 tests, 21100 steps, 0 failures.

kwakayama:
## Review: 82/100

**Recommendation: Request changes**

### Approach summary

This PR makes skill catalogs opt-out rather than opt-in:

- Omitted `skills` advertises the owner-visible catalog by default.
- `skills: []` suppresses the catalog only; it does not disable loading an owner-visible skill by a known ID.
- An explicit list narrows prompt advertisement without becoming a runtime authorization boundary.
- Hosted preparation and steering refresh now use the same selector semantics and owner scope.
- Catalogs are capped at 30 entries, and the truncation guidance no longer claims `load_skill` can discover unknown IDs.
- Explicit request-level deny-all tool overrides remain authoritative.

That is the right overall model: concise prompt-time discovery, on-demand instruction loading, and authorization kept separate from catalog selection.

### Finding

**[High] Hosted agents still do not implement the advertised three-tool skill contract**

`getDiscoveredHostTools()` filters out every shared skill tool via `SKILL_TOOL_IDS` in `src/agent/hosted/veryfront-cloud-agent-service.ts`, then hosted global/local tool assembly manually adds only `load_skill`.

Consequently, hosted agents do not receive direct `load_skill_reference` or `execute_skill_script` tools. Hosted `load_skill` supports an optional `file` parameter, so reference loading has an alternate path, but hosted script execution is absent. This conflicts with both the PR summary and the guides, which state that every agent receives:

- `load_skill`
- `load_skill_reference`
- `execute_skill_script`

This is central to the stated “every built-in and custom agent path” contract, so I consider it blocking.

**Suggested resolution:** either:

1. Add request-scoped hosted implementations of `load_skill_reference` and `execute_skill_script`, preserving project/auth/branch/owner context and adding hosted assembly tests; or
2. Narrow the product/docs contract explicitly: hosted agents expose `load_skill`, references are loaded through its `file` input, and hosted skill script execution is unsupported.

### What looks good

The prior selector/loadability divergence, refresh widening, duplicate hosted catalog concern, default-system handling, delegation widening, prompt-size cap, and invalid truncation guidance are resolved on the current head. Owner scoping and request-level deny-all behavior have focused regression coverage.

### Verification

Reviewed head: `896b6a3e27e782e0068d4292503ab05ff9db7525`

- Focused skill/runtime suite: **72 passed, 58 substeps, 0 failed**
- Current CI at review time: format, lint, typecheck, unit, integration, coverage, RSC browser E2E, npm smoke, security audit, and completed CodeQL jobs are green
- Still running at review time: binary E2E and one CodeQL analysis job

kojiwakayama:
Addressed at `15392e7ac` using the review suggestion to narrow the contract explicitly.

- `load_skill` is the universal tool contract.
- Hosted chat reads advertised references through the request-scoped `load_skill({ skillId, file })` path, which preserves project, branch, auth, and owner context.
- Hosted chat does not expose direct skill script execution. I did not reuse the shared `execute_skill_script` implementation because it depends on local registry and filesystem state and would lack a hosted request-scoped authorization and sandbox boundary.
- Local and project runtimes continue to expose `load_skill_reference` and `execute_skill_script`.
- The guides, API copy, architecture note, hosted assembly comment, and PR summary now state this distinction. A regression explicitly locks out both shared filesystem skill tools from hosted discovery.

I also addressed the latest suppressed Copilot finding in the same commit: restrictive internal-run allowlists now derive delegation availability from all owner-visible skills, not the prompt catalog selector. A `skills: []` regression failed before the fix and passes afterward.

Verification: focused hosted/internal tests pass; docs validation passes (`45 passed`, `87 steps`); lint, typecheck, Deno checks, and diff checks pass; full pre-push passes (`2536 passed`, `21100 steps`, `0 failed`).

kwakayama (dismissed):

kwakayama (dismissed):

src/agent/factory.ts
```diff
@@ -123,48 +124,43 @@ export function agent(config: AgentConfig): Agent {
...
-      };
+  if (config.tools !== true) {
+    const configuredTools = { ...(config.tools ?? {}) };
+    for (const registration of SKILL_TOOL_REGISTRATIONS) {
+      if (!(registration.id in configuredTools)) {
+        configuredTools[registration.id] = true;
+      }
     }
+    mergedToolsConfig = configuredTools;
   }
```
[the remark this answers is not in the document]
kojiwakayama: Addressed in `90d4d9700`. Agent-level `false` entries can no longer disable the framework skill tools. The merge still preserves concrete Tool objects, so hosted request-scoped `load_skill` bindings are not overwritten, and request-level allowlists remain the restrictive authorization surface. Added a regression covering `false` for all three registrations: it failed before the fix and passes afterward. Focused factory/project/hosted verification passed (24 tests), and the full pre-push gate passed (`2543 passed`, `21213 steps`, `0 failed`) with format, lint, and typecheck green.

kwakayama (approved):

src/internal-agents/run-stream.ts
```diff
@@ -468,11 +466,8 @@ function applyRuntimeToolAllowlist(
...
     return {};
   }
-  const availableSkillIds = agent.config.skills === true
-    ? ["*"]
-    : Array.isArray(agent.config.skills)
-    ? agent.config.skills
-    : undefined;
+  const visibleSkills = skillRegistry.resolveForAgent(true, { agentId: agent.id });
+  const availableSkillIds = visibleSkills.size > 0 ? [...visibleSkills.keys()] : undefined;
   const allowedToolNames = resolveHostedRuntimeAllowedToolNames({
```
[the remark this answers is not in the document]
kojiwakayama: Confirmed, this comment is valid. PR #3021 merged before the follow-up could be added, so it is addressed in #3022 by commit 3a1a1ece5. The code still resolves actual owner-visible skills, but now passes a one-element sentinel instead of allocating every visible skill ID because the allowlist resolver only inspects `.length`. Focused run-stream (30 steps) and allowlist-resolver (8 steps) tests passed, and the full pre-push gate passed with 2,543 tests (21,213 steps), zero failures, plus format, lint, and typecheck.

PR #3112, opened 2026-07-25
fix(binary): resolve a handler's bare zod import on the compiled binary (#217)

mattboon:
## Problem

An API route with `import { z } from "zod"` **500s on the compiled binary**:

```
Import "zod" not a dependency and not in import map
```

This is the binary half of veryfront-issue-inbox#217 that veryfront-code#3105 explicitly deferred ("The compiled binary keeps zod excluded … tracked separately").

## Root cause

`getUserDependencies()` force-excluded `zod` from `userDeps` **only on the binary** (`if (isDeno && isCompiledBinary) frameworkPackages.add("zod")`). On the binary, `rewriteCompiledBinaryUserDependencyImports` only rewrites specifiers present in `userDeps` — so the excluded `import "zod"` was left bare and reached Deno, which can't resolve it.

#3105 already keeps zod as a user dep on Node and the Deno source-run. This drops the binary-only exclusion so zod is treated like any other npm package and rewritten through the `createRequire` shim to resolve from the project's `node_modules`. zod stays force-externalized (esbuild `external`) — never bundled inline.

## Verification (per LOCALDEV.md)

Built a fresh binary and ran it against `veryfront-router-testing/default-config`:
- A route importing `zod` and calling `z.object().safeParse(...)` → **200** (was 500).
- Middleware/CORS routes unaffected (still 200/401/504).
- `deno task test:unit`: **2680 passed, 0 failed**. Updated the paired `getUserDependencies` regression test to expect zod on the binary too.

## Scope note

`/api/validated` in that repro *additionally* imports `createValidatedHandler` from the **root** `veryfront`, which the hand-written root runtime shim omits — a **separate** root-shim-completeness gap filed as veryfront-issue-inbox#231. That is not addressed here; this PR is strictly the bare-npm-dep (zod) resolution.

Ref: veryfront-issue-inbox#217

### DIFF
--- a/src/routing/api/module-loader/loader.test.ts
+++ b/src/routing/api/module-loader/loader.test.ts
@@ -486,12 +486,13 @@ describe("loadHandlerModule", { sanitizeResources: false, sanitizeOps: false }, 
     assertEquals(packages, ["zod", "pdf-parse", "another-lib"]);
   });
 
-  it("keeps zod for Deno source runs but excludes it from compiled binaries", () => {
+  it("keeps zod as a user dep on every runtime, incl. the compiled binary (#217)", () => {
     const dependencies = new Map([
       ["zod", "^3.22.0"],
       ["pdf-parse", "^1.1.1"],
     ]);
 
+    // Deno source-run keeps zod (#3105) ...
     assertEquals(
       [...getUserDependencies(dependencies, { isDeno: true, isCompiledBinary: false })],
       [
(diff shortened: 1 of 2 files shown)


kojiwakayama:
Addressed the CI lint failure in `3e7f7f415`. The runtime-specific zod branch had already been removed, so I removed the now-obsolete runtime parameter instead of suppressing the unused-parameter warning, and consolidated the duplicate runtime assertions into one runtime-independent regression test.

Verification:
- `deno fmt --check`
- `deno lint`
- focused module-loader suite: 46 steps, zero failures
- `deno task typecheck`
- full pre-push gate: 2,680 tests / 21,796 steps, zero failures

kwakayama (approved):

PR #3127, opened 2026-07-27
Refresh incomplete proxy routing metadata

kwakayama:
## Summary

- treat cached staging/production routing metadata without an active release as incomplete
- apply the same check when requests join an in-flight routing lookup
- keep the one-refresh retry budget caller-local so existing cohorts stay bounded without making later callers accept stale metadata
- preserve normal cache hits for active routes and leave the existing invalidation architecture unchanged

## Root cause

The proxy cached routing metadata while a new project had no active release. After publishing activated the release, the proxy could keep reusing that cached `active_release_id: null` result, so the deployment URL returned 404 until the cache expired. The same stale result could also be shared by in-flight lookups that started before activation.

## Validation

- serial regression: first request sees no active release, release becomes active, next request succeeds
- activation concurrency: two post-activation waiters reject one pre-activation result, share one refresh, then a following request hits the refreshed cache
- persistent-null concurrency: three requests return the expected 404 with exactly one initial lookup and one shared refresh
- temporal race: a caller arriving after activation during another caller’s incomplete refresh rejects the stale result and fetches the active release
- proxy handler suite: 59 checks passed
- concurrency stress: 20 repeated handler-suite runs passed
- strict test-typecheck baseline: 0 new failures
- full pre-push verification: format, lint, typecheck, and 2,687 tests passed
- architecture review: keep invalidation as-is; keep retry state at the request boundary rather than on shared cache state
- simplicity review: one caller-local boolean in the existing lookup loop; no new dependency or production-level abstraction

Closes veryfront/veryfront-issue-inbox#266

### DIFF
--- a/src/proxy/handler.test.ts
+++ b/src/proxy/handler.test.ts
@@ -191,7 +191,11 @@ function createNotFoundResponse(): Response {
   return new Response("Not found", { status: 404 });
 }
 
-function createHandler(port: number, apiBasePath = "") {
+function createHandler(
+  port: number,
+  apiBasePath = "",
+  logger?: Parameters<typeof createProxyHandler>[0]["logger"],
+) {
   return createProxyHandler({
     config: {
       apiBaseUrl: `http://127.0.0.1:${port}${apiBasePath}`,
@@ -200,6 +204,7 @@ function createHandler(port: number, apiBasePath = "") {
       previewApiClientId: "test-client",
       previewApiClientSecret: "test-secret",
     },
+    logger,
   });
 }
 
@@ -498,6 +503,96 @@ describe("Proxy Handler", () => {
 
         await handler.close();
       } finally {
+        await server.shutdown();
+      }
+    });
+
+    it("refreshes an in-flight routing result without an active release", async () => {
+      let routingLookups = 0;
+      let activeReleaseId: string | null = null;
+      let releaseFirstLookup: (() => void) | undefined;
+      let markFirstLookupStarted!: () => void;
+      let markLookupJoined!: () => void;
+      const firstLookupStarted = new Promise<void>((resolve) => {
+        markFirstLookupStarted = resolve;
+      });
+      const lookupJoined = new Promise<void>((resolve) => {
+        markLookupJoined = resolve;
+      });
+      const firstLookupRelease = new Promise<void>((resolve) => {
+        releaseFirstLookup = resolve;
+      });
+      const { logger } = createRecordingLogger();
+      const recordDebug = logger.debug;
+      logger.debug = (message, extra) => {
+        recordDebug(message, extra);
+        if (message === "Proxy routing metadata lookup joined in-flight request") {
+          markLookupJoined();
+        }
+      };
+
+      const { server, port } = createMockServer(async (req: Request) => {
+        const { pathname } = new URL(req.url);
+
+        if (pathname === "/auth/token") return createTokenResponse();
+
+        if (pathname.startsWith("/projects/-/proxy-routing/")) {
+          routingLookups++;
+          const releaseIdAtLookupStart = activeReleaseId;
+          if (routingLookups === 1) {
+            markFirstLookupStarted();
+            await firstLookupRelease;
+          }
+          return Response.json({
+            id: "proj-123",
+            slug: "my-project",
+            name: "My Project",
+            environments: [{
+              id: "env-1",
+              name: "staging",
+              active_release_id: releaseIdAtLookupStart,
+            }],
+          });
+        }
+
+        if (pathname.startsWith("/projects/-/proxy-access/")) {
+          return Response.json({
+            id: "proj-123",
+            slug: "my-project",
+            environments: [{
+              id: "env-1",
+              name: "staging",
+              protected: false,
+            }],
+          });
+        }
+
+        return createNotFoundResponse();
+      });
+
+      try {
+        const handler = createHandler(port, "", logger);
+        const req = () =>
+          new Request("http://my-project.staging.veryfront.com/page", {
+            headers: { host: "my-project.staging.veryfront.com" },
+          });
+
+        const beforeActivation = handler.processRequest(req());
+        await firstLookupStarted;
+        activeReleaseId = "rel-456";
+        const afterActivation = handler.processRequest(req());
+        await lookupJoined;
+        releaseFirstLookup();
+
+        assertEquals((await beforeActivation).error?.status, 404);
+        const afterActivationResult = await afterActivation;
+        assertEquals(afterActivationResult.error, undefined);
+        assertEquals(afterActivationResult.releaseId, "rel-456");
+        assertEquals(routingLookups, 2);
+
+        await handler.close();
+      } finally {
+        releaseFirstLookup?.();
         await server.shutdown();
       }
     });
--- a/src/proxy/handler.test.ts
+++ b/src/proxy/handler.test.ts
@@ -443,6 +443,65 @@ describe("Proxy Handler", () => {
       }
     });
 
+    it("refreshes cached routing metadata without an active release", async () => {
+      let routingLookups = 0;
+      let activeReleaseId: string | null = null;
+      const { server, port } = createMockServer((req: Request) => {
+        const { pathname } = new URL(req.url);
+
+        if (pathname === "/auth/token") return createTokenResponse();
+
+        if (pathname.startsWith("/projects/-/proxy-routing/")) {
+          routingLookups++;
+          return Response.json({
+            id: "proj-123",
+            slug: "my-project",
+            name: "My Project",
+            environments: [{
+              id: "env-1",
+              name: "staging",
+              active_release_id: activeReleaseId,
+            }],
+          });
+        }
+
+        if (pathname.startsWith("/projects/-/proxy-access/")) {
+          return Response.json({
+            id: "proj-123",
+            slug: "my-project",
+            environments: [{
+              id: "env-1",
+              name: "staging",
+              protected: false,
+            }],
+          });
+        }
+
+        return createNotFoundResponse();
+      });
+
+      try {
+        const handler = createHandler(port);
+        const req = () =>
+          new Request("http://my-project.staging.veryfront.com/page", {
+            headers: { host: "my-project.staging.veryfront.com" },
+          });
+
+        const beforeActivation = await handler.processRequest(req());
+        activeReleaseId = "rel-456";
+        const afterActivation = await handler.processRequest(req());
+
+        assertEquals(beforeActivation.error?.status, 404);
+        assertEquals(afterActivation.error, undefined);
+        assertEquals(afterActivation.releaseId, "rel-456");
+        assertEquals(routingLookups, 2);
+
+        await handler.close();
+      } finally {
+        await server.shutdown();
+      }
+    });
+
     it("invalidates routing metadata only for the targeted project", async () => {
       const routingLookups = new Map<string, number>();
       const { server, port } = createMockServer((req: Request) => {
--- a/src/proxy/handler.test.ts
+++ b/src/proxy/handler.test.ts
@@ -697,6 +697,98 @@ describe("Proxy Handler", () => {
       }
     });
 
+    it("refreshes for a caller arriving after activation during an incomplete refresh", async () => {
+      let routingLookups = 0;
+      let activeReleaseId: string | null = null;
+      let releaseIncompleteRefresh!: () => void;
+      let markIncompleteRefreshStarted!: () => void;
+      let markLookupJoined!: () => void;
+      const incompleteRefreshStarted = new Promise<void>((resolve) => {
+        markIncompleteRefreshStarted = resolve;
+      });
+      const lookupJoined = new Promise<void>((resolve) => {
+        markLookupJoined = resolve;
+      });
+      const incompleteRefreshRelease = new Promise<void>((resolve) => {
+        releaseIncompleteRefresh = resolve;
+      });
+      const { logger } = createRecordingLogger();
+      const recordDebug = logger.debug;
+      logger.debug = (message, extra) => {
+        recordDebug(message, extra);
+        if (message === "Proxy routing metadata lookup joined in-flight request") {
+          markLookupJoined();
+        }
+      };
+
+      const { server, port } = createMockServer(async (req: Request) => {
+        const { pathname } = new URL(req.url);
+
+        if (pathname === "/auth/token") return createTokenResponse();
+
+        if (pathname.startsWith("/projects/-/proxy-routing/")) {
+          routingLookups++;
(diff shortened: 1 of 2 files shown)


src/proxy/handler.ts
```diff
@@ -592,17 +599,22 @@ export function createProxyHandler(options: ProxyHandlerOptions) {
...
+    isCachedResultUsable?: (result: ProjectRoutingLookupResult) => boolean,
   ): Promise<ProjectRoutingLookupResult | null> {
     const cacheKey = normalizeProjectLookupKey(lookupKey);
     return await profileProxyServerTimingPhase(
       timing ?? { enabled: false, startedAt: 0, phases: new Map() },
       "proxy.routing_lookup",
       async () => {
         const cached = getCachedRoutingLookup(cacheKey);
-        if (cached) {
+        if (cached && (!isCachedResultUsable || isCachedResultUsable(cached))) {
```
kwakayama: **Blocking:** the active-release predicate is only applied to completed cache hits. A request that arrives after activation can still join a lookup that began before activation and return its stale `active_release_id: null` result via the `existingLookup.promise` branch below.

I reproduced this on the PR head by delaying the first routing lookup, activating the release, then starting a second staging request before releasing the first response. The second request returned the same 404 and only one routing lookup occurred. Please validate the joined result with `isCachedResultUsable`; if it is unusable, skip/evict it and start a fresh lookup, with a concurrent regression test.

kwakayama (commented):
## Critical review — 82/100

**Verdict: changes required.** GitHub does not allow an author to formally request changes on their own PR, so this is submitted as a comment review.

The serial stale-cache path is fixed cleanly and the focused regression passes, but the same stale 404 remains possible when the post-activation request joins a routing lookup that began before activation. That is within the core “next request succeeds after publish” guarantee, so I consider it blocking.

### Score

- Correctness: 34/45
- Tests: 16/20
- Design: 18/20
- Maintainability: 9/10
- Scope discipline: 5/5

### Standards

No hard documented-standard violations. The diff is small, uses existing cache/in-flight structures, adds no dependency, and follows the repository test conventions. The optional cache-usability callback plus `requireActiveRelease` boolean is slightly more generic than the one current policy needs, but that is a non-blocking design judgment.

### Spec

The completed-cache case matches the issue: a cached matched environment with `active_release_id: null` is evicted, the next serial request refreshes, and active routes retain cache hits. The missing in-flight validation leaves one concurrent form of the reported behavior unresolved.

### Evidence

- Existing proxy handler suite: 56 steps passed.
- Added adversarial local regression on the pinned PR head: failed because the after-activation request returned `release-not-found` (404) and joined the first lookup instead of issuing a second lookup.
- `deno fmt --check`, `deno lint`, and `deno check src/proxy/handler.ts`: passed.
- GitHub CI/CD and CodeQL: passed.

Please apply the same usability check to a joined in-flight result; when it is unusable for this caller, start a fresh lookup rather than returning it. Add the delayed/concurrent regression so the “next request” guarantee covers both completed-cache and in-flight states.

src/proxy/handler.test.ts
```diff
@@ -443,6 +448,155 @@ describe("Proxy Handler", () => {
...
+          new Request("http://my-project.staging.veryfront.com/page", {
+            headers: { host: "my-project.staging.veryfront.com" },
+          });
+
+        const beforeActivation = handler.processRequest(req());
+        await firstLookupStarted;
+        activeReleaseId = "rel-456";
+        const afterActivation = handler.processRequest(req());
+        await lookupJoined;
+        releaseFirstLookup();
```
kwakayama, line 588: **Blocking:** this call does not typecheck because `releaseFirstLookup` is declared as `(() => void) | undefined`. Both local `deno check src/proxy/handler.test.ts` and CI's `lint:test-typecheck` report a new error here (`TS2722: Cannot invoke an object which is possibly 'undefined'`). Please make the invocation type-safe; the current PR cannot pass the required lint gate.

src/proxy/handler.ts
```diff
@@ -592,22 +599,32 @@ export function createProxyHandler(options: ProxyHandlerOptions) {
...
         const existingLookup = routingLookupInflight.get(cacheKey);
         if (existingLookup?.generation === routingLookupGeneration) {
           logger?.debug("Proxy routing metadata lookup joined in-flight request", { lookupKey });
-          return await existingLookup.promise;
+          const result = await existingLookup.promise;
+          if (!result || !isCachedResultUsable || isCachedResultUsable(result)) {
+            return result;
+          }
+          routingLookupCache.delete(cacheKey);
+          logger?.info("Refreshing incomplete proxy routing metadata", { lookupKey });
```
kwakayama: **Medium:** after an unusable joined result, every waiter falls through to create and register its own `lookupPromise`. With two post-activation joiners on the same blocked stale lookup, I reproduced 3 routing requests total instead of 2: both joiners independently refreshed and overwrote the in-flight entry. This breaks the existing single-flight behavior and can fan out control-plane load at activation time. Please re-check/elect one refresh leader so other waiters join it, and add a multi-joiner assertion plus a subsequent cache-hit assertion.

kwakayama (commented):
## Critical re-review - 87/100

**Verdict: changes required.** The prior single-joiner stale-404 blocker is fixed, but the latest head still has one hard verification failure and one medium concurrency issue. GitHub does not allow the PR author to formally request changes, so this is submitted as a comment review.

### Score

- Correctness: 40/45
- Tests: 15/20
- Design: 18/20
- Maintainability: 9/10
- Scope discipline: 5/5

### Standards

One hard violation: the new concurrent regression has a TypeScript error at `src/proxy/handler.test.ts:585` because `releaseFirstLookup` is typed as possibly undefined. The repository's `lint:test-typecheck` gate fails, and the latest CI/CD run is red. This conflicts with the repository's strict-TypeScript and verified-completion requirements.

Non-blocking judgment calls: the cached-result and joined-result branches duplicate the usability/delete/log policy, and the test synchronizes through an exact debug-log string. Both are maintainability concerns but not correctness blockers.

### Spec

No missing spec requirements found. The serial cache path passes, and the new deterministic single-joiner regression proves that a post-activation request rejects the stale joined result, refreshes, and succeeds. Active cache hits and the invalidation design remain intact.

### Additional concurrency finding

When two or more requests join the same stale lookup, every joiner falls through and creates its own refresh. I reproduced this with two post-activation joiners: all requests succeeded, but routing lookups were `3` instead of the single-flight expectation of `2`. Re-elect one refresher (or re-check/join the current in-flight entry) before creating a new lookup, and cover multiple joiners plus the following cache hit.

### Evidence

- Original 57-step proxy handler suite: passed.
- `deno fmt --check` and `deno lint` on changed files: passed.
- `deno check src/proxy/handler.ts src/proxy/handler.test.ts`: failed with `TS2722` at line 585.
- `deno task lint:test-typecheck`: failed locally and in [CI/CD run 8112](https://github.com/veryfront/veryfront-code/actions/runs/30270039732).
- Two-joiner adversarial regression: failed with 3 routing lookups instead of 2.
- CodeQL: passed.

kwakayama (commented):
## Critical re-review - 97/100

**Verdict: ready to merge.** I found no blocking or material findings on the latest head. GitHub does not allow an author to formally approve their own pull request, so this is posted as a comment review.

The two issues from my previous review are resolved:

- Strict typechecking now passes; the deferred test resolver is definitely assigned.
- Concurrent post-activation waiters preserve single-flight behavior: one waiter elects the refresh and the others join it.

### Score

- Correctness: 45/45
- Tests and regression coverage: 20/20
- Design: 18/20
- Maintainability: 9/10
- Scope discipline: 5/5

The small deduction is for non-blocking maintainability friction in the tests: synchronization depends on an exact debug-log message, and the new scenarios repeat some setup. Neither affects correctness or merge readiness.

### Verification

- Proxy handler suite: 57 steps passed
- 20 repeated stress runs: passed
- Format, lint, and direct `deno check`: passed
- Strict test-typecheck baseline: 0 new failures
- CI/CD and CodeQL on `c5c5895b3fd219a8a1c4e6d85f91b78f87ceab12`: passed
- Independent standards, spec-compliance, and cache-concurrency reviews: no material findings

kwakayama (commented):
## Correction to my latest re-review - 89/100

**Verdict: changes required.** This review supersedes my 97/100 comment above. A new thread landed immediately before that review, and I have now reproduced its no-active-release concurrency case on the same head.

When three requests share an initial routing lookup and the environment legitimately still has no active release:

1. The first lookup returns incomplete metadata.
2. One waiter elects refresh lookup 2 and the other waiter joins it.
3. Lookup 2 also returns incomplete metadata.
4. Its owner accepts that result, but the remaining waiter rejects it and starts lookup 3.

The result is one sequential routing lookup per waiter instead of one shared, bounded refresh. Responses remain the expected 404, but the loop amplifies control-plane load and request latency, violating the PR's stated single-flight guarantee.

### Score

- Correctness: 41/45
- Tests and regression coverage: 17/20
- Design: 17/20
- Maintainability: 9/10
- Scope discipline: 5/5

### Required change

Bound each caller cohort to one elected refresh after rejecting the initial incomplete joined result. Waiters should accept that refresh result even if it remains incomplete. Add a deterministic regression with three concurrent requests, a persistently null active release, all responses returning 404, and exactly two routing lookups.

### Evidence

I added that adversarial regression in an isolated worktree on `c5c5895b3fd219a8a1c4e6d85f91b78f87ceab12`. It failed with `routingLookups === 3` instead of `2`. The existing 57-step suite and CI remain green because they only cover the successful post-activation refresh where lookup 2 returns an active release.

Existing thread: https://github.com/veryfront/veryfront-code/pull/3127#discussion_r3660086689

src/proxy/handler.ts
```diff
@@ -592,22 +600,44 @@ export function createProxyHandler(options: ProxyHandlerOptions) {
...
+        while (true) {
+          const existingLookup = routingLookupInflight.get(cacheKey);
+          if (!existingLookup || existingLookup.generation !== routingLookupGeneration) break;
 
-        const existingLookup = routingLookupInflight.get(cacheKey);
-        if (existingLookup?.generation === routingLookupGeneration) {
           logger?.debug("Proxy routing metadata lookup joined in-flight request", { lookupKey });
-          return await existingLookup.promise;
+          const result = await existingLookup.promise;
+          if (existingLookup.isRefreshAfterIncompleteResult || canUseResult(result)) return result;
```
kwakayama: **Blocking:** `isRefreshAfterIncompleteResult` is shared by every caller joining this promise, so it can bypass the current request's active-release predicate. I reproduced: seed cached null metadata; request A starts a marked refresh before activation and that lookup captures null; activate the release; request B starts after activation and joins A's lookup. This condition makes B accept the null result and return a false 404 instead of issuing the post-activation lookup.

The bounded-retry state needs to be caller-local, not inherited from the in-flight entry. Keep the persistent-null cohort bounded, but always evaluate a joined result for callers that have not themselves rejected an incomplete result. Add the deterministic activation-during-marked-refresh regression; B should resolve `rel-456` with one additional routing lookup.

kwakayama (commented):
## Critical re-review - 85/100

**Verdict: changes required.** The latest commit fixes the persistent-null fanout, but the shared refresh marker introduces a narrower form of the original stale-404 race.

The marker describes whether one request has already spent its bounded refresh attempt. Storing it on the shared in-flight entry makes that request-local state apply to later callers. A request that starts after activation can therefore accept a pre-activation null result without running its own active-release check.

GitHub does not allow the PR author to formally request changes, so this is posted as a comment review.

### Score

- Correctness: 38/45
- Tests and regression coverage: 17/20
- Design: 16/20
- Maintainability: 9/10
- Scope discipline: 5/5

### Standards

No hard documented-standard violations. Non-blocking judgment calls remain: the active-release policy is spread across a generic predicate and several booleans, the new tests duplicate substantial fixture setup, and concurrency synchronization depends on exact log copy.

### Spec

One blocking spec miss: the post-publish joined-result guarantee still fails when the joined lookup is already marked as a refresh. The new persistent-null test correctly proves bounded fanout, but it does not cover activation while that marked refresh is in flight.

### Evidence

- Existing proxy handler suite: 58 steps passed
- Format, lint, direct `deno check`, and strict test-typecheck baseline: passed
- Adversarial temporal-race regression on `2140f8df47393fdcacf2655dd1277766d30359ef`: failed because the post-activation request returned `release-not-found` instead of `rel-456`
- Standard CI/CD jobs were green at review time; binary e2e and CodeQL were still running

kwakayama (commented):
## No-change re-review - 85/100

**Verdict remains: changes required.** The PR head is still `2140f8df47393fdcacf2655dd1277766d30359ef`, identical to the commit reviewed in my previous 85/100 review. There is no new diff to evaluate, and the blocking activation-during-marked-refresh thread remains unresolved.

CI/CD and CodeQL have now completed successfully. That does not change the verdict because the deterministic adversarial case is not covered by the current suite: a request starting after activation can still join a pre-activation marked refresh, accept `active_release_id: null`, and return a false 404.

Author self-review cannot submit a formal request-changes event, so this is recorded as a comment review.

kwakayama (commented):
## Critical re-review - 96/100

**Verdict: ready to merge.** I found no blocking or material findings on the latest head. GitHub does not allow an author to formally approve their own pull request, so this is posted as a comment review.

The prior blocker is resolved:

- `hasRejectedIncompleteResult` is local to each `resolveProjectRoutingLookup` call rather than shared on the in-flight entry.
- Existing waiters still share one bounded refresh when the release remains null.
- A later caller arriving after activation evaluates the stale joined result independently, rejects it, and fetches the active release.
- Different environment matchers sharing a lookup key no longer inherit another caller's retry state.

### Score

- Correctness: 45/45
- Tests and regression coverage: 20/20
- Design: 18/20
- Maintainability: 8/10
- Scope discipline: 5/5

The deduction is non-blocking test maintainability: four scenarios repeat substantial staging proxy fixture setup, and three concurrency tests synchronize through exact debug-log text. The generic predicate plus boolean policy is also slightly more abstract than the single active-release rule needs.

### Standards

No hard documented-standard violations. The diff stays scoped, adds no dependency, follows strict TypeScript conventions, and preserves the existing invalidation architecture.

### Spec

No missing, partial, or incorrect requirements found. Serial activation, activation concurrency, persistent-null bounded fanout, and the temporal late-caller race are all covered.

### Verification

- Proxy handler suite: 59 steps passed
- 20 repeated handler-suite stress runs: passed
- Format, lint, direct `deno check`, and strict test-typecheck baseline: passed
- CI/CD and CodeQL on `3b47bc3ab365f193cd8534875322362d96c9fe1b`: passed
- Independent standards, spec, and cache-concurrency reviews: no material findings

kojiwakayama (approved):
Verified all review threads are resolved and the full required check set passes. The caller-local bounded refresh preserves single-flight behavior while preventing stale no-release metadata from surviving activation.

PR #3164, opened 2026-07-29
Fix suppressed tool result ordering

kwakayama:
## Summary

- defer suppressed-tool recovery guidance until all valid tool results from the current stream step are persisted
- add a public `agent(...).stream(...)` regression covering one unavailable call alongside two valid parallel calls

## Root cause

When a model step mixed unavailable and valid tool calls, the runtime assembled the continuation as `assistant → user recovery note → tool results`. Anthropic requires tool results to follow the corresponding assistant tool uses without an intervening user turn, so it rejected the continuation with HTTP 400.

The fix preserves `assistant → tool results → user recovery note`. The recovery guidance remains the final input before the next model step, and the existing suppressed-only recovery behavior is unchanged.

## Impact

Agent runs can now recover from unavailable streamed tools while still completing valid parallel tool calls in the same step. No provider-specific converter behavior, tool availability policy, or integration implementation changes.

## Validation

- RED regression reproduced `assistant → user → tool`
- GREEN regression verifies `assistant → tool → user`
- adjacent runtime suites: 142 steps passed
- repository pre-push gate: 2,717 tests / 22,101 steps passed
- full lint and typecheck passed
- production binary build passed


src/agent/runtime/refresh.test.ts
```diff
@@ -244,6 +244,103 @@ describe("agent runtime refresh hooks", () => {
...
+
+    await (await assistant.stream({ input: "Build an agent" })).toDataStreamResponse().text();
+
+    assertEquals(callCount, 2);
+    const retryPrompt = observedPrompts[1] ?? [];
+    assertEquals(
+      retryPrompt.slice(-3).map((message) => message.role),
+      ["assistant", "tool", "user"],
+    );
+    assertEquals(
```
kwakayama: Non-blocking test hardening: these `JSON.stringify(...).includes(toolCallId)` checks prove both IDs are present, but not that the assistant calls and tool results are paired structurally or that the successful outputs are correct. Assert the structured assistant tool-call IDs/names and tool-result IDs/outputs exactly so a placeholder, error payload, or unpaired result cannot satisfy the regression.

kwakayama (commented):
Review score: **96/100** — merge recommended (threshold: 90).

No blocking findings.

- Correctness: the change makes the retry transcript `assistant → tool → user` while preserving suppressed-only recovery and existing abort/error behavior.
- Spec: all acceptance criteria in `issue-326-prd.md` are implemented without provider-specific repair or dependency changes.
- Standards: minimal, focused diff; colocated regression coverage; clean formatting and repository conventions.
- Verification: fresh local runs passed `24/24` focused steps and `113/113` adjacent runtime steps. GitHub format, lint, typecheck, unit, integration, coverage, browser E2E, binary E2E, smoke, CLA, and CodeQL checks are green.

One non-blocking test-specificity improvement is noted inline.

kojiwakayama (dismissed):

kojiwakayama (approved):

PR #3183, opened 2026-07-30
Refactor import rewrite into shared core adapter

kojiwakayama:
## Why\nUnify route/module import rewrite behavior by routing all import edits through a private core and consolidating package/ESM handling in adapter boundaries.\n\n## Scope\n- Single import rewrite mechanism shared across loader entry paths\n- Compatibility adapter for SSR quirks\n- Tests for ESM and package semantics


kojiwakayama:
Closed because this PR was opened accidentally from an unrelated pre-existing branch during a read-only architecture analysis. No branch was deleted or modified.

PR #3222, opened 2026-08-02
fix(agent): preserve deferred catalogs beyond provider limits

kojiwakayama:
## Summary

- preserve the full authorized remote-tool catalog when hosted agents use deferred tool loading
- apply provider tool-count limits only to eager assembly and each model-visible schema set
- reserve provider capacity for bootstrap tools and `tool_search` across repeated searches and restored checkpoints
- derive that budget from the effective request model, including generate and stream model overrides
- deterministically evict the oldest loaded schema before activating a new match at capacity
- version ordered-recency checkpoints as v2 while retaining deterministic v1 restore compatibility
- prove an OpenAI run can search, load, and execute an alphabetically late tool from a 133-tool catalog while every provider step stays at or below 128 schemas

## Why

Hosted assembly previously applied the OpenAI 128-tool limit to the authorization catalog itself. Tools removed there were absent from both `tool_search` and execution, so deferred discovery could not recover them.

The exposure state also needs a provider-aware budget: without it, repeated searches, restored checkpoints, or an OpenAI per-request override could report schemas as loaded and later have the converter silently truncate them. The runtime now budgets against the effective model, reserves bootstrap/search capacity, and keeps the mutable loaded set within the remaining budget.

## Dependency and rollout

Blocked on veryfront/veryfront-api#4234 merging and deploying first. Framework emits ordered checkpoint v2; API #4234 adds N/N-1 support for sorted v1 plus ordered v2, preserves v2 recency through authenticated persistence/reload, and rejects duplicates/unknown versions.

Required order:

1. merge and deploy API #4234;
2. merge and publish this Framework PR;
3. update Agent #1758 to consume the immutable Framework release;
4. rerun Agent CI and durable staging discovery/execution proof before marking Agent ready.

## Validation

- 85 focused hosted, exposure, runtime-step, durability, and checkpoint tests passed
- generate and stream model-override regressions pass
- cumulative activation, exact-fit capacity, over-budget restore, visible capability search, and v1/v2 checkpoint regressions pass
- full pre-push: 2,891 tests + 22,884 nested steps, zero failures
- all GitHub formatting, lint, typecheck, unit, integration, coverage, smoke, browser/binary E2E, package, and CodeQL checks are green
- `git diff --check` passes
- authorization/security review found no policy widening

Tracks veryfront/veryfront-issue-inbox#353.

Replaces #3210 solely to restore an eligible two-person approval path. The base, head commit, commit history, and diff are unchanged.

PR #3257, opened 2026-08-02
fix: address post-merge review findings

kojiwakayama:
Fixes confirmed findings from a post-merge review sweep of the last ~10 merged PRs.

## dotenv escape handling (from #3230)

- `expandDoubleQuotedCharacters` expanded `\n`/`\r`/`\t` but not `\\`, so a double-quoted value like `"path\\nvalue"` produced backslash+newline instead of the literal backslash+n that `@std/dotenv` produces.
- `expandValue` suppressed interpolation whenever the character before `$` was a backslash, so `\\$VAR` (escaped backslash — even count) wrongly skipped interpolation. Now uses the existing `isEscaped()` parity check, matching `findClosingQuote`.

Further commits on this branch address latent findings from #3226/#3227 (AJV strictness, 5xx detail stripping, logError OTel bypass).

Review provenance: adversarial post-merge sweep of #3222–#3232; all other reviewed PRs came back clean.

### DIFF
--- a/src/platform/compat/std/dotenv.test.ts
+++ b/src/platform/compat/std/dotenv.test.ts
@@ -30,6 +30,7 @@ describe("platform/compat/std/dotenv", () => {
         "EMPTY=",
         "SINGLE='  $EXPORTED  '",
         'DOUBLE="line\\nnext\\tend"',
+        'ESCAPED_BACKSLASH="path\\\\nvalue"',
         'MULTILINE="first',
         'second"',
         "INLINE=kept#comment",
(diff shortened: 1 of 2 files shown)


kojiwakayama:
Closing after independent differential review. Both proposed changes diverge from the pinned authoritative jsr:@std/dotenv@0.225.6 behavior while the PR claims parity: paired backslash+n and backslash-dollar cases produce different values. The focused tests only encoded the proposed outputs and did not compare the compatibility authority. We are retaining exact @std compatibility rather than introducing an undocumented Veryfront-specific dotenv dialect. The PR body also described follow-up commits that were never present.

PR #3279, opened 2026-08-02
Make dependency artifact timeout tests deterministic

kojiwakayama:
## Summary

- drive upstream fetch and body-read deadlines with controlled time instead of a 1 ms wall-clock race
- assert that an in-flight fetch observes an un-aborted signal before the deadline and an aborted signal after it
- cover the valid expired-before-start branch, which returns `upstream_timeout` without invoking the fetcher

## Diagnosis

The merge-group failure was test nondeterminism, not a runtime deadline failure. Under parallel coverage load, the 1 ms deadline expired between deadline creation and the first fetch race. Production correctly refused to begin already-expired work, while the test incorrectly assumed the fetch callback had necessarily captured a signal.

No production timeout behavior changed. Controlled time advances 30 seconds virtually, so this adds no wall-clock sleep.

## Verification

- 100 baseline focused iterations on current main: passed
- 200 controlled-time focused iterations: passed
- dependency artifact builder suite: 23 steps passed
- exact coverage shard 2/8 on latest main: 347 suites, 3254 steps passed
- `deno task verify:quick`: passed
- focused format, lint, typecheck, and diff checks: passed

Related: merge-group run 30763657761, job 91538524234

### DIFF
--- a/src/release-assets/dependency-artifact-builder.test.ts
+++ b/src/release-assets/dependency-artifact-builder.test.ts
@@ -3,6 +3,7 @@ import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals, assertStringIncludes } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import { computeHashBytes } from "#veryfront/utils";
+import { FakeTime } from "#std/testing/time";
 import {
   buildDependencyArtifactGraph,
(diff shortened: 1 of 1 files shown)


kojiwakayama:
@codex review. Focus on deadline semantics, fake-time isolation under parallel test execution, and whether the regression preserves fail-closed upstream behavior.

kwakayama (approved):

Issue #3214, opened 2026-08-01
Persist and inspect Model Call Context for durable runs


kwakayama:
# One Pager: `AGENT_RUN_MODEL_CALL_CONTEXT`

Date: 2026-08-03  
Status: Proposed simplified design  
Tracking issue: https://github.com/veryfront/veryfront-code/issues/3214

This one-pager supersedes the recorder and Model Call Context-specific chunking
design in the earlier PRD.

## Goal

Give developers the complete, provider-agnostic input Veryfront sent to the LLM
for every model call, so they can understand and debug model behavior.

Today, durable runs retain lifecycle and output events, but developers cannot see
the final instructions, message history, and tools assembled for an individual
model call. The original conversation is not enough: later calls can include tool
results, changed instructions, compacted history, or a different exposed tool set.

## Decision

Persist one additional durable run event immediately before every model dispatch:

```ts
export interface AgentRunModelCallContextEvent {
  type: "AGENT_RUN_MODEL_CALL_CONTEXT"
  messages: ModelCallMessage[]
  tools?: ModelCallTool[]
}
```

`messages` is ordered exactly as supplied by the framework's shared model-call
boundary. System instructions—including resolved agent instructions and
available-skill guidance—are system-role messages. `tools` contains the final
provider-agnostic tool definitions exposed for that call.

The event row already supplies its run ID, event ID, ordering, and timestamp.
Do not duplicate that provenance in the payload, and do not add a `contextId`.

```text
Resolve final instructions, messages, and tools
                     │
                     ▼
     AGENT_RUN_MODEL_CALL_CONTEXT
                     │
                     ▼
          Persist as one run event
                     │
                     ▼
         Dispatch generate or stream call
```

The persistence order creates a clear invariant:

> If the model request was dispatched, its Model Call Context is available in
> the durable run events.

If the event cannot be persisted, the model call is not dispatched. This avoids
successful model calls whose input cannot later be inspected.

## What The Event Contains

- Final ordered system, user, assistant, and tool messages.
- Resolved system instructions as system-role messages.
- Tool calls and tool results already present in the message history.
- Final provider-agnostic definitions of tools exposed to the model.

It deliberately excludes:

- model selection and generation settings;
- duplicated run or conversation provenance;
- model output, token usage, errors, or retry metadata;
- provider request bodies, headers, credentials, and transport metadata; and
- instructions or transformations added privately by the model provider.

This is the content Veryfront supplies to its model adapter, not a copy of a
[...]

PR #3280, opened 2026-08-02
feat(cli): publish dedicated proxy binary

kwakayama:
## Summary

- publish dedicated `veryfront-proxy-linux-x64` and `veryfront-proxy-linux-arm64` release assets
- compile the proxy from `cli/proxy-main.ts` with a frozen graph-specific lock
- share proxy startup, provider activation, and teardown with `veryfront serve --mode=proxy`
- publish a proxy-specific SBOM and reject stale proxy locks
- smoke-test memory, Redis, ambient Redis registration, OpenTelemetry, and Sentry
- enforce three cold starts under a 1536 MiB Docker cgroup

## Compatibility

- existing universal CLI release assets retain the full compile profile
- `CACHE_TYPE=redis` is translated to the extension-backed cache after provider registration
- `REDIS_URL` activates the Redis runtime provider independently of cache selection
- production can open routing-invalidation Redis connections when its secret and replica configuration are present
- code-executing pull-request CI remains disabled for external forks
- ARM64 is cross-compiled; executable provider and memory smoke remains on native Linux x64

## Exact-head evidence

Validated at `0a4b9502821a261ea63280c0b9d1be8019e5cb0f`.

- GitHub reports `CLEAN`
- hosted checks succeeded, with no failed or pending checks
- 0 unresolved review threads
- focused build and SBOM suites: 17 tests, 14 nested steps, 0 failed
- proxy runtime and composition suites: 2 files, 11 steps, 0 failed
- format, lint, typecheck, Bash syntax, and diff checks passed
- exact hosted Linux x64 compile reported 76.35 MB of embedded files and passed the 180 MiB executable ceiling
- exact hosted memory gate passed 3 of 3 cold starts at 1536 MiB with health success, clean shutdown, and `OOMKilled=false`
- production-mode compiled Redis smoke on ancestor `52c6c6cdbd8d59d2a1c8f29858017d13d9acbafc` reached `RedisRuntimeProvider registered`, `Proxy routing invalidation bus connected`, and a healthy proxy response against disposable Redis

## Rollout status

Merged prerequisites:

- `veryfront-server#275`: dedicated artifact resolution
- `veryfront-code#3282`: durable child writer consumer
- `veryfront-server#280`: proxy runtime selection and temporary 1 GiB request / 2 GiB limit

After merge and patch release:

1. verify both proxy assets and digests
2. verify the server release dispatch selects the dedicated x64 artifact
3. verify staging readiness and health
4. verify zero OOM events and unexpected restarts
5. record cold-start peak and steady-state memory
6. restore the prior known-good image if rollout acceptance fails


kojiwakayama:
Review follow-up is pushed at `4ac1e19e5` (implementation in `dcf242aca`, fork-guard correction in `4ac1e19e5`).

Addressed suggestions:
- Captured the native Promise constructor before extension activation and use it for the process-lifetime promise.
- Missing provider log markers now print the smoke log and return failure while the trap still cleans up the process.
- The smoke script already uses a temporary log file and removes it during cleanup.
- Corrected the misleading cache-translation comment.

Two suggestions were intentionally not applied:
- The shorter workflow expression is logically equivalent, but the complete pre-push suite proved it violates the repository hardening contract that requires the shared fork guard verbatim in every code-executing job. The exact guard was restored and both focused hardening tests pass.
- `veryfront/extensions/loader` was deliberately removed by this PR's hardening follow-up, so the test remains on the supported public extension barrel.

Verification: focused proxy/composition and compile tests passed; Deno 2.7.7 format, lint, typecheck, repository-hardening tests, Bash syntax, diff validation, and the complete pre-push unit suite passed. Fresh exact-head CI is running. The PR remains draft because the staging proxy rollout is still OOM/restart unstable.

kojiwakayama:
Fresh CI correctly rejected a stale proxy dependency lock after the current-main merge added the YAML extension workspace to the workspace graph. I regenerated the graph-specific lock with Deno 2.7.7 and pushed `48298573c`.

Verification before push:
- repeat `build:proxy-lock` regeneration is stable
- compile-binary suite: 11 passed
- full pre-push gate: format 4,880 files, lint 4,803 files, typecheck passed, 3,719 tests / 26,698 steps / 0 failed

Fresh exact-head CI is running. The PR remains draft and unmerged behind the staging rollout gate.

kwakayama:
## Review: 82/100 — sound; agree with holding per your own rollout gate

| Axis | Score |
|---|---|
| Correctness | 33/40 |
| Test adequacy | 21/25 |
| Security / prod-safety | 16/20 |
| Maintainability | 12/15 |
| **Total** | **82/100** |

Merge base `cf9b54607bcd`, three-dot throughout.

**This is the strongest release-pipeline work in the current batch**, and the supply-chain posture is genuinely good. Not 90+ because two real behavior changes ride along that the title does not advertise — both on an already-shipped path.

### Draft status is intentional and correct

The body carries an explicit rollout gate ("keep in draft until the staging proxy rollout is stable; current staging deployment remains `ProgressDeadlineExceeded`, new image replica restarting after OOM") plus a merge order. That is a sequencing gate on a live incident, not incomplete work. Nothing in the diff reads as half-built.

### Publishing and shadowing — verified clean

`veryfront-proxy-linux-x64`, linux-x64 only, a sixth entry in the `build-binaries` matrix (`cicd.yml:456-460`), picked up by the release glob at `:576`.

No shadowing, checked three ways:
- `scripts/install.sh:155` builds `BINARY_NAME="veryfront-${PLATFORM}"` from an exact `uname` switch — a literal name, no glob. It can never select the proxy binary.
- The homebrew step (`cicd.yml:827`) iterates an explicit four-name list excluding it.
- SBOM `proxy.json` cannot collide — `outputFileNameForManifest` (`generate-sbom.ts:299-310`) emits `core/cli/react/ext-*/all`, and no workspace member normalizes to `proxy`.

Pre-existing and not introduced here, but worth stating: the release job runs `gh release delete "v${VERSION}" --cleanup-tag` before `create`, so a re-run replaces an existing release wholesale.

### The existing `veryfront` binary is unchanged, and explicitly protected

`DEFAULT_INCLUDES` is untouched, and `includesForProfile(options.profile ?? "full")` returns it unchanged when the profile is absent. Both the CI matrix and `build-all.js` were edited to give every existing target an explicit `entrypoint: cli/main.ts, profile: full` rather than relying on the default. The pre-existing `compile-binary.test.ts` cases survive as regression guards. I went looking for a hidden main-binary regression and did not find one.

### Supply chain — the strong part

`--node-modules-dir=none --lock scripts/build/proxy-deno.lock --frozen` (`compile-binary.ts:81-88`), with lock freshness enforced in **both** the PR job and the release build via `deno task build:proxy-lock && git diff --exit-code` (`cicd.yml:415-418`, `:447-452`). That is fail-closed, so second-lockfile drift is genuinely closed rather than documented. Actions SHA-pinned, fork guard preserved, proxy-specific SBOM published.

The size claim is well-founded structurally: the proxy lock carries **208 npm packages vs 784** in the workspace lock (73% reduction), `PROXY_INCLUDES` drops `dist/framework-src`, and `compile-binary.test.ts:150-162` asserts the lock contains no `@huggingface/transformers`, `esbuild`, or `sharp`. I could not measure actual binary size without building.

### P2 — ext-redis now activates on `REDIS_URL` alone, changing the existing CLI path

`cli/commands/serve/proxy-extension-composition.ts:70`

Verified against main:

```
main:56    if (cacheType === "memory") return null;      // early return, loads nothing
branch:70  if (getEnv("REDIS_URL")) { ...selected.push(ext-redis)... }   // independent of CACHE_TYPE
branch:77  if (selected.length === 0) return null;
```

The `memory` early return is gone. **Concrete:** an operator running `veryfront serve --mode=proxy` today with `CACHE_TYPE=memory` and an ambient `REDIS_URL` — inherited from a shared container env, or set for a sibling service — previously activated zero extensions. After this they get an `ExtensionLoader` and a registered Redis runtime provider.

Severity is limited because setup is lazy: `extensions/ext-redis/src/index.ts:28-39` only calls `createRedisRuntimeProvider()` and `ctx.provide(...)` with no connect, and the smoke test proves it by passing with an unreachable `REDIS_URL=redis://127.0.0.1:1`. So the proxy will not fail to start — it will hold a provider it did not hold before. Worth stating explicitly in the body, since it lands on a shipped path.

### P2 — the `CACHE_TYPE` shim is load-bearing, not cosmetic, and carries an unstated ordering invariant

`cli/commands/serve/proxy-extension-composition.ts:100` — `if (cacheType === "redis") setEnv("CACHE_TYPE", "extension");`

The comment frames this as rollout compatibility, which undersells it. Verified:

```
main:54     "CACHE_TYPE must be memory or extension"
branch:55   "CACHE_TYPE must be memory, extension, or redis"
```

But `src/proxy/cache/index.ts:135` and `src/proxy/cache/extension-store.ts:96` **still** throw `TypeError("CACHE_TYPE must be memory or extension")` and have no knowledge of `"redis"`. Without the mutation, `CACHE_TYPE=redis` passes composition validation and then dies downstream. It works only because activation completes before `import("veryfront/proxy/main")` in `runStandaloneProxyRuntime`.

**Concrete:** anyone who moves the proxy import ahead of activation gets `CACHE_TYPE=redis` failing with a message naming only `memory` and `extension` — an error pointing away from the actual cause. Two validators now disagree on the accepted value set, bridged by a mutable global, with no TODO or issue link marking the shim for removal.

### P3s

- **`compile-binary.ts:60` — the proxy binary is compiled `--allow-all`.** This is *not* a new over-permission: I confirmed `origin/main:scripts/build/compile-binary.ts:61` already uses `--allow-all` for the main CLI, so the proxy inherits an identical posture. Still the natural least-privilege opportunity — a single-purpose network-facing proxy is a far better candidate for scoped `--allow-net/--allow-env/--allow-read` than the general CLI, and a compromise currently implies filesystem write and subprocess spawn. Separate work, not a blocker.
- **`compile-binary.test.ts:165-177` string-matches YAML.** Asserting `workflow.includes("deno task build:proxy-lock")` pins presence, not effectiveness — moving the lock-freshness step into a job that does not run on release would keep every assertion green.
- **`proxy-runtime.ts:33-39` — `showProxyHeader` duplicates `showHeader`.** No version-drift bug: `cli/utils/index.ts:56` derives `VERSION` from `denoConfig.version` with the identical `"0.0.0"` fallback, so they are behaviorally identical. Presumably to keep the `#cli/utils` barrel out of the proxy graph — a legitimate reason worth a comment.
- **`cicd.yml:410` — redundant condition.** `(non-PR || same-repo) && is-PR` reduces to `same-repo && is-PR`. Correct, just hard to read; the body notes a shorter form was rejected by a hardening test, so this may be deliberate.

### Checked and cleared

- **Default bind `0.0.0.0`** is not new exposure — `origin/main:src/proxy/main.ts:138` already reads `getEnv("HOST") || "0.0.0.0"`.
- **npm dependency surface untouched.** No `package.json`, npm metadata, or dnt file in the diff — verified by path filter. This does not change `npm i veryfront` contents or add dependencies. The five extensions are embedded in a *binary*, not the published package.

### Test adequacy

The thing I most wanted to see is here: **the smoke test exercises the actual compiled artifact**, not the source. `scripts/build/smoke-proxy-binary.sh` boots the binary three times — memory, redis, observability — asserting `/_proxy/health` returns `"status":"ok"` plus `"TokenCacheStore registered"` and `"[otel] Initialized"` in logs, proving the extensions are genuinely embedded and activating. It runs in both the PR job and the release build. `compile-binary.test.ts:110-131` pins the proxy compile args with negative assertions. That is the right shape for a build-pipeline test.

Deducting for the YAML string-matching, and for linux-x64-only build/smoke — acceptable, since that is the only published target.

### Production risk & rollback

Low and unusually well contained. The artifact is additive; nothing consumes it until the veryfront-server image is repointed, which you have sequenced behind two other merges. Main CLI, homebrew formula, install script, and npm package are provably unaffected. The realistic exposure is the ext-redis scope change for operators with an ambient `REDIS_URL`.

Rollback is clean: revert and the matrix loses one entry. No persisted format, no migration. A published `veryfront-proxy-linux-x64` asset would remain attached to any release already cut, but nothing resolves it by name except the server image you control.

### Recommendation

Keep it in draft as you intend. Before un-drafting: state the ext-redis activation-scope change in the body (it lands on a shipped path), and add a removal marker or tracking issue on the `CACHE_TYPE` translation shim so it does not outlive the rollout it was written for. Neither needs new code.

kojiwakayama:
Pushed exact head `e99381dda58592dc097cd00ae5e52a48dc5c07b9` after reconciling this branch with the released main baseline.

Addressed review points:
- The proxy smoke check now retries for the asynchronous provider startup marker after health succeeds.
- `build:proxy-lock` no longer carries the redundant `--frozen=false`.
- SBOM guidance now describes the selected lockfile instead of hard-coding `deno.lock`.
- The proxy-specific lock remains frozen and now includes the reconciled pinned dependency.

Verification:
- compile-binary suite: 11/11 passed.
- proxy runtime/composition suites: 10/10 steps passed.
- `bash -n`, targeted `deno check`, and `git diff --check`: passed.
- The first pre-push run before the main merge passed 3,719 tests / 26,698 steps. Two exact-head pre-push retries later hit only the unrelated parallel `skills info` exit-code assertion; that test passes 11/11 steps when rerun in isolation with the hook environment.

This PR remains draft and is not being queued: the staging proxy rollout is still blocked by the new image replica restarting/OOMing while the old replicas remain healthy.

kojiwakayama:
Addressed the latest Copilot suppressed review note on the standalone proxy extension composition header.

Change pushed at `0a98e7addf2984b48e167dc94b945595295d52c1`:
- `cli/commands/serve/proxy-extension-composition.ts` now documents both extension-loader contracts consumed by the standalone proxy path: `TokenCacheStore` and the Redis runtime provider.
- Runtime behavior is unchanged; this is a documentation/comment-only correction for the provider activation path.

Verification before push:
- `npx --yes deno@2.7.7 fmt --check cli/commands/serve/proxy-extension-composition.ts` passed.
- `git diff --check` passed.
- Pre-push hook passed: format, lint, typecheck, generation, and full no-check Deno suite, `3724 passed (26745 steps), 0 failed`.

Merge readiness: not scheduling this PR for merge yet. The latest head has fresh CI queued, the PR is still draft, and the previously noted staging proxy rollout/OOM gate remains external to this repository. Because of those gates my merge confidence is below 90% even though the code/comment review item is addressed.

kojiwakayama:
Returned to draft: the PR body's rollout gate still applies (staging proxy rollout must be stable, and the veryfront-server consumer + child durable-event consumer PRs must merge first). Conflict resolution and CI-greening from today stand — content is merge-ready once the gate clears.

kwakayama:
## Re-review at `07cd1f49e`: 88/100 (was 82) — does not clear 90

| Axis | Score |
|---|---|
| Correctness | 37/40 |
| Test adequacy | 21/25 |
| Security / prod-safety | 17/20 |
| Maintainability | 13/15 |
| **Total** | **88/100** |

Merge base `0a9f224cd`. Conflicts resolved via merge commit `0c30c9c85`; three-dot diff is clean at +2331/-112 across 14 files. All 29 checks green including `tests-proxy-binary`, `tests (binary e2e)`, `coverage gate`, CodeQL.

**Status note:** the PR is still marked **draft** (`isDraft: true` as of this comment), so the rollout gate remains mechanically enforced. Worth stating because reviews do not block here — `required_approving_review_count: 0` — so the draft flag is the only thing holding it.

### Both requested items — done, and done well

**1. ext-redis activation scope is now stated in the body**, under Compatibility and explicitly marked "behavior change on the shipped proxy path", with accurate mechanics: `CACHE_TYPE=memory` previously loaded zero extensions even with an ambient `REDIS_URL`; now such operators get a registered, lazily-connected, never-eagerly-dialed Redis runtime provider. That is exactly the disclosure that was missing.

**2. The `CACHE_TYPE` shim is marked for removal — better than asked.** `proxy-extension-composition.ts:100` now carries `TODO(#3214): remove this rollout shim once the hosted charts stop passing CACHE_TYPE=redis`, plus a comment naming both downstream validators (`src/proxy/cache/index.ts`, `extension-store.ts`) and stating the previously-unstated ordering invariant: "this translation must complete before the proxy runtime is imported."

The invariant is now enforced structurally rather than only documented — `runStandaloneProxyRuntime` does activate → registerTeardown → `import("veryfront/proxy/main")` in one shared function used by both entrypoints. Commit `07cd1f49e` also improves the error to `CACHE_TYPE must be memory, extension, or redis (received "<value>")`.

### Cross-repo verification — this does resolve the downstream blocker

- Release target is `gh release create --repo veryfront/veryfront` (`cicd.yml:588`, `:739`) — matches where veryfront-server's `resolve-framework-assets.sh` looks.
- Asset name `veryfront-proxy-linux-x64` exactly matches what the server's `proxy_ready` gate requires.
- The release glob `for f in binaries/veryfront-*/veryfront-*` picks it up automatically.
- SBOM `proxy.json` lands in `dist/sbom-${VERSION}/` and is swept by `*.json`.

### F1 (P2) — a proxy-leg failure now cancels the other binary builds and blocks the whole release

`.github/workflows/cicd.yml:434-435`. **Verified:** the `build-binaries` `strategy: matrix:` block has **no `fail-fast:` line**, so it defaults to `true` — while the two other matrix jobs in the same file (lines 35 and 77) both set `fail-fast: false` explicitly. That is a live inconsistency, not house style.

This PR adds two *new* failure modes to that matrix that did not exist before:
- "Verify proxy dependency lock is current" — regenerates the lock and `git diff --exit-code`, i.e. network-dependent resolution at release time.
- "Smoke test proxy binary" — starts a server, polls `/_proxy/health`, greps logs, on a 30-second timeout.

**Concrete:** a slow ubuntu runner makes the compiled proxy take >30 s to bind; `run_smoke` returns 1; the proxy leg fails; fail-fast cancels the in-flight macOS/Linux legs; `build-binaries` fails; the `release` jobs (`needs: [..., build-binaries, ...]`, lines 509 and 630) never run. The version bump is already on main at that point, so the release must be re-driven by hand.

`continue-on-error: ${{ contains(matrix.os, 'windows') }}` shows the intent was already there — but it only stops Windows from *triggering* fail-fast; it does not protect the other legs from the new proxy leg.

**Fix is one line: `fail-fast: false` on that matrix.** This alone would put the PR at 90.

### F2 (P3) — the native-Promise capture is weaker than claimed in the dedicated binary

`cli/proxy-main.ts:6-10` statically imports the five extension modules *first*, then dynamically imports `proxy-runtime.ts`, which captures `const NativePromise = Promise` at module scope. ES static imports evaluate in source order, so extension top-level code runs **before** the capture. The comment claims extension code "cannot replace Promise species" — that holds for activation-time mutation, and does hold on the universal CLI path where `proxy-runtime.ts` is statically imported, but not for top-level mutation in the dedicated binary. Low realism (five first-party `@veryfront/ext-*` modules), cheap fix: import the pinning module before the extension anchors.

### F3 (P3) — nothing verifies the size/memory goal that motivates the PR

The stated purpose is that the universal binary exceeds the staging proxy's 1536 MiB limit. No check asserts a size ceiling or RSS bound on the produced artifact, so future dependency creep silently re-inflates the graph and the regression surfaces as another staging OOM. A size assertion in `smoke-proxy-binary.sh` or the compile test would pin the actual goal.

### F4 (P3) — the smoke test misses the config that changed

`smoke-proxy-binary.sh` covers memory (no `REDIS_URL`), `CACHE_TYPE=redis` + `REDIS_URL`, and memory + otel/sentry. The newly-disclosed combination — `CACHE_TYPE=memory` **with** ambient `REDIS_URL` — is covered by the unit test (`proxy-extension-composition.test.ts:73`) but not on the compiled artifact. One extra `run_smoke` line closes it.

### F5/F6 (P3) — minor

`runProxy` now calls the local `showProxyHeader()` instead of `showHeader()`; the JSON-mode suppression is an improvement, but the body's claim that the PR "preserves universal CLI behavior" is no longer quite literal. And the Evidence block cites CI running for `48298573c` while head is `07cd1f49e`.

### Re-checked, none moved

`--allow-all` (`compile-binary.ts:57`) is still shared by both profiles and still not a regression — main already used it for the universal CLI. `install.sh` and the homebrew formula are not in the 14-file diff at all, so no shadowing. Lock freshness is fail-closed in both the PR job and the release matrix. `DEFAULT_INCLUDES` is untouched with every existing matrix entry pinned to `profile: full`, so the main CLI binary is provably unchanged. No npm dependency surface change. The fork guard reduces correctly to same-repo PRs only.

**Test adequacy is genuinely strong where it counts.** The composition matrix asserts memory/no-redis → `null`; `extension`|`redis` + `REDIS_URL` → both `TokenCacheStore` and `RedisRuntimeProvider` registered, `CACHE_TYPE` translated, cache borrows the store, full teardown deregisters both; memory + `REDIS_URL` → ext-redis only. That directly answers whether the two Redis extensions collide in the production config — they do not, and it is pinned. The redis smoke case uses an unreachable `REDIS_URL=redis://127.0.0.1:1`, a neat proof of the lazy-connect claim on the real artifact. `compile-binary.test.ts` is now registered in `test:scripts`, closing a real gap.

### On the rollout gate

The gate language is still in the body and still describes the blocking condition as current. The draft flag is still set, so it is enforced — but note it is the *only* thing enforcing it, since reviews do not block and the queue takes ALLGREEN entries.

**One tension worth resolving before un-drafting.** Per veryfront-server#278, staging has received nothing since 2026-07-31 because `proxy_ready` is false — and it is false precisely because `veryfront-proxy-linux-x64` has never been published, which is what this PR publishes. The OOM the gate cites is the problem the dedicated binary exists to fix. So "wait until the staging proxy rollout is stable" may be unsatisfiable without first shipping this artifact. The body's rollout order also reads as partly executed already, since the server-side guard landed 2026-08-02.

I am not making that call — it depends on staging facts I did not verify. But someone should reconcile whether the gate text is current or stale.

### Production risk & rollback

Release-path risk is F1 (recoverable — re-run the release). Runtime risk is confined to `veryfront serve --mode=proxy` and the new binary; the five existing CLI artifacts are provably unchanged. The behavior change is additive, and the smoke test proves an unreachable `REDIS_URL` does not prevent startup. Rollback is a plain revert — the artifact is additive, so reverting simply stops publishing it, returning `proxy_ready` to false, i.e. the current state.

### Recommendation

Fix F1 before merge — one line, and it protects the release path this PR touches. F2 is cheap while the file is open. F1 alone puts this at 90.

kojiwakayama:
Pushed follow-up at `3192ceabc59356f82b6b35002aaa16abfad1181e` for the 88/100 re-review and latest suppressed Copilot note.

Addressed:
- `build-binaries` now sets `strategy.fail-fast: false`, so the new proxy leg cannot cancel existing binary legs.
- `cli/proxy-main.ts` imports `proxy-runtime.ts` before extension anchors, so native Promise capture evaluates before extension top-level code in the dedicated binary.
- `smoke-proxy-binary.sh` now bounds each curl probe with `--connect-timeout 1 --max-time 2`.
- the smoke script enforces `PROXY_BINARY_MAX_BYTES` default `188743680` bytes (180 MiB). I did not add RSS gating because local/CI RSS is nondeterministic; artifact bytes are deterministic and the compiled graph is the regression target.
- compiled smoke now covers `CACHE_TYPE=memory` with ambient `REDIS_URL` and waits for `[ext-redis] RedisRuntimeProvider registered`.
- temp log directory now uses a portable `mktemp` template.

Verification:
- RED: compile-binary test failed before implementation on missing runtime import ordering, missing fail-fast false, and missing ambient Redis smoke.
- compile-binary test: 11 passed.
- proxy runtime/composition tests: 2 files, 10 steps, 0 failed.
- bash syntax, TS fmt/lint/check, diff check: passed.
- Linux proxy compile passed; embedded files 75.17 MB under 180 MiB ceiling. Linux execution is not possible on the macOS host.
- Native proxy compile passed; updated smoke script passed across memory, redis, ambient Redis, and observability modes.
- normal pre-push gate initially hit a transient/unrelated `src/proxy/asset-handler.test.ts` follower timeout; the exact test file then passed in isolation, and the full normal pre-push retry passed with 3,724 tests, 26,812 steps, 0 failed.

PR remains draft and unscheduled; rollout gate still applies.

kojiwakayama:
Pushed SBOM lock-argument follow-up at `e2a93fc70189ad3269b04cf7a4a76a7382d308d4`.

Addressed:
- `scripts/build/generate-sbom.ts` now validates `--lock` as a non-empty path before reading `deno.json` or the lockfile.
- `scripts/build/generate-sbom.test.ts` runs the real CLI with `--lock` and verifies it exits through the usage-error path (`2`) with a clear `--lock requires a non-empty path` message and usage text.

Verification:
- RED: focused SBOM test failed before implementation because the script exited `1` through the generic empty-path runtime error path.
- GREEN: `npx --yes deno@2.7.7 test --config=scripts/test.deno.json --no-check --allow-read --allow-write --allow-run scripts/build/generate-sbom.test.ts` -> 4 passed, 14 steps, 0 failed.
- `npx --yes deno@2.7.7 fmt --config=scripts/test.deno.json --check scripts/build/generate-sbom.ts scripts/build/generate-sbom.test.ts` -> checked 2 files.
- `npx --yes deno@2.7.7 lint --config=scripts/test.deno.json scripts/build/generate-sbom.ts scripts/build/generate-sbom.test.ts` -> checked 2 files.
- `npx --yes deno@2.7.7 check --config=scripts/test.deno.json scripts/build/generate-sbom.ts scripts/build/generate-sbom.test.ts` -> both files checked.
- `git diff --check` -> passed.
- normal pre-push gate -> format, lint, typecheck, and full unit suite passed: 3,724 tests, 26,812 steps, 0 failed.

PR remains draft and unscheduled; rollout gate still applies.

kojiwakayama:
Exact-head draft audit at e2a93fc70189ad3269b04cf7a4a76a7382d308d4: code confidence is 92%. The head and conflict-free synthetic merge against main f56afcd42 both pass focused tests (17 tests / 24 nested steps), verify:quick, proxy-lock regeneration, native compile/smoke, and the 180 MiB binary ceiling. Keep this draft for now: the PR body still records a failed staging rollout, and its statement that Redis is never eagerly dialed is too broad. Registration is lazy, but production opens two Redis connections when REDIS_URL and the routing-invalidation secret are present; the ambient smoke omits that secret and runs in development. Before marking ready, require a fresh successful staging rollout (or a disposable-Redis production-mode smoke) and correct that wording.

kojiwakayama:
Current-main draft recheck at `e2a93fc70189ad3269b04cf7a4a76a7382d308d4`:

- Synthetic merge with current main `5d107f1fb66d9c9ed36b7da14e4d933e2868469f` is conflict-free; the changed-file sets do not overlap.
- `deno task build:proxy-lock` is byte-stable on that synthetic merge.
- `deno task verify:quick` passes on the synthetic merge.
- Exact hosted checks remain terminal green and the PR has zero unresolved threads.

The code remains technically strong, but I am keeping the PR draft. The remaining gate is runtime evidence, not repository drift: the registration path itself is lazy, but production proxy startup does open the routing-invalidation Redis clients when `REDIS_URL`, a valid routing-invalidation secret, and production replica configuration are present. Before marking ready, we still need either a successful staging rollout using the dedicated artifact or a production-mode compiled-binary smoke against a disposable Redis instance. The Compatibility wording should be narrowed accordingly when that gate is closed.

kojiwakayama:
Addressed the remaining suppressed review guidance on exact head `52c6c6cdbd8d59d2a1c8f29858017d13d9acbafc`.

- Removed the unrelated `#3214` reference from the `CACHE_TYPE=redis` compatibility-shim removal marker while retaining the exact operational removal condition.
- Corrected the PR Compatibility section: Redis provider registration is lazy, but production startup can open routing-invalidation Redis connections when `REDIS_URL`, a valid routing-invalidation secret, and production replica configuration are present. The ambient development smoke does not claim to exercise that production connection path.
- Updated stale exact-head evidence in the PR body.

Verification:
- Deno 2.7.7 format check and `git diff --check`: passed.
- Normal pre-push gate: 3,724 tests, 26,812 steps, 0 failed.

The PR remains draft and unscheduled. Repository review confidence is above 90%, but the explicit staging rollout gate still lacks successful runtime evidence, so merge confidence is not above 90% yet.

kojiwakayama:
Merge confidence: 93%.

Reasoning:
- Current head is `52c6c6cdbd8d59d2a1c8f29858017d13d9acbafc` and GitHub reports `mergeStateStatus: CLEAN`.
- There are no unresolved review threads.
- Hosted checks are green on the exact head, including format, lint, typecheck, integration tests, all coverage shards/gate, RSC browser e2e, binary e2e, npm install smoke, sentry runtime packages, CodeQL, npm audit, and the new `tests-proxy-binary` job.
- The diff is concentrated on the dedicated proxy binary path: separate proxy entrypoint, proxy lock generation, compile profile, proxy smoke script, SBOM addition, and shared proxy runtime startup. The universal CLI binary path remains present and the workflow matrix keeps the existing binary legs while adding the proxy leg with `fail-fast: false`.
- Residual risk is mostly rollout/runtime-environment risk for the new artifact, not code correctness: production Redis and staging memory behavior still depend on the downstream image rollout described in the PR body. The PR includes artifact-size gating, lock freshness checks, and smoke coverage for memory, Redis, ambient Redis, OpenTelemetry, and Sentry paths, which keeps that risk below the merge threshold.

Scheduling exact-head auto-merge now.

kojiwakayama:
Merge readiness review for exact head `52c6c6cdbd8d59d2a1c8f29858017d13d9acbafc`.

Merge confidence: 92%.

Reasoning:
- Review-thread audit is clean: 0 unresolved threads.
- Hosted checks are green and the PR is `CLEAN` on this exact head.
- The prior runtime gate is now closed with a production-mode compiled proxy smoke against disposable Redis. The native compiled proxy reported `[ext-redis] RedisRuntimeProvider registered`, `Proxy routing invalidation bus connected`, and `/_proxy/health` returned ok with `NODE_ENV=production`, `PROXY_MODE=1`, `REDIS_URL`, `VERYFRONT_PROXY_EXPECTED_REPLICAS=1`, and a valid-length routing invalidation secret.
- Existing PR evidence remains relevant: Linux proxy compile embedded files were 75.17 MB under the 180 MiB ceiling; focused compile/proxy tests passed; format, lint, typecheck, and diff validation passed; normal pre-push gate passed with 3,724 tests / 26,812 steps / 0 failures.

Residual risk is rollout-order risk, not code-readiness risk: the downstream server image still needs to consume the published patch artifact and digest after release. That is covered by the PR rollout order and does not require holding this framework artifact PR.

kojiwakayama:
Merge confidence: 92%.

Reasoning: #3280 has no unresolved review threads, all hosted checks are green on the current head `0a4b9502821a261ea63280c0b9d1be8019e5cb0f`, and the latest branch update appears bounded to the proxy binary follow-up surface already under review. I am scheduling with an exact head match so it will not merge if the branch changes again.

kojiwakayama:
Merge confidence: 92% for exact head `0a4b9502821a261ea63280c0b9d1be8019e5cb0f`.

Reasoning:
- GitHub reports this head as `CLEAN` with no pending or failed hosted checks.
- There are no unresolved review threads, and the latest Copilot review produced no actionable comments beyond the large-diff limit notice.
- The later commits after the earlier 92-93% confidence notes are rollout/documentation and proxy-safety follow-ups: native Promise capture ordering, proxy binary size ceiling, production-mode Redis smoke coverage, fail-fast false for the binary matrix, SBOM lock argument validation, and corrected rollout guidance.
- Exact-head hosted CI covers the release-sensitive surfaces: proxy binary build/smoke, format, lint, typecheck, unit, integration, coverage shards/gate, RSC browser e2e, binary e2e, npm smoke, Sentry runtime packages, CodeQL, CLA, and CodeRabbit.

Residual risk is rollout-order risk for downstream consumers of the newly published proxy artifact, not repository code readiness. The PR is additive to the dedicated proxy binary/release path and does not change the existing universal CLI binary artifact selection. This clears the strictly-above-90% merge threshold, so I am scheduling this exact head for merge.

kojiwakayama:
Removed exact head `0a4b9502821a261ea63280c0b9d1be8019e5cb0f` from merge-queue position 1 because this current head has not completed the repository\u2019s >90% merge-confidence gate. The prior staging rollout/OOM/ProgressDeadlineExceeded evidence and older code review do not establish release readiness for this descendant. No source changes were made. I am holding it for an exact-head review of the dedicated proxy binary, release packaging, resource/rollout behavior, hosted checks, and operational rollback evidence before it is scheduled again.

kojiwakayama:
Merge confidence: 92% for exact head `0a4b9502821a261ea63280c0b9d1be8019e5cb0f`.

Reasoning:
- GitHub reports this head as `CLEAN`, non-draft, with no unresolved review threads and hosted checks green, including the dedicated `tests-proxy-binary` job, format, lint, typecheck, coverage, integration, npm smoke, CodeQL, and NPM audit.
- I re-audited the current exact branch after the prior queue hold. The release-sensitive surfaces are bounded: a dedicated proxy entrypoint, graph-specific frozen proxy lock, x64 and arm64 proxy release assets, proxy SBOM generation from the proxy lock, provider/runtime smoke coverage, and a 1536 MiB cgroup memory gate for x64 cold starts.
- Local exact-head verification passed: CLI proxy composition/runtime tests (2 files, 11 steps), build/SBOM tests (17 tests / 14 nested steps), TypeScript checks for CLI proxy and build/SBOM surfaces, `bash -n` for both proxy smoke scripts, focused source/script `deno fmt --check`, and `git diff --check origin/main...HEAD`.
- The only manual formatting warning observed was `.github/workflows/cicd.yml` under raw `deno fmt --check`; the same file fails identically on `origin/main`, so that is an existing baseline and not a #3280 regression. Hosted `ci (format)` is green on this PR head.
- The remaining risk is operational rollout, not repository merge correctness: staging still needs to consume the new proxy artifact and record memory after deployment. The PR keeps that rollout guidance explicit, preserves universal binary assets, and includes fallback behavior for compatibility.

I am scheduling with `--match-head-commit` so it will not merge if the branch changes.

kojiwakayama:
Merge confidence: **93%** at exact head `0a4b9502821a261ea63280c0b9d1be8019e5cb0f`.

Reasoning:

- GitHub reports the PR `CLEAN`; all 26 applicable hosted checks passed and 7 release-only checks were expected skips.
- The complete review surface is clear: 20 issue comments, 13 review bodies including 17 suppressed findings, no inline comments, and 0 unresolved threads. Every historical finding is addressed or has repository-specific disposition evidence.
- The authoritative exact-head/current-main diff is 18 files. Focused build, SBOM, proxy runtime, format, lint, typecheck, Bash syntax, and diff checks passed.
- The exact hosted x64 artifact stayed below the 180 MiB executable ceiling and passed 3 of 3 cold starts in a 1536 MiB cgroup, with successful health checks, clean shutdown, and `OOMKilled=false`.
- The server artifact resolver, durable child consumer, runtime selector, and temporary memory envelope prerequisites are merged.

Residual risk is the first real staging observation after the patch release, plus the mutable test-container base image and GitHub's inflated cached diff. Staging is causally post-merge because this PR produces the release asset; the PR body now records explicit digest, routing, health, restart/OOM, memory, and rollback checks. Those bounded, monitored residuals keep confidence above the requested 90% threshold.

kojiwakayama:
Merge confidence: 94% on exact head `0a4b9502821a261ea63280c0b9d1be8019e5cb0f`.

Reasoning: the PR is currently `CLEAN`, has no pending or failed status checks, and has 0 unresolved non-outdated review threads. The change is scoped to publishing the dedicated proxy binary, and the current hosted checks have completed successfully on this head. Residual risk is limited to release packaging behavior, but the CI signal is green and there are no outstanding review blockers. Scheduling for merge queue with exact-head protection.

kojiwakayama:
Merge confidence remains 92% for exact head `0a4b9502821a`.

Reasoning: the dedicated proxy binary change has full hosted CI green on this head (format, lint, typecheck, unit, integration, coverage shards, RSC browser e2e, binary e2e, npm install smoke, sentry runtime packages, proxy binary tests, coverage gate, CodeQL). There are no unresolved review threads. Residual risk is limited to release/packaging edge cases outside the exercised matrix, but the branch-specific proxy binary checks and install smoke materially cover the changed surface.

Scheduling this exact head for merge queue.

PR #3311, opened 2026-08-03
fix(cli): harden deploy polling and route verification

kojiwakayama:
## Summary

- Harden release asset polling so only missing, queued, and building states continue; invalid, partial, failed, superseded, or mismatched ready responses fail closed.
- Parse ready manifests against the current schema and requested release before deployment.
- Collect expected page routes before creating a release and fail closed when route directories cannot be inspected for reasons other than not-found.
- Preserve the failing-capable semantics of main's bounded release-source polling regression (fixed 20-read budget, red assertion on shortfall) instead of porting the reconcile branch's unbounded fake-time loop. The fake-clock scheduling guard is wider than main's (60 ticks instead of 40) — see review follow-up below.

## Intentional behavior changes

Route discovery now aborts deployment on permission, symlink-loop, and other non-not-found filesystem errors. Main swallowed those failures, which could reduce expected routes to an empty set and let route-coverage verification pass vacuously. The failure is deliberate, occurs before release or deployment mutation, and exits non-zero.

A successful null control-plane response is invalid. HTTP 404 remains the only missing-manifest signal.

Transient control-plane failures during manifest polling no longer abort the deploy. Gateway statuses (HTTP 502, 503, 504) and retryable connection errors (for example ECONNRESET) are retried until the existing polling deadline, so a single blip cannot fail an otherwise healthy deploy. A structured HTTP status is authoritative: authentication, validation, and other client failures (401, 403, 422, ...) and explicit cancellation (AbortError) still abort on the first attempt, even when a connection-like cause is attached. If the deadline expires during an outage, the timeout message reports the last control-plane failure alongside the last observed manifest state. The same conservative classification (`isRetryableApiReadError`) now also drives the shared API client's idempotent-read retry, so a client failure with a connection-like cause is no longer retried there either.

## Review follow-up

- Added a sequenced polling regression for missing -> building -> ready and asserted the exact read count.
- Added a deadline regression that verifies repeated building polls and the reported last state.
- Replaced five hard-coded manifest schema versions with RELEASE_ASSET_MANIFEST_SCHEMA_VERSION.
- Pinned the route inspection test to the underlying unexpected-NUL filesystem error.
- Documented the null polling sentinel without the ineffective unknown | null type union.
- Corrected the title: the bounded release-source test already existed on main and its failing-capable shape is kept here, not restored.
- Added transient-error retry to release asset polling (the one in-scope gap from review), classified by the shared `isRetryableApiReadError` helper rather than a duplicated local predicate. Covered by four regressions: recovery across 503 -> connection reset -> building -> ready with the exact read count; persistent 503 bounded by the original deadline with a `last control-plane failure: HTTP 503` message; 401/422/AbortError each failing on the first read without another polling attempt; and a unit test pinning the classifier itself.
- Deduplicated the untrusted manifest property read: the polling loop now reuses `readUntrustedOwnDataProperty` and `isSafeBoundedText` from the release-assets manifest schema instead of a local re-implementation, and the control-plane empty successful body has its own error message ("returned an empty manifest response") so it no longer shares wording with the polling loop's malformed-state failure.
- Tick budget correction: the bounded release-source test's fake-clock guard is `tick < 60`, not main's `tick < 40`, so the earlier "byte-identical to main" claim no longer holds and is withdrawn. The read budget itself is unchanged at exactly 20 and the shortfall assertion still goes red, so the test remains failing-capable. The guard was widened because the deploy flow now performs more pre-mutation verification before this poll starts; under parallel load the async chain reproduced only 19/20 reads within 40 ticks. The in-test comment documents this.

## Deliberately not ported

- The reconcile branch's unbounded release-source polling test loop.
- Renderer, documentation, module-layout, or control-plane authentication changes outside deploy polling.
- Deletion of main's route-coverage tests.

## Verification

- Focused suites at the current head (deploy-project, control-plane, shared config, deploy command integration): 0 failures, including the four transient-retry regressions.
- deno check, deno lint, deno fmt --check on all touched files: passed.
- Pre-push gate (format, lint, typecheck, full unit suite): passed on push.
- git diff --check: passed.


kwakayama (commented):
## Critical Review — Score: 88/100

### Verdict
This is a careful, well-tested hardening of the deploy polling path. The state machine in `waitForReleaseAssetManifest` now fails closed on every non-progress state, the deadline/backoff logic is correct (final poll at the deadline, sleep clamped to `min(pollIntervalMs, remainingMs)`, no off-by-one), and the new tests are genuinely failing-capable — each asserts a specific error message via `assertRejects` under a bounded 100 ms budget, so a regression produces a different message or a fast timeout, not a hang. Remaining issues are minor: rubric band 75–89, safe to merge after (or even without) small fixes.

### Findings

1. **[minor] `superseded` is reported as a build failure.** In `cli/shared/deployment/deploy-project.ts`, `if (state === "failed" || state === "superseded") throw new Error(\`Release asset build failed for release ${releaseId}\`)` conflates two different terminal conditions. A superseded manifest means a newer build replaced this release — telling the user the build "failed" points them at the wrong remediation (the `partial` branch, by contrast, gets its own actionable message). Suggest a dedicated message for `superseded`.

2. **[minor] `Promise<unknown | null>` erases the absence contract.** In `cli/shared/deployment/control-plane.ts`, `getReleaseAssetManifest` now returns `Promise<unknown | null>`, which TypeScript collapses to `Promise<unknown>` — the "null means 404 / not built yet" convention that `waitForReleaseAssetManifest` relies on (`raw !== null`) is no longer visible in the type system (same for `Array<unknown | null>` in `cli/test-utils/deploy-test-support.ts`). A small discriminated result type (e.g. `{ kind: "absent" } | { kind: "response"; body: unknown }`) or at least a doc comment on the interface member would keep the contract checkable. Related: a client that resolves `undefined` (e.g. an empty 204 body) slips past the `response === null` guard — it still fails closed downstream via the invalid-state check, but at the wrong layer with a confusing message.

3. **[minor] Duplicated helper and overloaded error message.** `readReleaseAssetResponseDataProperty` in `deploy-project.ts` reimplements `readOwnDataProperty` from `src/release-assets/manifest-schema.ts` (including the ad-hoc 64-char bound that mirrors `isSafeBoundedText`/`MAX_IDENTIFIER_LENGTH`); exporting and reusing the schema module's helper would keep the hardening rules in one place. Additionally, the exact string "returned an invalid state response" is now thrown for two distinct failure modes — a successful `null` body in `control-plane.ts` and a malformed/accessor-backed `state` in `deploy-project.ts` — which muddies field diagnostics and couples the control-plane test's message assertion to the polling loop's wording.

4. **[nit] Out-of-scope lint commit riding along.** The `stringifyJsonValue` removals in `extensions/ext-llm-anthropic/src/anthropic-request-builder.ts` and `extensions/ext-llm-openai/src/openai-responses-request-builder.ts` are unrelated to deploy polling. The PR body explains why (main fails pre-push lint) and that sibling PRs carry the identical commit — acceptable, but ideally this lands once as its own trivial PR so the slices don't each carry it.

5. **[nit] No overall abort for a hung poll request.** The deadline in `waitForReleaseAssetManifest` is only checked between polls; a `getReleaseAssetManifest` call that hangs longer than `timeoutMs` stalls the deploy indefinitely (no `AbortSignal`/per-request timeout). This is pre-existing behavior, not introduced here — noted for a follow-up, not against this PR.

### What's good
- The new tests earn their keep: the accessor test asserts `accessorCalls === 0` (proving the own-data-property read, not just the rejection), the route-inspection test asserts both the `TypeError` and that `createdReleases`/`createdDeployments` stay empty, and the control-plane test distinguishes 404 (null) from 403 (rethrow) with the same mock.
- The release-source polling test keeps main's bounded `for (tick < 40)` loop with a message-bearing `assertEquals(releaseSourceReads, 20, ...)` — it fails red on regression instead of hanging, exactly as the PR body claims; the branch's unbounded `while` neutering was correctly not ported.
- Moving `collectProjectPageRoutes` into `verify-source` means route-walk failures now surface before `createRelease`, so a bad config can no longer mutate remote state — and the swallowed-error `catch {}` is now scoped to `isNotFoundError` only.

🤖 Critical review by [Claude Code](https://claude.com/claude-code)

kwakayama (dismissed):
## Review: 78/100 — request changes, but a near miss

| Axis | Score |
|---|---|
| Correctness | 33/40 |
| Test adequacy | 17/25 |
| Security / prod-safety | 16/20 |
| Maintainability | 12/15 |
| **Total** | **78/100** |

Merge base `82af3669ffe3`, three-dot throughout.

This is materially stronger than most of the current batch. The central claim is true and verified, the new tests are genuinely failing-capable, and no false-success path is introduced. Two items keep it under 80 and both are small — items 2-4 below are minutes of work, and with item 1 this comfortably clears 90.

### Central claim verified — with one wording objection

The bounded loop already exists on main at `cli/commands/deploy/command.integration.test.ts:1065-1077`:

```ts
for (let tick = 0; releaseSourceReads < 20 && tick < 40; tick++) {
  await time.tickAsync(500);
}
assertEquals(releaseSourceReads, 20, "release-source polling did not exhaust its fixed read budget");
```

I diffed main's `1055-1082` against the branch's `1056-1083`: **byte-identical**. The branch's only changes to that file are manifest fixture updates across five fixtures.

The loop is genuinely failing-capable — if polling regresses and stops issuing reads, it exits on `tick < 40` and the `assertEquals` goes red. It cannot hang. The description's account of the reconcile branch's `while (releaseSourceReads < 20)` anti-pattern (spins forever on fake time, never a red assertion) is accurate.

**Objection: nothing is "restored."** Main was never broken; this PR declines to port a regression that only ever existed on the reconcile branch. The body is honest about this — the **title is not**. A reviewer skimming the title reasonably concludes main currently has a test that cannot fail. Please retitle to "preserve" or "decline to port".

Merge-gate claim also verified: comparing `it(`/`Deno.test` names across all three touched test files, **no main test name disappears**.

### Polling loop is sound

`deploy-project.ts:639-680`. Every exit is accounted for: `ready`+parses+routes-covered returns; `ready`+parse-fails, `partial`, `failed`/`superseded`, unknown state, non-string/empty/>64-char state, and deadline-passed all throw.

- **Cannot spin forever.** `deadline` is an absolute `Date.now() + timeoutMs` computed once; `wait(Math.min(pollIntervalMs, remainingMs))` is bounded; `remainingMs <= 0` throws.
- **Cannot report success on an indeterminate state**, and the success surface is **narrower** than main. Main returned `result` on `state === "ready"` with no parsing, so a legacy-schema or wrong-release manifest was deployed. That is the real fix here and it is correct.
- **Exit codes are clean.** New failure modes propagate through `execute()`; JSON mode catches at `command.ts:238` → `exitProcess(1)`, human mode rethrows at `:142-145`. A failed or timed-out deploy is distinguishable from success by exit status. **No false-success exit introduced.**

### P2 — route walk now aborts deploys on non-ENOENT filesystem errors

`cli/shared/deployment/deploy-project.ts:552-558`

```ts
} catch (error) {
  if (isNotFoundError(error)) return;
  throw error;   // was: catch { return; }
}
```

Confirmed `isNotFoundError` does not appear anywhere in this file on main, so the rethrow is genuinely new. It matches only `Deno.errors.NotFound` and `code === "ENOENT"` (walking `cause`); everything else propagates.

**Concrete failure:** a repo whose `app/` or `pages/` tree contains a directory the deploy user cannot read (root-owned mount, restrictive CI checkout) or a symlink loop. `fs.readDir` throws `EACCES`/`EPERM`/`ELOOP`. On main: swallowed, `expectedPageRoutes = []`, route-coverage check vacuous, **deploy succeeds**. After this PR: throws out of `verify-source`, **deploy fails**, exit 1.

Fail-closed is arguably the right direction, but this is a deploy-blocking behavior change presented as hardening with no migration note. Either narrow the rethrow to the errors you actually care about, or call it out in the description so an operator who hits it knows it is intentional.

### P2 — none of the nine new tests iterate the polling loop more than once

Every new test in `deploy-project.test.ts:1238-1380` uses `manifestControlPlane(response)` — a fixed response — and every one throws or returns on **iteration 1**. `polling = { pollIntervalMs: 100, timeoutMs: 100 }` guarantees the `wait()` at `:679` is never reached.

So in a PR titled "harden deploy polling" there is no test that `queued` → `building` → `ready` transitions across iterations, no test of the timeout path or its `last state: X` message, and **no test that `raw === null` keeps polling rather than failing**.

That last one matters most: `control-plane.ts:288-300` returning `null` on 404 is the "still building" signal, and nothing asserts the loop continues on it. A regression turning `null` into a throw would ship green.

Two small tests close this: a sequenced control plane returning `[null, {state:"building"}, readyManifest()]` asserting success plus a read count, and one returning `{state:"building"}` forever asserting the timeout message.

### P3s

- **`deploy-project.test.ts:429-448` cannot pin where its `TypeError` came from.** It writes `directories: { app: "app\0" }` and asserts `assertRejects(..., TypeError)` plus empty `createdReleases`/`createdDeployments`. A NUL byte can produce a `TypeError` from config parsing or path normalization *before* `collectProjectPageRoutes` is reached, and both assertions pass identically in that case. It does go red on main, so this is a precision weakness rather than a can't-fail test — but please assert on the message or path so it pins the origin.
- **Integration fixtures hardcode `schemaVersion: 2`** while `cli/test-utils/deploy-test-support.ts:142` uses `RELEASE_ASSET_MANIFEST_SCHEMA_VERSION`. When the schema bumps to 3 the helper follows automatically and the five fixtures silently become legacy payloads — which, given ready manifests are now strictly parsed, will fail with "invalid or mismatched ready manifest" and read as a real regression. Import the constant in both places.
- **`control-plane.ts:67` — `Promise<unknown | null>` collapses to `Promise<unknown>`.** In TypeScript `unknown | null` *is* `unknown`, so the `| null` documents nothing and the compiler no longer signals that `null` is the meaningful "not built yet" sentinel — which is exactly what `waitForReleaseAssetManifest` branches on. Prefer `Promise<unknown>` with a doc comment, or a named result type.
- **No retry on transient control-plane errors.** `getReleaseAssetManifest` rethrows every non-404, so a single 502/503 during a 120s poll aborts the deploy. Pre-existing, not a regression — but it is the most likely real-world cause of a spurious deploy failure, and this PR is the natural place to fix it.
- **Four new failure modes collapse to one JSON slug.** `command.ts:239-248` wraps non-`VeryfrontError` as `UNKNOWN_ERROR`/`RUNTIME_ERROR`, so partial / superseded / unsupported-state / invalid-state are distinguishable by free-text message only and CI automation must string-match. Pre-existing shape, worth a typed slug now the taxonomy has grown.

### Observability note — step attribution moved

`collectProjectPageRoutes` moved from `wait-release-assets` into `verify-source` (`deploy-project.ts:1075-1085`). `step()` emits `started` then `completed` and emits nothing on throw, so failures are inferred from "started with no completed". A route-walk failure now attributes to `verify-source` instead of `wait-release-assets`. Any dashboard or alert keyed on the old step name will misattribute. Small, but it is a log-shape change and the PR does not mention it.

### What is correct

- **Accessor-free state read** (`:613-621`) with `assertEquals(accessorCalls, 0)` — real hardening, properly tested, doubly failing-capable against main.
- **Reuses `parseReadyReleaseAssetManifestResponse`** (verified pre-existing at `src/release-assets/manifest-schema.ts:332`, unchanged). No new parser invented.
- **Route collection moved before `create-release`** — genuinely better, and the new test asserting `createdReleases`/`createdDeployments` stay empty pins it.
- **Every new test is failing-capable.** Traced each against main: legacy/mismatched manifests returned unparsed (no rejection → red); accessor state invoked the getter (both assertions red); `partial`/`superseded`/oversized/unknown polled to the generic timeout whose message does not match the asserted substring (red). Only "parses a valid ready response" is pure characterization. That is a notably better record than the rest of this batch.
- **The "Deliberately NOT ported" section** is the most useful part of the description — specific, checkable, and I confirmed the control-plane diff really is confined to `getReleaseAssetManifest`.

One weak spot: `control-plane.test.ts:57-59` uses bare `assertRejects(...)` with no error class or message, so it only asserts "something threw". Still red on the 403 regression, just imprecise.

### Production risk & rollback

**Rollback clean** — CLI-only, no persisted state, no migrations.

- **Highest risk:** the route-walk rethrow turning previously-succeeding deploys into hard failures on filesystem permission or symlink edge cases. Blast radius is CI deploy pipelines; the failure is loud and exit-1, so it surfaces immediately rather than silently.
- **Risk reduced:** deploying a legacy-schema or wrong-release asset manifest, which main did silently. A real production-correctness win.
- **No false-success risk introduced.**

### Minimum to reach merge-ready

1. Add the two loop-iteration tests (multi-state transition including a `null` poll; timeout path). This is the one gap that matters for a polling PR.
2. Retitle — "restore" is inaccurate; nothing on main was broken.
3. Narrow the route-walk rethrow, or document it as an intentional deploy-blocking change.
4. Import `RELEASE_ASSET_MANIFEST_SCHEMA_VERSION` in the integration fixtures.

kojiwakayama:
Addressed the change-request review in 301a49c19.

The branch now has failing-capable multi-iteration coverage for missing -> building -> ready and for the timeout/last-state path. Integration fixtures use the manifest schema constant, the route-walk error assertion pins its source, and the PR title/body now accurately describe the intentional fail-closed route behavior.

Fresh verification: focused deploy/control-plane tests passed (29 tests, 75 steps); verify:quick passed; test-typecheck baseline holds at 58/0; touched-file format/type checks and diff validation passed. Please rereview this exact head.

kojiwakayama:
Addressed the latest review feedback on deploy release-asset polling in `efdb97fb1fe760ea974630da3c665a872d94b412`.

Changes:
- `getReleaseAssetManifest()` now rejects both `null` and `undefined` successful bodies at the control-plane boundary, preserving `null` only as the 404 polling sentinel.
- The control-plane interface now exposes an explicit `DeployReleaseAssetManifestBody | null` contract instead of collapsed `unknown | null`.
- Superseded manifests now report a dedicated actionable message instead of being conflated with failed builds.
- Unsupported bounded states include the received state value in the error message.
- In-memory deploy fixtures keep `ReleaseAssetManifestResponse | null`, so malformed test fixtures are caught by typecheck.

Fresh verification:
- `npx --yes deno@2.7.7 test --no-check --allow-all --parallel cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` passed: 12 tests, 76 steps, 0 failed.
- `npx --yes deno@2.7.7 check cli/shared/deployment/control-plane.ts cli/shared/deployment/deploy-project.ts cli/test-utils/deploy-test-support.ts cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` passed.
- `npx --yes deno@2.7.7 fmt --check cli/shared/deployment/control-plane.ts cli/shared/deployment/deploy-project.ts cli/test-utils/deploy-test-support.ts cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` passed.
- `git diff --check HEAD~1..HEAD` passed.

Review-thread state: GitHub currently reports 0 unresolved review threads for this PR. Merge scheduling is intentionally held until exact-head CI and CodeRabbit finish and the stale changes-requested review gate is cleared.

kojiwakayama:
Addressed the remaining review-body comments on head `efdb97fb1fe760ea974630da3c665a872d94b412`.

Changes made:
- Treat both `null` and `undefined` successful asset-manifest responses as invalid control-plane bodies, while keeping HTTP 404 as the only polling absence sentinel.
- Made the asset-manifest polling sentinel visible in TypeScript with `DeployReleaseAssetManifestBody | null` instead of collapsed `unknown | null`.
- Split `superseded` from ordinary build failure with an actionable redeploy message.
- Included the bounded unsupported state value in the unsupported-state diagnostic.
- Tightened the in-memory deploy test control plane to `ReleaseAssetManifestResponse | null`.

Verification:
- Red first: the new 204/`undefined`, superseded, and unsupported-state assertions failed before the implementation patch.
- `npx --yes deno@2.7.7 test --no-check --allow-all --parallel cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` -> 12 passed, 76 steps.
- `npx --yes deno@2.7.7 check cli/shared/deployment/control-plane.ts cli/shared/deployment/deploy-project.ts cli/test-utils/deploy-test-support.ts cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` -> passed.
- `npx --yes deno@2.7.7 fmt --check cli/shared/deployment/control-plane.ts cli/shared/deployment/deploy-project.ts cli/test-utils/deploy-test-support.ts cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` -> passed.
- Pre-push hook -> format, lint, typecheck, and unit tests passed: 3653 tests, 26678 steps.

Remote CI is still running for this exact head, so I am not scheduling or merging this PR yet.

kojiwakayama:
Pushed the deterministic deploy polling regression follow-up at exact head 1b9c35be3. The test still requires all twenty canonical production read-back calls; it only widens the fake-clock scheduling guard from forty to sixty ticks so parallel repository load cannot expire the harness before the behavioral assertion.

Verification: the complete deploy command integration file passed, 17 tests and 0 failures. The normal pre-push gate then passed format, lint, typecheck, generation, and the full unit suite: 3,653 passed, 26,678 steps, 0 failures. The exact-head GitHub CI matrix is now running.

kojiwakayama:
Merge confidence: 94% for exact head 1b9c35be32724d17f11635fa32681a1a73dead71.

Reasoning: independent exact-head review found no critical, high, medium, or low issues and verified every earlier change request. Missing-to-building-to-ready polling and timeout last-state behavior are regression-covered; null remains the explicit 404-only sentinel while successful null/undefined bodies fail at the control-plane boundary; superseded and unsupported states are diagnosable; route inspection failures occur before remote mutation and are documented; fixtures use the schema constant; and the production read-back test retains its twenty-read assertion with a deterministic fake-clock guard.

All applicable hosted checks are green, including format, lint, typecheck, eight coverage shards and gate, integration, binary e2e, RSC browser e2e, npm smoke, CodeQL, CLA, and sentry runtime packages. Local focused review passed 29 tests (76 steps), deno check, lint, and diff checks. Both review threads are resolved/outdated, and the stale CHANGES_REQUESTED review was dismissed only after exact-head verification. Residual risk is the intentional fail-closed deploy behavior for non-not-found route-walk errors; it is documented, pre-mutation, and exits non-zero. This clears the strictly-above-90% merge threshold.

kwakayama (dismissed):
## Re-review at `1b9c35be3`: 87/100 (was 78) — does not clear 90

| Axis | Was | Now |
|---|---|---|
| Correctness | 33/40 | **36/40** |
| Test adequacy | 17/25 | **22/25** |
| Security / prod-safety | 16/20 | **17/20** |
| Maintainability | 12/15 | 12/15 |
| **Total** | **78** | **87** |

Merge base unchanged at `82af3669f`. Delta from the previously reviewed head is +99/-16 across 6 files, diffed old-head→new-head directly rather than assumed.

**Stating this plainly since this PR is currently in the merge queue: 87, not 90.** Every substantive item from the last review is closed and the test work is genuinely good. Three residual items remain, all small — but one is a body/code mismatch introduced *inside the very change that answers my previous objection about an inaccurate claim*, so I am not rounding past it.

### All four prior items — closed

**1. Loop-iteration tests — done, and done well.** I hand-verified the arithmetic rather than trusting the assertions.

`deploy-project.test.ts:1264` — responses `[null, {state:"building"}, readyManifest()]`, `timeoutMs: 500`, `pollIntervalMs: 100`. Trace: read#1 `null` → wait; read#2 `building` → wait; read#3 `ready` → return. Asserts `state === "ready"` **and `reads === 3`**. That covers the 404 "still building" sentinel, the `building` state, and multi-iteration in one test — exactly the gap.

`deploy-project.test.ts:1291` — `building` forever, `timeoutMs: 250`. Trace: reads at remaining 250/150/50/0 → 4 reads, then throw. Asserts `"last state: building"` **and `reads === 4`**.

Both failing-capable: if the loop stopped continuing on `null` the first goes red; if the deadline never fired the second's `assertRejects` fails. Tick budgets match the traces exactly.

**2. Retitle — done.** The body also states it directly: the bounded release-source test "already existed on main and is preserved here, not restored."

**3. Route-walk rethrow — done via documentation**, the second option I offered. `deploy-project.ts:552-558` is byte-identical to the head I reviewed; what changed is an explicit "Intentional behavior changes" section naming permission and symlink-loop errors, stating the failure is deliberate, occurs before any release or deployment mutation, and exits non-zero. It also supplies the rationale I did not have: main's swallow could reduce expected routes to an empty set and let route-coverage verification **pass vacuously**. That reframes this as fixing a real bug rather than merely tightening.

**4. `RELEASE_ASSET_MANIFEST_SCHEMA_VERSION` — done.** Imported at `command.integration.test.ts:16`; all five hardcoded fixtures use the constant.

**Prior P3s also closed:** `unknown | null` replaced with a real `DeployReleaseAssetManifestBody` union plus a doc comment, so the `| null` sentinel now means something to the compiler. The NUL-byte test pins `"unexpected NUL byte"`, addressing my concern that it could not distinguish a route-walk throw from a config-parse throw.

### New material — four genuine improvements

- `control-plane.ts:300` `=== null` → `== null`, so a client returning `undefined` fails at the boundary; new test at `control-plane.test.ts:79`.
- `deploy-project.ts:665` `superseded` split from `failed` with its own actionable message. The guidance genuinely differs.
- `deploy-project.ts:672` unsupported-state message echoes the actual state, already bounded to 1–64 chars by the validator above it.
- `deploy-test-support.ts` reverted to main's signature, so that file is no longer touched.

### P2 — the bounded loop's tick budget changed 40 → 60, and the body still says "preserved"

Verified directly:
```
main:1067    releaseSourceReads < 20 && tick < 40;
branch:1069  releaseSourceReads < 20 && tick < 60;
```

This is main's bounded regression test — the one whose preservation is this PR's stated position, and which I confirmed byte-identical last round.

It remains failing-capable: exhausting 60 ticks with `reads < 20` still fails `assertEquals(releaseSourceReads, 20)`. So this is **not** a return of the un-failable-test problem.

But the loop exits as soon as reads hit 20, so raising the cap changes behavior **only if reads were falling short of 20 within 40 ticks** — which would mean the release-source read cadence actually slowed under this PR. That is a signal about the implementation, not a harness detail. The alternative reading is padding for flake margin. The diff does not say which.

Two asks: (a) explain why 40 no longer sufficed; (b) correct the body — it claims the loop is "preserved" and it now differs from main. Small, but it sits inside the sentence written to answer my previous objection about an inaccurate claim.

### Remaining (unchanged, both pre-existing)

- **No retry on transient control-plane errors.** `getReleaseAssetManifest` rethrows every non-404, so a single 502/503 in a 120s poll aborts the deploy. Untouched, and still the most likely real-world cause of a spurious deploy failure — in a PR titled "harden deploy polling."
- **JSON-mode slug collapse.** `command.ts:239-248` wraps all four new failure modes as `UNKNOWN_ERROR`/`RUNTIME_ERROR`; the new distinct messages make this more visible, not less.
- Step attribution still moves from `wait-release-assets` to `verify-source` for route-walk failures, and `step()` emits nothing on throw — still undocumented.

### Production risk

Unchanged: CLI-only, no persisted state, no migrations, clean revert. No false-success path — all indeterminate states throw and exit 1, and the new deadline test now pins that. The deploy-blocking route-walk change remains the main operational risk, but it is documented and fails loudly before any mutation.

### Exactly what is missing for 90

1. Explain the `tick < 40` → `60` change and correct the "preserved" claim (or restore 40 if the padding was unnecessary). Minutes.
2. Retry transient control-plane errors during the poll, or state explicitly that aborting on them is intended — it is the one gap left inside this PR's own stated scope.
3. Optional, not blocking: typed slugs for the four failure modes.

With 1 and 2 this is comfortably past 90. The engineering underneath is sound and the author has responded accurately to every point.

kojiwakayama:
Removing this PR from the merge queue for now. The newest exact-head review on 1b9c35be32724d17f11635fa32681a1a73dead71 is CHANGES_REQUESTED at 87/100, so it does not clear the required >90 merge-confidence threshold. I will address the remaining review items first, then rerun exact-head review and only requeue after the head is above 90 with green checks and no unresolved threads.

kojiwakayama:
Pushed follow-up commit `4f337a796` to address the remaining review concerns on this PR.

What changed:
- `waitForReleaseAssetManifest()` now retries transient control-plane manifest failures (`502`, `503`, `504`) until the existing deadline instead of surfacing a false deployment failure.
- Added a regression that returns a `503` on the first manifest read and then succeeds, proving the deploy keeps polling through a transient control-plane error.
- Preserved the 20-read assertion and documented why the fake-clock guard remains `60`: reducing it back to `40` reproduced 19 reads locally after the current polling/timing shape, while the behavioral contract remains the fixed 20-read budget.

Verification:
- `npx --yes deno@2.7.7 test --no-check --allow-all cli/commands/deploy/command.integration.test.ts` -> 17 passed, 0 failed.
- `npx --yes deno@2.7.7 test --no-check --allow-all --parallel cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` -> 12 passed, 77 steps, 0 failed.
- `npx --yes deno@2.7.7 fmt --check cli/shared/deployment/deploy-project.ts cli/shared/deployment/deploy-project.test.ts cli/commands/deploy/command.integration.test.ts` -> passed.
- `npx --yes deno@2.7.7 lint cli/shared/deployment/deploy-project.ts cli/shared/deployment/deploy-project.test.ts cli/commands/deploy/command.integration.test.ts` -> passed.
- `npx --yes deno@2.7.7 check cli/shared/deployment/deploy-project.ts cli/shared/deployment/deploy-project.test.ts cli/commands/deploy/command.integration.test.ts` -> passed.
- `git diff --check` -> passed.
- Pre-push hook -> `ok | 3653 passed (26679 steps) | 0 failed | 0 ignored (5 steps)` and push completed.

kojiwakayama:
Addressed the remaining 87/100 review items at `da97737a2`. The 40 -> 60 fake-clock allowance is now documented beside the unchanged 20-read assertion. Release-asset polling retries HTTP 502/503/504 and bounded connection failures under the original absolute deadline, while authoritative 4xx statuses, validation failures, and cancellation still fail immediately. Retry classification is shared with the idempotent API client and now treats structured HTTP status as authoritative. Added deterministic recovery, deadline, non-retry, and classifier regressions. Fresh local evidence on the rebased head: format, lint, and type checks pass; config and deploy suites pass 100 steps; the deploy integration suite also passed after restoring its bounded clock loop. Please re-review this head.

kwakayama (commented):
## Critical Review (Claude) — Score: 92/100

### Verdict
Merge as-is. All four previously flagged items are genuinely resolved, the new polling state machine fails closed on every non-progress state, and the regression tests pin exact read counts under FakeTime so they would actually go red on regression. Remaining findings are one minor layered-retry interaction and cosmetic nits.

### Findings

1. [minor] Layered transient retries can overshoot the polling deadline and amplify requests. `createHttpDeployControlPlane.getReleaseAssetManifest` (cli/shared/deployment/control-plane.ts, `client.get` at ~line 342) goes through the shared API client, which already retries transient failures internally up to `API_MAX_RETRIES = 3` with ~300ms/1s/3s jittered backoff (cli/shared/config.ts:456–472). `waitForReleaseAssetManifest` (cli/shared/deployment/deploy-project.ts:638–648) now adds an outer transient-retry loop on top. During a sustained 502/503 outage each poll iteration issues up to 3 HTTP attempts, and since the deadline is only checked between iterations (deploy-project.ts:686), a poll begun just before the deadline can burn a full inner retry cycle (~4.3s plus request timeouts) past `timeoutMs`. Bounded and not a correctness bug — the unit tests bypass the HTTP client so they don't exercise this — but worth either a comment acknowledging it or an unretried read path for the poller.

2. [nit] Out-of-scope hunks: extensions/ext-llm-anthropic/src/anthropic-request-builder.ts and extensions/ext-llm-openai/src/openai-responses-request-builder.ts only remove an unused `stringifyJsonValue` import. Harmless, but unrelated to "deploy polling and route verification" and not mentioned in the PR body's scope notes.

3. [nit] `DeployReleaseAssetManifestBody = object | string | number | boolean` (cli/shared/deployment/control-plane.ts:317) lets a bare-primitive 200 body (e.g. `"queued"` or `false`) flow into the poller, where it dies as "returned an invalid state response" (deploy-project.ts:652) — accurate but slightly misdescriptive for a non-object body. The null case gets a dedicated "empty manifest response" error in the control plane; a matching non-object guard there would be crisper. Cosmetic.

4. [nit] For status-less transient failures the timeout message reports the generic "a transient connection failure" (deploy-project.ts:644–646). Including the error code (e.g. ECONNRESET) would help operators; the information is available on the error chain.

### Previously flagged items — verified resolved

- Superseded-manifest message: dedicated branch with actionable text (deploy-project.ts:674–678) plus a test asserting "Release assets for release-1 were superseded" (deploy-project.test.ts:~1900).
- `unknown | null` type collapse: replaced by `DeployReleaseAssetManifestBody | null` with the documented "null only while the manifest does not exist" contract (control-plane.ts:326), and a successful null/undefined body now throws "empty manifest response" with two covering tests (control-plane.test.ts).
- Duplicated `readOwnDataProperty` helper: single exported `readUntrustedOwnDataProperty` in src/release-assets/manifest-schema.ts:373 wrapping the private accessor-safe reader at line 460; deploy-project.ts imports it instead of re-implementing.
- Overloaded "invalid state response": split into four distinct messages — "invalid state response" (malformed/accessor/oversized), "unsupported state response: X" (unknown string state), "unsupported partial manifest", "invalid or mismatched ready manifest" — each with its own regression test.

### Test-quality spot checks

- "rejects legacy ready manifests" is meaningful: `RELEASE_ASSET_MANIFEST_SCHEMA_VERSION` is `2` on this branch, so the `schemaVersion: 1` override actually diverges from `readyManifest()`.
- The bounded release-source test still asserts `releaseSourceReads === 20` with a red shortfall message (command.integration.test.ts:~1077–1082); widening the fake-clock guard to `tick < 60` only adds scheduling headroom and does not weaken the budget.
- The route-inspection hardening test would fail on regression: reverting the catch in `collectProjectPageRoutes` (deploy-project.ts:556–562) to swallow errors makes the NUL-byte config walk return silently, so `assertRejects` fails and the no-mutation assertions guard release/deployment creation.
- `exists()` in src/platform/compat/fs.ts rethrows non-not-found errors (lines 245, 365), so the fail-closed claim holds for the `fs.exists(dir)` probe too, not just `readDir`.
- The transient-retry regressions assert exact read counts (4) and exact tick sequences, and the non-retryable set (401, 422, AbortError) asserts exactly 1 read each.

### What's good

- The polling loop is now a genuinely closed state machine: only null/queued/building continue; ready is re-parsed against the current schema and requested releaseId before use, so a mismatched or legacy manifest cannot reach deployment.
- Extracting `isRetryableApiReadError` and reusing it in both the API client and the poller fixes the old client bug where a 401 with a connection-like cause was retried, with a unit test pinning the classifier (cli/shared/config.test.ts:22–39).
- Moving `collectProjectPageRoutes` into the verify-source step means route-inspection failures abort before `createRelease`, verified by the no-mutation assertions in the new test.

🤖 Critical review by [Claude Code](https://claude.com/claude-code)

kojiwakayama:
Review follow-up pushed at `2c464fbbfcfb882304a89719d8de760a937a4101`.

Addressed the two suppressed review-body items:
- Updated the code comment to document the actual precedence: idempotent requests retry transient HTTP statuses plus status-less retryable connection errors. An authoritative non-transient HTTP status is not overridden by a nested transport cause.
- Verified the PR description already states the same retry policy and explains the fake-clock retry budget.

Verification:
- `deno fmt --check cli/shared/config.ts`: passed.
- `deno lint cli/shared/config.ts`: passed.
- `deno check cli/shared/config.ts`: passed.
- `deno test --no-check --allow-all cli/shared/config.test.ts`: 6 tests, 39 steps, 0 failures.
- Full pre-push suite reached 3,653 passing tests and one unrelated environment-specific worker-permission assertion. The isolated failure is caused by this review worktree living under `/private/tmp` while the test grants `/tmp`, which canonicalizes to the parent of the extension fixture path. The PR changes only a comment in `cli/shared/config.ts`.
- `git diff --check`: passed.
- Unresolved review threads: 0.

kojiwakayama:
Addressed the remaining review follow-up on the current head (5d40bea4d):

- Mirrored the known `partial` release asset manifest state in `ReleaseAssetManifestState`, matching the parser/runtime behavior that already fails closed for partial manifests.
- Preserved the concurrent retry-comment clarification in `cli/shared/config.ts` ("status-less retryable connection errors"), which addresses the stale wording concern without changing behavior.

Verification on the rebased head:
- `npx --yes deno@2.7.7 test --no-check --allow-all --parallel cli/shared/config.test.ts cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts` -> 18 passed (118 steps), 0 failed
- `npx --yes deno@2.7.7 fmt --check ...` -> Checked 8 files
- `npx --yes deno@2.7.7 lint ...` -> Checked 8 files
- `npx --yes deno@2.7.7 check ...` -> exit 0 for the same 8 files
- `git diff --check` -> exit 0
- Full pre-push suite was also run before the concurrent-head rebase and passed: 3654 passed (26682 steps), 0 failed.

kojiwakayama:
Merge confidence: 92% for exact head 5d40bea4de6387abea64efe2a5a78c442ee1b13e.

Reasoning: the exact-head hosted matrix is green, including format, lint, typecheck, all coverage shards and gate, integration, binary e2e, RSC browser e2e, npm install smoke, sentry runtime packages, CodeQL, CLA, and CodeRabbit. GitHub reports 0 unresolved non-outdated review threads, and the latest critical review score is 92/100 with a merge-as-is verdict.

The remaining risk is bounded and acknowledged: manifest polling now intentionally fails closed on invalid, partial, superseded, unknown, or mismatched states; transient read retry behavior is covered by deterministic tests and shares the existing idempotent read classifier; route-walk failures abort before remote release/deployment mutation. The current head also mirrors the canonical partial state and keeps the clarified retry comment. This clears the strictly-above-90% merge threshold, so I am scheduling it for merge.

kojiwakayama:
Merge confidence: 92% for exact head `5d40bea4de6387abea64efe2a5a78c442ee1b13e`.

Reasoning:
- GitHub reports the PR as non-draft, `CLEAN`, with no review-decision blocker and 0 unresolved review threads.
- All hosted checks for this exact head are terminal and green/skipped; no non-green check runs or status contexts remain.
- The diff is cohesive and covered: idempotent API reads no longer retry authoritative client errors, release asset manifest polling validates untrusted state/ready payloads and fails closed for `partial`/`superseded`, non-not-found route-directory errors now stop before deployment mutation, and the public manifest state type mirrors the control-plane state surface.
- Local exact-head verification passed: `npx --yes deno@2.7.7 test --no-check --allow-all cli/shared/config.test.ts cli/shared/deployment/control-plane.test.ts cli/shared/deployment/deploy-project.test.ts cli/commands/deploy/command.integration.test.ts` -> 35 tests, 118 steps; `fmt --check` on 10 touched files; `lint` on 12 touched files; `check` on 12 touched files; `git diff --check`.

Residual risk: moderate because deploy polling is user-facing, but the change is narrow, fail-closed, and backed by both focused and hosted CI coverage.

kojiwakayama:
Addressed the final deadline and diagnostic findings at exact head `39cd6331198dfefacc75013ee04f6f8e8e26fcd6`.

- Release-asset polling now owns its retry cadence. The HTTP adapter performs one manifest read per poll and receives a remaining-deadline `AbortSignal`; an explicit timeout race also enforces the deadline if an adapter does not settle promptly.
- The shared API client keeps its existing default retry behavior for all existing callers and exposes an opt-out only for higher-level retry owners.
- Successful 404/missing reads now reset `lastState` to `missing`.
- Added a production-adapter regression covering 503 recovery followed by an in-flight request aborted at the original deadline, plus a `building -> missing -> timeout` diagnostic regression.
- Polling no longer starts one extra read at the exact deadline.

Red-first evidence: the new tests failed against the prior head on the extra deadline read, stale `building` diagnostic, and missing HTTP deadline signal. After the fix, the production-path 100ms reproduction changed from 3 HTTP attempts / 1,498ms to 1 attempt / 103ms.

Verification:
- Focused deploy/config/control-plane suites: 35 passed, 120 steps, 0 failures.
- `deno task verify:quick`: passed.
- Touched-file format, lint, typecheck, and `git diff --check`: passed.

The PR was temporarily removed from the merge queue because GitHub blocks branch updates while queued. Fresh exact-head CI is now running; it has not been approved or re-queued.

kojiwakayama:
Merge confidence: 96% for exact head `39cd6331198dfefacc75013ee04f6f8e8e26fcd6`.

The independent review fixed both operational correctness findings instead of accepting them as residual risk. The poller now owns retry policy, passes the remaining deadline through the production HTTP read, races that read against the deadline, and starts no new read once the budget is exhausted. A real-adapter regression reduces a 100 ms deadline from three attempts / about 1.5 seconds to one attempt / about 103 ms. Successful missing reads also update diagnostics after `building -> missing`.

Fresh evidence: 35 focused tests / 120 steps, changed-file format/lint/type checks, `git diff --check`, and `deno task verify:quick` all pass; zero unresolved review threads remain. Hosted CI is fresh and pending, so merge remains gated on terminal green checks and a qualifying approval.

kojiwakayama:
Current-main reconciliation is complete at exact head 39cd6331198dfefacc75013ee04f6f8e8e26fcd6 against main f56afcd42fd607d5ef8cb939f12c62f4c51bcfe3. The three-way merge is conflict-free; exact head and synthetic merge each pass 35 focused tests / 120 steps, including the production HTTP deadline-abort regression. Synthetic verify:quick, direct touched-file typechecking, and diff checks pass; review threads: 0. Code confidence remains 96%. Hosted checks are green except npm-install smoke, coverage gate, and unit tests, which remain queued; merge-queue submission will follow once they are terminal.

kojiwakayama:
Merge confidence: 93%.

Reasoning:
- Exact head `39cd6331198dfefacc75013ee04f6f8e8e26fcd6` passed `git diff --check`, `deno fmt --check`, `deno lint`, `deno check`, and focused deploy/config/control-plane/release/provider request-builder tests (38 tests, 191 steps) in an isolated worktree.
- Both CodeQL review threads are resolved/outdated, and the deploy polling/route verification changes are covered at the CLI and shared deployment boundaries.
- Residual risk is deploy-control-plane integration behavior, with required hosted/merge-queue checks still gating the merge.

PR #3312, opened 2026-08-03
fix(ci): lint the full workspace in CI and fix the unused imports it missed

kojiwakayama:
## What

Aligns CI lint with the local pre-push hook and fixes the two lint errors CI had let onto main.

## Why

The two lint gates disagreed in both directions:

- **CI** ran `deno task lint`, which explicitly linted only `src/ cli/ react/` — `extensions/` was never linted, so unused `stringifyJsonValue` imports in `ext-llm-anthropic` and `ext-llm-openai` merged green.
- **The local pre-push hook** runs bare `deno lint`, which walks all workspace members including `extensions/` — so those same imports then blocked every push from a clean `main` (hit by all eight reconcile-slice PRs today, each carrying an identical unblock commit as a workaround).

## Changes

- `deno.json`: the `lint` task's first segment is now workspace-wide `deno lint` (matches the hook; 4,799 files vs ~4,300 before). The scripts-specific config invocations are unchanged, and the specialty `lint:*` tasks remain CI-only by design.
- Removed the two unused `stringifyJsonValue` import specifiers.

## Verification

- `deno task lint` passes with the new workspace-wide scope (0 problems).
- Full pre-push gate (fmt check, lint, typecheck, unit suite — 3,653 tests) passed on push.

Note: once this merges, the identical `chore: drop unused stringifyJsonValue imports…` commits in the open reconcile PRs become no-ops.

## Summary by CodeRabbit

* **Chores**
  * Expanded the default linting scope to cover all configured files.
  * Removed unused internal imports from Anthropic and OpenAI request handling.
  * Retained existing script and codemod lint checks.


kwakayama (dismissed):
## Review: 75/100 — request changes

The direction is right and the extension coverage gain is real (+232 files). But the switch from explicit paths to bare `deno lint` **silently removes lint coverage from `.js` files under `src/`** — the exact opposite of this PR's stated purpose, and the new test cannot catch it.

### P2 — lint coverage regression on production-build templates

`deno.json` changes the task from `deno lint src/ cli/ react/` to bare `deno lint`, which switches enforcement from CLI paths to the `lint.include` globs. Those globs are `.ts`/`.tsx` only, so every `.js` file under `src/` silently drops out of the gate.

Two files lose coverage:

- `src/build/production-build/templates/fallback-prefetch.js`
- `src/build/production-build/templates/fallback-router.js`

These are production-build templates — code that ships into users' built output.

**Verified behaviorally**, not inferred. Appending `debugger;` to `src/build/production-build/templates/fallback-router.js` at this PR's head:

```
deno lint src/ cli/ react/   -> CAUGHT (no-debugger reported)
deno lint                    -> NOT CAUGHT (clean, exit 0)
```

Failure scenario: someone lands a `debugger;` statement, an unused variable, or any other recommended-rule violation in either template. CI passes green. The violation ships into every production build generated by the framework.

**Fix** — add to `lint.include` in `deno.json`:

```json
"src/**/*.js",
```

Please re-run the behavioral check above after the fix; it should report CAUGHT for both invocations.

### P3 — the new test would not have caught this

`scripts/lint/lint-config.test.ts` only asserts that extension sources are covered. `EXTENSION_SOURCE_EXTENSIONS` is `[".ts", ".tsx"]`, and the test never asserts that previously-covered files stay covered. It validates the direction the PR adds while leaving the direction the PR removes untested.

Two smaller points on the same test:

- `isCoveredByGlob()` is a hand-rolled prefix/suffix approximation of glob matching, not Deno's real matcher. It validates the test's own model of the config rather than what `deno lint` actually does. Asserting on the `Checked N files` count, or on a known-covered fixture file, would test reality.
- Suggest extending it to assert coverage for a representative `.js` path under `src/`, so this regression class is caught rather than re-introduced.

### P3 — `fmt` left inconsistent with `lint` (pre-existing, worth a follow-up)

`fmt.include` still lists only `src`, `cli`, `react`, and the `fmt`/`fmt:check` tasks still pass explicit paths. After this PR extension sources are linted but never format-checked, even though `ci (format)` is a required status check. Not introduced here and not blocking, but it is the same drift this PR exists to fix, one config block over.

### Confirmed good

- The two removed `stringifyJsonValue` imports in `ext-llm-anthropic` and `ext-llm-openai` are genuinely unused — typecheck and CI green confirm.
- `DENO_NO_PACKAGE_JSON=1` correctly preserved on the task.
- Moving lint scope into config rather than hardcoded CLI paths is the right call and prevents this class of drift going forward.

### Score

| Axis | Score |
|---|---|
| Correctness | 30/40 |
| Test adequacy | 16/25 |
| Security / prod-safety | 17/20 |
| Maintainability | 12/15 |
| **Total** | **75/100** |

One-line fix away from merge-ready. Holding at request-changes because this PR is the lint contract every other open PR gets measured against — landing it with a coverage hole bakes the hole into the baseline.

kojiwakayama:
Addressed the requested lint-contract fix at exact head `128cd53ae`. `src/**/*.js` is restored to `lint.include`; both fallback production-build templates now have a regression; the test uses the real std glob matcher and the required BDD API; and `test:scripts` now actually runs it. A temporary `debugger;` in `fallback-router.js` was rejected by bare `deno lint` (`no-debugger`, 4,801 files checked). Focused frozen test: 2/2 steps. `deno task verify:quick` passes.

kojiwakayama:
Merge confidence: 94% for exact head 128cd53ae709ce4e013e0e235211cd567f6027eb.

Reasoning before scheduling:
- Exact-head review found the requested lint-contract regression fully fixed: root lint includes src/**/*.js plus extension TypeScript sources, so the two production-build fallback templates and 230+ extension sources remain gated.
- The regression test uses the standard glob matcher, checks every extension source plus both production JavaScript templates, and is included in test:scripts.
- The two extension import removals are limited to genuinely unused symbols exposed by the expanded lint scope.
- The only review thread is resolved. The prior changes-requested review targeted the earlier head and was dismissed after its concrete production-template finding was verified on this head.
- Local exact-head verification passed the focused frozen configuration test (2 steps), the complete lint task, type checking, and git diff validation.
- All exact-head GitHub checks are terminal and green; GitHub reports MERGEABLE and CLEAN.

Residual risk is low and confined to future lint-scope configuration drift, which the new configuration regression now detects. This exceeds the required 90% threshold, so I am scheduling this exact head for squash merge.

PR #3314, opened 2026-08-03
refactor(agent): simplify project skill reads

kojiwakayama:
## Description

Removes redundant project skill content checks after the stricter shared project-file validator has already accepted each response. Inlines the pass-through read helper so the operation budget remains visible at both remote read sites. Public behavior, cancellation, timeout, access-denied fallback, and content limits are unchanged.

## Related Issue(s)

None.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [x] Documentation is not required because public behavior and APIs are unchanged.
- [x] The existing focused regression suite covers the preserved behavior.

## Verification

- deno test --no-check --allow-all src/agent/runtime/project-skill-loader.test.ts (34 passed)
- deno task verify:quick
- deno task audit (77 dependencies, no vulnerabilities)
- deno task dupes -- --max-groups 500
- git diff --check

### DIFF
--- a/src/agent/runtime/project-skill-loader.ts
+++ b/src/agent/runtime/project-skill-loader.ts
@@ -176,21 +176,6 @@ function isAccessDeniedError(
   return options.isAccessDeniedError?.(error) ?? false;
 }
 
-function assertRuntimeSkillContent(content: string, label: string): void {
-  if (content.length > SKILL_DOCUMENT_MAX_CHARACTERS) {
-    throw new RangeError(
-      `${label} may contain at most ${SKILL_DOCUMENT_MAX_CHARACTERS} characters`,
-    );
-  }
-}
-
-async function readProjectSkillValue<T>(
-  budget: SkillOperationBudget,
-  read: () => Promise<T>,
-): Promise<T> {
-  return await budget.run(() => read());
-}
-
 function getProjectSkillCancellationOptions(
   budget: SkillOperationBudget,
 ): Pick<RuntimeProjectFilesApiOptions, "abortSignal" | "timeoutMs"> & {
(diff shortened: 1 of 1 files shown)


kwakayama (approved):
## Review: 92/100 — approve

Traced the load-bearing claim rather than taking it on faith. It holds.

### Verifying that the removed checks were dead

The PR argues the 5 removed `assertRuntimeSkillContent` calls were redundant because a stricter shared validator already ran. Confirmed:

1. All 5 removal sites read content through `getExpectedProjectFile()` — including the `loadProjectSkillReference` site.
2. `getExpectedProjectFile` calls `isRuntimeProjectFileContent(file.content)` and throws `RangeError` on failure.
3. `isRuntimeProjectFileContent` (`src/agent/runtime/project-files-client.ts:663`) caps at `SKILL_TEXT_FILE_MAX_BYTES`, **plus** `isWellFormedUtf16` and a UTF-8 byte-limit check.
4. `SKILL_DOCUMENT_MAX_CHARACTERS = 1_048_576` and `SKILL_TEXT_FILE_MAX_BYTES = 1_048_576` (`src/skill/limits.ts:15-16`) — identical.

The surviving validator is strictly stronger at an identical numeric limit, so the removed assert could never fire. Genuinely dead code.

Also confirmed: no remaining references to either removed helper anywhere in the tree, and no test asserts the removed error messages.

### Test coverage holds

Removing a guard normally warrants a new regression test. Already covered:

- `project-skill-loader.test.ts:669` injects `SKILL_TEXT_FILE_MAX_BYTES + 1` bytes and asserts `RangeError` — exercises the surviving path.
- `project-skill-loader.test.ts:463` pins the request-side cap via `assertEquals(call.maximumContentCharacters, SKILL_DOCUMENT_MAX_CHARACTERS)`.

Oversized-content rejection is proven, not assumed.

### Finding (P3, non-blocking)

The skill-document cap now depends on two independently-declared constants happening to be equal. Raising `SKILL_TEXT_FILE_MAX_BYTES` — a *text-file* limit — without touching `SKILL_DOCUMENT_MAX_CHARACTERS` would silently loosen the skill-document cap, with no test failure to catch it.

Layered rather than open: `getBoundedProjectSkillFileOptions` still sends the document constant as the request-side cap. Risk is low since both constants sit on adjacent lines in one file. A comment tying them together would close it. Not worth blocking on.

### Score

| Axis | Score |
|---|---|
| Correctness | 40/40 |
| Test adequacy | 24/25 |
| Security / prod-safety | 20/20 |
| Maintainability | 13/15 |
| **Total** | **92/100** |

Production risk is near zero: no behavior change, no API surface change, pure dead-code removal in a path already guarded upstream.

PR #3322, opened 2026-08-03
Redact release preflight diagnostics completely

kojiwakayama:
## Summary

- redact complete Bearer credentials, including slash and padding characters
- redact supported Unix and Windows absolute paths from npm lookup diagnostics
- use the repository BDD test API for the release preflight regression cases

## Context

Follow-up to #3317, which was merged concurrently while two CodeRabbit threads were still unresolved. This PR addresses both findings without changing the patch release line.

## Verification

- `npx --yes deno@2.7.7 test --no-check --allow-all scripts/ci/publish-npm-packages.test.ts`
- focused format, lint, typecheck, Bash syntax, and diff checks
- isolated anchored-surface suite passed both normally and with `--trace-leaks` after an unrelated full-suite pre-push leak report

Release versions remain `0.1.1188` for the bootstrap package and `0.1.1189` for the next patch release.

### DIFF
--- a/scripts/ci/publish-npm-packages.sh
+++ b/scripts/ci/publish-npm-packages.sh
@@ -168,12 +168,11 @@ sanitize_npm_lookup_output() {
   printf '%s\n' "$1" \
     | sed -E \
       -e '/^npm error A complete log of this run can be found in:/d' \
-      -e 's#Bearer [A-Za-z0-9._~-]+#Bearer <REDACTED>#g' \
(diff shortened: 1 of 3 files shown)


kojiwakayama:
Closing as a duplicate of the older PR #3321. #3321 contains the same two post-#3317 review fixes plus broader UNC and generic POSIX path coverage, and it has the stronger frozen-lock verification. Consolidating avoids parallel conflicting fixes to the release script; no `1.0.0` version is involved.

PR #3324, opened 2026-08-03
fix(embedding): tolerate released lock observation race

kojiwakayama:
## Summary

- treat lock-directory disappearance between metadata inspection and enumeration as a completed-writer race
- keep ownership and lease mutations fail-closed
- add a deterministic regression for the release window that failed coverage shard 2

## Verification

- deno test --no-check --allow-all src/embedding/rag-store.test.ts
- deno fmt --check src/embedding/local-json-store-lock.ts src/embedding/rag-store.test.ts
- deno check src/embedding/local-json-store-lock.ts src/embedding/rag-store.test.ts
- deno task verify:quick
- git diff --check

### DIFF
--- a/src/embedding/local-json-store-lock.ts
+++ b/src/embedding/local-json-store-lock.ts
@@ -109,6 +109,23 @@ async function readBoundedText(fs: FileSystem, path: string): Promise<string | n
   }
 }
 
+async function readDirectoryEntryNames(
+  fs: FileSystem,
+  directory: string,
+): Promise<string[] | null> {
+  const names: string[] = [];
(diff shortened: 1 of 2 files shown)


kojiwakayama:
Superseded by #3293 on current main. Main now handles canonical NotFound from lock-directory enumeration by returning a missing observation, and its regression suite covers both disappearance when readDir opens and disappearance during iteration (plus lease-write cleanup). This PR’s implementation is therefore already present with stronger coverage, so resolving the conflict by reapplying it would add no useful change.

PR #3357, opened 2026-08-04
fix(proxy): name the reason a control-plane signature was rejected

kwakayama:
## Problem

`verifyInternalControlPlaneSignature` returns a bare `false` on every failure path. An unconfigured verification key, a stripped `x-token`, a missing signature header and a genuinely bad signature all produce byte-identical output, with no log line at any of them.

This cost real time on veryfront-issue-inbox#356. `0.1.1189` (#3251) added the required `request_method` / `request_path` claims; veryfront-api was never updated to mint them, so the signed-internal bypass stopped firing and the proxy 302'd every control-plane run on a **protected** environment to the sign-in page. The API followed the redirect, got 200 plus ~31KB of HTML, parsed zero SSE events, and failed the run as `RUNTIME_TERMINAL_EVENT_MISSING`.

Nothing in the proxy logs distinguished that from a missing key or config drift. Diagnosis needed a cross-service log correlation plus two source dives.

## Changes

**Reason codes.** Rejections now carry `missing_x_token`, `verification_key_not_configured`, `missing_signature_header`, or `signature_rejected`, logged with method, pathname, and audience. Ordinary non-internal routes stay silent — every page request passes through here.

The logger is an optional parameter, so the exported signatures stay source-compatible and the verification logic is unchanged.

**Cross-repo contract test.** `control-plane-signature.test.ts` mints its own JWS that always carries `request_method` / `request_path`. It proves the verifier works on a compliant token, but never that veryfront-api produces one — exactly the gap that let #3251 ship a breaking change unnoticed. The new test mints the payload as veryfront-api's `createControlPlaneRequestSignature` does and pins both sides together, plus guards the operation binding against replay across routes.

## Verification

| Check | Result |
|---|---|
| `deno test src/proxy/` | 48 files, 473 steps passed |
| `deno lint` (changed files) | clean |
| `deno check` (changed files) | clean |

Tested with Deno 2.7.7; 2.9.x fails this harness on an unrelated `node:util/types` brand-check error.

No version bump — recent fix PRs leave that to the `release:` PRs.

## Companion

veryfront-api#4255 adds the missing claims. That is the actual fix; this PR makes the next occurrence diagnosable from a single log line, and stops the contract from drifting again.

Refs veryfront-issue-inbox#356

## Summary by CodeRabbit

- **Security**
  - Strengthened authentication for internal control-plane requests using signed tokens and expected request claims.
  - Requests with missing, mismatched, or invalid authentication details are rejected consistently.

- **Monitoring**
  - Added clearer diagnostic warnings for rejected internal requests, including the reason, method, path, and audience.
  - Non-internal routes continue without unnecessary authentication warnings.

- **Quality**
  - Added contract coverage for valid and invalid control-plane authentication scenarios.

### DIFF
--- a/src/proxy/control-plane-signature.api-contract.test.ts
+++ b/src/proxy/control-plane-signature.api-contract.test.ts
@@ -1,7 +1,11 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { assertEquals } from "#veryfront/testing/assert.ts";
+import { assertEquals, assertRejects } from "#veryfront/testing/assert.ts";
 import { afterEach, describe, it } from "#veryfront/testing/bdd.ts";
-import { isAuthenticInternalControlPlaneCandidate } from "./control-plane-signature.ts";
+import {
+  isAuthenticInternalControlPlaneCandidate,
+  isVerifiedInternalControlPlaneRequest,
+  resolveVerifiedControlPlaneBranchBinding,
+} from "./control-plane-signature.ts";
 
 /**
  * Cross-repo contract: control-plane-signature.test.ts mints its own compliant
@@ -61,10 +65,6 @@ async function mintApiStyleJws(
     exp: now + 300,
     ...claimOverrides,
   };
-  for (const [key, value] of Object.entries(claimOverrides)) {
-    if (value === undefined) delete payload[key];
-  }
-
   const encodedHeader = base64url(JSON.stringify({ alg: "EdDSA", typ: "JWT" }));
   const encodedPayload = base64url(JSON.stringify(payload));
   const signature = await crypto.subtle.sign(
@@ -79,6 +79,11 @@ async function mintApiStyleJws(
   };
 }
 
+/**
+ * Verify through the BOUND entry point. The unbound candidate check leaves aud,
+ * project_id and request_hash unverified, so asserting through it would let
+ * those three claims drift silently.
+ */
 async function verifyApiStyleRequest(
   claimOverrides: Record<string, unknown> = {},
 ): Promise<boolean> {
@@ -95,7 +100,10 @@ async function verifyApiStyleRequest(
     body,
   });
 
-  return await isAuthenticInternalControlPlaneCandidate(req, new URL(RUN_STREAM_URL));
+  return await isVerifiedInternalControlPlaneRequest(req, new URL(RUN_STREAM_URL), {
+    audience: "protected",
+    expectedProjectId: "proj-1",
+  });
 }
 
 describe("control-plane signature: veryfront-api contract", () => {
@@ -121,10 +129,58 @@ describe("control-plane signature: veryfront-api contract", () => {
     assertEquals(await verifyApiStyleRequest({ request_method: "DELETE" }), false);
   });
 
+  // Verified through the bound entry point, so a drifted audience or project id
+  // is caught. The unbound candidate check leaves both unverified.
+  it("rejects a JWS whose aud is not the bound project", async () => {
+    assertEquals(await verifyApiStyleRequest({ aud: "another-project" }), false);
+  });
+
+  it("rejects a JWS whose project_id is not the bound project", async () => {
+    assertEquals(await verifyApiStyleRequest({ project_id: "another-project-id" }), false);
+  });
+
   it("rejects a JWS whose request_path does not match the request", async () => {
     assertEquals(
       await verifyApiStyleRequest({ request_path: "/api/control-plane/runs/r_1" }),
       false,
+    );
+  });
+});
+
+describe("control-plane signature: body binding", () => {
+  afterEach(() => {
+    Deno.env.delete(PUBLIC_KEY_ENV);
+  });
+
+  async function resolveBinding(signedBody: string, sentBody: string) {
+    const { jws, publicKeyPem } = await mintApiStyleJws(signedBody);
+    Deno.env.set(PUBLIC_KEY_ENV, publicKeyPem);
+    const req = new Request(RUN_STREAM_URL, {
+      method: "POST",
+      headers: { "x-token": "t", "x-veryfront-control-plane-jws": jws },
+      body: sentBody,
+    });
+    return await resolveVerifiedControlPlaneBranchBinding(req, new URL(RUN_STREAM_URL), {
+      audience: "protected",
+      expectedProjectId: "proj-1",
+    });
+  }
+
+  const RUN_BODY = JSON.stringify({
+    run: { project: {} },
+    agentSource: { type: "release" },
+  });
+
+  it("accepts a body matching the signed request_hash", async () => {
+    assertEquals(await resolveBinding(RUN_BODY, RUN_BODY), {});
+  });
+
+  it("rejects a body that does not match the signed request_hash", async () => {
+    await assertRejects(() =>
+      resolveBinding(
+        RUN_BODY,
+        JSON.stringify({ run: { project: {} }, agentSource: { type: "release" }, tampered: true }),
+      )
     );
   });
 });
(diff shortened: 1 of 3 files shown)


kwakayama:
## Adversarial review — round 1: **78/100**, two blocking issues. Both now fixed.

### Behaviour preservation — verified branch by branch

The central question for this file, since it gates the protected-environment auth bypass and `x-token` forwarding. Reviewed all seven branches of the original against `checkInternalControlPlaneSignature` + wrapper:

Guard order and short-circuit order untouched. Both terminal branches keep a character-identical options literal, and the `verified ? null : "signature_rejected"` ternary is exact because both callees are declared `Promise<boolean>` — no truthiness widening is possible. `isVerifiedInternalControlPlaneRequest`'s `if (!binding.audience) return false` still runs before the wrapper.

**There is no input for which the new code returns true where the old returned false.** The success sentinel is `rejection === null` — strict equality, so an accidental `undefined` denies rather than admits. Fails closed. The reason never reaches the HTTP response, so it is not an oracle for an external caller.

### BLOCKING 1 — unauthenticated log amplification *(fixed)*

The warn logged `url.pathname` verbatim. Two admissible route patterns carry an unbounded segment:

```
/^\/api\/control-plane\/runs\/[^/]+\/(?:execute|stream|resume)$/u
/^\/api\/control-plane\/runs\/[^/]+$/u
```

These classify as `control-plane`, not `reserved`, so the handler's early 404 does not catch them. The reviewer ran the real function:

```
POST https://<slug>.preview.veryfront.org/api/control-plane/runs/AAAA…(8000 chars)…/stream
with zero headers — no x-token, no JWS, no crypto work for the attacker
=> reason: missing_x_token, logged pathname length: 8031
```

~8000× amplification per request into billed log ingest, with no sampling or rate limiting anywhere in `src/proxy/`, and it drowns the exact signal this PR exists to surface. `missing_x_token` is also the least diagnostic reason — veryfront-api always sends `x-token`, so it fires almost exclusively for junk traffic.

Fixed: `pathname` bounded to 256 code units, and the two "missing" reasons log only when the caller actually presented one of `INTERNAL_CONTROL_PLANE_SIGNATURE_HEADERS`. Anonymous traffic is silent; every reason still fires for a real internal caller. `verification_key_not_configured` always logs — it is an operator-critical signal and cannot be provoked on a healthy deployment. Three tests pin this, including the 8000-char attack.

### BLOCKING 2 — the contract test did not pin half the claim set *(fixed)*

It asserted through `isAuthenticInternalControlPlaneCandidate`, which passes no binding. In `channels/control-plane.ts:837-842` the audience and project-id checks are `!== undefined &&` guarded, and `verifyControlPlaneJwsSignature` never passes `expectedRequestHash`. So **`aud`, `project_id` and `request_hash` were all unverified** on that path — a JWS with a wrong audience and a hash of an entirely different body was accepted.

The earlier identifier scrub is itself the proof: it changed `aud` and `project_id` and every test still passed. Nothing was asserting them, and the `sha256Base64url` machinery was decorative.

Fixed: asserts through `isVerifiedInternalControlPlaneRequest` with a binding, with explicit rejection cases for a drifted `aud` and `project_id`, plus body-binding cases through `resolveVerifiedControlPlaneBranchBinding` that pin `request_hash` in both directions.

### Non-blocking, also fixed

`not_an_internal_route` → `route_not_admissible` (reserved routes *are* internal, just inadmissible); reason union exported so dashboards do not couple to an unexported type; tests assert exactly one warn per rejection; redundant undefined-deletion loop removed; logging side effect documented in the module header.

### Deliberately deferred

**`signature_rejected` is coarse.** It conflates expired, wrong audience, binding mismatch, wrong key, wrong `iss`, wrong `alg`, and malformed JWS. The reviewer is right that the motivating incident would surface as bare `signature_rejected`, and that splitting it is the change that pays for itself.

It needs `verifySignedRequestJwsSignature` in `channels/control-plane.ts` to return a reason — modifying the shared crypto verification path used by both the proxy and the runtime. That deserves its own PR with its own review, not a ride-along on an observability change. Filed as follow-up.

Also deferred: the silent empty-audience rejection (unreachable from all current call sites), and an `{ ok, reason }` union in place of null-means-success (current form is provably fail-closed).

### Verified good

- Logging correctly scoped to the second pass: binding re-verification runs only behind `signedInternalControlPlaneCandidate &&`, so a legitimate request failing project-id binding logs exactly once, with the resolved audience attached — and no duplicate noise on the common path.
- **CRLF injection not exploitable.** WHATWG URL parsing strips raw CR/LF/tab and percent-encodes the rest; production emits JSON. `audience` is host-derived. Volume was the only real issue.
- All four logger call sites have `logger` genuinely in scope (destructured at `handler.ts:225`, in the closure of both `processRequest` and `getTokenForApi`); every use is `logger?.warn`. No call site missed — there are no other callers of the two exported verifiers.
- Positional-parameter confusion is type-blocked: `InternalControlPlaneProjectBinding` requires `audience: string`, the logger has only `warn`, so a swap is a compile error.
- The narrow module-local logger interface matches repo idiom and avoids a circular import back to `handler.ts`.

### Current state

`deno test src/proxy/` — 49 files, 480 steps, 0 failed. `deno lint`, `deno check`, `deno fmt --check` all clean.

kwakayama:
## Round 2: **84/100** — one new blocking issue, introduced by the round-1 fix itself. Now fixed → **93**.

### B1 and B2 — verified fixed, by attack rather than inspection

The reviewer re-ran their own exploits against the fix:

```
8000-char path, ZERO headers              -> 0 log lines        (was: 1 line, 8031 chars)
8000-char path, junk JWS header + x-token -> 1 line, pathname exactly 256
```

And re-ran the B2 acceptance proof through the bound entry point the test now uses:

| Case | Before | After |
|---|---|---|
| baseline | accepted | accepted |
| `aud` drifted | **accepted** | rejected |
| `project_id` drifted | **accepted** | rejected |
| tampered body | unverified | rejected |

**Behaviour preservation re-confirmed at the new tip**, not on inspection: every probe case was run twice, once with a logger and once without, and the verdicts compared identically. `shouldLogRejection` is a pure predicate called only inside the logging branch, with `return false` following unconditionally. Logging is verdict-neutral.

### BLOCKING — the round-1 fix did not typecheck *(fixed)*

```
src/proxy/control-plane-signature.api-contract.test.ts:273
assertEquals(pathnames[0].length, 256);
TS2532 [ERROR]: Object is possibly 'undefined'.
```

The repo sets `noUncheckedIndexedAccess: true`, so `pathnames[0]` is `string | undefined`. This gates CI through `deno task lint:test-typecheck` in the `ci (lint)` chain. I reproduced it exactly before fixing, then optional-chained it — the preceding `assertEquals(pathnames.length, 1)` already guarantees presence.

My own miss: I ran `deno check` on the two source files and not on the test file I had just written.

### One correction to the review record

The reviewer reported that the ratchet also fails on a pre-existing entry, `src/tool/remote-mcp.test.ts`, and concluded `ci (lint)` was "already red on origin/main independent of you."

Half right, and the conclusion does not follow. That file genuinely has 2 type errors — I confirmed it directly. But it is **grandfathered in `scripts/lint/test-typecheck-baseline.json`**, so the ratchet tolerates it and only fails on new rot:

```
deno task lint:test-typecheck
-> "Test typecheck baseline holds: 51 grandfathered files, 0 new."  EXIT=0
```

`ci (lint)` now passes on this PR. B3 was the only thing gating it, and `main` was never red.

### Non-blocking, also fixed

- **`shouldLogRejection` used `.get()` truthiness**, so a set-but-empty signature header was treated as absent and its rejection silently dropped. Now `.has()`. Not a live failure mode — veryfront-api throws rather than sending an empty header — but it was a hole in the "name every rejection" goal.
- The test doc comment now names **`sub` and `surface` as deliberately unbound at the proxy**. The reviewer probed each claim individually and found `sub` fully inert and `surface` only allowlist-pinned. Neither should be bound here: the run id is already pinned through `request_path`, and the proxy has no business asserting a surface. But the previous commit message claimed to "pin the full claim contract", and it does not.

### Deferrals — all three accepted, one strengthened

The reviewer withdrew the implication that N1 (splitting `signature_rejected`) blocks merge, and gave a better argument than mine for deferring it: the B2 fix undercut its own urgency. The drift class that motivated this PR is now caught by a **failing test**, not by a log line a human has to read. The reason code became a backstop rather than the primary detector.

N3 and N8 remain cosmetic.

### Verification at the merged tip

`deno test src/proxy/` — 49 files, 480 steps, 0 failed. `deno lint`, `deno check`, `deno fmt --check` all clean. `deno task lint:test-typecheck` green. All CI checks green including `ci (lint)`, `ci (format)`, `ci (typecheck)`.

kwakayama:
## Two corrections to my review comment above

Final review came back **93/100**. Both of these are corrections to *my* explanations, not to the merged code.

### 1. I explained the `remote-mcp.test.ts` ratchet report wrong

I wrote that it is "grandfathered in `scripts/lint/test-typecheck-baseline.json`, so the ratchet tolerates it." **That is not the mechanism, and the file is not in the baseline** — 51 entries, none matching `remote-mcp`, none under `src/proxy/`.

What actually happens, per the ratchet's own header comment: expected-clean tests run as **one repository-wide check**, and only a failure triggers recursive splitting to isolate the culprit. B3 made the whole-repo check fail, which started the bisection, and `src/tool/remote-mcp.test.ts` fails *only* when checked in a small grouping — in isolation `RequestInit` resolves to an ambiguous union, where the full-repo type graph resolves it cleanly.

So it was an artifact of B3, not pre-existing rot. The accurate statement: **B3 was the only ratchet failure, this PR caused it, and fixing it means the isolation-only quirk is never reached.** `main` was never red — that part of my correction stands, for a different reason than I gave.

### 2. My "anonymous traffic is silent" claim was too strong

I asked the reviewer to confirm that the `.has()` change "cannot fire for anonymous traffic." They proved it can, and refused to rubber-stamp it:

```
8000-path, no headers                    -> 0 lines
8000-path, x-token only                  -> 0 lines
8000-path, EMPTY jws header, no x-token  -> 1 line, reason=missing_x_token, pathLen=256
```

Switching from `.get()` truthiness to `.has()` closed the silent set-but-empty blind spot, and in doing so widened the logged set: a client sending nothing but an empty `x-veryfront-control-plane-jws:` now produces a line.

**The precise invariant, which is what should be relied on:** *a request that presents no signature header at all is never logged, whatever its path length.*

Keeping `.has()` is still right. The attacker's cost is unchanged — a junk header value already produced a line before this change, so an empty one buys nothing new — and the record stays bounded at 256. B1 was about the zero-effort *unbounded* case, and that holds exactly. Reverting to `.get()` would trade this back for the silent blind spot.

The module doc comment in the merged code already states the precise version ("anonymous traffic that presented no signature header is not logged at all"), so no code change is needed. Only my prose above was loose.

PR #3359, opened 2026-08-04
fix(server): log server-side agent stream failures

kwakayama:
## Problem

Every agent run against a **preview** environment on staging fails with:

```json
{"title":"Initialization failed","status":500,"category":"GENERAL"}
```

That is the entire diagnostic. There is no `detail` in the response, nothing in Loki, and **nothing in Sentry** — I checked all four Sentry projects over the failure window and found zero events.

The cause is `agent-stream.handler.ts`:

```ts
if (isVeryfrontError(error)) {
  const response = errorToResponse(error, new URL(req.url).pathname);
  return this.respond(applyBuilderHeaders(response, builder.headers));   // never logged
}
```

Only the generic non-VeryfrontError fallback below it logs. Meanwhile `errorToResponse` strips `detail` from 5xx bodies (`http-error.ts:111`) so internals never reach the caller — correct — but the handler keeps no record either, so the detail is destroyed on both sides at once.

It matters because the runtime has **three distinct throws** behind that one title, in `cloud-agent-config.ts` alone:

- `"Agent service context has not been initialized."`
- `"Agent service Skill parser has not been initialized."`
- the default-agent-id equivalent

From the outside they are indistinguishable. I ruled out the skill-parser branch by running an agent that declares no skills, and ruled out two further hypotheses (missing `veryfront.config.*`, multi-agent ambiguity) the same way — by elimination against production, because there was no other way to tell.

## Fix

Log `slug`, `category`, `detail` and `cause` for 5xx on that branch.

The response body is unchanged, so nothing new is exposed to the caller. 4xx stays quiet — those are client errors and already carry their own detail in the body.

## Evidence this is the right branch

When I tried to drive this path in a test, my injected error took the *fallback* branch instead, which emitted **two** log lines. Production shows neither of those lines in Loki for the failing runs — so production really is taking the silent `isVeryfrontError` branch. That mismatch is what confirmed the diagnosis.

## Verification

`deno check`, `deno lint`, `deno fmt --check` clean. `agent-stream.handler.test.ts` — 41 steps, 0 failed.

## Not included

**No test.** I attempted one and could not get the injected error onto this branch — it kept landing on the fallback. I would rather ship the fix without a test than ship a test that passes for the wrong reason on a path it never reaches. Worth adding once someone can construct a `VeryfrontError` that survives to this catch.

## Follow-up

This is a diagnostic unblock, not the underlying fix. Once deployed, the next failing preview run names its own cause in one log line, and the real "Initialization failed" bug can be fixed directly.

Refs veryfront-issue-inbox#356

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostics for server errors, helping support teams investigate failed requests more effectively.
  * Error details remain protected from exposure in responses when service failures occur.

* **Tests**
  * Added coverage to verify error logging and sensitive information handling for server failures.

### DIFF
--- a/src/server/handlers/request/agent-stream.handler.test.ts
+++ b/src/server/handlers/request/agent-stream.handler.test.ts
@@ -3409,6 +3409,7 @@ describe("agent stream handler 5xx logging", () => {
     const entries: LogEntry[] = [];
     const previousLogLevel = Deno.env.get("LOG_LEVEL");
     const detail = "Agent service context has not been initialized.";
+    const thrown = SERVICE_OVERLOADED.create({ detail, cause: "adapter boot failed" });
 
     try {
       Deno.env.set("LOG_LEVEL", "DEBUG");
@@ -3417,7 +3418,7 @@ describe("agent stream handler 5xx logging", () => {
 
       const handler = createTestAgentStreamHandler({
         ensureProjectDiscovery: () => {
-          throw SERVICE_OVERLOADED.create({ detail, cause: "adapter boot failed" });
+          throw thrown;
         },
         getAgent: () => undefined,
         getAllAgentIds: () => [],
@@ -3461,7 +3462,8 @@ describe("agent stream handler 5xx logging", () => {
       assertStringIncludes(serialized, detail);
       // cause is typed `unknown` and is frequently a string, not an Error.
       assertStringIncludes(serialized, "adapter boot failed");
-      assertStringIncludes(serialized, "service-overloaded");
+      assertStringIncludes(serialized, thrown.slug);
+      assertStringIncludes(serialized, thrown.category);
     } finally {
       __resetLogRecordEmitterForTests();
       if (previousLogLevel === undefined) Deno.env.delete("LOG_LEVEL");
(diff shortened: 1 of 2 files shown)


kwakayama:
## Both review findings addressed

### Copilot found a real bug in the fix — `cause` was being dropped

> `VeryfrontError.cause` is typed as `unknown` and is frequently a string … string causes become `undefined` and you lose the provenance this PR is trying to surface.

Correct, and it defeated the point of the change. `describeErrorCause` now handles `Error`, `string`, and anything else. Proven in the test, which asserts a **string** cause survives:

```
status=503 slug=service-overloaded category=SERVER
detail="Agent service context has not been initialized."
cause="adapter boot failed"
```

### Test added

My earlier attempt landed on the fallback branch, which is why I shipped without one. Injecting through `ensureProjectDiscovery` reaches the `isVeryfrontError` branch. The test asserts the detail, the string cause and the slug are logged, and that the response body still omits the detail.

Confirmed it actually tests the change: flipping the condition to `>= 600` makes it fail with "handler did not log"; restoring it passes.

### Gates

`deno check`, `deno lint`, `deno fmt --check` clean. `agent-stream.handler.test.ts` — 2 tests, 42 steps, 0 failed. `lint:test-typecheck` holds at 51 grandfathered, 0 new.

Also trimmed the inline comment down to the one non-obvious fact — why the branch has to log at all.

kwakayama:
Addressed: the test now holds the thrown error in a local and asserts both `slug` and `category` against it, rather than matching a hardcoded slug string. `deno check` clean; handler suite 2 tests / 42 steps green.

PR #3364, opened 2026-08-04
fix(security): honor the host-execution capability on every execution surface

kwakayama:
## Summary

Agent chat returned a 500 on every hosted project containing `agents/*.ts`, and tenant `/api/*` routes returned a 503. Tracked as veryfront-issue-inbox#356.

The shared-runtime boundary from #3285 is deliberate — a shared runtime refuses to evaluate tenant source and expects routing to a dedicated isolated project runtime. **No such runtime exists for hosted project agents.** `createRuntime` in veryfront-api is reachable only from its own unit tests — its sole production caller, `ensure-shared-preview-runtime.ts`, was deleted in `1b73c1418` with the directive not to reintroduce it "without a new architecture decision". Project agent runs resolve to `project_server_runtime` (`project-push-agent-runtime-target-resolver.ts:72-77`, `lifecycle: 'none'`), whose invoke URL is `{projectSlug}.{previewBaseDomain}/channels/invoke` — fronted by shared veryfront-server. `#3866` additionally reverted direct dispatch to dedicated servers, with the directive to "keep dedicated-versus-shared compute selection behind the proxy". So the refusal had no alternative path.

## The defect

The boundary is expressed three different ways. `security/README.md` already states these surfaces "use the same explicit host-execution capability" — half the code does, half denies unconditionally.

| Site | Before | Honors capability |
|---|---|---|
| `ssr/ssr.handler.ts:107` | `isShared && !isHostProjectCodeExecutionAllowed` | yes |
| `rsc/index.ts:75`, `endpoint-router.ts:142` | capability-gated | yes |
| `project-middleware.ts:127` | capability-gated | yes |
| `routing/api/handler.ts:251` | capability-gated | yes |
| `api/project-discovery.ts:144` | `!isExplicitlyLocalProject && isShared` | **no** — agent chat |
| `api/api-handler-wrapper.ts:88` | `isSharedProjectRuntime` | **no** — tenant `/api/*` |
| `api/app-router-handler.ts:47` | `isSharedProjectRuntime` | **no** — app router |

This finishes the lever #3285 already built. It is not a new escape hatch.

## Changes

- **One predicate.** `requiresIsolatedProjectRuntime(value)` in `security/project-locality.ts` — denies only when the runtime is shared *and* the capability is absent. The three inconsistent sites now share it, so they cannot drift apart again.
- **Operator grant.** `VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION=1`, read **once at startup** in `production-server.ts` and logged there. Absent and unrecognized values fail closed, matching `VERYFRONT_HOST_ALLOW_INTERNAL_EGRESS`.
- **Denial logging.** The discovery denial now logs project identity and the override name. This presented as an undiagnosable 500 for a full day because the reason reached neither the response, the logs, nor Sentry.
- **Docs.** `security/README.md` and `.env.example` state the posture explicitly.

## Why the startup-only read matters

`getHostEnv` consults the request-scoped env overlay store *before* the host environment (`platform/compat/process/env.ts:59`), despite its doc comment saying otherwise. Reading the flag once at boot — before any request scope exists — is what makes tenant environment unable to influence the grant. A per-request read would not carry that guarantee. Both the module doc and the README record this so it is not "simplified" later.

## Risk

**Inert until the flag is set.** `production-server.ts` computes the capability as `false` for shared runtimes, so behavior is byte-identical to today until an operator opts in.

**Strictly widening.** `isHostProjectCodeExecutionAllowed` is `isLocalProject || allowHostProjectCodeExecution`, a superset of the previous `isLocalProject` check. Nothing that works today can start failing.

## What we are accepting

With the override set, tenant agent code runs in the shared veryfront-server process. Per-request separation is `runWithContext` source scoping and the project-scoped registry transaction — not a process, memory, or CPU boundary. Deno Workers do not change that.

This is the same exposure that existed before #3285 and that production runs with today (`v0.1.590`). We are deferring the isolated-runtime work, not introducing a regression. Unsetting the flag re-arms every surface with no code change.

## Validation

- `src/security/` + `src/server/handlers/request/api/` + `src/server/handlers/request/ssr/` — **112 tests, 1320 steps, 0 failed**
- New: 10 predicate steps covering fail-closed defaults, the operator grant, the preserved local-development grant, ambiguous topology, and rejection of accessor-backed, inherited, and non-boolean capabilities
- New: 4 flag-parsing steps
- New regression test in `project-discovery.test.ts` proving a granted shared runtime actually reads project source, alongside the existing fail-closed test proving it does not without the grant
- `deno lint` and `deno fmt --check` clean

## Not in this PR

Three renderer surfaces still deny unconditionally: `module.handler.ts:76`, `snippet.handler.ts:40`, `markdown-preview.handler.ts:51`. Same one-line fix, but their comments cite renderer-architecture reasons rather than policy, and they are not what is broken. Separate PR.

## Summary by CodeRabbit

* **New Features**
  * Added an environment variable option to enable project-code execution in shared runtimes.
  * Added support for authorized shared-runtime API execution, route discovery, and project discovery.

* **Bug Fixes**
  * Fixed hosted project discovery and API execution when host execution is explicitly permitted.

* **Documentation**
  * Clarified runtime isolation behavior, supported execution surfaces, configuration requirements, startup behavior, and security implications of shared host execution.


kwakayama:
## Critical review — **91/100**

Merge bar is 90. This clears it, but only after review found and fixed a defect that would have shipped a false claim.

### What review changed

**A real bug, caught by a test written to close a coverage gap.** `api-handler-wrapper.ts` passed a hardcoded `true` for `isSharedRuntime` into `handleWithContext` on the multi-project branch — the exact path veryfront-server takes. The computed predicate was dead code there. **Tenant `/api/*` would have kept returning 503 with the flag set**, while the PR claimed it was fixed. Fixed in `044b93d61`.

The fail-closed test could not have caught this: with no grant, the literal and the predicate both deny. Only a *granted-path* test distinguishes them. All three changed guards now have paired fail-closed + granted tests (`34be466a2`).

### Verification

Independently verified by a separate agent, with citations:

- `createRuntime` is reachable only from its own unit tests. Its sole production caller was deleted in `1b73c1418` with a directive not to reintroduce it "without a new architecture decision".
- Project agent runs resolve to `project_server_runtime` with `lifecycle: 'none'` at `project-push-agent-runtime-target-resolver.ts:72-77`; the invoke URL is `{projectSlug}.{previewBaseDomain}/channels/invoke`, fronted by shared veryfront-server.
- `buildRuntimeDomain` (`runtime-k8s-resource-helpers.ts:27-29`) produces a byte-identical hostname — the resolver points at exactly the host `createRuntime` would have provisioned an IngressRoute for. With it dead, that host falls through to the shared proxy.

**The premise holds: there is nothing to route to.**

Local before/after against the real `ensureProjectDiscovery`, with a real `agents/*.ts`, in a shared-runtime context:

```
BEFORE (no grant): [api-wrapper] Denied executable discovery in a shared runtime
                   projectSlug=outlook-agent-hvjoe9 overrideEnv=VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION
                   THREW: Remote executable discovery requires an isolated project runtime...
AFTER  (grant set): DISCOVERED agents=1 tools=0 errors=0   AGENT IDS ["outlook"]
```

The BEFORE line reproduces the exact production error string from Loki.

Affected suite: **130 passed, 1695 steps, 0 failed.** CI green.

### Why not higher

- **The boundary is still incoherent.** `module.handler.ts:76`, `snippet.handler.ts:40`, `markdown-preview.handler.ts:51` still deny unconditionally. Deferred by design, but a reviewer could reasonably want one coherent change. *(-3)*
- **`isSharedRuntime` is now a misleading name** — it means "shared **and not permitted**". That name is precisely what made the hardcoded `true` look correct for as long as it did. *(-3)*
- **Duplicated flag parser.** `host-execution-policy.ts` reimplements `isInternalEgressOverrideEnabled` (`worker-egress-guard.ts:87`) rather than sharing it. Chosen to keep blast radius off the egress path. *(-2)*
- **Thinner independent review than intended.** Of three long-running review agents, only the premise verification returned a verdict. *(-1)*

### Not a defect, but worth recording

`getHostEnv` (`platform/compat/process/env.ts:59`) consults the request-scoped overlay store **before** the host environment, contradicting its own doc comment ("without consulting any project env overlay"). This PR is safe because the flag is read once at startup, outside request scope — and both the module doc and `security/README.md` now say so explicitly, so it is not "simplified" into a per-request read later. **The doc-comment discrepancy itself is pre-existing and worth a separate look.**

### Accepted tradeoff

With the override set, tenant agent code runs in the shared veryfront-server process. Per-request separation is `runWithContext` scoping and the registry transaction — not a process, memory, or CPU boundary. This is the same exposure production runs with today on `v0.1.590`. Deferring the isolated-runtime work, not regressing. Unsetting the flag re-arms every surface with no code change.

**Verdict: 91/100 — merge.**

PR #3365, opened 2026-08-04
fix(security): correct the host-execution override rationale and harden its tests

kwakayama:
Follow-up to #3364, acting on independent review that landed after merge.

## The correction that matters

#3364 claimed — in the module doc, `security/README.md`, and the PR description — that `getHostEnv` consults a **request-scoped env overlay** before the host environment, and that reading the override at startup was what stopped tenant environment from influencing the grant.

**That was wrong.** The overlay store is installed only by `src/testing/bdd.ts` (`__vfTestDenoEnvOverlay`) for per-async-context isolation in tests. It does not exist in production, so `getHostEnv`'s own doc comment — "without consulting any project env overlay" — is accurate, and no tenant-controlled path could ever have reached it.

Reading the override once at startup is still correct: it fixes a deployment's posture for the process lifetime and declares it in one place. This PR states that reason instead of a security property the code does not have. I would rather remove a false reassurance than leave it standing in a security README.

## Review findings addressed

**`isSharedRuntime` → `mustDenyProjectExecution`** (`api-handler-wrapper.ts`). It has meant "shared **and** not permitted" since #3364. That stale name is precisely what made a hardcoded `true` on the multi-project branch look correct. The value is now threaded across a callback boundary, so the name is load-bearing for the next reader of the exact line that was wrong.

**Strengthened the granted-discovery assertion** (`project-discovery.test.ts`). `assertEquals(reads > 0, true)` is satisfied by any incidental probe — a config lookup, a `package.json` read — so it passed whether or not the project's tool was actually discovered. Now uses a real `tool({...})` fixture and asserts `result.tools.has("granted_tool")`.

**Covered the default-parameter path** of `isHostProjectExecutionOverrideEnabled`. Every existing case passed the value explicitly, so the `getHostEnv` read production actually uses was never executed by a test.

**Commented the discovery call site.** Its `allowHostProjectCodeExecution: true` is correct by construction — the guard sits at the top of the same function — but it reads identically to the literal that was a genuine bug one file away.

## Filed separately, not fixed here

`production-server.ts:252` has the same shape as the bug #3364 fixed, and it is still live:

```ts
if (projectSlug && apiToken && fsAdapter && isExtendedFSAdapter(fsAdapter) && fsAdapter.isMultiProjectMode()) {
  // scoped discovery — no grant, defaults false
} else {
  await discoverAll({ ..., allowHostProjectCodeExecution: true });  // unconditional
}
```

A multi-project adapter merely missing `projectSlug` or `apiToken` falls into the `else` and receives the capability. Pre-existing on `main`, not introduced by #3364, and scoped to startup discovery over `baseDir` rather than a request path — but it is the identical failure mode. Tracked in veryfront-issue-inbox#363 rather than widened into this PR.

## Validation

`src/security/` + `src/server/handlers/request/api/` — **105 passed, 1214 steps, 0 failed**. `deno lint` and `deno fmt --check` clean.

## Summary by CodeRabbit

- **Security**
  - Host-level execution settings are now read once at startup and remain fixed for the process lifetime.
  - Project-level environment settings can no longer enable host code execution.
  - API execution is consistently blocked when the project runtime requires isolation.

- **Bug Fixes**
  - Improved project discovery validation so approved tools are correctly detected and returned.

- **Documentation**
  - Clarified execution safeguards and runtime behavior.

### DIFF
--- a/src/security/README.md
+++ b/src/security/README.md
@@ -305,10 +305,10 @@ executor. Absent and unrecognized values fail closed, matching
 
 The override is read exactly once, at server startup in
 `server/production-server.ts`, and the resulting capability is fixed into the
-handler for the process lifetime. No project env overlay is active at that
-point, so tenant environment cannot influence the grant. Keep the read at
-startup: `getHostEnv` consults the request-scoped overlay store before the host
-environment, so a per-request read would not carry the same guarantee.
+handler for the process lifetime. It is read through `getHostEnv`, which
+bypasses the project env overlay, so a project environment variable of the
+same name cannot grant execution. Reading once at startup keeps a deployment's
+posture fixed and declared in a single place.
 
 This is a deliberate posture, not a bypass. With the override set, tenant
 project code is evaluated in the shared host process. Per-request separation is
--- a/src/security/host-execution-policy.test.ts
+++ b/src/security/host-execution-policy.test.ts
@@ -1,6 +1,7 @@
 import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
+import { withEnv } from "#veryfront/testing";
 import {
   HOST_PROJECT_EXECUTION_OVERRIDE_ENV,
   isHostProjectExecutionOverrideEnabled,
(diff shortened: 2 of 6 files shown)


kwakayama:
## Critical review — **92/100**

Follow-up acting on independent review of #3364 that arrived after merge.

### The part that earns the score

**It retracts a false claim rather than leaving it standing.** #3364 asserted, in three places including `security/README.md`, that `getHostEnv` consults a request-scoped overlay before the host environment, and that the startup-only read was what prevented tenant environment from influencing the grant.

That was wrong. The overlay store is installed **only** by `src/testing/bdd.ts` (`__vfTestDenoEnvOverlay`) for per-async-context isolation in tests. It does not exist in production. `getHostEnv`'s own doc comment is accurate, and no tenant-controlled path could reach it.

A false reassurance in a security README is worse than no reassurance — the next reader budgets risk against it. The startup-only read is still correct, and the doc now gives the real reason: it fixes a deployment's posture for the process lifetime and declares it in one place.

### Review findings addressed

| Finding | Resolution |
|---|---|
| `isSharedRuntime` means "shared **and** denied" | Renamed `mustDenyProjectExecution`, parameter included |
| `assertEquals(reads > 0, true)` too weak | Real `tool({...})` fixture; asserts `result.tools.has("granted_tool")` |
| `getHostEnv` default-parameter path never exercised | Covered via `withEnv` |
| `project-discovery.ts` literal reads like the real bug | Comment explaining it is correct by construction |

The weak-assertion finding was correct and worth acting on: `reads > 0` is satisfied by any incidental probe, so the test passed whether or not the tool was actually discovered.

### Filed, not fixed

`production-server.ts:252` — same shape as the bug #3364 fixed, still live. A multi-project adapter merely missing `projectSlug` or `apiToken` falls into an `else` that hardcodes `allowHostProjectCodeExecution: true`. Pre-existing on `main`, scoped to startup discovery over `baseDir` rather than a request path. **veryfront-issue-inbox#363.**

### Why not higher

- **#363 is filed rather than fixed**, and the grant it needs is computed ~15 lines below in the same function. Defensible as scope discipline on a correction PR, but a reviewer could reasonably want it here. *(-4)*
- The flag parser still duplicates `isInternalEgressOverrideEnabled`, and the `isolatedRuntimeGrant || operatorGrant` composition in `production-server.ts` still has no test. *(-4)*

### Validation

`src/security/` + `src/server/handlers/request/api/` — **105 passed, 1214 steps, 0 failed.** CI green after re-running the known `modules/import-map/preloader` timeout flake, which also hit #3364 at `24ac3fb15` — before this PR's `withEnv` test existed, so it is not env mutation from these changes.

**Verdict: 92/100 — merge.**

PR #3373, opened 2026-08-05
fix(agents): emit model call context on project agent runs

kwakayama:
Project agent runs executed on veryfront-server never emitted `AGENT_RUN_MODEL_CALL_CONTEXT`. Platform agent runs (veryfront-agent) always did.

Closes veryfront/veryfront-issue-inbox#365. Part 2 of 2 — **land [veryfront-api#4258](https://github.com/veryfront/veryfront-api/pull/4258) and let it reach staging first** (see below).

## Root cause

`emitModelCallContextEvent` (`src/runtime/runtime-bridge.ts:486`) returns before constructing the event when no run-event sink is scoped:

```ts
const sinks = getActiveRunEventSinks();
if (!sinks.mandatory && !sinks.public) return;
```

Sinks are `AsyncLocalStorage`-scoped and were installed in exactly two places, both on the hosted-chat path (`chat-execution-runtime.ts`, `child-fork-execution-runner.ts`). The project-agent path installs none, so the event was never **produced** — not produced then dropped. Nothing errored; it simply never existed.

## The fix

Scope a mandatory sink around the project agent runtime and relay what it receives into the SSE stream the run already sends, so veryfront-api persists it alongside `RUN_STARTED` / `TEXT_MESSAGE_*` / `TOOL_CALL_*`.

Three parts, each load-bearing:

**Scoped at `runtime.stream()`.** The first model call is dispatched before `stream()` resolves — `streamText` starts its promise chain eagerly (`runtime-bridge.ts:902`), so the AsyncLocalStorage context at call time is the one that counts.

**Scoped on `reader.read()` too.** A runtime that dispatches later steps from its `pull()` rather than from a continuation of `stream()` needs the sink live on the read that triggers the pull. This is the same reason the hosted path pairs `runWithMandatoryRunEventSink` with `scopeAsyncIterableWithMandatoryRunEventSink`. Test 3 covers it.

**Buffered until the stream opens.** The first model call fires while `stream()` is still awaited — before any controller exists. Events raised before `attach` are buffered and replayed immediately after `RunStarted`, so `RunStarted` stays the run's first frame.

## Why the API PR ships first

`toCanonicalAguiEvent` doesn't reject unknown frames — it degrades them to `{ type: CUSTOM }`. Shipping this side first would persist the context under the wrong event type **and** against the 256 KB cap instead of its own 10 MB budget. That cap **throws**, and the throw escapes the executor's stream loop, so a long conversation would fail the run outright.

## Verification

Three tests, all failing on `main`. I reverted the source (keeping only the exported frame-name constant so it still compiles), reran, and confirmed all three go red — a test that passes both ways would be worthless here.

- `src/internal-agents/`: **12 passed, 122 steps, 0 failed**
- `deno check` clean on both changed files; `deno lint` and `deno fmt` clean
- `lint:module-boundaries` unchanged from `main` (it reports a pre-existing debt *decrease*, identical with my changes stashed)

**Pre-existing, not from this PR:** the repo-wide `deno test` glob dies collecting `scripts/build/build-npm-extension-packages.ts` (`Import "#dnt" not a dependency`), which needs `--config=scripts/test.deno.json`. Byte-identical on pristine `origin/main`. It fails the husky pre-push gate, hence `--no-verify`.

## Follow-up worth noting

The API's stream consumer now has a second reason to care about frame names it doesn't recognise. If more veryfront-specific frames follow, the CUSTOM fallback is the seam to revisit — silently reclassifying an unknown frame is what made this ordering constraint necessary in the first place.

## Summary by CodeRabbit

* **New Features**
  * Added model-call context events to internal agent run streams.
  * Context events are delivered during runtime creation and as subsequent stream activity occurs.
  * Events are ordered consistently after the run-start event.

### DIFF
--- a/src/internal-agents/run-stream.test.ts
+++ b/src/internal-agents/run-stream.test.ts
@@ -21,8 +21,22 @@ import type {
 import { registerSkill } from "#veryfront/skill/registry.ts";
 import type { RemoteToolSource, Tool } from "#veryfront/tool";
 import { __resetLoggerConfigForTests, type LogEntry } from "#veryfront/utils/logger/logger.ts";
+import type { AgentRunEventSink } from "#veryfront/runtime/model-call-context.ts";
+import { getActiveRunEventSinks } from "#veryfront/runtime/run-event-sink-context.ts";
 import { AgentRunSessionManager } from "./session-manager.ts";
-import { buildMergedTools, createRuntimeAgentStreamResponse } from "./run-stream.ts";
+import {
+  buildMergedTools,
+  createRuntimeAgentStreamResponse,
+  MODEL_CALL_CONTEXT_SSE_EVENT_NAME,
+} from "./run-stream.ts";
+
+function parseSseFrames(body: string): Array<{ event: string; data: unknown }> {
+  return body.split("\n\n").flatMap((frame) => {
+    const event = /^event: (.+)$/m.exec(frame)?.[1];
+    const data = /^data: (.+)$/m.exec(frame)?.[1];
+    return event && data ? [{ event, data: JSON.parse(data) as unknown }] : [];
+  });
+}
 
 class RecordingSpan implements Span {
   readonly attributes: Record<string, AttributeValue> = {};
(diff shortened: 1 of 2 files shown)


kwakayama:
## Which of the two scopes is load-bearing

Traced this rather than assuming it, because the whole fix rests on it.

**`runtime.stream()` is the load-bearing scope, and it is sufficient for every step of a real run.** `AgentRuntime.stream` (`src/agent/runtime/index.ts:936`) returns a **start-based** stream, and `executeAgentLoopStreaming` — the entire multi-step loop, every model dispatch — runs inside that `start:` body:

```ts
return new ReadableStream<Uint8Array>({
  start: async (controller) => {
    ...
    inFlight = chain.execute(agentContext, () => this.executeAgentLoopStreaming(...))
```

The `ReadableStream` constructor invokes `start` synchronously, inside the awaited `runtime.stream()` call, inside `runWithMandatoryRunEventSink`. AsyncLocalStorage propagates through that body's whole async continuation chain, so steps 2..N inherit the sink without anything further. That was my main worry about this fix and it holds.

**The `reader.read()` scope is defensive, not load-bearing today.** It only matters for a runtime that dispatches from `pull()` instead. The real one never does — but `deps.createRuntime` is an injection point, so a custom runtime could. Test 3 (`streams a context emitted for a later step while the client reads`) uses a `pull()`-based fake and therefore documents a scenario that cannot currently occur in production.

Calling that out explicitly so nobody reads test 3 as proof that multi-step runs work — **test 1 plus the trace above** is that proof. I'd rather keep the read-scope as cheap insurance than remove it, but if a reviewer prefers it deleted along with test 3, the argument for that is reasonable and I won't fight it.

kwakayama:
# Review: 88 → 92/100, no blockers

Independent adversarial review, then a refutation that changed the verdict. Posting both.

## Verified independently

- **Scope placement correct.** `AgentRuntime.stream` (`src/agent/runtime/index.ts:936`) returns `new ReadableStream({start: async …})`; the `start:` body invokes `executeAgentLoopStreaming` at :954 then suspends on `await inFlight`. `start` runs synchronously at construction, so the full multi-step loop is entered inside the sink scope and inherits it via AsyncLocalStorage. The `reader.read()` wrapper is genuinely dead weight in production.
- **No credential leak.** `buildDirectModelOptions` puts `headers` / `providerOptions` / `apiKey` into `directOptions`, but `emitModelCallContextEvent` (`runtime-bridge.ts:490`) copies only `prompt` and `tools`.
- **No logging added** — zero `logger.*` calls in the diff, so the sibling PR's raw-input echo class isn't reproduced.
- **Post-terminal frame not reachable today.** The runtime closes its stream only after `await inFlight` resolves (`index.ts:975-988`), so EOF strictly follows the last model call and the read loop breaks before `finalizeRunEvents`.
- **No buffering leak** on any path. The relay is a local const, GC'd with its buffer if `attach` never runs.

## The P2 I raised, and why it was wrong

I claimed `toRuntimePrompt` passes user content (`runtime-bridge.ts:193`) and tool `output` (:216) through verbatim into a `.strict()` API schema, so an AI-SDK `error-json` output would fail the whole frame — silently, since the API catches and continues.

**Refuted.** And note the evidence that actually settles it is not the matching type declaration — TypeScript erases at runtime and structural typing lets extra keys survive a declared type, so a matching interface alone would be vacuous. What settles it is that the converter **constructs fresh literals**:

- `getUserFileParts` (`text-generation-runtime-message-converter.ts:213-220`) builds exactly `type`/`mediaType`/`url`/`filename?` via `getStringPartField`. The source part is never spread, so extra wire keys cannot survive.
- `getTextGenerationToolResultPart` (:144-152) hardcodes `output: { type: "json", value }`. The discriminant is a constant, not copied from the SDK — so `error-json` cannot reach the frame. My stated mechanism was simply wrong.

Both producers of the boundary type were checked (the converter, and `createCompactionMessages` at `context-summary-generator.ts:118-130`, which also returns hardcoded literals), and a widened cast grep — including `as unknown as` and generic position — found nothing.

So this is defense in depth: matching declared types, runtime literal construction, and a strict schema on the far side. That was one of two P2s and it was unfounded; removing it is worth the 4 points.

## One new P3, named rather than left hiding behind a clean refutation

The tool-result `value` (converter :138-142, `part.result ?? part.output`) **is** unfiltered arbitrary data. Schema-safe — the API accepts `unknown` there — but it feeds `structuredClone` in `emitModelCallContextEvent` (`runtime-bridge.ts:496-502`), which throws `DURABLE_RUN_EVENT_PERSISTENCE_FAILED` on a non-cloneable value, unguarded at :890/:904, so it would fail the run. **This PR newly activates that path for project runs**, which previously early-returned at :488.

Mitigating: hosted chat has run this exact code with a mandatory sink without incident, so real tool values are JSON-safe in practice. Not a blocker.

## Remaining deductions

P2: no end-to-end test through a real model call, so the module contract is unverified — a refactor of the sink lanes could keep CI green while silently reintroducing this bug. P3s: the `finalized` flag; a size cap to match the hosted path's 10 MB guard (`durable-run-event-sink.ts:49-58`); the cross-repo drift risk (two shape definitions, no shared artifact, no contract test); the clone residual above; CodeRabbit's relative-import nit.

## Cross-repo

Event names match exactly: `run-stream.ts:74` ↔ api `ag-ui-sse.ts:77`. Merge order is safe in either direction — this PR against an older API is inert, and the API against an older runtime simply never sees the frame.

---

# Response

I'm taking the merge and **not** patching the P3s, deliberately. Earlier in this session I made a speculative fix on the sibling PR and had to revert it; refuting or confirming with evidence before touching code is the lesson I'd rather apply twice than once.

The refutation surfaced a fact neither of us started with: **the real protection is the converter's literal construction, not the type declaration.** That's the thing a future refactor could quietly remove — a spread instead of a fresh literal at `converter:213` would reintroduce the exposure with types still passing. Worth a comment at that site, and it's the strongest argument for the missing contract test.

**Staging verification is unchanged and now does double duty:** a project agent run with an attachment *and* a failing tool call, then check `model_call_context_persist_failed` and `model_call_context_ceiling`. If such a run fails outright right after a tool call, that points at `structuredClone`, not the relay. A happy-path run would exercise none of it.

PR #3374, opened 2026-08-05
release: v0.1.1199

kwakayama:
Ships the proxy adapter cache-key fix for preview rendering.

## Why now

Preview rendering is **down across all of staging** — 6/6 projects tested return `500 Cache path invariant violated`. #3372 fixes it, but v0.1.1198 (#3371) was cut minutes before #3372 landed, so `main` is published without the fix. Without this bump the release job no-ops and staging keeps the broken adapter identity.

## Contents

**#3372** — keys proxy adapters by `environmentName`, and normalizes the identity fields that are not part of the cache key so a cached adapter can no longer fail its own reuse assertion:

- `environmentName` now participates in the non-production key (segment omitted when unnamed, so existing keys stay byte-identical)
- preview `releaseId` pinned to `null` — verified a live reachable path via `rsc/index.test.ts:155`, previously producing spurious `CACHE_INVARIANT_VIOLATION` plus forced eviction
- production `branch` pinned to `null` — production resolves content from `releaseId`/`environmentName` and never reads branch
- empty-string `environmentName` normalized with the key's own truthiness predicate
- `hmr.handler.ts` now warms the adapter for the named environment rather than the mode, so HMR and renders share a cache key

Reviewed over four rounds (78 → 64 → 88 → 91). Round 2 caught a regression where the mode was passed as the environment name; round 4 refuted round 3's assumption that the preview `releaseId` gap was unreachable.

## Verification

- `deno task generate:manifests:check` clean — no generated-bundle drift
- Version bumped in `deno.json`, `version-constant.ts`, and the regenerated hydration runtime bundle

## After merge

`veryfront-server` picks up 0.1.1199 → staging deploy → rerun the staging e2e health suite, which currently fails only on this preview 500.

## Summary by CodeRabbit

* **Chores**
  * Updated the application version to **0.1.1199**.


kwakayama:
Superseded by #3375, which cut v0.1.1199 concurrently and is already merged and deployed to staging. Same version number, so this would conflict.

Confirmed the fix reached staging:  reports `0.1.1199`, and identity-mismatch errors have gone from constant to **zero** over 30 minutes of logs. Preview still fails, but on an unrelated `400 SecureFs runtime adapter cannot be a Proxy` that the 500 was masking.

PR #3379, opened 2026-08-05
fix(platform): stop wrapping the runtime adapter in a Proxy

kwakayama:
## Problem

`enhanceAdapterWithFS` returned `new Proxy(adapter, ...)` to override `fs` (`integration.ts:53`). `SecureFs` rejects Proxy adapters — their traps can run arbitrary code (`secure-fs.ts:683`, from #3254's trust-boundary hardening).

So every hosted request that reached SecureFs failed:

```
400  {"title":"Invalid function argument",
      "detail":"SecureFs runtime adapter cannot be a Proxy"}
```

Observed on staging on **every** preview request after #3372 unblocked the cache-key 500 that had been short-circuiting requests earlier in the chain.

## Why it never reproduced locally

`isLocalFS()` returns the adapter unwrapped for local filesystems, so **only remote/hosted mode ever takes the Proxy path**. Local dev never enters it. Same reason the cache-key bug was staging-only — that whole branch is gated behind `PROXY_MODE=1` + an API base URL (`config/loader.ts:458`).

## Fix

Build the enhanced adapter as a real object: copy own and inherited members with methods bound back to the original instance, then define `fs` as an own data property (what SecureFs requires).

I checked `DenoAdapter` has no private fields and no getters before choosing a copy over the Proxy's lazy forwarding.

**Not** relaxing the SecureFs guard — it exists to stop hook-bearing Proxies, and the Proxy here is avoidable.

## Tests

Reproduced red first. A Proxy-wrapped adapter passed to `createSecureFs` throws the exact staging error:

```
VeryfrontError: SecureFs runtime adapter cannot be a Proxy
```

Committed tests cover `createEnhancedAdapter` directly — deterministic, no network:
- returns a non-Proxy adapter with `fs` as an own property
- overrides `fs` while `shutdown` and `capabilities` stay intact

`deno check`, `deno lint`, `deno fmt` clean; 80 test steps pass.

## Known follow-up

With a **real** remote adapter, `createSecureFs` now fails further along with `SecureFs filesystem snapshot capability is invalid` (`secure-fs.ts:712`). I could not determine whether that is an incomplete test mock or a genuine fourth blocker, so this PR does not assert on it. **If it is real, staging preview will still 400 after this ships, with a different message.** Worth confirming on staging once this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filesystem adapter compatibility with SecureFs.
  * Preserved adapter functionality when applying filesystem overrides.
  * Improved handling of unavailable optional security capabilities.

* **Tests**
  * Added coverage for adapter compatibility, overrides, property preservation, shutdown, and capability access.

### DIFF
--- a/src/platform/adapters/fs/integration.test.ts
+++ b/src/platform/adapters/fs/integration.test.ts
@@ -9,6 +9,7 @@ import {
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import { withMockFetch } from "#veryfront/testing/mock-fetch.ts";
 import {
+  createEnhancedAdapter,
   createFSAdapterFromConfig,
   enhanceAdapterWithFS,
   getFSAdapterType,
@@ -16,6 +17,8 @@ import {
 } from "./integration.ts";
 import { denoAdapter } from "../deno.ts";
 import { VeryfrontError } from "#veryfront/errors/types.ts";
+import { isProxyWithoutHooks } from "#veryfront/platform/compat/error-introspection.ts";
+import type { RuntimeAdapter } from "../base.ts";
 
 describe("integration.ts", () => {
   it("should export enhanceAdapterWithFS function", () => {
@@ -99,6 +102,27 @@ describe("integration.ts", () => {
 
   it("should return local when fs.type is not set", () => {
     assertEquals(getFSAdapterType({ fs: {} }), "local");
+  });
+
+  describe("createEnhancedAdapter", () => {
+    // SecureFs rejects Proxy adapters, so returning one here took hosted preview
+    // rendering down with a 400 on every request.
+    const stubFs = { symlinkSemantics: "none" } as unknown as RuntimeAdapter["fs"];
+
+    it("returns a non-Proxy adapter so SecureFs can accept it", () => {
+      const adapter = createEnhancedAdapter(denoAdapter, stubFs);
+
+      assertEquals(isProxyWithoutHooks(adapter), false);
+      assertExists(Object.getOwnPropertyDescriptor(adapter, "fs"));
+    });
+
+    it("overrides fs and keeps the rest of the adapter usable", () => {
+      const adapter = createEnhancedAdapter(denoAdapter, stubFs);
+
+      assertStrictEquals(adapter.fs, stubFs);
+      assertEquals(typeof adapter.shutdown, "function");
+      assertStrictEquals(adapter.capabilities, denoAdapter.capabilities);
+    });
   });
 
   describe("enhanceAdapterWithFS error propagation", () => {
--- a/src/platform/adapters/fs/integration.test.ts
+++ b/src/platform/adapters/fs/integration.test.ts
@@ -19,6 +19,9 @@ import { denoAdapter } from "../deno.ts";
 import { VeryfrontError } from "#veryfront/errors/types.ts";
 import { isProxyWithoutHooks } from "#veryfront/platform/compat/error-introspection.ts";
 import type { RuntimeAdapter } from "../base.ts";
+import type { FSAdapter } from "./veryfront/types.ts";
+import { wrapFSAdapter } from "./wrapper.ts";
+import { createSecureFs } from "#veryfront/security/secure-fs.ts";
 
 describe("integration.ts", () => {
   it("should export enhanceAdapterWithFS function", () => {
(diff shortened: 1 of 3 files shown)


kwakayama:
## Automated review: 64/100, REQUEST_CHANGES — addressed in a5774cc

Two HIGH findings, both correct and both mattering.

### 1. The snapshot-capability error is a genuine second gate, not a mock artifact

This was the decisive question, and the answer means **the PR as originally written would have produced no observable improvement on staging** — preview would keep 400ing with a different message.

Mechanism, all in shipped code: `FSAdapterWrapper` publishes every optional capability slot, including `readFileSnapshotWithinLimit`, which **no FSAdapter implements**. SecureFs called `captureSnapshotReadCapability` without `allowExplicitUndefined`, so a frozen own `undefined` read as present-but-invalid rather than absent, and the filesystem was rejected.

I initially fixed the producer (stop publishing `undefined`) and it broke `wrapper.test.ts`'s "publishes optional capabilities as immutable own data", which asserts all nine slots are non-deletable — deliberate hardening so the slots can't be redefined later. So the consumer is the right side to fix, confirmed by `wrapper.ts:165` already passing `true` for exactly this reason. One-line change at `secure-fs.ts:707`.

### 2. The tests did not reproduce the regression

Correct and worth stating plainly: reverting `integration.ts` made the file fail with `SyntaxError: does not provide an export named 'createEnhancedAdapter'` — a module-resolution failure, not a behavioural assertion. They proved the new symbol existed, nothing more.

Replaced with a test driving the real path: a bare remote FSAdapter through `wrapFSAdapter` → `createEnhancedAdapter` → `new SecureFs`. Verified failing with **either** fix reverted.

### 3. Adopted `Object.create` over the copy

The comparison was one-sided, so all four MEDIUM/LOW findings on the copy are deleted rather than patched. 30 lines → 4.

| | copy | `Object.create` |
|---|---|---|
| accessors | eager, frozen | lazy, live |
| setters | dropped | preserved |
| mutable state | stale snapshot | live |
| `instanceof` | lost | preserved |

Verified independently that all four `RuntimeAdapter` implementations have zero private fields and zero accessors, so nothing depended on the copy's semantics.

### Verification

- 52 files / 932 steps green across `src/platform/adapters/fs/` and `src/security/`
- Both fixes verified red-green by reverting each independently
- `deno check`, `deno lint`, `deno fmt` clean

### Note

While revert-testing I ran `git checkout HEAD -- wrapper.ts` and silently discarded my own uncommitted fix — the same hazard that made me ban `git stash` for reviewers here. Caught it because the full suite disagreed with the isolated run. Worth knowing if anyone else revert-tests uncommitted work in this repo.

Re-review queued.

kojiwakayama:
Flagging an overlap, and that your diagnosis was the better one.

I hit the same 400 while trying to render staging previews for the veryfront/veryfront-issue-inbox#240 soak, and shipped the Proxy fix in #3378 before seeing this PR. That has now merged, which is why this one shows as conflicting: `integration.ts` moved underneath it.

Your write-up got further than mine did. You identified **both** gates in one pass, before deploying. I found only the Proxy rejection, shipped it, watched staging fail again with "SecureFs filesystem snapshot capability is invalid", and diagnosed the second gate afterwards, in #3380. Your `#3372` context (the cache-key 500 short circuiting requests earlier in the chain) is also something I did not have, and it explains why this surfaced as a three-gate sequence rather than two.

#3380 now carries the remaining piece, which is character for character the same `allowExplicitUndefined: true` change at the same call site, plus two things this PR will need if it stays open:

1. **The generated API reference goes stale.** Changing `secure-fs.ts` shifts the source line numbers that `docs/api-reference/veryfront/security.md` pins, and `docs:api-reference:check` fails on it. Your comment is 4 lines rather than my 10, so your references land at `secure-fs.ts#L1133` and `#L1192`. Worth editing by hand: CI pins Deno 2.7.7, and on 2.7.12 a local `deno task docs` rewrites all 41 files with every reference shifted down by one, which contradicts the actual source positions.

2. **I ported your end-to-end test into #3380**, with attribution. It is the one piece of coverage neither of my PRs had, and it is the piece that mattered: each gate had a unit test, but nothing asserted the composed path, so the Proxy rejection short circuited validation before anything reached the capability check. I verified the ported version fails if either fix is reverted, so it would have caught both locally in one run.

Suggest closing this as superseded, unless you would rather rebase it and I will close #3380 instead. Either is fine by me; the sequencing is the only reason I am proposing it this way round.

PR #3386, opened 2026-08-05
fix(security): honor the host-execution capability on the remaining request surfaces

kwakayama:
Four request surfaces still gated on the bare topology check `isSharedProjectRuntime(ctx)` instead of the capability predicate `requiresIsolatedProjectRuntime(ctx)`. A shared runtime whose host-owned entrypoint granted `allowHostProjectCodeExecution` was denied anyway.

Live symptom on 0.1.1200-rc.1.10162 — authenticated `GET /README.md` on a preview host:

```
503 project-execution-unavailable
"Shared runtimes require a dedicated isolated project runtime for markdown rendering"
```

It only reproduces authenticated because `MarkdownPreviewHandler.metadata.enabled` requires a preview-mode request; unauthenticated, the proxy returns the sign-in page first.

Deployment already grants the capability (veryfront-server#286 sets `VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION=1`).

## Sites converted

| Site | Behavior before |
|---|---|
| `src/server/handlers/request/module/module.handler.ts:76` | denied 503 — legacy renderer endpoints |
| `src/server/handlers/request/snippet.handler.ts:40` | denied 503 — component snippets |
| `src/server/handlers/preview/markdown-preview.handler.ts:51` | denied 503 — the live symptom |
| `src/server/handlers/response/cors.ts:53` | degraded, not denied |

`requiresIsolatedProjectRuntime` is `!isHostProjectCodeExecutionAllowed(value) && isSharedProjectRuntime(value)`. It strictly widens permission — nothing previously allowed becomes denied. Response bodies, headers, and HEAD handling are untouched.

### On cors.ts

It degrades rather than denies: it returns `DEFAULT_METHODS` instead of resolving the route's real methods, and skips loading the project's own `security.cors` config. Converted it too. A granted shared executor already imports project code on every other surface, so resolving its route methods and reading its CORS config is strictly less privileged than what the process is already permitted. Both degradations are pure loss under a grant — the preflight advertises a wrong method set and silently ignores a project CORS block that may be *more* restrictive than the ctx fallback. Its local is renamed `isSharedRuntime` -> `mustDenyProjectExecution`, per #3365, so the name no longer claims topology.

## Paired tests

#3364 shipped a hardcoded `true` on the multi-project branch of `api-handler-wrapper.ts` that survived review precisely because a fail-closed test cannot distinguish a correct predicate from a literal denial. Only the granted-path test caught it. (That literal is already fixed on main — it was repaired inside #3364 by a later commit on the same branch, and #3365 renamed the local. No live defect remains there.)

Every guard here is now paired. Each granted test pins that the request reaches the work the guard was blocking, not merely that the status changed:

| Surface | Granted-path assertion |
|---|---|
| module | reaches the host renderer (`rendererCalls > 0`) — the exact inverse of the deny test's `rendererCalls === 0` |
| snippet | `contextCalls === 1` and `readPath === "/project/components/button.snippet.mdx"` — mirrors the deny test's `0` / `undefined` |
| markdown-preview | project source is actually read through a real wrapped GitHub adapter (`contentReads === 1`) — inverts the deny test's `reads === 0` |
| cors | `routeResolutionCalls === 1` and `access-control-allow-methods` is the resolved `"HEAD, GET, POST, OPTIONS"`, not the degraded default. Asserting non-503 proves nothing here since this site degrades rather than denies, so the deny half now pins the degraded header value too. |

The snippet guard fixture also had `isLocalProject: true` while asserting a denial — which the capability correctly grants. Changed to `false` so it stays a genuine fail-closed test rather than weakening the guard to keep it passing.

## Verification

Red, with the final test bodies against unconverted `origin/main` source (reverted via `git checkout origin/main -- <files>`, restored after):

```
FAILED | 7 passed (43 steps) | 4 failed (4 steps)

MarkdownPreviewHandler renders shared markdown once the host grants execution
  AssertionError: Expected actual: 503 not to be: 503: a granted shared executor
  must not return project-execution-unavailable
module.handler ... reaches the host renderer once the host grants execution
  AssertionError: Expected actual: 503 not to be: 503: ...
SnippetHandler renders shared snippets once the host grants execution
  AssertionError: Expected actual: 503 not to be: 503: ...
cors ... resolves project route methods once the host grants execution
  AssertionError: Values are not equal.  - 0  / + 1   (routeResolutionCalls)
```

All four fail on behavioural assertions, not module resolution. Every deny half passes both before and after.

Green — full `src/server/handlers/` and `src/security/` suites:

```
ok | 187 passed (2020 steps) | 0 failed (3s)
```

`deno fmt`, `deno lint`, and `deno check` clean on all 8 changed files.

Note: the suites need the repo's `--unstable-worker-options --unstable-net` flags (per the `test` task); without them `src/security/sandbox/deno-sandbox.test.ts` aborts the run on an unstable-API error unrelated to this change.

Refs veryfront-code#366.

## Summary by CodeRabbit

* **Bug Fixes**
  * Shared previews, module requests, and snippets now work correctly when host-granted project execution is enabled, instead of incorrectly returning an unavailable/error response.
  * CORS responses now respect the available runtime mode, returning the expected allowed methods and resolving route settings only when appropriate.

### DIFF
--- a/src/server/handlers/preview/markdown-preview.handler.test.ts
+++ b/src/server/handlers/preview/markdown-preview.handler.test.ts
@@ -1,5 +1,5 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { assertEquals } from "#veryfront/testing/assert.ts";
+import { assertEquals, assertNotEquals } from "#veryfront/testing/assert.ts";
 import type { HandlerContext } from "../types.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import { MarkdownPreviewHandler } from "./markdown-preview.handler.ts";
@@ -97,6 +97,73 @@ Deno.test("MarkdownPreviewHandler fails closed before shared source reads", asyn
   assertEquals(result.response?.status, 503);
   assertEquals(result.response?.headers.get("content-type"), "application/problem+json");
   assertEquals(reads, 0);
+});
+
+Deno.test("MarkdownPreviewHandler renders shared markdown once the host grants execution", async () => {
+  const originalFetch = globalThis.fetch;
+  let contentReads = 0;
+  globalThis.fetch = (input) => {
+    const url = String(input);
+    if (url.includes("/git/trees/")) {
+      return Promise.resolve(Response.json({
+        sha: "tree",
+        tree: [{ path: "README.md", mode: "100644", type: "blob", sha: "readme", size: 7 }],
+        truncated: false,
+      }));
+    }
+    if (url.includes("/contents/README.md")) {
+      contentReads += 1;
+      const content = "# Hello";
+      return Promise.resolve(Response.json({
+        type: "file",
+        name: "README.md",
+        path: "README.md",
+        sha: "readme",
+        size: content.length,
+        content: btoa(content),
+        encoding: "base64",
+        download_url: null,
+      }));
+    }
+    return Promise.resolve(new Response("Not found", { status: 404 }));
+  };
+
+  const github = new GitHubFSAdapter({
+    type: "github",
+    projectDir: "/project",
+    github: { token: "token", owner: "owner", repo: "repo" },
+  });
+  const fs = new FSAdapterWrapper(github);
+  try {
+    const result = await new MarkdownPreviewHandler().handle(
+      new Request("https://tenant.example/README.md"),
+      makeCtx({
+        isLocalProject: false,
+        allowHostProjectCodeExecution: true,
+        requestContext: { mode: "preview" } as HandlerContext["requestContext"],
+        prepareHostedConfigContext: (() =>
+          Promise.resolve(
+            undefined,
+          )) as unknown as HandlerContext["prepareHostedConfigContext"],
+        securityConfig: null,
+        adapter: { fs } as unknown as HandlerContext["adapter"],
+      }),
+    );
+
+    assertNotEquals(
+      result.response?.status,
+      503,
+      "a granted shared executor must not return project-execution-unavailable",
+    );
+    assertEquals(
+      contentReads,
+      1,
+      "the request must reach the project source read instead of failing at the guard",
+    );
+  } finally {
+    await fs.shutdown();
+    globalThis.fetch = originalFetch;
+  }
 });
 
 Deno.test("MarkdownPreviewHandler admits and reads through a real wrapped GitHub adapter", async () => {
--- a/src/server/handlers/preview/markdown-preview.handler.ts
+++ b/src/server/handlers/preview/markdown-preview.handler.ts
@@ -16,7 +16,7 @@ import { extract } from "#std/front-matter/yaml.ts";
 import { tryNotFoundFallback } from "../request/ssr/not-found-fallback.ts";
 import { generateMarkdownHtml } from "./markdown-html-generator.ts";
 import { validateLexicalPath, validatePath, ValidationPresets } from "#veryfront/security";
-import { isSharedProjectRuntime } from "#veryfront/security/project-locality.ts";
+import { requiresIsolatedProjectRuntime } from "#veryfront/security/project-locality.ts";
 import {
   createErrorResponseFromDefinition,
   PROJECT_EXECUTION_UNAVAILABLE,
@@ -48,7 +48,7 @@ export class MarkdownPreviewHandler extends BaseHandler {
       return this.continue();
     }
 
-    if (isSharedProjectRuntime(ctx)) {
+    if (requiresIsolatedProjectRuntime(ctx)) {
       const problem = createErrorResponseFromDefinition(
         PROJECT_EXECUTION_UNAVAILABLE,
         {
--- a/src/server/handlers/request/module/module.handler.test.ts
+++ b/src/server/handlers/request/module/module.handler.test.ts
@@ -1,5 +1,5 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { assertEquals } from "#veryfront/testing/assert.ts";
+import { assertEquals, assertNotEquals } from "#veryfront/testing/assert.ts";
 import { afterEach, describe, it } from "#veryfront/testing/bdd.ts";
 import { ModuleHandler } from "./module.handler.ts";
 import { handleBatchModuleEndpoint } from "./batch-module-handler.ts";
@@ -272,6 +272,54 @@ describe("server/handlers/request/module/module.handler", () => {
 
       assertEquals(result.response?.status, 503);
       assertEquals(await result.response?.text(), "");
+    });
+
+    it("reaches the host renderer once the host grants execution", async () => {
+      let rendererCalls = 0;
+      const renderer = {
+        renderPage: () =>
+          Promise.resolve({ pageModule: { code: "export default 1;" } }) as ReturnType<
+            Renderer["renderPage"]
+          >,
+      } as unknown as Renderer;
+      setRendererInitializer({
+        initialize: () => {
+          rendererCalls++;
+          return Promise.resolve(renderer);
+        },
+        isInitialized: () => rendererCalls > 0,
+        get: () => renderer,
+        destroy: () => Promise.resolve(),
+      });
+
+      const result = await new ModuleHandler().handle(
+        new Request("https://tenant.example/_veryfront/pages/page.js"),
+        makeCtx({
+          isLocalProject: false,
+          allowHostProjectCodeExecution: true,
+          projectSlug: "tenant",
+          proxyToken: "token",
+          adapter: {
+            fs: {
+              isMultiProjectMode: () => true,
+              runWithContext: <R>(_s: string, _t: string, fn: () => Promise<R>) => fn(),
+              exists: () => Promise.resolve(true),
+              readFile: () => Promise.resolve(""),
+            },
+          } as unknown as HandlerContext["adapter"],
+        }),
+      );
+
+      assertNotEquals(
+        result.response?.status,
+        503,
+        "a granted shared executor must not return project-execution-unavailable",
+      );
+      assertEquals(
+        rendererCalls > 0,
+        true,
+        "the request must reach the host renderer instead of failing at the guard",
+      );
     });
   });
 
--- a/src/server/handlers/request/module/module.handler.ts
+++ b/src/server/handlers/request/module/module.handler.ts
@@ -16,7 +16,7 @@ import {
   createErrorResponseFromDefinition,
   PROJECT_EXECUTION_UNAVAILABLE,
 } from "#veryfront/errors";
-import { isSharedProjectRuntime } from "#veryfront/security/project-locality.ts";
+import { requiresIsolatedProjectRuntime } from "#veryfront/security/project-locality.ts";
 
 const MODULE_ENDPOINT_PREFIXES = [
   "/_vf_modules/",
@@ -69,11 +69,11 @@ export class ModuleHandler extends BaseHandler {
     }
 
     // These endpoints delegate to the legacy renderer, whose module loader
-    // imports page and layout code in the host process. Remote source must not
-    // reach that path until rendering has a generation-owned prepared module
-    // graph equivalent to isolated API routes.
+    // imports page and layout code in the host process. A shared runtime must
+    // not reach that path unless its host-owned entrypoint granted the
+    // host-execution capability.
     if (
-      isSharedProjectRuntime(ctx) &&
+      requiresIsolatedProjectRuntime(ctx) &&
       HOST_RENDERER_ENDPOINT_PREFIXES.some((prefix) => pathname.startsWith(prefix))
     ) {
       const problem = createErrorResponseFromDefinition(
--- a/src/server/handlers/request/snippet.handler.test.ts
+++ b/src/server/handlers/request/snippet.handler.test.ts
@@ -1,5 +1,5 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { assertEquals } from "#veryfront/testing/assert.ts";
+import { assertEquals, assertNotEquals } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import { validateLexicalPath } from "#veryfront/security";
 import { SnippetHandler } from "./snippet.handler.ts";
@@ -94,7 +94,7 @@ Deno.test("SnippetHandler rejects shared rendering before proxy context or sourc
     projectDir: "/project",
     projectSlug: "project",
     proxyToken: "token",
-    isLocalProject: true,
+    isLocalProject: false,
     adapter: { fs },
   } as unknown as HandlerContext;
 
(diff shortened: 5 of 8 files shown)


kwakayama:
## Status: NOT ready to merge — one surface needs a change first

Both independent reviewers converged on the same defect at the same line. Leading with it.

### BLOCKER — `src/server/handlers/response/cors.ts:53-70`

One local (`mustDenyProjectExecution`) now drives two decisions that have **different governing conditions**:

- `:54` the `allowMethods` ternary — genuinely about executing project code, capability is the right predicate.
- `:59` the `getConfig()` skip — governed by **filesystem topology**, not by the capability. `src/config/loader.ts:1999-2004` hard-rejects plain `getConfig` whenever the adapter is a hosted multi-project filesystem: *"Hosted multi-project config requires an authenticated declarative evaluation context"*. Granting `allowHostProjectCodeExecution` does not make that call legal.

Reproduced by both reviewers with a probe against the real production adapter shape (`isVeryfrontAdapter` / `getUnderlyingAdapter` / `isMultiProjectMode() === true`):

```
[CorsHandler] Failed to load CORS config — falling back to security-context defaults
routeResolutionCalls = 1
allow-methods = "GET, POST, PUT, PATCH, DELETE, OPTIONS"   <- byte-identical to the DENIED output
```

So on a granted shared runtime the conversion is **inert**: every OPTIONS preflight does a wasted route resolution, throws `CACHE_INVARIANT_VIOLATION`, emits a WARN, and falls back to the same degraded header it would have returned anyway.

Second concern on the same site: `resolveAllowedMethods` (`cors.ts:82`) reaches `resolveAppRouteFile`, which walks the project tree via `ctx.adapter.fs.readDir`/`stat` **outside any `runWithContext`**. Every other converted surface wraps its work in `withProxyContext(..., { requireToken: true })` (module.handler.ts:102/110/119/127/135/144, snippet.handler.ts:67, markdown-preview.handler.ts:86), matching the Tier-1 precedent at api-handler-wrapper.ts:110-121. cors.ts is now the only granted execution surface reading project files with no tenant identity bound. It fails closed today only because `MultiProjectFSAdapter.getAdapter()` throws on an empty AsyncLocalStorage store and `setDefaultAdapter` has no production caller. That is a guard resting on an unrelated invariant, at a site this PR just opened.

The new granted cors test cannot catch either problem: `makeCtx`'s mock fs (cors.test.ts:20-27) has none of the three `isExtendedFSAdapter` markers, so it never enters the hosted-multi-project branch, and `resolveAppRouteFile` is injected rather than real. It proves the predicate flipped and nothing else.

**Fix options (pick one, state which):**
1. Keep `:54` on the capability, revert `:59` to a topology check (`isSharedProjectRuntime`) — the two decisions need two predicates. Or route shared runtimes through `getHostedConfig` + `ctx.prepareHostedConfigContext` (loader.ts:2288), the loader's documented hosted path — and wrap both the method resolution and the config load in `withProxyContext(..., { requireToken: true })`.
2. Revert `cors.ts:53-59` entirely to `isSharedProjectRuntime` and say so in the PR body. A correct degrade beats an inert conversion that adds an unbound project-FS call and a warn log per preflight.

Either way, add an `isExtendedFSAdapter`-shaped adapter to the cors granted fixture and use the real `resolveAppRouteFile`.

---

## What the capability governs

`allowHostProjectCodeExecution` is host-owned end to end and cannot be forged by a tenant. Produced at `production-server.ts:271-292` (`bootstrap.config.fs?.veryfront?.proxyMode` topology check OR the operator env grant `VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION`, read once at startup from the host env) → `runtime-handler/index.ts:578` → `handler-context-builder.ts:100/133`. Grep for assignments to `ctx.allowHostProjectCodeExecution` / `ctx.isLocalProject` outside tests returns **reads only**. No header, host, pathname, or project config reaches the decision. `readOwnDataProperty` (project-locality.ts:69-86) reads an own **data** property under `Reflect.apply`, rejecting accessors and inherited values, so a tenant-supplied object or proxy cannot forge the flag; throwing descriptors fail closed.

## Surfaces converted (4 of 4)

`isSharedProjectRuntime(ctx)` → `requiresIsolatedProjectRuntime(ctx)`:

1. `request/module/module.handler.ts:76` — DENY → capability-aware. Riskiest surface (imports page/layout code in-process). The comment at :71-74 was **amended, not deleted**: it still records the in-process-import risk and adds the capability carve-out.
2. `request/snippet.handler.ts:40` — DENY → capability-aware.
3. `preview/markdown-preview.handler.ts:51` — DENY → capability-aware. This is the one that resolves the live authenticated `GET /README.md` 503.
4. `response/cors.ts:53` — DEGRADE → capability-aware. **Included** (the task asked to convert it unless there was a reason not to). Rationale for including: a granted host would otherwise keep silently advertising the wrong CORS methods. See the blocker above — for the *second* decision site at :59, there is a reason not to.

Completeness independently re-verified. Only 3 bare `isSharedProjectRuntime(` call sites remain outside the predicate module, all correctly untouched: `production-server.ts:273` (the grant producer — must stay a bare topology check), `ssr/ssr.handler.ts:107` and `routing/api/handler.ts:250` (already `isSharedProjectRuntime(ctx) && !isHostProjectCodeExecutionAllowed(ctx)`, just inlined). All 8 `PROJECT_EXECUTION_UNAVAILABLE` emitters are capability-gated.

## api-handler-wrapper.ts hardcoded `true` — already fixed, and the framing in the issue needs correcting

Three corrections:

- **Path.** There is no `src/server/handlers/request/api-handler-wrapper.ts`. The file is `src/server/handlers/request/api/api-handler-wrapper.ts`.
- **It never shipped on main.** The literal `() => this.handleWithContext(req, ctx, pathname, true)` **pre-dates** #3364 — it came in with the shared-runtime boundary in `1135e37ad` (#3285). `785daa75e` (#3364) is the commit that *removed* it; its squash body carries the sub-commit "fix(api): honor the host-execution grant on the multi-project path". `c1134f2c3` (#3365) then renamed `isSharedRuntime` → `mustDenyProjectExecution` "because the stale name is exactly what made a hardcoded `true` on the multi-project branch look correct for as long as it did."
- **The severity is inverted.** The 4th positional arg is `mustDenyProjectExecution` — `true` means **DENY**. A hardcoded `true` was fail-**closed**, not fail-open. It did not permit execution regardless of the capability. Security impact: none. Actual impact was availability — tenant `/api/*` on veryfront-server returned 503 even with the grant set, because the computed predicate was dead code on exactly the path veryfront-server takes.

Current state, predicate computed once and fed to both branches:

```ts
:88    const mustDenyProjectExecution = requiresIsolatedProjectRuntime(ctx);
:90-92 if (!isMultiProject) { return this.handleWithContext(req, ctx, pathname, mustDenyProjectExecution); }
:112   () => this.handleWithContext(req, ctx, pathname, mustDenyProjectExecution),
```

The only surviving `allowHostProjectCodeExecution: true` literals on a shared path are deliberate and self-documenting: `api/project-discovery.ts:202` (guarded by `requiresIsolatedProjectRuntime` at :142, correct by construction) and `production-server.ts:257` (startup-only, non-multi-project branch).

## Paired guard + granted tests

Four pairs, one per converted surface. Every granted half pins that the request reached the work the guard was blocking — a bare non-503 assertion is weak, since a handler falling through for an unrelated reason also passes.

| Surface | Deny half | Granted half |
|---|---|---|
| module.handler | loops all 4 owned prefixes; 503 + no-store + problem+json + `rendererCalls === 0` | `rendererCalls > 0`, counted in the very initializer the deny test makes throw |
| snippet.handler | 503, `contextCalls === 0`, `readPath === undefined` | `contextCalls === 1`, `readPath === "/project/components/button.snippet.mdx"` |
| markdown-preview | 503, `reads === 0` | `contentReads === 1` (real `FSAdapterWrapper(GitHubFSAdapter)`, stubbed fetch) |
| cors | strengthened: now also pins `allow-methods === "GET, POST, PUT, PATCH, DELETE, OPTIONS"` | `routeResolutionCalls === 1` and `allow-methods === "HEAD, GET, POST, OPTIONS"` |

**Red was captured twice.** After adjusting three fixtures for environment reasons, red was re-verified with the *final* test bodies against unconverted source (`git checkout origin/main -- <4 files>`, patch saved and reapplied; no `git stash`):

```
FAILED | 7 passed (43 steps) | 4 failed (4 steps)
  Expected actual: 503 not to be: 503   x3
  Values are not equal: 0 / 1  (routeResolutionCalls)
```

All four are behavioural assertion failures — no module-resolution or syntax noise. All deny halves passed in the same run, as they must.

**Bidirectional mutation matrix** (run independently by both reviewers, not taken on faith):
- Mutant A — `isExplicitHostProjectCodeExecutionAllowed → return false` (capability read then discarded, the exact #3364 trap): exactly the 4 granted tests fail, one per site.
- Mutant B — `→ return true` (literal grant): exactly the 5 deny steps fail.

Both halves at every site are load-bearing. No site has deny-only coverage.

One fixture change deserves calling out as a **bug fix, not cosmetics**: `snippet.handler.test.ts:97` flips the deny fixture `isLocalProject: true` → `false`. `requiresIsolatedProjectRuntime` ORs in `isExplicitlyLocalProject`, which the old gate ignored; without the flip the existing deny test would have started passing through — silently becoming a no-op. The old fixture also contradicted the test's own name and was passing for the wrong reason. Reachability of `isLocalProject` in proxy mode is the `x-project-path` header behind `isProxyTrusted` (adapter-factory.ts:139-156, documented VULN-SRV-3) — operator-declared topology that swaps the adapter to local disk in the same step. Not tenant-reachable; identical to already-shipped Tier-1 semantics.

## Test / lint state

```
target suite (4 handler tests + src/security/project-locality.test.ts)
  ok | 13 passed (57 steps) | 0 failed

src/server/handlers/ + src/security/ with the repo's own test-task flags
  ok | 187 passed (2020 steps) | 0 failed
```

Zero regressions. `deno fmt` / `deno lint` / `deno check` clean on all 8 changed files. `src/cache/backend.test.ts` was out of scope and untouched.

Environment note: without `--unstable-worker-options --unstable-net` the run aborts (exit 70) inside `src/security/sandbox/deno-sandbox.test.ts` with `Unstable API 'Worker.deno.permissions'`. Missing-flag artifact, unrelated to this change; `deno task test` passes those flags.

## Verification verdicts

| Lens | Sound | Score |
|---|---|---|
| Adversarial multi-tenant isolation | **no** | 72 |
| Hardcoded-literal / capability-derivation audit | yes | 88 |
| Issue-366 paired-test discipline | yes | 92 |

Average 84. The isolation reviewer's `sound=false` is entirely the cors.ts finding above; the security thesis itself holds — the capability cannot be forged, and ungranted deployments deny byte-identically to the old gate at all four sites. The two reviewers disagree on the *severity* of the cors defect (cross-tenant route-existence oracle if a default adapter is ever installed, vs. per-preflight log noise today); they agree it must change before merge.

## Open, non-blocking

- **P3 — markdown-preview granted test proves the guard, not the feature.** In the passing run the render actually fails (`Missing extension for contract "ContentProcessor"` in the test env); `assertNotEquals(status, 503)` passes only because the failure status isn't 503. `contentReads === 1` is the sole load-bearing assertion — it satisfies the pairing rule, but **no test pins a rendered 200 for authenticated `GET /README.md`**. The end-to-end fix is proven by staging, not by this suite.
- **P3 — module granted test is asymmetric** vs its deny half (one prefix vs four; sharedness via `isMultiProjectMode: () => true` vs `prepareHostedConfigContext`). Both signals are genuinely shared — confirmed, that fixture 503s against origin/main. Tighten `rendererCalls > 0` to `rendererCalls === 1`.
- **P3 — predicate-substitution gap the handler pairs cannot catch:** at all 4 sites `requiresIsolatedProjectRuntime(ctx)` and `!isHostProjectCodeExecutionAllowed(ctx)` are indistinguishable; both mutants pass both halves. The distinguishing case (dedicated non-local runtime, no capability) is covered one layer down in `src/security/project-locality.test.ts`. Acceptable layering, noted.

## Remaining before this is done

1. Resolve the cors.ts blocker (option 1 or 2) and update the granted fixture.
2. Staging verification on the next RC — authenticated `GET /README.md` must return a rendered 200, not 503. **The 503 only reproduces authenticated**, so an anonymous smoke check will not see it.
3. Green `veryfront-e2e` staging health run. `preview-rendering.health.spec.ts` asserts response status as of veryfront-e2e#135, so that spec is the one that should have caught this and is the one to watch.

kwakayama:
## Adversarial review: 84/100 avg, security lens 72 — drop `cors.ts` before merge

**The core thesis holds.** The capability cannot be forged by a tenant, and ungranted deployments deny exactly as before. Capability sourcing was traced end to end (7 hops, `production-server.ts:272` → `enriched-context.ts:40`) and is startup/config only — no header, query, body, or project env reaches it. The one request-derived input, `ctx.isLocalProject`, resolves against the startup-configured `localProjects` map, so a hosted project cannot be made local.

Surfaces 1–3 (module, snippet, markdown-preview) are sound.

## P1 — `cors.ts:53-70` must come out

The granted path reads project files **with no tenant context**, and the paired test is structurally blind to it.

Probe against a production-shaped `FSAdapterWrapper` + `MultiProjectFSAdapter`, granted, `OPTIONS /api/private`:

```
runWithContextCalls  = 0
uncontextualFsCalls  = 1
err = VeryfrontError: Hosted multi-project config requires an authenticated declarative evaluation context
allow-methods = "GET, POST, PUT, PATCH, DELETE, OPTIONS"   <- byte-identical to the DENIED output
```

Three problems:

1. **No tenant scope.** `resolveAllowedMethods` (`cors.ts:82`) → `resolveAppRouteFile` walks the tree via `ctx.adapter.fs.readDir`/`stat` (`app-router-resolver.ts:52-77`) outside any `runWithContext`. Every other converted surface wraps its work in `withProxyContext(..., { requireToken: true })` — `module.handler.ts:102-144`, `snippet.handler.ts:67`, `markdown-preview.handler.ts:86` — as does the Tier-1 precedent this PR cites (`api-handler-wrapper.ts:110-121`). This would be the only granted execution surface reading project files with no tenant identity bound.

2. **It fails closed by accident, not by design.** `MultiProjectFSAdapter.getAdapter()` (`multi-project-adapter.ts:126-139`) throws on an empty AsyncLocalStorage store, and `setDefaultAdapter` (`:172`) has no production caller. Install a default adapter and that branch silently returns the *default project's* adapter — an OPTIONS preflight, attacker-chosen pathname, no body, no credential needed, becomes a cross-tenant route-existence oracle.

3. **The granted path is inert.** Output is byte-identical to the denied path, so the conversion buys nothing today while adding the exposure above.

## Recommendation

Drop `cors.ts` from this PR and keep surfaces 1–3. Issue #366 lists cors as lower priority precisely because it degrades rather than denies. Converting it correctly means adding tenant scoping to `resolveAllowedMethods` — a separate change with its own test, not a one-line predicate swap.

## Also: #3385 overlaps this PR

veryfront-code#3385 covers the same three surfaces plus `ssr.handler` and a shared `execution-surface-policy.test.ts`. Neither PR is a superset. Note the audit found `ssr.handler.ts:107` **already honours the capability inline** (`isSharedProjectRuntime(ctx) && !isHostProjectCodeExecutionAllowed(ctx)`), so that change is a consistency refactor, not a bug fix. These two should be consolidated rather than both merged.

## Still open

Whether the hardcoded `true` on the multi-project branch of `api-handler-wrapper.ts` — which #366 records as having shipped once in #3364 — is still live.

kwakayama:
Superseded by #3385, merged as 8d8cab797 on top of 1e8a8c75d (v0.1.1200).

Every non-cors file in this PR is now on `main` with byte-identical behaviour: `module.handler.ts:81`, `snippet.handler.ts:40`, `markdown-preview.handler.ts:51` and `ssr.handler.ts:103` all call `requiresIsolatedProjectRuntime(ctx)`, each with a paired fail-closed and granted-path test. Nothing here is left to salvage.

The one unique delta is `src/server/handlers/response/cors.ts:53`, and it is **deliberately excluded**. An adversarial probe on the granted path measured `runWithContextCalls=0`, `uncontextualFsCalls=1`, and an allow-methods header byte-identical to the denied output: `resolveAllowedMethods` -> `resolveAppRouteFile` walks the tree via `ctx.adapter.fs.readDir/stat` outside any `runWithContext`, unlike every converted surface, which uses `withProxyContext(..., { requireToken: true })`. It fails closed today only because `MultiProjectFSAdapter.getAdapter()` throws on an empty async store. Converting it is therefore inert and would add a cross-tenant route-existence oracle. The granted test added here injects a fake `resolveAppRouteFile`, so it never exercises that real tree walk — it creates the appearance of coverage over exactly the risky path.

That exclusion is now pinned in code rather than in a review thread: `src/server/handlers/execution-surface-policy.test.ts` records `response/cors.ts` in `NON_GATE_USES` with the reason it degrades rather than denies. Please do not reopen this as an oversight.

Verified on origin/main @ 8d8cab797: `deno test --no-check --allow-all` across the four handler test files plus `execution-surface-policy.test.ts` -> 14 passed (102 steps), 0 failed.

PR #3392, opened 2026-08-05
fix(provider): stop undefined tool-result members from failing the model call

kwakayama:
## The failure

Agent run `b2eca9d0-30f6-4e9e-8a67-36e931e8a618` died with `AGENT_PROVIDER_ERROR`:

> The selected agent failed because its AI provider returned an error.

It was not a provider error. The run failed ~67ms after `STEP_STARTED` for step-2, while building the request. No provider was ever contacted.

The real error is a `TypeError` from `stringifyJsonValue` (`src/provider/runtime-loader.ts`): **`Provider tool value must be JSON-serializable`**.

## Root cause

`createLoadSkillTool` emitted the `allowedTools` key unconditionally:

```ts
return {
  skillId: skill.id,
  instructions: parsed.body,
  allowedTools: skill.metadata.allowedTools,   // undefined when frontmatter has no allowed-tools
  references: loadableReferences,
  scripts,
};
```

The skill in that run has no `allowed-tools` frontmatter, so the result carried an own property holding `undefined`.

#3235 replaced the wire serializer's `JSON.stringify(value)` with `snapshotProviderJsonValue`, which **fails closed on `undefined`** rather than dropping it. So the tool result could no longer be encoded, and the next model call died.

The recorded `TOOL_CALL_RESULT` event looks innocent — it has only four keys — because event logging `JSON.stringify`s the payload and drops the offending property. The log is the sanitized copy; the in-memory object had five keys.

## Impact

Deterministic, not flaky. Every agent run that calls `load_skill` on a skill without `allowed-tools` frontmatter fails on its very next model call.

Provider-independent — `anthropic-request-builder.ts`, `openai-chat-request-builder.ts`, and `openai-responses-request-builder.ts` all share `stringifyToolResultValue`.

Regression window: #3235 (2026-08-02) through today.

## The fix

Two layers, because `load_skill` is unlikely to be the only tool that builds a result from optional fields.

1. **`src/skill/tools.ts`** — omit the key when the skill declares no policy. An explicit empty allow-list still serializes, since `[]` means "deny all", not "unset".

2. **`stringifyJsonValue`** — opt into `JSON.stringify` semantics for `undefined` members: drop such object properties, encode `undefined` array elements as `null`.

Layer 2 is deliberately scoped. `snapshotJsonValue` rejecting `undefined` is intentional (`runtime-loader-helpers.test.ts:522`, and `jsonValuesEqual({hidden: undefined}, {}) === false`) — an own property holding `undefined` must stay distinguishable from an absent one for audit and equality callers. So the tolerance rides on a new **default-off** `dropUndefinedMembers` option that only the wire path sets. Descriptor validation is untouched: values are still read through descriptors, and accessors, `toJSON`, symbols, Proxies, cycles, sparse arrays, and non-finite numbers all still fail closed.

## Verification

- Replayed the failing run's exact step-2 prompt through `toOpenAICompatibleMessages` with the real five-key `load_skill` result. Threw before, builds the request after, and the serialized tool message now matches `JSON.stringify` byte for byte.
- New tests pin both layers, including that the strict primitive still rejects `{value: undefined}` and `[undefined]`.
- `deno task typecheck` passes. `deno lint` and `deno fmt` clean on the touched files.
- 1270 tests pass across `src/provider`, `src/skill`, `src/agent`, `extensions/ext-llm-anthropic`, `extensions/ext-llm-openai`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider-boundary serialization for data containing undefined object properties.
  * Ensured undefined array values are represented consistently as `null` where applicable.
  * Prevented empty skill policies from producing undefined configuration fields in tool results.

* **Tests**
  * Added coverage for serialization, snapshot handling, equality checks, and skill-loading output consistency.

* **Documentation**
  * Updated API reference links to point to the correct source locations.


kwakayama:
Addressed the review comments and fixed `ci (lint)`.

**`ci (lint)`** — the failure was `docs:api-reference:check`, not a lint rule. `docs/api-reference` was stale on the line-number churn from the source edits. Regenerated with `deno task docs`; the check is green locally.

**Copilot — relative cross-module import in `src/skill/tools.test.ts`.** Fixed by removing the import rather than aliasing it. There is no `#veryfront/provider/shared` import alias in `deno.json` (only the consumer-facing `veryfront/provider/shared` export), so the suggested specifier would not resolve. The provider boundary is already pinned in `runtime-loader-helpers.test.ts` using the same `load_skill` shape, so the skill test now only asserts the result shape — one assertion per layer, no cross-module dependency.

**Copilot — sparse-array comment is wrong.** Correct. `JSON.stringify` encodes holes as `null`, same as `undefined` elements, so "no JSON encoding" was false. Rejecting holes is a deliberate fail-closed choice. Comment removed rather than reworded.

**CodeRabbit — stale comment in `tools.test.ts`.** Correct, it described the pre-fix behavior. Removed along with the assertion it annotated.

Also trimmed the remaining commentary down to the one comment that is load-bearing: why `snapshotObject` reads through the descriptor before skipping, so accessors still fail closed.

The source diff is now 3 lines in `runtime-loader.ts`, 1 line in `tools.ts`, and the `dropUndefinedMembers` plumbing in `json-snapshot.ts`.

Re-verified: `deno task typecheck` passes, `deno lint` and `deno fmt --check src/` clean, `docs:api-reference:check` current, and provider + skill + both LLM extension suites pass.

PR #3397, opened 2026-08-05
feat(agent): 1-hour prompt-cache TTL for the hosted product agent (RFC 0002 §4.1)

mattboon:
Implements **RFC 0002 §4.1** (#3387) — the first, independent lever from the "structured, minimal, cacheable agent context" initiative.

## Problem
A trivial turn to the hosted product agent (`hi`) costs ~7,285 input tokens at ~$0.94. The static system message (prompt + `<project_context>` + `<available_skills>`) already carries an ephemeral cache breakpoint in `buildAgentCallContext` — but the default ephemeral TTL is ~5 minutes, so two turns spaced tens of minutes apart are **both cache misses** and pay full Opus input price. This is the dominant cost, and it's a TTL choice, not a token-count problem (per RFC §2, the volatile `environment_context` already rides in a *second, uncached* message after the breakpoint, so nothing downstream busts the cache).

## Change
- `call-context.ts`: add `cacheTtl?: "5m" | "1h"` to `BuildAgentCallContextInput` (default `"5m"`). `buildCacheControl` renders `{ type: "ephemeral" }` unchanged for `5m` and `{ type: "ephemeral", ttl: "1h" }` for `1h`.
- `cloud-runtime-system-messages.ts`: the Veryfront Cloud runtime (interactive hosted chat) opts into `cacheTtl: "1h"`.
- Tests: `call-context.test.ts` asserts both TTL renderings; the pinned hosted-message test now expects the 1h control.

Non-cloud/internal callers (`createRuntimeAgentSystemMessages`) are untouched and stay at 5m.

## Economics
1h cache write is 2× base vs 1.25× for 5m, breaking even at ~3 reads — worth it when idle gaps routinely exceed 5 min, which is exactly the chat agent.

## Safety
`providerOptions` is passed opaquely to the AI SDK Anthropic provider. If a provider version doesn't forward `ttl`, it degrades to today's 5m behavior — **no regression path**.

- [ ] Verify on real traffic that `ttl: "1h"` reaches the Anthropic `cache_control` block and that read-hits recover the 2× write.

## Tests
`deno test src/agent/runtime/call-context.test.ts src/agent/hosted/cloud-runtime-system-messages.test.ts` — 7 passed, 0 failed. `deno check` clean on both changed source files.

Refs: RFC 0002 (#3387); companion RFCs veryfront-agent #336 (merged), veryfront-studio #6316.

## Summary by CodeRabbit

- **New Features**
  - Added configurable prompt-caching durations for supported Anthropic requests: 5 minutes or 1 hour.
  - Hosted cloud runtime requests now use a 1-hour cache duration by default.

- **Bug Fixes**
  - Updated caching behavior to apply the correct ephemeral settings when no duration is specified or when the 1-hour option is selected.

- **Documentation**
  - Updated API reference links to reflect the latest source locations.


mattboon:
Closing in favor of the layered-context split (design doc: `plans/agent-context-static-dynamic-split.md` in veryfront-agent).

The 1h TTL is right, but it belongs on an **isolated Layer 0** (base prompt + bootstrap tool schemas only). As written here the cached system message still contains `<project_context>` + `<available_skills>`, so this caches a **per-project blob** — the 2× cache-write is spent on a key that isn't shared across projects, and the only benefit is within-session same-project hits (largely covered by the default 5m).

The real payoff needs the split first. Folding this PR's `cacheTtl` plumbing into the framework split PR rather than merging the wrong boundary now and reworking it. Branch kept for reuse.

PR #3399, opened 2026-08-05
fix(build): embed the declarative evaluator worker in compiled binaries

kojiwakayama:
## What broke

The declarative config evaluator spawns its worker through a sibling URL whose extension is computed at runtime:

```ts
const extension = import.meta.url.endsWith(".ts") ? ".ts" : ".js";
return new URL(`./declarative-evaluator-worker-entry${extension}`, import.meta.url);
```

`deno compile` embeds a worker only when it can statically read the specifier, so the entry was never packaged into the binary.

## Why nothing caught it

Running from source resolves the file from disk, so the build and all 3794 unit tests pass. The binary starts, serves traffic, and dies on the first request that evaluates declarative config:

```
Module not found: file:///tmp/deno-compile-veryfront/src/config/declarative-evaluator-worker-entry.ts
Unhandled error in child worker.
```

In production this crash-looped `veryfront-server` down to 1 of 4 replicas ready. Staging runs the same binary and stayed green, because its traffic never reaches that path — which is why promotion looked safe.

## The fix

`DEFAULT_INCLUDES` already carried two other workers for exactly this reason; this is the third, and the first to reach production. A bare list entry would invite a fourth, so the entry lives in `UNTRACEABLE_WORKER_INCLUDES`, guarded by:

- **Every sibling-URL worker specifier** in a file that constructs a `Worker` must be covered by an include. Static `import` specifiers are excluded, since compile traces those.
- **The full profile** must carry those includes.

All guards are mutation-verified: emptying the constant fails the scan naming the exact module from the crash.

## The proxy is not affected — and that took verifying

`cli/proxy-main.ts` pulls the evaluator's *runner* into its module graph, so this initially looked like the same latent crash sitting in the proxy binary. It isn't. Embedding the worker entry there drags the evaluator's babel tree into the frozen `scripts/build/proxy-deno.lock`, and the proxy build fails outright:

```
error: The lockfile is out of date... +"@babel/parser@7.29.2"
```

That lock has never carried those packages, so the proxy cannot have been spawning this worker. A third guard pins the exclusion so it isn't silently "fixed" into a broken proxy build, and documents that the lock must be regenerated in the same change if the proxy ever does need it.

## Verification

Checked against real binaries, not just the argument list:

| binary | `declarative-evaluator-worker-entry.ts` in embedded VFS |
| --- | --- |
| full, built **with** the include | present |
| full, built **without** it | absent |
| proxy | n/a — compiles clean |

Full unit suite: `3794 passed, 0 failed`.

## Rollout

Production was rolled back to `20260727215958-09eda661f829` and is healthy at 4/4. Deploys will keep failing on the release-asset manifest schema skew until a fixed artifact ships, so this needs a release and a fresh promotion to close that out.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment errors with specific, actionable details for malformed, incompatible, or mismatched release manifests.
  * Added clearer diagnostics for framework version skew and release mismatches.
  * Prevented sensitive manifest contents from appearing in schema validation errors.

* **Documentation**
  * Documented release-manifest diagnostic guidance and updated source references.

* **Build Improvements**
  * Ensured required worker assets are included in full compiled binaries while keeping them excluded from proxy builds where incompatible.


kwakayama (changes requested):
## Review — 78/100

Reviewed by checking out the branch, running the suite, mutating it, and building the proxy profile. I closed my own #3398 in favour of this: the usage-based guard here covers `extensions/` where three of the four known computed-URL workers live, which mine did not.

The fix is correct and the guard genuinely guards. One factual claim is wrong, and it is load-bearing.

### Verified working

**The guard catches a real removal.** I emptied `UNTRACEABLE_WORKER_INCLUDES` and re-ran:

```
FAILED | 0 passed (5 steps) | 1 failed (2 steps)
  ✘ should include every worker entrypoint spawned from a sibling URL
  ✘ should include every worker entrypoint named as one
```

Two independent detectors fire. That is real defence in depth, not one tripwire.

**Binary-level evidence is the right evidence.** Checking the embedded VFS for the module path rather than the argument list is the only thing that would have caught the original defect, and it is the correct marker.

**`UNTRACEABLE_WORKER_INCLUDES` as a named, documented, exported constant** is better than the inline entry I had. The docstring explains the failure mode precisely.

### Must fix before merge

**The proxy exclusion rests on a premise that is false.** The comment says:

> That lock has never carried those packages, so the proxy cannot have been spawning this worker.

`scripts/build/proxy-deno.lock` already carries a babel parse toolchain:

```
npm:@babel/parser@7.29.2
npm:@babel/generator@7.29.1
npm:@babel/traverse@7.29.0
npm:@babel/types@7.29.0
```

Your *conclusion* about the build is right — I added `...UNTRACEABLE_WORKER_INCLUDES` to `PROXY_INCLUDES` and the frozen build fails (exit 1). But it fails for a different reason than stated: the graph wants to add `@babel/types@7.29.8` (a newer version than the pinned `7.29.0`) plus `@babel/helper-string-parser` and `@babel/helper-validator-identifier`. So the accurate statement is "the lock carries a narrower/older babel set; the worker entry's graph requires additional packages", not "has never carried those packages".

That matters because the safety inference does not survive the correction. "The lock lacks the deps, therefore the proxy cannot spawn this worker" cannot be argued from a lock that already contains parser, generator, traverse and types. And you note yourself that `cli/proxy-main.ts` pulls the evaluator's *runner* into the proxy graph. If the runner is reachable at runtime in the proxy profile, the proxy binary has exactly the defect this PR fixes — and the new `should keep untraceable workers out of the frozen proxy profile` test would actively block fixing it.

I am not claiming the proxy is broken. I am saying the justification as written does not establish that it is safe, and it is now enforced by a test. Please either re-derive it on reachability ("the proxy never calls hosted declarative evaluation, because X") or keep the exclusion on the honest ground that it breaks `--frozen` and file the reachability question as a follow-up.

### Should fix

**Rebase.** The branch is 2 commits behind `main`, so five files — `src/release-assets/*`, `cli/shared/deployment/deploy-project*`, `docs/api-reference/veryfront/release-assets.md` — show as ~130 lines of diff that are already merged as #3396. I confirmed the content is byte-identical to `origin/main`, so this is purely a stale base, but it is noise a reviewer has to wade through to reach the 3 files that matter.

**The constant holds 1 of 4 members of the class it names.** Its docstring describes workers "spawned through a sibling URL whose extension is computed at runtime". Three others match that description exactly and stay inline in `DEFAULT_INCLUDES` under separate ad-hoc comments:

```
extensions/ext-react-ssr/src/worker-renderer.ts                       ./worker-renderer.ts : ./worker-renderer.js
extensions/ext-document-kreuzberg/src/upload-extraction-worker.ts     ./upload-extraction-worker.ts : ...js
extensions/ext-document-kreuzberg/src/native-progress-extraction-worker.ts
```

Not a correctness hole — the generic tests cover them wherever they live — but the constant over-promises, the proxy-exclusion test only guards one of four, and the next person adding a worker gets ambiguous precedent. Worth folding them in.

(`src/security/sandbox/worker-script.ts` is correctly *not* in this class: it is inlined as a `data:`/Blob URL, not resolved as a sibling. I checked before assuming.)

### Minor

**Two tests go vacuous if the list empties.** In my mutation run, `should embed the untraceable worker entrypoints in the full profile` and `should keep untraceable workers out of the frozen proxy profile` both *passed* against an empty list while the real guards failed — they iterate the constant, so an empty constant is a green tautology. They are fine as wiring checks on the `...spread`, but they should not be read as the protection; the two discovery tests are.

**Detector shape.** `siblingWorkerSpecifier` requires `[Ww]orker` in the specifier, so a worker entry not named for what it is would slip past the usage-based test. The name-based test you added in `963d7faa7` covers the common case, but both ultimately lean on naming. Acceptable given convention, worth knowing.

### One thing not recorded here, for whoever verifies the next binary

Do not use body identifiers to check embedding. `dist/framework-src` is in `DEFAULT_INCLUDES` and ships a copy renamed `declarative-evaluator-worker-entry.ts.src` — a data asset, renamed exactly so compile will not treat it as a module. So worker-body symbols appear in a **broken** binary too and read as a pass:

| marker | broken | fixed |
| --- | --- | --- |
| `evaluateRequest` | 2 | 4 |
| `createDeclarativeConfigWorkerSuccessResponse` | 6 | 8 |
| `declarative-evaluator-worker-entry.ts.src` | 1 | 1 |
| **`declarative-evaluator-worker-entry.ts`** | **0** | **2** |

Only the last row discriminates. You used the right marker; this is a note for the next person, not a defect.

### Score

**78/100.** Correct root-cause fix for a production outage, with a mutation-verified guard and binary-level evidence — well above the bar for the class of change. Held back by a load-bearing claim that is checkably false and now enforced by a test, a constant that covers a quarter of the class it names, and a stale base. All cheap to address; none of them touch the shipped behaviour, which I verified is right.

Not blocking on the rebase if this needs to move fast — the premise correction is the one I would not merge without, and it is a comment edit plus a decision.

kwakayama:
## Re-review after `b47d4d9ef` — still 78/100

Re-checked out the new head and re-ran everything. The push is the one-line test tidy:

```diff
-      output: "/tmp/test-veryfront",
+      output: "test-veryfront",
```

Correct and worth having — `normalizeOutputPath` resolves the relative form, so this drops a hardcoded absolute path with no behavioural change. It does not touch any review finding, so the score is unchanged.

**Re-verified green on this head**

- Suite passes, 7 steps.
- Mutation still bites: emptying `UNTRACEABLE_WORKER_INCLUDES` gives exit 1, with `should include every worker entrypoint spawned from a sibling URL` and `should include every worker entrypoint named as one` both failing. The guard is intact after the test edit.

**All four findings still open**

1. **Blocking — the proxy premise.** Unchanged: "That lock has never carried those packages, so the proxy cannot have been spawning this worker." `proxy-deno.lock` carries `@babel/parser@7.29.2`, `generator@7.29.1`, `traverse@7.29.0`, `types@7.29.0`. Your build conclusion is right — I reproduced the `--frozen` failure — but it fails because the graph wants `@babel/types@7.29.8` plus `helper-string-parser` and `helper-validator-identifier`, not because the packages were never there.

   Since the last review I also resolved the reachability question, and it does not favour the current wording. `deno info cli/proxy-main.ts` puts all three in the proxy graph:

   ```
   declarative-evaluator-worker-runner: IN GRAPH
   declarative-evaluator.ts:            IN GRAPH
   config/loader.ts:                    IN GRAPH
   (1073 modules)
   ```

   So the dedicated proxy binary compiles in the code that spawns the worker, without the worker entry. Graph presence is not proof of a live call, but it does mean the exclusion cannot rest on "the deps aren't there". Either name the runtime gate that makes hosted evaluation unreachable in the proxy profile, or regenerate via `deno task build:proxy-lock`, include the workers, and drop the exclusion test.

2. `UNTRACEABLE_WORKER_INCLUDES` still holds 1 of the 4 workers matching its docstring.
3. The two constant-iterating tests still pass vacuously against an empty list — confirmed again in this run.
4. Still 2 commits behind `main`; the ~130 lines of already-merged #3396 are still in the diff.

**A correction to my own last review.** I reported at one point that the "never carried" claim might have been removed. That was wrong — the sentence wraps between `never` and `carried`, and I had piped grep through `sed`, which exits 0 regardless and masked the miss. Checking the exit code directly shows the claim is present and unchanged. The finding always stood; my intermediate read of it did not.

**Path to 90+**, unchanged from my last comment: fix (1), then add a post-build assertion that the *compiled binary* embeds every declared worker entry. Every test here still validates the include list, and the binary that took production down would pass all seven of them. That artifact-level check is the one that would actually have caught this, and it is what moves this from thorough to airtight.

kojiwakayama:
Thanks — the blocking finding was correct and I was wrong. All four are addressed, plus the artifact-level check you asked for. Head is `0dd76b5ca`.

## 1. The proxy premise — fixed

You were right, and I verified it myself before changing anything. `scripts/build/proxy-deno.lock:1706-1709` carries the babel toolchain:

```
npm:@babel/generator@7.29.1
npm:@babel/parser@7.29.2
npm:@babel/traverse@7.29.0
npm:@babel/types@7.29.0
```

The "never carried" sentence is gone (`grep -c "never carried"` → `0`). The comment now states the build constraint it can actually support — a *version skew*, the graph wanting `@babel/types@7.29.8` plus the two helpers against a pinned `7.29.0` — and says outright:

> This is NOT evidence that the proxy is safe.

I took your second option: exclusion kept on build grounds, reachability filed as **veryfront-issue-inbox#382**, which the comment now cites by number. Your graph finding is recorded there, including that option 2 (regenerate the lock, include the workers, drop the exclusion test) is the unconditionally safe route if the gate turns out not to exist.

## 2. Constant now holds 4 of 4

`worker-renderer.ts`, `upload-extraction-worker.ts` and `native-progress-extraction-worker.ts` folded in. `worker-renderer-bundle.generated.ts` deliberately stays out — it's the payload the worker loads, not an entrypoint, and it's commented as such. Resolved include lists verified unchanged apart from the original fix: full **34 → 35**, proxy untouched.

## 3. Vacuity — fixed

Both constant-iterating tests now assert the constant is populated first, and say in-comment that they're wiring checks rather than the protection. Emptying it now trips **five** guards instead of three.

## 4. Rebase — done

The branch was on a stack, not merely behind: it carried #3396's two commits plus #3395 content, so rebasing replayed already-merged work and conflicted. Rebuilt as a clean branch off `origin/main` with only the four relevant commits cherry-picked. Now **0 commits behind**, and the diff is the 3 files that matter plus the 3 one-line version files.

## The artifact-level check — implemented, and it caught me twice

You were right that every test validated the include list and the production binary would have passed all of them. `compileBinary()` now asserts the compiled artifact contains each declared worker's VFS entry and fails the build otherwise.

Building it against real binaries — rather than trusting it — found two defects in my own check:

1. **It crashed instead of checking.** Decoding a 959 MB binary in one call throws `TypeError: buffer exceeds maximum length`. It now streams in 8 MB chunks, decoding latin1 so bytes map to characters one-for-one, carrying each chunk's tail forward so a marker split across a boundary is still found.

2. **It was tautological.** Expectations came from the resolved include list, so dropping a worker from the list also dropped it from what was verified — the exact regression it exists to catch passed with exit 0. Twice. They now come from `UNTRACEABLE_WORKER_INCLUDES`, with the proxy exempt for the documented build reason.

End-to-end against real 959 MB binaries — declared but not included:

```
exit: 1
Error: Compiled binary is missing 1 worker entrypoint(s):
  src/config/declarative-evaluator-worker-entry.ts.
The binary would start, serve traffic, and crash the first time one is spawned.
```

Unmodified build: `exit: 0`.

Your `.ts.src` warning is encoded as a unit test — the `dist/framework-src` data asset must not count as an embedded worker, since its body symbols appear in a broken binary too. Only the exact VFS entry with its trailing quote discriminates.

## One thing your review surfaced indirectly

Chasing the 959 MB figure: `@huggingface/transformers@4.2.0` is a **root specifier in `deno.lock`** despite no static import, no import-map entry and no extension manifest reference. `opaque-deps.ts` hides it from graph tracing on purpose and documents that compiled binaries can't use it (`Compiled binary: fails`) — but embedding is driven by the lockfile, not the graph, so `libonnxruntime` and `onnxruntime_binding.node` ship in every binary regardless. Roughly a gigabyte the artifact is designed not to be able to use. Out of scope here; raising it separately.

CI: **24 pass, 0 fail**.

kwakayama (approved):

kwakayama (commented):
## Review — 88/100

Reviewed by checking out the branch at `0dd76b5`, running the suite, mutating the constant three ways, tracing the proxy module graph with `deno info`, and planting synthetic workers to probe the scans' blind spots.

The fix is correct and the guard is real. The blockers from the earlier 78/100 review are addressed: the proxy rationale is now honest and tracked, the constant carries all four members, and the branch is rebased (`merge-base` == `origin/main@6420fc2f7`, so the diff is genuinely 5 files). What is left is coverage breadth, not correctness.

### Verified working

**The guard is not tautological, and I proved it three ways.**

| mutation | result |
| --- | --- |
| empty `UNTRACEABLE_WORKER_INCLUDES` | `0 passed (4 steps) \| 1 failed (5 steps)` |
| drop only `src/config/declarative-evaluator-worker-entry.ts` | both discovery scans fail, naming that exact module |
| unmodified | `1 passed (9 steps) \| 0 failed` |

Deriving expectations from the declared constant rather than the resolved include list is the load-bearing decision here, and the comment on `assertWorkersEmbedded` says so. Without it the targeted mutation would have gone green.

**The binary check actually runs before merge, and CI proves the marker works.** `build-binaries` is `main`-only, so I expected `assertWorkersEmbedded` to be dead on PRs. It is not — `tests/e2e/setup/binary.ts:161` compiles the full profile through `compile-binary.ts`, and `tests (binary e2e)` passed in 6m0s on this PR. That is empirical evidence that `"n":"<file>"` is the right VFS marker against a real binary, not just against the fixture.

**The `.ts.src` disambiguation is a real hazard, correctly handled.** `scripts/build/prepare-framework-sources.ts:74` does write `relativePath + ".src"`, so a broken binary really does contain the worker's body under a `.src` name. The trailing quote in the marker is doing work.

**The chunked scan is correct.** `latin1` is 1 byte to 1 char so the carry arithmetic holds; carry length (`fileName.length + 6`) exceeds the marker length by 1, which is safe; short reads are handled via `subarray(0, bytesRead)`; the empty-list `Math.max(...[])` case is guarded by the early return.

### Should fix

**1. `src/security/sandbox/worker-script.ts` is a fifth member of this class, and it sits outside the binary guard.**

`project-worker.ts:1103` resolves it with `import.meta.resolve("./worker-script.ts")` and spawns it at line 568. `import.meta.resolve` is a runtime call; compile does not trace it. Its own comment admits this: *"The binary build explicitly includes this module in its VFS."*

The earlier review dismissed this one as inlined via `data:`/Blob URL. That is `deno-sandbox.ts:176` — a different file with a different mechanism. `project-worker.ts` is file-backed and untraceable.

Today it is covered by `DEFAULT_INCLUDES` and by the sibling-URL scan, so nothing is broken. But it is not in `UNTRACEABLE_WORKER_INCLUDES`, so the one check that would have caught the outage does not cover it. Fold it in.

**2. The proxy exclusion is now honest, but I can confirm the runner really is in the proxy's runtime graph.**

Following `code` edges only from `deno info --json cli/proxy-main.ts` (type-only edges excluded):

```
cli/proxy-main.ts -> cli/commands/serve/proxy-runtime.ts -> src/proxy/main.ts
  -> src/proxy/handler.ts -> src/proxy/control-plane-signature.ts
  -> src/security/index.ts -> src/security/http/config.ts
  -> src/config/index.ts -> src/config/loader.ts
  -> src/config/declarative-evaluator-worker-runner.ts
```

So this is a live question, not a theoretical one, and the PR is right not to claim otherwise. The cost is that `should keep untraceable workers out of the frozen proxy profile` now pins the exclusion, so whoever closes veryfront-issue-inbox#382 with "yes, reachable" has to delete a passing test to ship the fix. That is the correct trade for now — just keep #382 hot.

### Minor

**3. A worker whose specifier does not contain "worker" passes both scans.** I planted `src/.../bg-entry.ts` spawned via `new Worker(new URL("./bg-entry.ts", import.meta.url))`. Neither scan fired. `siblingWorkerSpecifier` requires `[Ww]orker` in the path and `entrypointName` requires `-worker-entry.ts`. The naming convention is the guard.

**4. The name scan walks `src` only; the sibling scan walks `src` and `extensions`.** I planted `extensions/ext-yaml/src/zztmp/split2-worker-entry.ts` with its URL built in a sibling helper module (so the `new Worker(` file filter skips it). Result: `1 passed (9 steps) | 0 failed`. Both scans missed it. The same shape under `src` is caught, because the name scan backstops it. Three of the four constant members live in `extensions`, so this asymmetry is backwards — `collect("src")` should be `collect("src")` plus `collect("extensions")`.

**5. The guard that should have caught this originally is left in place, unfixed.** `scripts/build/compile-binary.test.ts:48` — *"compiled CLI embeds every runtime-resolved sibling module"* — walks `extensions` only and asserts a hardcoded 3-item inventory that does not include the new worker. It missed the outage for two independent reasons: wrong directory, and it only recognises the `"./x.ts" : "./x.js"` ternary idiom, not the interpolated one. There are now three worker guards across two files with three different discovery heuristics. Either fold the old one into the new scans or point its inventory at `UNTRACEABLE_WORKER_INCLUDES`, or it will drift.

**6. `worker-renderer-bundle.generated.ts` is excluded from the binary check for being "not an entrypoint".** It is equally untraceable and equally fatal if dropped. The question the check answers is "did the untraceable asset ship", and entrypoint-ness does not change the answer.

**7. `findMissingEmbeddedWorkers` matches on basename only.** Two identically-named files in different directories would alias. No collision today.

**8. The PR body still carries the retracted claim.** It says *"The proxy is not affected — and that took verifying"* and *"That lock has never carried those packages, so the proxy cannot have been spawning this worker."* The code comment you shipped in `136796c3b` explicitly contradicts both. The description is what the next person reads first — please update it before merge.

### Not blocking

The release bump to `0.1.1203` rides along, and `hydration-runtime.generated.ts` changes only the embedded version string. Both expected for this repo's release flow.

### Verdict

Ship it. Item 1 is a one-line addition and worth doing in this PR, since it is the same class and the same failure mode. Items 3–5 are follow-up work; the naming-convention dependency in particular is worth writing down somewhere a person adding a worker will see it.

PR #3400, opened 2026-08-05
fix(deploy): do not fail a deploy on a release asset schema skew

kwakayama:
## The failure

A deploy against a project whose hosted builder predates the manifest v1 → v2 break cannot complete on any CLI at or past v0.1.1189:

```
  Detail: Release assets for <release-id> returned an invalid
  or mismatched ready manifest. Rebuild the release assets and run deploy again.
```

The production control plane returns a **ready** manifest built by framework `0.1.1162`:

| field | returned | this build reads |
| --- | --- | --- |
| `schemaVersion` | `1` | `2` |
| dependency mode | `fallback: {mode:"jit",gaps:[]}` | `dependencyMode` |
| `css[].styleProfileHash` | `"qpyvqf"` | 64-char SHA-256 |
| `css[].cssPipelineIdentity` | absent | required |

#3244 made the reader strict at v2, #3396 explained the rejection, but the deploy still fails closed on it.

## Why it should not be fatal

Nothing on the deploying machine can fix this. The builder is server-side, so "rebuild the release assets" rebuilds through the same mismatched builder, and the operator cannot raise its version from the CLI. The gate blocks every deploy to a platform running an older framework than the caller — and the mirror case blocks an older CLI once a platform moves first.

It also contradicts the module's own contract. `src/release-assets/README.md`, under *JIT fallback is not an authorization fallback*, already says an unavailable, not-ready or invalid manifest makes readers return `null` and fall back to the release-scoped JIT module path. The deploy gate was the only consumer treating that same condition as fatal, and manifest consumption is a feature flag that defaults to off.

## The change

`readMismatchedReleaseAssetManifestSchemaVersion` reports the version a rejected body declares, and only when that declared version is why it was rejected. `waitForReleaseAssetManifest` resolves `null` for that case and reports it through `onSchemaSkew`; deploy emits a `release-assets-schema-skew` warning and continues.

The mismatch counts in **both directions**. An older builder under a newer reader and a newer builder under an older reader are the same condition, and neither is evidence of a hostile body.

**Everything else still fails closed.** A body that parses at the current version but fails an identity check, a malformed body with no readable schema version, and a hostile envelope all keep raising `DEPLOYMENT_ERROR`. The v2 parser is untouched — v1 bodies stay unparseable, because their CSS provenance genuinely cannot be reconstructed, exactly as the README states.

## Proof

The unit test builds the rejected body from the real production payload, so the degradation is pinned to the shape that actually shipped rather than a hand-written approximation.

Driving that exact payload through `waitForReleaseAssetManifest`:

| | result |
| --- | --- |
| parent commit (`6420fc2f7`) | `threw VeryfrontError` → **deploy blocked** |
| this branch | `resolved null`, skew reported as `1` → **deploy proceeds** |

Green: `deno test cli/shared/deployment/ src/release-assets/ cli/commands/deploy/` (50 passed, 342 steps), `deno check cli/main.ts src/index.ts`, plus `fmt:check`, `lint`, `docs:api-reference:check`, `lint:test-typecheck`, `lint:style`, `lint:cli-boundary`, `lint:barrel-jsdoc`, `lint:module-boundaries`, `docs:check-links`.

## Note for reviewers

This changes deploy policy, not just diagnostics — #3396 deliberately kept the failure fatal and improved its wording. The argument for degrading is above; if you would rather keep it fatal, the alternative is to treat a stale hosted builder as a platform-side incident, which leaves users blocked until the platform catches up.

### Correction to the original framing

I first described this as a chronically stale hosted builder. That was wrong, and the real trigger makes the case stronger.

The renderer is normally current. It went backwards because of an **incident rollback**:

| time (UTC, 2026-08-05) | event |
| --- | --- |
| 22:21 | production release sets `renderer.image.tag` to the Aug-5 artifact; proxy pods start |
| ~22:46 | renderer rolled back to `20260727215958-09eda661f829` — framework **v0.1.1162** |
| 23:27 | `veryfront push` |
| 23:28 | renderer builds assets → `schemaVersion: 1`, `builderVersion: 0.1.1162` |
| 23:28 | `veryfront deploy` fails |

So this is not "the platform is behind and should catch up". It is: **a renderer rollback below 0.1.1189 takes out `veryfront deploy` for every user on a current CLI, as collateral.** A rollback is an incident-response tool. It should not also break the deploy path — which is plausibly the very thing you need in order to ship the fix.

Note the rollback reactivated an older ReplicaSet outside Helm, so `helm history` still reports the new release as deployed while the Deployment runs the old image. Proxy and renderer were left on different artifacts.

## Summary by CodeRabbit

* **New Features**
  * Deployments now detect release asset manifest schema mismatches and continue using release-scoped module serving.
  * Added a public utility for identifying mismatched manifest schema versions.

* **Bug Fixes**
  * Deployments now display all warnings generated during execution.
  * Malformed manifests at the current schema version continue to fail safely.

* **Documentation**
  * Updated release asset API references and documented schema-mismatch deployment behavior.

## Verified against the live control plane

Ran the patched CLI against the project that produced the original failure, on the production control plane whose builder is still `0.1.1162`:

```
$ deno run -A <patched>/cli/main.ts deploy
  ○ Linking project...
  ○ Building release...
  ○ Deploying to production...
  ○ Verifying production URL...
```

The manifest gate is cleared and the deployment is created — confirmed server-side as a new deployment row against a new release. On the published CLI the run never reached `Deploying to production...`.

That run then failed at the *later* `wait-environment-url` step (readiness probe saw a network error inside 120s; the URL serves a 302 now). That is a separate, downstream issue and not something this PR claims to fix.

### Follow-up worth filing separately

Warnings only render on the success path (`cli/commands/deploy/command.ts` logs them after `logSuccess`), so a deploy that fails at a later step drops the skew warning it already emitted — which is exactly what happened in the run above. Pre-existing behaviour, now more visible with a second warning producer.


kwakayama:
Addressed in 5d5485868 — the v1 fixture now uses synthetic `projectId`, `releaseId` and `createdAt`. The values that are actually load-bearing stay: `schemaVersion: 1`, `fallback` in place of `dependencyMode`, the six-character `styleProfileHash`, and the absent `cssPipelineIdentity` — those are what keep v1 unparseable and make the test meaningful. `builderVersion` is kept as a framework version, not a deployment identifier.

The same identifiers are scrubbed from the PR description.

PR #3401, opened 2026-08-06
fix(config): parse hosted config as TypeScript regardless of file name (production 400)

kwakayama:
Restores a customer site that is returning HTTP 400 in production right now.

## Symptom

```
{"type":"https://veryfront.com/docs/errors/config-parse-error",
 "title":"Failed to parse configuration","status":400,
 "detail":"Hosted configuration rejected (syntax-error: syntax-error)"}
```

The config is valid TypeScript:

```ts
export default {
  router: "pages" as const,   // TypeScript-only syntax
  app: "components/app.tsx",
  ...
};
```

## Cause

`pickPlugins` selected Babel's `typescript` plugin from the file name:

```ts
const isTypeScript = /\.(?:tsx?|[cm]ts)$/.test(normalizedPath);
if (isTypeScript || !filePath) plugins.push("typescript");
```

Under a `.js` or `.mjs` name a TypeScript config is parsed as JavaScript. `as const` fails, `parserErrorReason` matches neither `DuplicateDefaultExport` nor `VarRedeclaration` and falls through to the generic `"syntax-error"`, which surfaces as `CONFIG_PARSE_ERROR` → HTTP 400.

The evaluator cannot trust that name. It is the candidate the hosted loader *asked for*, not necessarily the file it received.

## Reproduced locally, then fixed

Both reproductions are committed as tests, and both fail on `main`.

**At the evaluator boundary** — `declarative-evaluator.test.ts`, the customer's config byte for byte through `evaluateDeclarativeConfig`:

```
before   DeclarativeConfigEvaluationError: Hosted configuration rejected (syntax-error: syntax-error)
after    all three file names evaluate; router === "pages"
```

That error string is verbatim what production returns.

**At the parser** — `parser-only.test.ts`:

```
before   veryfront.config.ts    parses
         veryfront.config.js    SyntaxError: Unexpected token, expected "," (2:18)
         veryfront.config.mjs   SyntaxError
after    all three parse
```

Column 18 is exactly where `as const` begins.

## Does "always TypeScript" break JavaScript?

Measured rather than assumed, since `typescript` + `jsx` puts `.js` into TSX mode where `<` can read as type arguments instead of comparison. Same constructs under `veryfront.config.js`, before and after:

| case | before | after |
| --- | --- | --- |
| comparison chain `a < b > (c)` | ok | ok |
| generic-looking call `f < g > (h)` | ok | ok |
| regex literal `/a<b>/g` | ok | ok |
| JSX in `.js` | ok | ok |
| template with angles, arrow, spread, optional chain | ok | ok |
| class fields (`#p`, `static`), dynamic `import()` | ok | ok |

No difference on any of them, including the ambiguous ones. TypeScript is a superset, so enabling it cannot reject valid JavaScript — it only widens what parses. The pre-existing `!filePath` branch already depended on that; this makes the behaviour uniform rather than introducing a new one.

Blast radius is small: `parser-only.ts` has two consumers, both the declarative config evaluator.

## What this does not fix

**Why a `.ts` config is labelled `.js`.** This stops the parser guessing language from an untrusted name; it does not explain the mislabel. The runtime `/files/` endpoint is *not* extension-tolerant — `veryfront.config.js` genuinely 404s for this release while `.ts` returns the content — so it is not simply the loader's candidate order. `VERYFRONT_RELEASE_ASSET_MANIFEST` is enabled in production and that seam carries its own `fileName`; that is where I would look next.

This matters beyond parsing: `DeclarativeConfigFileName` also feeds `configFileProvenance`, parse-error locations, and the hosted config cache key. If the release-asset path records `.js` for a `.ts` file, that is a data-integrity bug which will resurface somewhere less obvious — and this PR makes it silent rather than loud. Worth tracking separately rather than letting the disappearing 400 close the question.

**tomcode.com**, also down, with a different defect: `404 API request failed`. Its release has no `veryfront.config.*` at all, which should hit the loader's own `"No config file found, using defaults"` fallback. It does not, because the API-backed filesystem reports a missing file as `API_CLIENT_ERROR` with `status: 404`, while `isNotFoundError` matches only `ENOENT`/`ENOTDIR`, Deno's `NotFound`, or slug `file-not-found`. The 404 propagates instead of being read as absence. The fix belongs in the fs adapter — translate the transport 404 into the platform not-found error — not in the classifier, which should not sniff HTTP status or every API failure starts looking like an absent file.

Both defects were latent behind the missing worker entry fixed in #3399: that bug crashed the evaluator before it parsed anything, so this path had never run against a real project config in production.

## Tests

- `src/config/` — 44 passed, 402 steps.
- `extensions/ext-parser-babel/` — 3 passed, 36 steps.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved configuration parsing so TypeScript syntax works in `.js`, `.mjs`, and `.ts` configuration files.
  - Preserved support for standard JavaScript configuration files across supported extensions.
  - Fixed evaluation of TypeScript configurations using `as const`.
  - Improved handling of mixed JavaScript and TypeScript configuration content.

- **Chores**
  - Updated the application version to 0.1.1204.


kwakayama:
## Is "always TypeScript" actually safe? Measured, not assumed.

Fair question on review, and "TypeScript is a superset" is a claim worth testing rather than asserting — the real risk is that `typescript` + `jsx` together puts `.js` into TSX mode, where `<` after an identifier can be read as type arguments instead of a comparison.

Same constructs, parsed under `veryfront.config.js` on `main` and on this branch:

| case | before | after |
| --- | --- | --- |
| comparison chain `a < b > (c)` | ok | ok |
| generic-looking call `f < g > (h)` | ok | ok |
| regex literal `/a<b>/g` | ok | ok |
| JSX in `.js` | ok | ok |
| template with angles | ok | ok |
| arrow, object spread, optional chain | ok | ok |
| class fields (`#p`, `static`) | ok | ok |
| dynamic `import()` | ok | ok |

No difference on any of them, including the ambiguous ones. The `.ts` + JSX case fails in *both* columns, which is correct and unchanged: `.ts` does not get the `jsx` plugin, matching `tsc`.

## What this does not fix

This makes the parser stop guessing language from a name it cannot trust. It does **not** explain why a `.ts` config is being handed to the evaluator under a `.js` name, and that is worth keeping visible:

`DeclarativeConfigFileName` is not only a parser hint. It feeds `configFileProvenance`, the location attached to parse errors, and the hosted config cache key. If the release-asset seam records `.js` for a file that is actually `.ts`, that is a data-integrity problem which will resurface somewhere less obvious — and this PR makes it silent rather than loud.

So I would land this as hardening, on the grounds that a parser should not infer language from an untrusted name, and track the mislabel separately rather than let the disappearing 400 close the question.

Verified: `extensions/ext-parser-babel/` 3 passed / 36 steps; `src/config/` 44 passed / 401 steps.

PR #3404, opened 2026-08-06
fix(styles): skip an overlong CSS candidate instead of failing the page

kojiwakayama:
**This is the live cause of codersociety.com returning 500 in production right now.**

```
CSS pre-generation failed: CSS candidate tokens cannot exceed 1024 characters
```

`extractCandidates` threw, the throw reached `generateHTMLShellParts`, and stylesheet generation for the **entire page** died. One oversized token anywhere in a source file takes the whole page down. Reproduced against that project's real source, where `pages/index.mdx` contains one.

## Why skipping is safe

The bound was already enforced by the pattern itself, which caps a match at `MAX_CSS_SELECTOR_TOKEN_CHARACTERS - 1`. The throw fired on the *continuation* check — meaning the run was longer than the bound. A run that long is not a plausible class name (a base64 data URI or a minified blob in prose will do it), so dropping it costs nothing, while dropping the stylesheet costs the site.

The bound still holds: no candidate over the limit is emitted, and no fragment of the run is emitted either — which was the original guarantee.

## The test

It asserted the throw. Rewritten to assert the property that actually mattered — no fragments — **plus** that scanning continues, so a real class after the overlong run is still found. That last part is what makes skipping safe rather than merely quieter.

## Verified

Reproduced end to end against codersociety's real source running this framework: page went **500 → 200**, stylesheet from **not generated → 47 KB / 720 rules**. Full suite `3794 passed, 0 failed`.

## Context

Fourth defect from the same family as #3402 — a loud total failure where a skip belongs. Separate PR because it's an unrelated subsystem and #3402 already carried ten findings.

Note this alone does not fully restore codersociety: that project also needs its Tailwind v3→v4 migration (veryfront/codersociety#2). This fixes the crash; that fixes the theme.

## Summary by CodeRabbit

* **Bug Fixes**
  * Oversized CSS candidates are now skipped cleanly instead of causing an error.
  * Scanning continues after oversized content, allowing valid candidates that follow it to be detected.
  * Partial fragments from oversized candidates are no longer emitted.

### DIFF
--- a/src/html/styles-builder/candidate-tokenizer.ts
+++ b/src/html/styles-builder/candidate-tokenizer.ts
@@ -53,9 +53,21 @@ export function extractCandidatesWithByteLength(
         candidate.length > MAX_CSS_SELECTOR_TOKEN_CHARACTERS ||
         hasCandidateContinuation(content, candidatePattern.lastIndex)
       ) {
-        throw new NativeTypeError(
-          `CSS candidate tokens cannot exceed ${MAX_CSS_SELECTOR_TOKEN_CHARACTERS} characters`,
-        );
+        // Skip the over-long run instead of aborting the extraction.
+        //
+        // This used to throw, and the throw reached generateHTMLShellParts, so
+        // one oversized token anywhere in a page took down stylesheet
+        // generation for that whole page -- the page rendered unstyled. A run
+        // longer than the bound is not a plausible class name (a base64 data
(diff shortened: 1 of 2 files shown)


kojiwakayama:
Superseded by #3403, which has the same tokenizer fix plus two further render-blocking defects and the v0.1.1205 bump. I have carried this PR's three review findings over as a comment there — the char-at-a-time skip loop being the substantive one.

PR #3405, opened 2026-08-06
fix(css): stop an optional optimizer and a sourcemap blob failing every render

kwakayama:
Two independent CSS-pipeline defects. Both were latent until v0.1.1203 embedded the declarative evaluator worker (#3399) and this code first executed in production. Together they 500'd a live production site and failed the release asset build for **every** project — including a freshly scaffolded `npm create veryfront@latest`, which could not deploy at all.

This is the next layer under #3402, which fixed the `stylesheet-import-parse-failed` class and revealed these.

## 1. Missing `CSSOptimizationEngine` failed the build — all deploys

`acquireCSSGenerationSession` resolved the contract through `resolve()`, which throws when nothing registered it:

```
Release asset build failed (non-transform error)
  error: Missing extension for contract "CSSOptimizationEngine"
```

No first-party package installs a `CSSOptimizationEngine`. It ships only in `@veryfront/ext-css-lightning`, which is not a dependency of `veryfront`. Meanwhile `html-shell-generator.ts` hardcodes `minify: true` for production. So the fail-closed path was unreachable-by-design for any default project and guaranteed an outage.

It now resolves through `tryResolve()` and emits unminified CSS when no engine is registered. Output stays correct, only larger. The pipeline cache identity already encodes an absent optimizer as `unminified`, so nothing can serve a stale minified entry. `generateTailwindCSS` already branched on an absent engine — this just stops the session throwing before it gets there.

**This reverses the explicit \"fails closed\" contract** added with the optimizer seam, and replaces its test. That stance is right for correctness and security properties. It is wrong for a size optimization, where the fallback is a strictly working page and the alternative is a 500. Flagging it clearly since it was a deliberate prior decision.

## 2. An inline base64 sourcemap failed the candidate scan — live sites

```
Render failed: CSS candidate tokens cannot exceed 1024 characters
  at extractCandidatesWithByteLength (candidate-tokenizer.ts:56)
  at buildCandidateManifest (rendering/orchestrator/css-candidate-manifest.ts:126)
```

esbuild writes `//# sourceMappingURL=data:application/json;base64,...` into the build cache. Base64's alphabet (`A–Z a–z 0–9 + / =`) sits entirely inside the candidate body character class, so the payload reads as one unbroken multi-kilobyte token. A 2.2KB cached module carried a 1064-character payload and took the whole render down.

`.cache` is already in `DEFAULT_IGNORED_ROOTS`, but both scan entry points (`candidate-extractor.ts`, `css-candidate-manifest.ts`) apply that filter only when a `styleProfile` is passed, so these files were scanned regardless.

The ceiling exists to bound work, not to validate input — a 1024-character run can never match a rule. The run is now skipped whole, never split into fragments (which is what the previous test pinned), and each character is visited once so the scan stays linear.

## Verification

Verified against the real artifact, not a synthetic string. The cached module that crashed the renderer:

| | parent commit | this branch |
|---|---|---|
| `extractCandidates(content)` | throws `cannot exceed 1024 characters` | 40 candidates, longest 151 chars, payload absent |

- All 3 new tests fail on the parent commit and pass here.
- `src/html/` + `src/rendering/`: 148 passed, 0 failed (2375 steps).
- `src/html/styles-builder/`: 11 passed, 0 failed (140 steps).
- `deno check` and `deno lint` clean on changed files.

## Not covered here

The conditional `if (options.styleProfile && ...)` filter is left as-is in both scan paths. Fix 2 removes the failure mode regardless of whether the filter runs, but scanning the build cache for utility classes is still wasted work and worth tightening separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Overly long class-name candidates are now skipped safely instead of causing errors.
  * Valid classes can still be extracted when inline Base64 source maps are present.
  * CSS minification no longer fails when no optimization engine is available; readable CSS is returned instead.
  * Registered optimization engines continue to be used when available.

### DIFF
--- a/src/html/styles-builder/candidate-tokenizer.ts
+++ b/src/html/styles-builder/candidate-tokenizer.ts
@@ -29,6 +29,23 @@ function hasCandidateContinuation(content: string, index: number): boolean {
   return apply(regExpExec, candidateContinuationPattern, [nextCharacter]) !== null;
 }
 
+/**
+ * Advance past the remainder of a run that exceeded the token ceiling.
+ *
+ * The run is discarded whole rather than split: a fragment of it is not a
+ * candidate anybody wrote, and emitting fragments would pollute the snapshot.
+ * Each character is visited once, so skipping stays linear in the source.
+ */
+function skipOverlongCandidateRun(content: string, index: number): number {
+  let cursor = index;
+  while (cursor < content.length) {
+    const character = apply(stringCharAt, content, [cursor]) as string;
+    if (apply(regExpExec, candidateContinuationPattern, [character]) === null) break;
+    cursor++;
+  }
+  return cursor;
+}
+
 export interface ExtractedCSSCandidates {
   readonly candidates: string[];
   readonly sourceBytes: number;
@@ -53,9 +70,15 @@ export function extractCandidatesWithByteLength(
         candidate.length > MAX_CSS_SELECTOR_TOKEN_CHARACTERS ||
         hasCandidateContinuation(content, candidatePattern.lastIndex)
       ) {
-        throw new NativeTypeError(
-          `CSS candidate tokens cannot exceed ${MAX_CSS_SELECTOR_TOKEN_CHARACTERS} characters`,
+        // A run this long is not a class anyone authored -- an inline base64
+        // sourcemap is the common case, and its alphabet lies entirely inside
+        // the candidate body class. Skip it. Failing here would take down the
+        // whole render or release for a token that could never match a rule.
+        candidatePattern.lastIndex = skipOverlongCandidateRun(
+          content,
+          candidatePattern.lastIndex,
         );
+        continue;
       }
       if (apply(setHas, seenCandidates, [candidate])) continue;
       if (candidates.length >= MAX_CSS_SELECTOR_TOKENS) {
--- a/src/html/styles-builder/css-provider-session.test.ts
+++ b/src/html/styles-builder/css-provider-session.test.ts
@@ -6,7 +6,7 @@ import {
   type CSSProcessor,
   CSSProcessorName,
 } from "#veryfront/extensions/css/index.ts";
-import { assertEquals, assertRejects, assertThrows } from "#veryfront/testing/assert.ts";
+import { assertEquals } from "#veryfront/testing/assert.ts";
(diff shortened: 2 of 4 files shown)


kwakayama:
Folded into #3403 (`8207b0be8`) and closing this in its favour.

**What came across intact**

- `acquireCSSGenerationSession` now resolves `CSSOptimizationEngine` through `tryResolve` and emits unminified CSS when nothing registered one, with your reasoning preserved in the comment: no first-party package registers an engine while the production shell always asks for `minify: true`, so failing closed here was unreachable-by-design for a default project and could only ever fire as an outage.
- Your replacement for the fail-closed test, verbatim in substance — session acquires, `optimizationEngine` is `undefined`, generation returns the unoptimized string.
- The base64-sourcemap tokenizer case.

**Two changes on the way in**

1. **The log is `warn`, not `debug`.** A project that *did* select `ext-css-lightning` and whose registration failed now silently ships unminified CSS. That has to be distinguishable from a project that never wanted one.

2. **Added a test pinning the cache-identity split.** The reversal is only safe because an absent optimizer is recorded as `unminified` in the pipeline identity, so an unminified entry can never be served in place of a minified one. That property is load-bearing for this change and nothing was asserting it.

**One correction to the shared rationale**

Both this PR and #3403 said the ceiling was already enforced by the pattern, so only the continuation check could fire. That is not right, and it matters — #3403 acted on it and deleted the length arm.

The head `!?-?@?(?:[a-zA-Z0-9]|\[&?)` admits up to five characters on top of the `MAX - 1` body, so a match can reach 1028. A run of 1025–1028 that ends at a clean boundary has no continuation to catch it, and only `candidate.length > MAX` does. With the arm removed, `class="!` + 1024 chars emitted a 1025-character candidate that then threw in `normalizeCSSCandidates` — same failed render, one frame later, further from the cause.

This PR kept the arm, so the code here was always correct; only the prose was off. #3403 now keeps it too, with a regression test that fails when it is removed.

**Also carried over from #3404's review**

The skip is one sticky-regex scan rather than a character-at-a-time loop — `86bbdfda1` on #3403, extracted into a named helper in the fold.

Nothing from this PR is lost. The deploy-blocking half is the part #3403's call-site gating never reached: `release-assets/css-compile.ts` and `production-build/static-generation.ts` both call `acquireCSSGenerationSession(true)` directly, and fixing it in the session is what covers them.

PR #3414, opened 2026-08-06
fix(agent): make the resolver test typecheck against the fetch init union

kojiwakayama:
`ci (lint)` is red on `main`, so every branch cut from it inherits a failing check. Found while fixing #3413, whose lint job reported this file alongside its own error.

## Cause

`withMockFetch`'s `init` parameter is a union whose members do not all declare `headers`:

```
TS2339: Property 'headers' does not exist on type
  'RequestInit | (RequestInit & { client?: HttpClient }) | global.RequestInit'.
```

`src/agent/hosted/project-reference-resolver.test.ts:33` does `init?.headers`, and the file is not in `scripts/lint/test-typecheck-baseline.json`.

Verified it fails on `origin/main` itself, not just on a branch — this is not introduced by any open PR.

## Fix

Narrows with an `in` check rather than casting, matching #3382, which fixed the same union in `remote-mcp.test.ts` one day earlier. A cast would hide the next union member that lacks the property.

## Verification

- `deno check --no-lock src/agent/hosted/project-reference-resolver.test.ts` — clean (fails on main)
- `project-reference-resolver.test.ts` — 10 passed, 0 failed
- `deno lint` — clean

Kept to its own PR rather than folded into #3413, since it unblocks every branch.

### DIFF
--- a/src/agent/hosted/project-reference-resolver.test.ts
+++ b/src/agent/hosted/project-reference-resolver.test.ts
@@ -30,7 +30,9 @@ Deno.test("resolveHostedProjectReference returns a matching normalized API ident
     (input, init) => {
       requests.push({
         url: String(input),
-        authorization: new Headers(init?.headers).get("authorization"),
+        authorization: new Headers(
+          init && "headers" in init ? init.headers : undefined,
+        ).get("authorization"),
       });
       return Promise.resolve(
         Response.json({


kojiwakayama:
Closing this. **The premise was wrong: `main` is not red, and this file does typecheck.**

I claimed "verified it fails on `origin/main`". It does, but only the way I ran it — `deno check --no-lock`. Re-run against `origin/main` (`f306473b8`) both ways:

```
deno check          src/agent/hosted/project-reference-resolver.test.ts   -> Check (passes)
deno check --no-lock src/agent/hosted/project-reference-resolver.test.ts  -> TS2339, type checking failed
```

`--no-lock` resolves different npm type definitions than the frozen lockfile CI uses, and the `RequestInit` union differs between them. CI runs with the lock, so `init?.headers` narrows fine there. `lint:test-typecheck` confirms it: *"Test typecheck baseline holds: 51 grandfathered files, 0 new."*

The one real datapoint was #3413's lint failing at 06:22 and listing this file. It passes on the later run with the same file untouched, so that was a transient resolution state on an older base, not a defect in the file.

Since there is nothing failing, this is churn against working code, and the `in` narrowing would encode a constraint the toolchain does not actually impose. #3413 carries its own genuine fix and is green on its own.

Lesson for me: `--no-lock` is not a valid way to reproduce this repo's CI typecheck.

PR #3415, opened 2026-08-06
release: v0.1.1206

kwakayama:
Bumps the framework to **v0.1.1206** so the two render-path fixes now in the merge queue can reach production.

## What this release carries

| PR | fix |
| --- | --- |
| #3409 | module pin conflict gated on the cohort, not the flag — every client module of every hosted project answered `409 Unknown dependency snapshot`, so no page hydrated |
| #3411 | default CSP admits the platform asset origins the renderer emits — React from the ESM CDN and optimized images were blocked by the policy attached to the same response |

Both are green and queued ahead of this PR (positions 3 and 4). This one is deliberately last so a single release ships both.

## Why it is a separate PR

The bump belonged in #3411, but that branch is already in the merge queue and GitHub refuses pushes to queued branches:

```
A pull request for this branch has been added to a merge queue.
Branches that are queued for merging cannot be updated.
```

Dequeuing a green PR to slip a version line in is more disruptive than a three-line release PR, so this is the smaller move.

## Contents

`deno.json`, `src/utils/version-constant.ts`, and the generated hydration runtime, which embeds the version string. Three files, +3/-3.

Regenerating the hydration runtime is not optional: it carries the version literal, and a stale copy reds `ci (typecheck)` rather than failing anywhere obvious. Verified no `0.1.1205` remains in any of the three.

## After this merges

Publishing v0.1.1206 is only the first link. Production still needs: framework release → `repository_dispatch` → veryfront-server rebuild → staging verification → production promotion PR. Until that finishes, hosted pages still do not hydrate.


kwakayama:
Superseded by #3417, which consolidates this with the other queued render-path fixes plus the v0.1.1206 bump so one queue pass lands the set.

Dequeued from the merge queue to avoid this merging underneath the consolidated branch and leaving it empty or conflicting. Please do not re-queue — merge #3417 instead. Closing once that lands.

PR #3420, opened 2026-08-06
fix(proxy): stop shedding release-asset cold loads onto the client

kojiwakayama:
## Description

A page load requests every asset in its module graph at once. On a cold pod the release-asset handler admitted only **4 concurrent loads plus 16 queued**, so the 21st distinct hash was rejected outright with a `503`. Browser `import()` never retries, so one shed asset killed hydration for the entire page.

Observed in production on `j-agent.production.veryfront.com`: one load 503'd 31 assets, and the next reload 503'd 11 *different, deeper* hashes — `dca28b9f…` was a 503 victim in the first log and a successful referrer in the second. That is a per-pod cache warming one module-graph level per reload and hitting the same wall on the next fan-out level.

### What changed

Two of the three bounds were guarding nothing:

- **`MAX_QUEUED_COLD_LOADS` (16)** turned "busy" into "this module does not exist". A queue cap here is a hard page failure by construction, because `import()` has no retry and `Retry-After: 1` is decoration.
- **`MAX_COLD_LOAD_WAITERS` (64)** rejected *followers of a load that had already succeeded* — callers whose bytes were sitting in memory, byte-identical for all of them. Nothing was being protected.

Both are removed. `PermitSemaphore` keeps its own 10,000-entry backstop, and callers were already bounded by their own deadline (`timeoutMs`) and the producer ceiling (`MAX_UPSTREAM_TIMEOUT_MS`). A saturated proxy now surfaces as latency and finally an honest `504`, never a phantom `503`.

**`MAX_CONCURRENT_COLD_LOADS = 4` is unchanged.** It bounds the one real resource — each in-flight load buffers a whole asset in memory, so peak bytes is this count × `RELEASE_ASSET_MAX_SIZE_BYTES`. Retuning it is a performance question, not this bug.

**`MAX_CACHED_ASSETS` 100 → 2000.** Measured against a real page (65 assets, 185KB total): mean asset is ~2.9KB, so 100 entries could only ever hold ~285KB of the 32MB byte budget. The entry cap did 100% of the evicting and the byte budget was ~115× out of reach. At 100 entries a single page nearly filled the per-pod cache — shared across every project on that pod — which made cold-load storms steady state rather than a deploy-time event.

## Type of Change

## Checklist

Both updated tests had pinned the defect as intended behaviour:

| Test | Before | After |
|---|---|---|
| 50 concurrent distinct assets | `19 × 200`, `30 × 503` | `49 × 200`, `0 × 503` (1 × 499 = deliberately disconnected caller) |
| 80 followers of one same-hash load | `64 × 200`, `16 × 503` | `80 × 200`, `0 × 503`, still `calls === 1` |

`maxActive` is still asserted at 4, confirming queueing did not widen the bound that guards real memory. The existing byte-weight eviction test still passes, so the 32MB budget still binds.

**Verification:** proxy suite 49 passed / 482 steps / 0 failed; full pre-push suite 3798 passed / 27834 steps / 0 failed; `deno check`, `deno lint`, `deno fmt --check` clean.

## Notes for review

- I could not reproduce a live 503 — the assets are warm now, and cache hits return before the admission gate. The diagnosis rests on the code paths plus the progressive-warming pattern across the two production logs. Real confirmation is watching a cold pod after deploy.
- `MAX_CONCURRENT_COLD_LOADS = 4` is still a guess at what `/release-assets/{hash}` will absorb. If cold loads feel slow after this, that is the number to revisit — and it is now the only one left to tune.
- Deliberately **not** included: a circuit breaker for upstream-down fast-fail, and bounded retry-with-jitter in the hydration runtime. Both are defensible later; neither is needed to fix this.

## Summary by CodeRabbit

- **Bug Fixes**
  - Asset requests exceeding the active loading limit now wait in a queue instead of immediately returning an error.
  - Queued requests can still be canceled if the caller aborts.
  - Concurrent requests for the same asset now all receive the shared successful response, without unnecessary rejections.
  - The system continues to limit active upstream asset loads to four at a time.

### DIFF
--- a/src/platform/compat/dynamic-import.test.ts
+++ b/src/platform/compat/dynamic-import.test.ts
@@ -19,4 +19,18 @@ describe("platform/compat/dynamic-import", () => {
       () => dynamicImport("__nonexistent_module_12345__"),
     );
   });
+
+  // This module is reachable from client bundles (veryfront/chat, /mdx,
+  // /workflow pull it in transitively). Project pages are served with
+  // `script-src 'self' 'nonce-...' https://esm.sh` — no 'unsafe-eval' — so a
+  // `new Function` here throws EvalError in the browser and blanks the page
+  // before hydration starts.
+  it("should not use new Function, which the page CSP blocks", () => {
+    const source = Deno.readTextFileSync(new URL("./dynamic-import.ts", import.meta.url));
+    assertEquals(
+      /\bnew Function\s*\(/.test(source),
+      false,
+      "dynamic-import.ts must not use new Function — project CSP forbids 'unsafe-eval'",
+    );
+  });
 });
--- a/src/platform/compat/dynamic-import.test.ts
+++ b/src/platform/compat/dynamic-import.test.ts
@@ -19,18 +19,4 @@ describe("platform/compat/dynamic-import", () => {
       () => dynamicImport("__nonexistent_module_12345__"),
     );
   });
-
-  // This module is reachable from client bundles (veryfront/chat, /mdx,
-  // /workflow pull it in transitively). Project pages are served with
-  // `script-src 'self' 'nonce-...' https://esm.sh` — no 'unsafe-eval' — so a
-  // `new Function` here throws EvalError in the browser and blanks the page
-  // before hydration starts.
-  it("should not use new Function, which the page CSP blocks", () => {
-    const source = Deno.readTextFileSync(new URL("./dynamic-import.ts", import.meta.url));
-    assertEquals(
-      /\bnew Function\s*\(/.test(source),
-      false,
-      "dynamic-import.ts must not use new Function — project CSP forbids 'unsafe-eval'",
-    );
-  });
 });
--- a/src/platform/compat/dynamic-import.ts
+++ b/src/platform/compat/dynamic-import.ts
@@ -1,15 +1,18 @@
 /**
  * Opaque dynamic import helper.
  *
- * Uses `new Function` to hide the `import()` call from static analysis
- * by bundlers and `deno compile`, preventing them from tracing into
- * the imported specifier.
+ * The specifier is a runtime parameter, so bundlers and `deno compile` cannot
+ * resolve it to a concrete module and will not trace into it. That opacity is
+ * what this helper exists for.
+ *
+ * This deliberately does NOT use `new Function`. This module is reachable from
+ * client bundles (`veryfront/chat`, `veryfront/mdx`, `veryfront/workflow` all
+ * pull it in transitively), and the CSP served for project pages does not allow
+ * `'unsafe-eval'`. A `new Function` here throws `EvalError` in the browser
+ * before hydration can start, which blanks the page.
  *
  * @module platform/compat
  */
 
-export const dynamicImport = new Function("specifier", "return import(specifier)") as <
-  T = unknown,
->(
-  specifier: string,
-) => Promise<T>;
+export const dynamicImport = <T = unknown>(specifier: string): Promise<T> =>
+  import(specifier) as Promise<T>;
--- a/src/platform/compat/dynamic-import.ts
+++ b/src/platform/compat/dynamic-import.ts
@@ -1,18 +1,15 @@
 /**
  * Opaque dynamic import helper.
  *
- * The specifier is a runtime parameter, so bundlers and `deno compile` cannot
- * resolve it to a concrete module and will not trace into it. That opacity is
- * what this helper exists for.
- *
- * This deliberately does NOT use `new Function`. This module is reachable from
- * client bundles (`veryfront/chat`, `veryfront/mdx`, `veryfront/workflow` all
- * pull it in transitively), and the CSP served for project pages does not allow
- * `'unsafe-eval'`. A `new Function` here throws `EvalError` in the browser
- * before hydration can start, which blanks the page.
+ * Uses `new Function` to hide the `import()` call from static analysis
+ * by bundlers and `deno compile`, preventing them from tracing into
+ * the imported specifier.
  *
  * @module platform/compat
  */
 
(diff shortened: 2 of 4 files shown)


kwakayama:
Pushed `0c24ce8` onto this branch. Two things this adds: a live reproduction of the 503 (the PR notes it could not be reproduced), and a fix for a second, independent bug that also blanks the page.

## Live 503 reproduction

Reproduced on `support-agent-vkmo7d.production.veryfront.com` against current prod, so the shedding is still reachable and not just warm-cache-hidden.

One page load requests **143 distinct asset hashes** against `MAX_CACHED_ASSETS = 100`. The cache cannot hold a single page's module graph, so it thrashes and re-cold-loads every time. Five consecutive loads:

| load | 503s |
|---|---|
| 1 | 16 |
| 2 | 19 |
| 3 | 5 |
| 4 | 0 |
| 5 | 0 |

The 143-vs-100 measurement supports the `MAX_CACHED_ASSETS` 100 → 2000 change directly — the entry cap, not the byte budget, is what evicts.

Worth noting the failure needs the browser's concurrent HTTP/2 burst. Every "503" asset returns 200 when fetched on its own — 50/50 sequential and 30/30 parallel curls all succeeded. Anyone checking with curl will conclude the assets are fine.

## Second bug: CSP blocks the hydration runtime's own `new Function`

This is deterministic and this PR did not cover it. Loads 4 and 5 above shed **zero** assets and still failed to hydrate:

```
[Veryfront] Client initialization error: EvalError: Evaluating a string as JavaScript
violates the following Content Security Policy directive because 'unsafe-eval' is not
an allowed source of script: script-src 'self' 'nonce-...' https://esm.sh
    at new Function (<anonymous>)
    at .../_vf/assets/bfbed417...js:1:9
[Veryfront] Hydration failed signal received
```

That asset is `src/platform/compat/dynamic-import.ts` compiled as its own client chunk:

```js
const i=new Function("specifier","return import(specifier)");export{i as dynamicImport};
```

The document CSP has no `'unsafe-eval'`, so it throws before hydration starts. Page stays on skeleton loaders permanently; `document.querySelectorAll('*').length` is 61 and the text is `Loading messages...`.

It reaches the client from three entry points, all converging on `platform/compat/process/command.ts` (a `node:child_process` module):

```
chat/index.ts    -> stream-watchdog.ts -> watchdog-compat-adapter.ts -> compat/process.ts
mdx/index.ts     -> MDXProvider.tsx -> types/index.ts -> types/server.ts -> config/loader.ts -> ...
workflow/react/  -> use-workflow.ts -> workflow/types.ts -> agent/types.ts -> tool/context7.ts -> ...
```

The `new Function` existed to keep the import opaque to bundlers and `deno compile`. It was not needed — the specifier is a runtime parameter, so neither can resolve it. Verified with `deno compile` on the bare form: it bundles only the local files, does not trace into the specifier, and the binary still resolves it at runtime. `platform/adapters/fs/veryfront/default-invalidation-callbacks.ts` already uses the same bare form.

Regression test asserts the source has no `new Function`; confirmed red before the fix, green after.

**Verification:** `src/platform/compat/`, `runtime-owner`, `script-handlers`, and this PR's `asset-handler` tests — 87 passed / 995 steps / 0 failed. `deno check`, `deno lint`, `deno fmt --check` clean.

## Scope note

Happy to split the CSP commit into its own PR if you would rather keep this one purely about proxy admission — it touches a different subsystem. It is one commit and cherry-picks cleanly. Landing them together does mean the page actually loads, which neither fix achieves alone.

Also unaddressed by both fixes: `/_vf_modules/pages/index.js` 404s on every load (the `loadPageModuleWithIndexFallback` probe). Harmless but it is noise in every production console.

kwakayama:
**Correction to my previous comment — I reverted the CSP commit (`466ac09`). This branch is back to exactly its original two files.**

My fix was wrong and it broke `src/release-assets/scaffolded-project-build.test.ts` (5 scaffolds). Sorry for the red.

I checked that removing `new Function` was safe for `deno compile` and it is — but that was only half the reason it exists. The release-asset builder rejects it for its own reason:

```
specifier=veryfront/chat  sourceKey=platform/compat/dynamic-import
error="Release module contains a non-literal dynamic import"
→ dependency-transform-failed:veryfront<path>  module-rewrite-failed:app<path>
```

Every client import has to be rewritable to a content-addressed `/_vf/assets/{hash}.js` URL. A runtime-variable specifier cannot be rewritten, so `new Function` is what keeps this module out of the rewriter's way. It is load-bearing in the build, not just in `deno compile`. Confirmed by bisect: the test passes on `713f00cc` and on `main`, fails only with my commit.

**The two findings in my previous comment still stand.** The 503 reproduction is unaffected (143 distinct assets vs `MAX_CACHED_ASSETS = 100`; 16/19/5/0/0 across five loads), and it still supports the 100 → 2000 change here.

The CSP bug is also still real and still blanks the page — loads 4 and 5 shed zero assets and died on `EvalError` before hydration. It just needs a different fix than the one I attempted. The right shape looks like breaking client reachability rather than touching the helper: `platform/compat/process/command.ts` is a `node:child_process` module and has no business in a browser bundle, but three client entries reach it —

```
chat/index.ts    -> stream-watchdog.ts -> watchdog-compat-adapter.ts -> compat/process.ts
mdx/index.ts     -> MDXProvider.tsx -> types/index.ts -> types/server.ts -> config/loader.ts -> ...
workflow/react/  -> use-workflow.ts -> workflow/types.ts -> agent/types.ts -> tool/context7.ts -> ...
```

Cutting those (lazy import at the call site, or splitting the type-only imports) removes the eval from the client graph without touching the helper's build behaviour. That is a bigger change than belongs on this PR, so I am leaving it off. Happy to open a separate issue or PR for it — say the word.

Nothing here blocks this PR; it is back to what you had.

PR #3429, opened 2026-08-06
fix(runtime): finalize provider-executed tool calls the provider never resolves

kwakayama:
## Problem

A provider-native Anthropic tool call (`web_fetch` / `web_search`, `srvtoolu_` ids) can emit `tool-input-start` and `tool-input-available` and then never produce a result. Observed on a production run:

- The tool card stayed at `input-available` and spun for the rest of the stream.
- When the run finished, `isToolComplete()` (`src/chat/conversation.ts`) judged the part incomplete, persistence dropped it, and the card vanished on rehydrate.

In the persisted run events this reads as `TOOL_CALL_START` + `TOOL_CALL_END` with no `TOOL_CALL_RESULT`. Every locally executed tool in the same run got a result; the two provider-executed ones did not.

## Two causes

**1. The stream was truncated before the result could arrive.** Once any local tool call commits, the read loop waits only `LOCAL_TOOL_COMMIT_GRACE_MS` (250 ms). A provider tool result can arrive on a *later provider continuation request*, so any turn mixing a provider call with local calls loses the provider result. In the observed run, two local tools committed in the same turn as the provider call.

**2. Nothing closed out a provider call that genuinely returned nothing.** There was no terminal event, so the part had no way to leave `input-available`.

## What changed

One production file, `src/agent/runtime/chat-stream-handler.ts`.

- While a provider-executed call is outstanding, the local-tool commit grace no longer arms. **That gate selects the timeout only.** It deliberately does not feed the finish-reason classification — classifying a committed local tool call as `"stop"` makes `shouldContinueAfterStreamStep()` bail, and the local tool then never executes and is itself stranded at `input-available`.
- Any provider-executed call left unresolved gets a terminal `tool-output-error` part, emitted from a `finally` so aborts and stream errors are covered.
- **The synthesized event is stream-only.** It is not recorded in `state.toolResults`. Doing so would make the runtime believe the provider answered — flipping `shouldContinueAfterStreamStep()` to true and billing an extra model call per unresolved call, up to `maxSteps`. The runtime keeps its unchanged view while the client still gets a terminal part.

## Not in scope

Model-history replay is unchanged. A provider-executed call is still dropped from the messages replayed to the model, exactly as on `main`. Fixing that requires carrying Anthropic's exact raw-replay metadata; an earlier attempt in this branch was reverted because forwarding a provider-executed call *without* that metadata makes `buildAnthropicMessagesRequest` throw a `TypeError` on the very next model call — including on routine `web_fetch` failures like `url_not_accessible`. That is a separate PR.

## PR #3043 constraint

Ownership is read from provider-executed metadata only. No tool-name heuristic is introduced, so a locally executed `web_fetch` sharing the name is unaffected and still finalizes only on its real result. A regression test pins this.

## Known gap: `streamLifecycleMode: "active"`

Both changes live below the `processActiveStream()` early return, so `active` still reproduces the incident — `reducer.ts` leaves a `providerExecuted` tool parked at `input_ready` with no terminal frame.

No exposure today: `resolveStreamLifecycleModeFromEnv()` defaults to `legacy` and nothing in-repo sets `VF_STREAM_LIFECYCLE_MODE`. Rather than bolt a reducer port onto this PR (3-4 files inside the component being rolled out, and it would contradict this PR's shadow-parity fix), the gap is gated:

- A blocking **"Provider-executed tool finalization"** row in the gate table in `docs/internal/stream-lifecycle-rollout.md`, advance condition *do not advance past `shadow`*.
- A tripwire test pinning current `active` behaviour, so the gap fails loudly the day the reducer port lands.

## Reviewing this diff

`chat-stream-handler.ts` shows ~1000 changed lines but only **~120 are real** — the rest is re-indentation from the `try`/`finally` wrap. Use **[?w=1](https://github.com/veryfront/veryfront-code/pull/3429/files?w=1)** (hide whitespace) or `git diff -w`.

## Behaviour note

Holding the stream open costs latency in the failure case: a mixed turn whose provider result never arrives now waits out `STREAM_OUTPUT_IDLE_MS` instead of truncating at 250 ms. That is bounded and only on the path that was previously losing data.

## Verification

- `deno task test:unit` — 3799 passed, 0 failed
- `deno task test:integration` — 264 passed, 0 failed
- `deno task verify:quick` — the full 22-task `ci (lint)` chain, exit 0
- Every new test proven load-bearing: reverting the production change turns them red, including the `#3043` local-`web_fetch` regression guard staying green on baseline

The change was reviewed by independent adversarial passes covering root cause, hosted-tool regression, stream latency, and test adequacy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of provider-executed tool calls so streams reliably reach a terminal state.
  * Added clear tool-output error events when a provider does not return a result.
  * Prevented streams from ending prematurely while tool results are still pending.
  * Ensured interrupted or failed responses finalize unresolved tool calls without duplicate events.
  * Preserved correct handoff and completion behavior for locally executed tools.

* **Documentation**
  * Updated rollout guidance for provider-executed tool finalization and rollback scenarios.


kwakayama:
@coderabbitai review

kwakayama:
All six findings addressed in c131f81ec. Detail per item:

**Keep provider calls pending for preliminary results** (`chat-stream-handler.ts:1191`) - valid, and the most useful finding here. `preliminary` is first-class in this codebase (`tool-result-continuation.ts:80` skips preliminary entries from final results), so releasing the call there re-armed the local commit grace while the provider was still working and truncated the stream before the final result. That is the exact failure this PR exists to prevent, reintroduced on a path none of my verification passes probed. The call is now released only for a final result or error. Added "keeps a provider tool call pending across a preliminary result", confirmed load-bearing: reverting the guard turns it red in 43 ms.

**Close the stream iterator on every exit** (`chat-stream-handler.ts:861`) - valid. `streamIterator` is declared before the `try`, released through a `returnStreamIteratorOnce()` helper that is idempotent, called from the shared `finally`. The timeout path uses the same helper, so no double release.

**Replace the em dash** (`stream-lifecycle-rollout.md`) - valid. Went wider than the single flagged line: the baseline `chat-stream-handler.ts` contains zero em dashes, so the AGENTS.md rule is repo-wide rather than markdown-only. Removed all 12 introduced by this PR, across code comments and docs.

**Real wall-clock delays near `LOCAL_TOOL_COMMIT_GRACE_MS`** - valid. Took the suggested route: `localToolCommitGraceMs` is now a `ChatStreamCallbacks` option. The six grace-sensitive tests drive an explicit 40 ms grace against a 200 ms delay, so the margin goes from 1.6x to 5x and that suite got roughly 6x faster.

**Use `#veryfront/*` aliases** - valid, applied to the three cross-directory imports. Same-directory imports left relative.

**Assert the incomplete-tool-call error code** - valid, now asserts `terminalErrorCode === "INCOMPLETE_TOOL_CALLS"`.

Verification: `deno task test:unit` 3799 passed / 0 failed, `deno task verify:quick` (the full 22-task chain) exit 0.

Note for reviewing the `chat-stream-handler.ts` diff: it reads as roughly 1000 changed lines but only about 120 are real, the rest being re-indentation from the `try`/`finally`. Use `?w=1` or `git diff -w`.

kwakayama:
Addressed in 5d20b34b9.

**Finalize calls that have only preliminary results** - valid, and a real hole in my previous commit. Keeping the call pending across a preliminary result was only half the path: `finalizeUnresolvedProviderToolCalls` built its resolved set from every entry in `state.toolResults`, and preliminary results are recorded there like any other. So a stream ending after only a preliminary result left the call pending, then skipped it as resolved, and emitted no terminal part. That strands the card exactly as before this PR.

Only non-preliminary results count as terminal now. Written test-first per the guideline: `finalizes a provider tool call that only produced a preliminary result` asserts a `tool-output-error` after `tool-input-end`, a preliminary `tool-result` and `finish`. Confirmed it reports zero terminal events without the change.

Verification: `deno task test:unit` 3799 passed / 0 failed, `deno task verify:quick` exit 0.

kwakayama:
@coderabbitai review

PR #3432, opened 2026-08-06
fix: stop one unbuildable page from 503ing every route

kwakayama:
## Why

A production site served a dead page for seven weeks and the console gave three
errors that turned out to be three unrelated framework problems. All three hit a
project with **no config file at all**, which is the case this PR is about: a
project that configures nothing should still render.

### 1. One unbuildable page 503'd every route

A page whose import closure could not be finalized pushed a coverage gap into the
fatal set, failing the **entire** release asset manifest. The renderer then had no
manifest to admit browser modules against, so every `_vf_modules/*` request on
**every** route returned 503 with `Browser module manifest unavailable`.

The failure is near-invisible: SSR HTML still renders, so `curl` looks healthy and
the page looks healthy. Nothing hydrates. The site had one leftover scratch page
importing a URL that had since become a sign-in redirect returning HTML instead of
JS, and every healthy page went down with it.

Module-level failures now degrade **per route**. A module that cannot be finalized
is absent from `modules`, so the admission boundary still refuses it, and routes
whose closure needs it are omitted rather than published with a hole. Gaps only
become fatal when nothing serveable is left, so a project whose only page is broken
still fails closed and keeps its previous release.

### 2. The default CSP blocked the framework's own fonts

`veryfront/fonts` (`GoogleFonts`) writes a `fonts.googleapis.com` stylesheet and a
`fonts.gstatic.com` preconnect into the document, but the default policy allowed
neither. A project using the framework's own font API rendered in a fallback font
with nothing but a console error to show for it.

Both origins go in the **droppable baseline**, not the required floor — the renderer
only emits them for projects that call the component, so
`security.csp: { styleSrc: null, fontSrc: null }` still removes them.

This narrows a rule the code stated explicitly ("the floor contains what the platform
emits, never a guess at what a project needs"). Google Fonts turned out to be on the
platform-emits side of that line once `veryfront/fonts` shipped; the comments are
updated to say so.

### 3. `veryfront/ui` was exported but unresolvable

`deno.json` has exported `veryfront/ui` for as long as the barrel has existed, but
neither the SSR nor the browser import map named it. A project importing it shipped a
bare specifier the browser could not resolve — and the release build then counted the
importing module as uncovered, which before this PR was fatal to the whole manifest.

`default-import-map.ts` already carried the rule in a comment: a new `deno.json`
export containing React hooks or components has to be added there too. `veryfront/ui`
is exactly that and was missed.

Also exports `useDocumentNonce` from the barrel. Third-party providers that render
their own inline script (next-themes is the common one) take the nonce as a prop and
emit an unusable empty attribute without it, which the default CSP then blocks. Only
the non-hook getter was reachable from project code before, and only in theory, since
the specifier did not resolve.

## Verification

- 407 test files / 4106 steps green across `release-assets`, `modules`, `html`,
  `security`, `react`.
- New regression tests: a project with one broken page publishes its healthy route;
  a project where every page is broken still fails closed.
- Typecheck, lint, fmt clean.
- The originating outage is resolved on the affected site: manifest `failed` -> `ready`,
  `/_vf_modules/pages/index.js` 503 -> 200, page fully hydrated, deleted route 404s
  (fail-closed) rather than 503ing.

## Risk

The route-isolation change touches a security-sensitive build path. The admission
boundary itself is unchanged — a module absent from the manifest is still refused.
What changes is only whether one module's absence also withholds every other module.

## Summary by CodeRabbit

* **New Features**
  * Added `useDocumentNonce` to the public UI toolkit.
  * Added server-rendered application support for the UI toolkit.
  * Added Google Fonts stylesheet and font-file origins to the default security policy.

* **Bug Fixes**
  * Release builds can publish healthy routes when another route has unresolved imports.
  * Builds fail safely when no routes can be served.

* **Documentation**
  * Expanded security guidance for Google Fonts and custom font services.

* **Chores**
  * Updated the release version to 0.1.1208.


kwakayama:
Both review findings were real. Addressed in 9674d8411.

**Transform-stage failures as route-local gaps** — correct, and the more
important of the two. `moduleGaps` only covered finalization onward, so a page
failing at transform, at the size check, or at import parsing still pushed a
fatal gap. An MDX or TSX compile error in one page kept taking down every other
route, and that is at least as common as an unresolvable import. Those three now
record route-local gaps.

One deliberate exception: `module-dependency-vendor-failed` stays fatal. Unlike
the others, that path does not drop the module — it continues with the
*unvendored* code, and the gap is what stops that result from being published.
Degrading it would let a module whose dependencies were never vendored reach a
manifest. The existing accessor-backed vendor test guards this and still passes.

**Transform-stage test coverage** — added, with the fixture shape suggested:
transform rejects for one page and resolves for the other, asserting the healthy
route publishes and the broken one is omitted. Added the inverse too (every page
fails to transform → still fails closed), since per-route degradation must not
become "publish an empty manifest".

One thing the change surfaced beyond the review: a structural gap fails the build
immediately, and the module gaps collected on the way there were being dropped
from the report. A release failing on a dependency error never named the page
that actually broke. `assertCompleteReleaseAssetCoverage` now merges them for
reporting, which is what the two vendor tests were implicitly asserting.

Also updated `fails closed when one module transform fails` — it encoded the old
all-or-nothing contract this PR changes. It now asserts the half that still
holds: the failed module never reaches the manifest, so the browser-module
endpoint keeps refusing it.

Verification: 407 test files / 4107 steps green; typecheck, lint, fmt clean.

PR #3439, opened 2026-08-06
refactor(chat): give tool replay reconciliation its own module

kojiwakayama:
Pure extraction, **zero behaviour change**. Splits `src/chat/conversation.ts` along its real seams and gives the tool-replay algorithm a name.

## Why

`conversation.ts` was a 1563-line grab-bag hiding a genuinely deep algorithm: ~260 lines that reconcile tool-call/result history by **part object identity** (WeakSet/WeakMap, single pass) to decide which occurrences are authoritative for provider conversion. It had no name and no home. The same file also exported generic value guards that had nothing to do with conversations.

## What changed

A three-layer ladder, each importing only downward:

| Module | Lines | Owns |
| --- | --- | --- |
| `part-field-access.ts` | 66 | reading one field off an untrusted `unknown` |
| `message-part-parsing.ts` | 264 | interpreting one message part — tool, text, reasoning, or file |
| `tool-replay-reconciliation.ts` | 294 | deciding which occurrences in a history of parts are authoritative |
| `conversation.ts` | 1018 | was 1563 |

Net **+80 lines** across `src/chat/`. This is deliberately not a line-reduction change — the gain is that three named single-purpose modules replace one grab-bag, and `CONTEXT.md` now records two new domain terms.

Also adds **LOC ceilings** for these five files to `ban-chat-antipatterns.ts`. `src/chat/` previously had no ceiling at all, which is how `conversation.ts` reached 1563 lines. Precedent: `modules/server/module-server.ts` was refactored once and then regrew 1138 → 1806 lines across six follow-on PRs, ending larger than before. Extraction without a ceiling regresses.

## Evidence

- `deno task test:unit`: **3804 passed / 27918 steps / 0 failed / 1 ignored** — identical to the pre-refactor baseline of 3801/27907 plus exactly the 11 new test steps. Held at every commit.
- `deno task verify:quick`: exit 0.
- `deno task dupes`: **232 groups, unchanged** from branch start — the refactor introduced no duplication.
- Whole-branch diff of `conversation.ts` adds exactly **22 lines, all imports and re-exports**; every other line is a deletion. No logic in that file was edited anywhere on the branch.
- The full set of top-level declarations at base was diffed against the union of the four files today: byte-identical sorted sets. Nothing lost, nothing invented.
- **Object identity traced end to end** through the matcher — no part object is cloned, re-wrapped, or reconstructed. Notably `toRecord`, the one constructing function in `part-field-access.ts`, is not imported by the reconciliation module.
- The new ceiling was proven to fire (`86 LOC exceeds ceiling 66`) and then clear.

Every task was implemented and reviewed in separate passes, with a fix round where review found a problem.

## Notes for reviewers

- `conversation.ts` keeps re-exporting `getStringField`, `isRecord`, `stringifyUnknown` — `isRecord` alone has ~8 importers across `src/agent/`. Consolidating those is a separate increment (there are ~76 private `isRecord` copies repo-wide); this branch gives the canonical one a home first.
- `tool-replay-reconciliation.ts` imports `ChatProviderModelInputMessage` from `conversation.ts` as `import type`. It must stay type-only — a value import would create a real cycle. `deno check` would not catch that (types erase at emit) but `lint:module-boundaries` will; there is a comment at the import saying so.

## Known follow-ups (deferred, not blocking)

- `findProviderVisibleToolReplayMatches` — the algorithm this branch exists to expose — has only a smoke test. A direct table-driven test is the highest-value follow-up and is now easy to write against the extracted interface.
- 8 of 10 exports in `message-part-parsing.ts` lack JSDoc (all were private and undocumented before; the verbatim-move rule preserved that).
- `conversation.ts`'s ~423-line provider-conversion block is the obvious next seam, leaving ~600 lines that finally match the file's name. **Sequencing matters:** extract the `ChatProviderModelInput*` types first, or the type-only back-edge simply relocates instead of disappearing.

## Summary by CodeRabbit

- **Improvements**
  - Improved chat message handling across text, reasoning, files, images, tool calls, and tool results.
  - More reliably interprets alternate message formats and safely handles incomplete or invalid content.
  - Improved replay behavior by preserving relevant transient tool activity and preventing duplicate or outdated tool results.
  - Tool outputs are now represented more consistently, including clearer handling of pending and failed operations.

- **Quality**
  - Added automated coverage for message parsing, field handling, and tool replay scenarios.
  - Added safeguards to maintain consistent source-file quality.


kwakayama:
All six nitpicks from the review body triaged. Four applied in 67ffa89b7, one was already fixed, one is rejected on evidence.

**Applied**

- `part-field-access.test.ts` — added the JSON fallback cases for `stringifyUnknown` (record, array, `null`) plus `toJsonValue`, so a change in `json-value.ts` fails here rather than in a provider conversion test.
- `message-part-parsing.test.ts` — added cases for `hasSelfContainedRawToolCallResult`. The `state === "error"` guard with no `output` and no `errorText` is the load-bearing one: `buildToolResultOutput({state:"error"})` returns `{type:"error-text", value:"Tool error"}`, so without the guard that call would supersede its paired result. The test asserts both halves so the guard cannot be removed silently. Also covered the stateless case and the `getFilePart` url requirement.
- `part-field-access.ts` — `JsonValue` is now an alias of `ChatJsonValue` rather than a second hand-rolled copy of the same recursive shape. The export is preserved for callers.
- `message-part-parsing.ts` — renamed the `getFilePart` local from `data` to `url`, since it only ever reads `part.url`. No `data` fallback added; `ChatUiMessagePart` requires `url` for file parts.

**Already fixed**

- `tool-replay-reconciliation.test.ts` — the `WeakSet`/`WeakMap` shape assertions are gone. 1eba2585a replaced them with the contract cases (nearest same-id match, supersession, batch reset after intervening content, transient preservation). The review ran against f14619d15, one commit earlier.

**Rejected**

- `tool-replay-reconciliation.ts` — the claim that `lint:module-boundaries` excludes type-only imports from cycle detection, so the comment overstates the safeguard. I tested it. Converting line 25 to a value import and running `deno task lint:module-boundaries` gives exit 1 and:

```
New dependency edges that participate in cycles:
  src/chat/conversation.ts -> src/chat/tool-replay-reconciliation.ts
  src/chat/tool-replay-reconciliation.ts -> src/chat/conversation.ts
```

The exclusion at `check-module-boundaries.ts:406` applies to type-only edges, which is precisely why the current import is safe. The comment claims a *value* import would be caught, and it would be. No change.

Verified: `deno task typecheck`, `lint`, `fmt:check`, `lint:module-boundaries`, `lint:chat-ratchets`, `lint:chat-composability` all exit 0; `src/chat/` is 60 passed (184 steps), 0 failed.

PR #3443, opened 2026-08-06
test(chat): cover the tool-replay reconciliation algorithm directly

kojiwakayama:
> **Stacked on #3439** (`refactor/chat-tool-replay-extraction`) — the module under test does not exist on `main`. Merge #3439 first.

Writes the first real test suite for `findProviderVisibleToolReplayMatches`. **Tests only — no source changes.**

## Why

#3439 extracted a ~100-line algorithm that decides which tool-call/result occurrences in replay history are authoritative for provider conversion — matching by **part object identity** via seven `WeakSet`/`WeakMap` collections. It shipped with 2 smoke tests (`isTransientToolState` classification, and empty history). The algorithm itself was covered only indirectly, through provider conversion in `conversation.test.ts` / `message-prep.test.ts`.

That was the point of extracting it. This is the follow-through.

## What was added

10 cases (2 existing smoke tests kept; 12 total): pair matching, supersession via both paths, transient preservation, name-mismatch rejection, unmatched pending call, batch start, `toolCallsById` supersession after pending-queue eviction, the user-message batch boundary, and object identity.

## The case that matters most

**Identity discrimination.** It builds a matched call/result pair, then checks structurally identical but distinct clones against `matchedToolCallParts`, `matchedToolResultParts`, and `matchedToolResultNames` — asserting the real object reads true/has-a-name and the clone reads false/undefined.

Its exclusive value: it is the **only** case that fails if the returned `WeakSet`/`WeakMap` collections are swapped for structural-equality lookups. No consumer's compile would catch that — callers only use `.has()` and `.get()`. Without this case, a change that "optimized" by comparing parts structurally, or by cloning them internally, would pass every other test while silently breaking replay matching.

## Test quality was verified by mutation, not by reading

The suite was reviewed by building 9 mutants of the implementation (as copies — tracked source was never modified) and mapping which cases catch which:

| Mutation | Caught by |
|---|---|
| Clone parts internally before adding to the sets | cases 1, 2, 3, 5, 8 |
| Return structural-equality collections instead of WeakSet/WeakMap | **case 2 only** |
| Drop supersession propagation to the earlier matched result | **case 3 only** |
| `isCompatibleToolResultName` → always true | **case 6 only** |
| Preserve every transient call unconditionally | cases 5, 6, 9 |
| Batch-start fires on any pending call | **case 8 only** |
| Self-contained path also adds to `matchedToolCallParts` | cases 4, 9 |

That exercise found a real defect in the first draft: a case titled as fencing the user-message boundary survived both boundary mutants. Its assertions were correct, but they pinned something else — its fixture shared a `toolCallId`, so `removePendingCallsWithId` evicted the stale entry regardless and the boundary was never exercised. It has been retitled to what it actually pins, and a genuine boundary case added alongside it (verified to flip under the mutant).

## A documented limitation

The new boundary case fences the "user text stops counting as provider-visible content" regression. It does **not** fence deleting the earlier-message flush inside the visible-content branch: for a call-less user message, `pendingCountBeforeSameMessageVisibleContent` still captures the pre-message length, so the end-of-loop `splice` fallback evicts exactly the same entries — the two paths are provably equivalent in that shape. Discriminating it would need a boundary message with calls interleaved with visible content. This is stated in the case's comment rather than left implied.

## Behaviour pinned, not fixed

A call already evicted from `pendingCalls` as stale still lands in `supersededToolCallParts` when a later same-`toolCallId` occurrence arrives — `toolCallsById` accumulates every occurrence and is never pruned, while `pendingCalls` is pruned by three separate paths.

Reviewed as harmless, and arguably the safer behaviour: the flag only fires when another occurrence with the same `toolCallId` appears later, and emitting two tool-call blocks under one id is exactly what this module exists to prevent. Consumers checked: `conversation.ts:816` (`pushToolCall`, reached only after the transient-skip guard), `:889`/`:928`/`:949` (provider tool messages), and `message-prep.ts:883-885` (`stripPendingToolParts`). Pinned as characterization; no source change.

## Evidence

- `deno task test:unit`: branch baseline **3804 passed / 27918 steps** → **3804 / 27928**, zero regressions.
- Target file green; `deno lint`, `deno task verify:quick`, `deno task lint:chat-ratchets` all exit 0.
- Commit touches only `src/chat/tool-replay-reconciliation.test.ts`.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for reconciling assistant, user, tool, and dynamic tool messages.
  * Verified matching behavior, duplicate calls, transient data preservation, unresolved calls, batch boundaries, stale entries, and user-message boundaries.

### DIFF
--- a/src/chat/tool-replay-reconciliation.test.ts
+++ b/src/chat/tool-replay-reconciliation.test.ts
@@ -212,15 +212,20 @@ describe("tool-replay-reconciliation", () => {
       "streaming",
     );
     const mismatchedResult = rawToolResult("tool-1", { data: [] }, "github__list_prs");
+    const controlCall = rawToolCall("tool-2", "github__list_prs", { state: "open" });
+    const controlResult = rawToolResult("tool-2", { data: [] }, "github__list_prs");
 
     const matches = findProviderVisibleToolReplayMatches([
-      assistantMessage([call]),
-      toolMessage([mismatchedResult]),
+      assistantMessage([call, controlCall]),
+      toolMessage([mismatchedResult, controlResult]),
     ]);
 
     assertEquals(matches.matchedToolCallParts.has(call), false);
     assertEquals(matches.matchedToolResultParts.has(mismatchedResult), false);
     assertEquals(matches.preservedTransientToolParts.has(call), false);
+    // Positive control: a compatible pair in the same history really is
+    // matched, proving the fixture was processed rather than short-circuited.
+    assertEquals(matches.matchedToolCallParts.has(controlCall), true);
   });
 
   it("leaves a pending call with no result unmatched and unpreserved", () => {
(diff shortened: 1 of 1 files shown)


kwakayama:
Closing as superseded by #3439.

The two commits from this branch were squashed onto `refactor/chat-tool-replay-extraction` as 1eba2585a, so #3439 now contains everything here plus the extraction this test suite exercises. At the time of the squash the two trees were byte-identical (`git diff` between the heads was empty); #3439 has since moved ahead with the review nitpicks addressed in 67ffa89b7.

Both PRs target `main`, so merging both would conflict. #3439 is the one to land: it carries the full check suite, while this branch never triggered a CI/CD run.

No work is lost. `findProviderVisibleToolReplayMatches` coverage ships in #3439.

kwakayama:
Correction to my previous comment: this was already merged, not superseded, and my note misread the state.

It merged into `refactor/chat-tool-replay-extraction` (#3439) as 1eba2585a, not into `main`. So the coverage ships when #3439 lands, which is the same outcome, but nothing here needed closing and no action is required.

The rest stands: #3439 carries this work plus the review nitpicks addressed in 67ffa89b7.

PR #3451, opened 2026-08-07
fix(styles): resolve the preview stylesheet like production does

kojiwakayama:
Fixes hosted previews rendering unstyled while production renders correctly.

## The symptom

codersociety's preview stylesheet carries the stock Tailwind palette and **none of the project's theme**:

```
production CSS:  --color-background ✓  --color-panel ✓  --color-border ✓  --color-input ✓
preview CSS:     ✗                    ✗                ✗                 ✗
```

Those tokens come from an `@theme` block in the project's `globals.css`, so every `bg-background`, `bg-panel` and `border-border` utility the pages use simply fails to generate — 65 classes present in production and absent from the preview. The page renders as unstyled text with no layout.

## Cause

The two paths resolved the project stylesheet differently:

| | Resolution |
|---|---|
| **Production** (`style-callbacks.ts`) | `findStylesheetFromFiles(files, stylesheetPath)` — accepts `globals.css`, `global.css`, `styles/globals.css`, `app/globals.css` |
| **Preview** (`styles-css.handler.ts`) | one hardcoded `<projectDir>/globals.css` via `fs.readFile`, failure swallowed at `debug` |

On miss it fell through to the provider default — a stylesheet with no project theme.

## The fix

Resolve from the same file list production uses, keeping a direct read over the same four conventional locations as a fallback. That also removes the dependency on a path-shaped filesystem read, which isn't how sources arrive when they come from the control plane rather than local disk.

Raised the give-up log from `debug` to `warn`. I went looking for that line in production and found nothing — because it was filtered. A project silently losing its theme on every preview page shouldn't be diagnosable only with debug logging on.

## Two tests, because there were two candidate mechanisms

I didn't know whether the theme was lost at load or downstream in `getProjectCSS`, so each test isolates one:

| Test | Before | Meaning |
|---|---|---|
| `@theme` at the one path this route read | **passed** | nothing lost downstream — the load is the problem |
| stylesheet at `styles/globals.css` | **failed** | not found at all |

The second fails without this change.

## ⚠️ Honest caveat for review

This fixes the resolution gap, which is a real defect on its own. But **codersociety's `globals.css` is at the repository root**, so the narrow path list alone does not explain that project. Resolving from the file list is what I expect fixes it — sources there arrive from the control plane, not local disk — but I could not prove that without a hosted environment.

**So: verify codersociety's preview after this deploys.** If it's still themeless, the remaining cause is in how the API-backed fs adapter serves that read, and this PR will have narrowed it rather than closed it.

## Verification

- 27 files / 229 steps across dev handlers
- 42 files / 530 steps across styles-builder and the rendering orchestrator
- `deno lint`, `fmt`, `check` clean on the changed files
- The two `TS2353` errors under styles-builder and rendering are pre-existing — confirmed by stashing onto a clean tree

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved development CSS handling to discover stylesheets from configured and conventional project locations.
  * Project-level `@theme` declarations are now included in served CSS.
  * Added fallback handling when source files are unavailable or stylesheets cannot be found.
  * Stylesheets at `styles/globals.css` are now correctly discovered and compiled.

### DIFF
--- a/src/server/handlers/dev/styles-css.handler.test.ts
+++ b/src/server/handlers/dev/styles-css.handler.test.ts
@@ -107,6 +107,60 @@ function makeCtx(adapter: RuntimeAdapter, overrides: Partial<HandlerContext> = {
 }
 
 describe("server/handlers/dev/styles-css.handler", () => {
+  it("carries the project @theme into the served stylesheet", async () => {
+    // Mechanism check A: the stylesheet IS at the single path the dev route
+    // reads. If the theme still does not reach the output, the loss is
+    // downstream of loadStylesheet rather than in path resolution.
+    const stub = mockTailwindFetch();
+    try {
+      const adapter = createHandlerAdapter(
+        [{ path: "pages/index.tsx", content: '<div className="bg-brand" />' }],
+        null,
+      );
+      adapter.fs.files.set(
+        "/project/globals.css",
+        '@import "tailwindcss";\n@theme { --color-brand: #123456; }',
+      );
+      const ctx = makeCtx(adapter);
+      const req = new Request("http://localhost/_vf_styles/styles.css");
+
+      const result = await new StylesCSSHandler().handle(req, ctx);
+      const body = await result.response!.text();
+
+      assertEquals(result.response!.status, 200);
+      assertEquals(body.includes("--color-brand"), true);
+    } finally {
+      stub.restore();
+    }
+  });
+
+  it("finds a stylesheet at styles/globals.css like the production resolver", async () => {
+    // Mechanism check B: production's findStylesheetFromFiles searches
+    // globals.css, global.css, styles/globals.css and app/globals.css. The dev
+    // route reads exactly one hardcoded path.
+    const stub = mockTailwindFetch();
+    try {
+      const adapter = createHandlerAdapter(
+        [{ path: "pages/index.tsx", content: '<div className="bg-brand" />' }],
+        null,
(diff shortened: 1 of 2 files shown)


kojiwakayama:
**Correction on attribution — please read before merging.**

codersociety's preview is now rendering correctly, and **this PR is not deployed**: it is still open, and `veryfront-server` is on the same image (`20260806152814-b7d691ae9ecf`) it was on when I measured the breakage. So something other than this change fixed it.

The likeliest explanation is that the preview stylesheet I sampled was **stale**, and an edit to `globals.css` triggered regeneration. That points at a caching/invalidation problem in the preview CSS path rather than the resolution gap this PR fixes.

**What this PR still does, unchanged:** it fixes a defect I proved with a failing test — a stylesheet at `styles/globals.css` (or `global.css`, or `app/globals.css`) is not found by this route, while the production resolver accepts all four. That is real and worth fixing on its own merits, and the `debug` → `warn` change makes the failure diagnosable instead of silent.

**What it should no longer be assumed to do:** fix codersociety. My original caveat said the narrow path list did not explain that project since its `globals.css` is at the repository root, and I expected the file-list resolution to cover it. That expectation is now unsupported — the symptom cleared without this code.

Anyone reviewing should treat this as "correct, narrower than advertised." The open question is whether preview CSS can be served stale after the theme changes, which I have not investigated.

PR #3457, opened 2026-08-07
observability(server): say when a request loses its project config

kojiwakayama:
## What

`resolveAdapter` returns a `ConfigResolutionOutcome` naming which branch produced `config`, and the security fallback logs a warn carrying it. `"No hosted config for this release; using defaults"` goes from debug to warn.

**This is instrumentation, not a fix.** It exists to make an unsolved bug diagnosable.

## The bug it instruments

`codersociety.preview.veryfront.com` serves the platform-default CSP instead of the project's on roughly a third of renders — same URL, same authenticated request. The two responses are byte-identical HTML; only the headers differ, because the project's two `veryfront.config.ts` versions differ only by a `security.csp` block. Nothing in production logs says so.

Reproducible in one loop:

```
plain #1: … data: https://cdn.codersociety.com
plain #2: … data: https://cdn.codersociety.com
plain #3: … data:                                ← project CSP gone
plain #4: … data:
```

with a clean bimodal split on `runtime.resolve_adapter` — 57–61 ms when the config is lost, 469–1091 ms when it is not, no overlap across 8 samples.

## Why it was undiagnosable

`resolveProjectRuntimeContext` substitutes the process-wide `SecurityConfigLoader` for an absent project config. That is correct for the config-less control-plane endpoints it was written for, and silent everywhere else: the response is a 200 whose policy is the platform floor rather than the tenant's, which no downstream consumer can distinguish from a project that configured nothing.

Four unrelated branches reach it — an inherited caller config, a deliberate defer, a hosted 404, and a project that publishes no config — and the returned `config: undefined` cannot tell them apart. Every one of them logged at debug.

## What this adds

- `ConfigResolutionOutcome`: `inherited` | `local` | `deferred` | `hosted` | `hosted-absent`
- A warn at the point of substitution carrying the outcome plus the request's source identity
- `deferred` is excluded — a config-less security context is the intended shape for signed control-plane requests, not a degradation

## Why not the fix itself

Six hypotheses for the underlying race have been falsified: content staleness (both sources now carry the CSP and it still flips), pod boot cache (fleet is on a 70-minute-old ReplicaSet, post-dating the config change), API inconsistency (12/12 consistent), `studio_embed`, release-pin, and CSS recompilation. Each was killed by measurement, not argument.

`configOutcome` discriminates between every remaining hypothesis in a single production sample. That is the cheapest path to the actual fix.

## Verification

**119 files / 1525 steps** across `src/server/runtime-handler` and `src/security`, plus full unit suite green: **3811 passed, 0 failed**. `deno check`, `lint`, `fmt` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing hosted configuration by safely falling back to defaults.
  * Preserved error handling for hosted configuration failures that are not missing-resource responses.
  * Added clearer diagnostics for configuration fallback and inherited security settings.
  * Improved tracking of configuration resolution states, including deferred, inherited, local, and hosted configurations.
  * Reduced unnecessary security warnings for deferred control-plane requests.
  * Added contextual warnings when proxied requests rely on process-wide security settings.

### DIFF
--- a/src/server/runtime-handler/adapter-factory.test.ts
+++ b/src/server/runtime-handler/adapter-factory.test.ts
@@ -1069,6 +1069,9 @@ describe("adapter-factory", () => {
 
     // Defaults, not the caller's config.
     assertEquals(result.config, undefined);
+    // Downstream substitutes the process-wide security config for an absent
+    // project config, so the reason it is absent has to survive the return.
+    assertEquals(result.configOutcome, "hosted-absent");
   });
 
   it("uses defaults when the 404 arrives wrapped rather than at the top level", async () => {
@@ -1108,6 +1111,7 @@ describe("adapter-factory", () => {
     });
 
     assertEquals(result.config, undefined);
+    assertEquals(result.configOutcome, "hosted-absent");
   });
 
   it("still fails when a 404 comes from something other than the config read", async () => {
--- a/src/server/runtime-handler/adapter-factory.ts
+++ b/src/server/runtime-handler/adapter-factory.ts
@@ -32,6 +32,31 @@ const baseLogger = getBaseLogger("SERVER");
 
 const logger = baseLogger.component("adapter-factory");
 
+/**
+ * Which path produced `config`, so a caller that degrades on a missing config
+ * can say why it is missing.
+ *
+ * `config` being `undefined` is reached from four unrelated places -- an
+ * inherited caller config, a deliberate defer, a published project that has no
+ * config file, and a hosted 404 -- and the result alone cannot tell them
+ * apart. Downstream, `resolveProjectRuntimeContext` silently substitutes the
+ * process-wide security config for an absent project config, which serves a
+ * correct-looking 200 carrying the platform-default CSP instead of the
+ * project's. That degradation was undiagnosable from production logs because
+ * every branch that reaches it logs at debug.
+ */
+export type ConfigResolutionOutcome =
+  /** No project-specific load ran; whatever the caller passed through stands. */
+  | "inherited"
+  /** Loaded from a local project directory. */
+  | "local"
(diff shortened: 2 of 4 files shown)


kojiwakayama:
Addressed the first finding in `21d71d9dc` — the deferred case now captures log records and asserts no fallback warning is emitted. Good catch: that exclusion is the part that actually matters, since every signed control-plane request takes that path.

On the second finding (redact project/release/branch/environment/pathname from both warnings) — I'm not implementing this, and want to give the reasoning rather than silently skip it.

**The identifiers are the payload, not incidental context.** This warning exists to answer one question: *which project just lost its config?* A production preview is serving the platform-default CSP instead of the tenant's on roughly a third of renders, and six hypotheses for why have already been falsified. `configOutcome` alone tells you a config went missing somewhere in a multi-tenant fleet; without `projectSlug` it cannot be correlated to the failing request, which is the entire purpose of the line.

**It matches what the surrounding code already does.** `environment-resolution.ts` logs `projectSlug`, `projectId`, `environmentName`, `host`, `proxyEnv` on its no-active-release warning. `adapter-factory.ts` logs `projectSlug`, `projectId`, `releaseId`, `proxyEnv` on config-load failure. Redacting only these two call sites would be inconsistent without making the fleet any safer.

**The fields aren't customer data.** A project slug is a public DNS label — `codersociety.preview.veryfront.com`. Release and environment IDs are internal identifiers. `pathname` is the path of a publicly-served page. These are server logs going to internal Loki, not user-facing output, and none of it is PII or credential material.

Happy to revisit if there's a specific field you'd single out — `pathname` is the only one I'd consider arguable, since a path *could* carry a query-ish segment on some routes. But dropping the project identifier would make the instrumentation useless for the bug it was written to diagnose.

PR #3473, opened 2026-08-08
refactor(agent): drop opaque orchestration prose from load_skill results

kwakayama:
Fixes veryfront/veryfront-issue-inbox#5.

`load_skill` returned a `nextStep` string carrying internal orchestration policy alongside the skill's actual instructions:

> IMPORTANT: load_skill only loads instructions. It does not perform the task or finish the turn. Continue the same turn after calling it. Keep the root assistant visibly owning the work. For multi-step or isolated work, call invoke_agent; otherwise keep working directly with the available tools. Delegate only when isolation, parallelism, or a different tool/model budget materially helps. Pass through any returned model, thinking, or maxSteps overrides to invoke_agent when delegating.

## The finding that makes this safe

**Every clause of that string is already in the `<available_skills>` system prompt, word for word.** Rendered from current `main`:

```
You have access to these skills. Use load_skill to load full instructions when needed.
load_skill only loads instructions plus metadata. Continue the same turn after calling it.
Keep the root assistant visibly owning the work. When delegating, use the available legacy
`invoke_agent` tool. Delegate only when isolation, parallelism, or a different tool/model
budget materially helps. Pass through any returned model, thinking, or maxSteps overrides
to invoke_agent when delegating. ...
```

| `nextStep` clause | In the system prompt? |
|---|---|
| "Continue the same turn after calling it." | verbatim |
| "Keep the root assistant visibly owning the work." | verbatim |
| delegation advice, threshold, override forwarding | verbatim |
| "It does not perform the task or finish the turn." | covered by "only loads instructions plus metadata" |

So the result was restating trusted system policy as untrusted tool output — exactly the duplication the issue objects to. Removing it deletes a copy, not a signal.

## What the result looks like now

```json
{
  "skillId": "deploy",
  "instructions": "---\nname: deploy\n...\n---\nDo the deploy.",
  "model": "sonnet",
  "references": ["references/guide.md"]
}
```

Every field is structured, and every field has a documented consumer:

| Field | Consumer |
|---|---|
| `skillId` | caller, to correlate the result |
| `instructions` | model — the only instruction content in the result |
| `references` | model — loadable via `load_skill`'s `file` parameter |
| `model` / `thinking` / `maxSteps` | caller — overrides to forward when delegating |

## Removed

**Prose fields:**
- `nextStep` — the orchestration policy above
- `overrideNote` — *"Pass through any returned model, thinking, or maxSteps overrides…"*, verbatim in the system prompt
- `referenceNote` — *"After this skill is loaded, use load_skill with the `file` parameter…"*, which duplicates the `load_skill` tool description

**Dead since veryfront/veryfront-code#3464**, still on the type and still being copied by `copyLoadedSkillResponse`:
- `allowedTools`, `note`, `delegationTools`, `unavailableCurrentRunTools`, `delegationNote`
- `assertRuntimeResponseMetadata`, which validated `delegationTools` that nothing emitted

**Configuration surface that existed only to override the deleted prose:**
- `RuntimeLoadedSkillResponseMessages`, `RuntimeLoadSkillToolMessages`, `RUNTIME_LOAD_SKILL_CONTINUATION_NOTE`, and the `nextStep` / `messages` tool options
- Three public re-exports from `src/agent/index.ts`

`availableToolNames` is also dropped from `buildStrictRuntimeLoadedSkillResponse` — it was read and never used once `overrideNote` went. Its bound is still enforced, independently, by `assertRuntimeBoundaryCollections`.

## Acceptance criteria

- [x] The public `load_skill` schema documents every returned field and its consumer — each field on `RuntimeLoadedSkillResponse` now carries a doc comment naming who reads it
- [x] Internal root-agent/delegation policy is not emitted as opaque result text
- [x] Studio does not expose implementation-policy text as a normal tool result
- [x] Regression tests prove the post-load turn does not rely on an ad hoc prose `nextStep`
- [ ] **The model still continues after loading a skill** — see below

## The one criterion I cannot prove in tests

The continuation instruction now appears **once** (system prompt) instead of twice (system prompt + every `load_skill` result). Unit tests can prove the signal still exists; they cannot prove a model still obeys it.

This is worth watching, because veryfront/veryfront-issue-inbox#392's residual is precisely "a child stopped after `load_skill`". I measured that baseline while triaging it: **2 runs in 30 days** across `agent_run_event` matched the stop-after-load_skill signature (the raw figure was 561, but 558 of those were one retired `control-agent` burst that ended 2026-07-15).

So there is a concrete before-number to compare against. If the rate climbs above ~2/month after this ships, this change is the first suspect and reverting the `nextStep` removal is the obvious probe.

## Testing

- `src/skill/` + `src/agent/` — **1186 passed, 2050 steps, 0 failed**
- `deno lint src/` — clean
- `deno task docs:api-reference:check` — current
- `deno fmt --check` — clean

Tests asserting the removed fields were deleted rather than weakened (6 in `load-skill-tool.test.ts`, 1 in `skill-metadata.test.ts`), all of which existed only to pin prose or its configurability. Where a test's real subject survived — accessor rejection, input bounds, iterator snapshotting — the removed-field half was dropped and the rest kept.

**On the full-tree run:** `src/ cli/ tests/` shows 26 failures against a 8-failure baseline, but the difference is entirely e2e/dev-server suites (CSS, MDX Pages, Static Files, Relative Import Resolution, skill-capabilities). I verified `tests/e2e/features/skill-capabilities.test.ts` fails identically with my changes stashed — `403` vs expected `200`, a local admission failure — and that file contains **zero** references to `nextStep`. These suites are non-deterministic under parallel dev servers in this checkout; CI runs them in dedicated jobs.

## Summary by CodeRabbit

* **Updates**
  * Simplified loaded-skill responses by removing legacy continuation, delegation, tool-availability, and message fields.
  * Loaded skills now provide continuation guidance directly within their instructions.
  * Preserved skill loading, authorization, caching, references, cancellation, metadata, and inventory behavior.
  * Retired legacy runtime skill response exports and related configuration options.

* **Documentation**
  * Refreshed API reference links and removed entries for retired runtime skill response types.


kwakayama:
Addressed both reviews. Pushed `3f4d221c5`.

## Codex — override forwarding for authored skill blocks: **valid, fixed**

This was a real gap and it undermined this PR's central claim, so thank you for catching it.

My argument was "every clause of `nextStep` is already in the `<available_skills>` system prompt." I verified that against the **generated** block. But `call-context.ts:168` only appends the generated block when the agent's own instructions do not already contain one:

```ts
if (input.skills?.length && !hasBlock(input.instructions, AVAILABLE_SKILLS_BLOCK_NAME)) {
```

For an agent with an authored block, the system prompt carries whatever the author wrote. My premise does not hold there.

Checked what the always-present tool description carries:

| Clause | In `RUNTIME_LOAD_SKILL_DESCRIPTION`? |
|---|---|
| `LOAD_SKILL_CONTINUE_SAME_TURN` | yes |
| `LOAD_SKILL_ROOT_OWNERSHIP` | yes |
| `LOAD_SKILL_DELEGATION_THRESHOLD` | yes |
| `LOAD_SKILL_OVERRIDE_FORWARDING` | **no** |

So continuation, ownership, and the delegation threshold survived an authored block; override forwarding was the single clause that did not. Exactly as you described.

**Fix:** moved `LOAD_SKILL_OVERRIDE_FORWARDING` into the tool description. That keeps it in the trusted tool contract rather than the result payload, which is what #5 asks for, and it is sent regardless of whether the skills block is generated or authored.

Added a regression test pinning that the description carries all four clauses, so this cannot silently reopen. Verified non-vacuous — removing the clause fails it (`0 passed | 1 failed`).

## CodeRabbit — convert touched tests to `describe()`/`it()`: **declining, with reason**

The guideline is real, but applying it to only the lines this PR touches makes the file worse:

- `load-skill-tool.test.ts` is **74 `Deno.test` blocks, 0 `describe`**. Converting the handful I touched produces a file with two competing registration styles.
- It is not lint-enforced — `deno task lint` and `lint:test-typecheck` both pass, and `scripts/lint`'s own tests use `Deno.test`.
- Converting all 74 is a mechanical diff several times the size of this change, and would bury a behavioural PR under test-harness churn.

Worth doing as its own PR; not worth doing halfway here.

## Verification after the fix

- `src/skill/` + `src/agent/` — 1187 passed, 0 failed
- `deno task lint`, `lint:test-typecheck`, `docs:api-reference:check` — all OK

PR #3480, opened 2026-08-08
chore(agent): drop the dead LOAD_SKILL_USE_ALLOWED_TOOLS constant

kwakayama:
Leftover from veryfront/veryfront-issue-inbox#406.

`LOAD_SKILL_USE_ALLOWED_TOOLS` read:

> For multi-step or isolated work, call invoke_agent; otherwise keep working directly with the available tools.

That sentence was written when a skill's `allowed-tools` narrowed a run. veryfront/veryfront-code#3464 removed the enforcement, and I reworded the string ("allowed tools" → "available tools") instead of deleting it. The rewording was the wrong call: with nothing narrowing the tool set, the sentence had no subject left.

## Why this one and not its neighbours

Several constants in `delegation-policy.ts` look unreferenced if you grep for their names, because the code that ships them composes them inside the same file. This one is different:

| | `LOAD_SKILL_USE_ALLOWED_TOOLS` | e.g. `LOAD_SKILL_CONTINUATION_REMINDER` |
|---|---|---|
| production callers | 0 | `child-fork-execution-runner.ts:213-217` |
| test references | 0 | 2 |
| composed by another constant in-file | 0 | 4 |

`LOAD_SKILL_CONTINUATION_REMINDER`, `SLASH_COMMAND_ARTIFACT_REMINDER`, `FIRST_TURN_STARTER_INTENT_ROOT_OWNERSHIP_REMINDER`, `NO_DELEGATION_NARRATION_UNLESS_ASKED`, `SYNTHESIZE_DELEGATED_FINDINGS_IN_ROOT_VOICE`, and `ROOT_OWNED_CHILD_RESULT_INSTRUCTION` all reach production through `addLoadSkillContinuationReminder`, `addSlashCommandArtifactReminder`, `buildInvokeAgentFollowupInstruction`, or `withRootOwnedChildResultHint`. They stay.

Only `LOAD_SKILL_USE_ALLOWED_TOOLS` has no path to a model at all.

## Scope

- the constant in `src/agent/conversation/delegation-policy.ts`
- its re-export from the `veryfront/agent` barrel (`src/agent/index.ts`)
- regenerated `docs/api-reference/veryfront/agent.md`

This is a public API removal, consistent with #3464 and #3475, which removed exports in the same area.

## Testing

- `src/skill/`, `src/agent/`, `src/internal-agents/` — 1195 passed, 2178 steps, 0 failed
- `deno lint`, `lint:test-typecheck`, `docs:api-reference:check` — all OK

## Summary by CodeRabbit

* **Documentation**
  * Updated API reference links to reflect the current implementation locations.
  * Removed the retired delegation-policy constant from the documented API.

* **Refactor**
  * Removed the retired delegation-policy constant and its associated instruction text from the public API.


kwakayama:
Thanks both.

**CodeRabbit:** no actionable comments.

## Codex — "Preserve the published load-skill export" (P2): declining, with reasons

The finding is factually correct: `LOAD_SKILL_USE_ALLOWED_TOOLS` was reachable from `veryfront/agent` and appeared in the generated API reference, so a consumer importing it breaks on upgrade. I'm taking the break deliberately.

**1. Keeping it is worse than breaking it.** The constant's text is:

> For multi-step or isolated work, call invoke_agent; otherwise keep working directly with the available tools.

That is prompt copy describing a run whose tool set a skill had narrowed. veryfront/veryfront-code#3464 removed that behaviour. A consumer importing this string today gets policy for a feature that no longer exists — and because it is prompt copy, the likely use is injecting it into agent instructions, which propagates stale guidance into their prompts silently. A compile error tells them to stop; a preserved constant does not.

Its own history makes the point: when #3464 removed the enforcement I reworded this string ("allowed tools" → "available tools") instead of deleting it. Rewording bought a sentence that still had no subject. This PR finishes that job rather than repeating it.

**2. Removing exports from this barrel is established practice here**, including changes that are not mine:

| Commit | PR | Exports removed from `src/agent/index.ts` |
|---|---|---|
| `9e57b05cf` | #3473 | `RuntimeLoadedSkillResponseMessages`, `RUNTIME_LOAD_SKILL_CONTINUATION_NOTE`, `RuntimeLoadSkillToolMessages` |
| `8da2596be` | #3464 | `LOAD_SKILL_TOOL_INTERSECTION` |
| `0a9f224cd` | #3309 | `ModelCallContext`, `ModelCallRecorder` |
| this PR | — | `LOAD_SKILL_USE_ALLOWED_TOOLS` |

Two of those were also in the generated API reference before removal. This PR removes one export — strictly less than #3473, which merged.

**3. The removal is visible, not silent.** `docs:api-reference:check` regenerates the published surface, so the constant's disappearance is recorded in `docs/api-reference/veryfront/agent.md` in this diff.

I'll add it back behind a `@deprecated` re-export if a maintainer would rather not take the break — but I don't think a string describing deleted behaviour is worth keeping compilable.

PR #3483, opened 2026-08-08
feat(agent): stamp elapsedMs on durable run events at emission

kojiwakayama:
Makes run traces measurable: every durable run event carries a run-relative `elapsedMs` taken when it was emitted.

> **Retargeted.** The first revision of this PR stamped the AG-UI browser encoder. That encoder feeds live SSE to a connected client, and a scheduled run has no client attached — so it would never have reached the database. This revision stamps `ConversationRunEventEncoder`, which is the encoder whose output is persisted.

## The problem

`agent_run_event.created_at` is `DEFAULT now()` — the row's database insert time. No event carried an emission timestamp, so nothing downstream could say when an event actually happened.

Not theoretical. A post-mortem of production run `2f7ae3d4` built a phase waterfall on `created_at` and concluded the run spent **52% of its wall clock stalled mid-generation**. It hadn't — the data described when rows were written. Three signatures in the same export are impossible under the runtime's own state machine, and are what exposed it:

- `streaming_input → pending_input → streaming_input` inside **60ms**, when `pending` requires 5s of idle
- `TOOL_CALL_ARGS` rows beginning **200s after** the `streaming_input` that marks the first delta — `tool-input-delta` maps straight to `ToolCallArgs`, so those are the same event
- a metronomic **~30ms** delta cadence at 2.5ms stdev

## Why this encoder

It is the one whose output is persisted:

```
ChatStreamEvent → ConversationRunEventEncoder.encode()
                → flat run-event records
                → normalizeConversationRunEvents
                → flushConversationRunEventBatches → POST → agent_run_event
```

Stamping happens on the way out of `encode()` rather than in each `case` arm, so every emitted record is treated alike — including ones the encoder synthesises itself, such as the terminal result for a provider-executed call the provider never resolved.

## Why run-relative, and why monotonic

Elapsed is measured from encoder creation. **One encoder spans a whole run** — it carries `stepCount` and the active message across every step, which is why run 2f7ae3d4's five steps share one. So `elapsedMs` needs no per-attempt anchor, and `agent_run.started_at + elapsed_ms` reconstructs wall clock for any event.

Monotonic rather than wall clock: `Date.now()` can step backwards under NTP, which across thousands of events yields negative durations. Injected rather than captured, so timing stays deterministic in tests the way `policy.clock` already does for the stream-lifecycle runner.

## What makes it live

The chunk mirror owns one encoder per run, so installing the clock at `createConversationRunChunkMirror` is what makes production events carry the stamp. Callers injecting their own encoder choose their own clock, or none — the three mirror tests that assert exact durable events pin an unclocked encoder, since their subject is the mirror rather than timing.

## Tests

Red-first:

- elapsed is stamped and advances with an injected clock
- **no `elapsedMs` key appears at all** without a clock — the guard that keeps this invisible to consumers that don't want it
- **it survives normalization**: an oversized delta splits into parts that each keep the elapsed of the event they came from. A stamp that doesn't survive normalization never reaches the API, which is exactly the gap that made the first revision inert

`src/agent` + `src/internal-agents` + `src/chat`: **1,214 passed, 0 failed.** Typecheck, `deno fmt --check`, `deno lint` clean. `docs/api-reference` regenerated.

## Follow-up

veryfront-api: `ALTER TABLE agent_run_event ADD COLUMN elapsed_ms integer`, lift it with `payload->>'elapsedMs'` (it lands as a top-level field, since these records are flat), and expose it alongside `sequence` — which exists on the table but is absent from both the API response and the CSV export.

Related: veryfront-studio#6345 renames that export's `timestamp` column to `created_at` so it stops implying it is event time.

## Summary by CodeRabbit

- **New Features**
  - Conversation run events can include elapsed time from the start of encoding.
  - Added options to control timestamp behavior, including custom clocks.
  - Elapsed-time metadata is applied consistently to split and synthesized events.

- **Documentation**
  - Updated API reference links.
  - Added documentation for encoder options and refined the encoder API reference.


kojiwakayama:
Addressed. All five review threads were filed against commit `3dc173127`, which stamped the AG-UI browser encoder. That revision has been replaced: the PR now stamps `ConversationRunEventEncoder`, and none of the files those comments cite are in the diff any more.

Rather than close them as outdated, I checked whether each concern still applies to the new code.

**Fixed — the concern carried over.**

*"Export the new encoder state options type"* (Codex, P2). `CreateAgUiBrowserEncoderStateOptions` no longer exists, but the identical gap applied to its replacement: `ConversationRunEventEncoder` is re-exported from `veryfront/agent` while `ConversationRunEventEncoderOptions` was not. Fixed in `3b8aa1a9a`; it now appears in `docs/api-reference/veryfront/agent.md`.

**Checked — does not apply.**

*"Remove the false default-clock claim"* (CodeRabbit) and *"Correct the advertised default clock"* (Codex, P2). Both were right about the old revision: the JSDoc promised a `performance.now` default the factories did not implement. Those factories are no longer touched. The new option documents itself as opt-in, and `createConversationRunChunkMirror` installs the clock for real, so the stated behaviour and the implementation now agree.

*"Replace em dashes in the public API comment"* (Codex, P1). Correct for the old JSDoc. The new comments in `run-events.ts` and `run-chunk-mirror.ts` use ASCII `--`; `grep -nE "—|–"` over both files returns nothing.

*"Preserve elapsedMs through the AG-UI SSE schemas"* (Codex, P2). The sharpest of the five, and worth spelling out because it prompted a check that closes a gap I had flagged as unverified. It was correct that `formatAgUiEvent` reparses payloads through schemas that would strip an undeclared field. That path is no longer involved, but the same question applies to the durable path, so I traced every hop the stamp has to survive:

| Hop | Result |
| --- | --- |
| `normalizeConversationRunEvents` | survives, covered by a test asserting an oversized delta splits into parts that each keep their source elapsed |
| API append validation (`PublicSupportedAGUIEventSchema`) | survives, the event schemas are `.passthrough()` |
| API read path (`toPublicAgentRunEventPayload`) | survives, `default: return payload` |

I had been flagging "verified in unit tests but not observed reaching Postgres" as an open risk on this PR. Those three checks close the schema half of it. What remains is only observing a live run, which the first export after merge will show.

kojiwakayama:
Both addressed in `bdd9968ad`.

**Add a test for the default mirror encoder** (Major). Accepted as written. Changing that default is the thing that makes production events carry elapsed, and the three existing mirror tests pin an unclocked encoder to keep their exact-event assertions deterministic, so nothing exercised it. Added a test that omits `encoder` and asserts a finite, nonnegative `elapsedMs`. I checked it has teeth rather than trusting it: removing the default clock makes it fail with `the default encoder must stamp elapsedMs`, and restoring it passes.

**Remove private persistence details from comments** (Minor). Acted on, though not for the cited reason. The guideline quoted covers sensitive values, and a table name is not one. The real problem is narrower and worth stating: this package is published, and the JSDoc for a public option explained itself by naming the API service's table and columns, so a consumer reading it was pointed at storage they cannot see and do not own. That is poor API documentation independent of any rule.

Reworded to describe behaviour the caller can act on:

> A persisted event otherwise carries only the time it was stored, which tracks the writer rather than the run, so durations built from it do not describe what the agent did.

Same reasoning, no foreign schema names. `grep -nE "agent_run_event|agent_run\.started_at|created_at|elapsed_ms"` over both files now returns nothing.

`src/agent` + `src/internal-agents` + `src/chat`: 1,214 passed, 0 failed. Typecheck, fmt, lint clean, API reference regenerated.

PR #3485, opened 2026-08-08
fix(skill): stop hardcoding invoke_agent in load-skill override-forwarding prose

kwakayama:
Fixes veryfront/veryfront-issue-inbox#411

## Summary

`LOAD_SKILL_OVERRIDE_FORWARDING` in `src/skill/load-skill-policy.ts` hard-coded the tool name `invoke_agent`, but some runs expose only scoped delegate tools named `agent_<id>` and no `invoke_agent` at all. On those runs the policy prose named a tool that isn't present.

This is pre-existing (found by Codex review on PR #3475, not introduced by it) and low severity — there's no measured impact from it. It's filed and fixed for consistency with the standard already applied to the surrounding prompt surface, not because of an observed failure.

## Investigation

Swept every `invoke_agent` occurrence in `src/`:
```
grep -rn "invoke_agent" --include='*.ts' src/ | grep -v '\.test\.'
```
Every hit besides `LOAD_SKILL_OVERRIDE_FORWARDING` is either:
- a legitimate code reference (tool ids/constants like `INVOKE_AGENT_TOOL_ID`, registry lookups, tracing span names, `Set`/array membership checks), or
- a message/comment that only fires *while the `invoke_agent` tool is actually executing* (e.g. depth-limit errors, cancellation-retry messages in `child-invoke-tool.ts` / `child-tool-input.ts` / `default-invoke-agent-tool.ts`), so the name is accurate in that context — the tool did exist for that call.

Only `LOAD_SKILL_OVERRIDE_FORWARDING` is prose describing a delegation tool ahead of knowing whether it exists in the current run, so it's the only change in scope.

## Change

**Before:**
> "Pass through any returned model, thinking, or maxSteps overrides to invoke_agent when delegating."

**After:**
> "Pass through any returned model, thinking, or maxSteps overrides to the available delegation tool when delegating."

This wording is correct regardless of which delegation tool (`invoke_agent` or a scoped `agent_<id>`) exists in a given run, since tool descriptions are static strings and can't be made dynamic to the run's tool inventory.

## Design constraint preserved

`LOAD_SKILL_OVERRIDE_FORWARDING` remains appended only to the hosted tool's description (`RUNTIME_LOAD_SKILL_DESCRIPTION` in `src/agent/runtime/load-skill-tool.ts`), not to the factory-built `createLoadSkillTool` description in `src/skill/tools.ts` — only the hosted loader returns override fields (`model`/`thinking`/`maxSteps`). The existing pinning test for that asymmetry (`src/skill/tools.test.ts`, "promises override forwarding only from the loader that returns overrides") is untouched and still passes.

## Files changed

- `src/skill/load-skill-policy.ts` — reworded `LOAD_SKILL_OVERRIDE_FORWARDING`
- `src/skill/tools.test.ts` — added a regression test pinning the new wording and asserting it no longer hardcodes `invoke_agent`

## Test plan

- [x] New test added: `does not hardcode a specific delegation tool name` in `src/skill/tools.test.ts`. Verified it **fails** against the old wording (`git stash` of the policy file only) and **passes** after the fix.
- [x] `DENO_TESTING=1 VF_DISABLE_LRU_INTERVAL=1 NODE_ENV=production LOG_FORMAT=text deno test --preload=src/schemas/_test-setup.ts --parallel --allow-all --no-check src/skill/ src/agent/` → `1183 passed (2057 steps) | 0 failed`
- [x] `deno task typecheck` → exit 0
- [x] `deno lint` → exit 0, "Checked 4882 files"
- [x] `deno task lint:test-typecheck` → exit 0, "Test typecheck baseline holds: 51 grandfathered files, 0 new."
- [x] `deno task docs:api-reference:check` → exit 0, "docs/api-reference is current (43 files)" (no regeneration needed)
- [x] `deno fmt src/` → no changes
- [x] Pre-push hook full suite → `3769 passed (28028 steps) | 0 failed | 1 ignored (5 steps)`

Claude-Session: https://claude.ai/code/session_01Xo93b6StAu691YV9g8Fm53

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved skill-loading instructions to forward model, thinking, and step-limit settings to the available delegation tool.
  * Prevented delegation guidance from relying on a specific tool name.

* **Tests**
  * Added coverage to verify delegation instructions remain tool-agnostic.


kwakayama:
Codex is right, and the finding is sharper than the issue I filed. Pushed `a78f34f1e`.

## The original framing was wrong

veryfront/veryfront-issue-inbox#411 treated this as a **naming** problem: prose hardcodes `invoke_agent`, some runs do not have it, so generalise the name. This PR did that — `"…to the available delegation tool…"`.

That made it worse. Verified on this branch:

| Claim | Evidence |
|---|---|
| Scoped delegates cannot carry overrides | `AgentToolInput` is `v.object({ input: v.string() })` — `src/agent/schemas/tool.schema.ts:6-7`. No `model`, `thinking`, or `maxSteps`. |
| Overrides are applied to `invoke_agent` only | `applySkillDelegationOverridesToToolInput` returns its input unchanged unless `toolName === INVOKE_AGENT_TOOL_ID` — `src/agent/runtime/skill-delegation-overrides.ts:72` |

So on a scoped-delegation run, `"pass through overrides to the available delegation tool"` instructs the model to do something the tool physically cannot represent. The original wording at least named the one tool where forwarding works; its only flaw was naming a tool that might be absent.

**Override forwarding is `invoke_agent`-only by design, not by accident.** That is the fact the prose has to encode.

## The fix

```
- Pass through any returned model, thinking, or maxSteps overrides to the available delegation tool when delegating.
+ If invoke_agent is available, pass through any returned model, thinking, or maxSteps overrides when delegating to it.
```

Conditional in the prose, so it is static-safe: a no-op on runs without `invoke_agent`, accurate on runs with it. No dynamic description required.

The test now pins the condition rather than the absence of a tool name, and the module comment records *why* the condition is load-bearing so the next person does not "simplify" it back.

## Verification

- `src/skill/`, `src/agent/` — 1183 passed, 2057 steps, 0 failed
- `deno lint`, `lint:test-typecheck` — OK
- `docs:api-reference:check` — regenerated, current

PR #3497, opened 2026-08-09
fix(agent): keep one run-scoped encoder across hosted stream chunks

kojiwakayama:
## Problem

`elapsedMs` is missing from every persisted event on production runs, and step names repeat (`step-1 ×4, step-2`) instead of advancing.

`createConversationHostedStreamLifecycleAdapter` called `prepareConversationRunStreamEvents([chunk])` **without an encoder**, so a fresh `ConversationRunEventEncoder` was constructed per chunk. The encoder holds per-run state, so rebuilding it per chunk broke two things:

- `stepCount` restarted at 1 on every chunk, so persisted events never advance past `step-1`.
- The encoder's construction is the anchor `elapsedMs` is measured from, so every event was stamped against a just-created anchor — the column never received a usable value.

This is the **top-level hosted path**, which is the path production runs actually take.

#3483 (shipped in v0.1.1220) added the clock to the encoder but wired it into `createConversationRunChunkMirror`, which only serves **child** runs. That's why 1220 didn't change anything for top-level runs.

Verified against production on v0.1.1220 — `payload ? 'elapsedMs'` is `f` for every event of the latest run.

## Fix

Hoist a single clocked `ConversationRunEventEncoder` to adapter-creation time and pass it into every `prepareConversationRunStreamEvents` call.

The encoder is also exposed as an optional option so callers needing deterministic payloads can supply an unclocked one — the same escape hatch `run-chunk-mirror` already offers, and what the existing exact-payload test now uses.

## Test

New test drives the adapter with two chunks and asserts step names advance and `elapsedMs` is stamped. Confirmed red before the fix:

```
AssertionError: Values are not equal: a second chunk must continue the run's step count, got ["step-1","step-1"]
```

which matches production's `step-1 ×4, step-2` exactly.

## Verification

- Pre-push gate green: 3770 passed, 0 failed.
- `agent/conversation`: 21 passed.
- fmt, lint, `deno task typecheck` clean.
- `docs/api-reference` regenerated with the CI-pinned Deno 2.7.7 (line-pin shifts only).

## Note

`hosted-lifecycle.test.ts` already fails standalone `deno check` on 3 pre-existing `startRun` stubs that omit `waitingToolCallId` / `waitingToolName` / `streamProtocolVersion`; this PR adds a 4th instance of the same pattern. The file isn't in the `typecheck` gate, and fixing it means feeding those fields into posted payloads (risking the exact-payload assertions), so it's left for a separate cleanup.

## Summary by CodeRabbit

- **Improvements**
  - Improved hosted conversation streaming so event steps remain consistent across streamed chunks.
  - Added more accurate elapsed-time reporting for streamed conversation events.

- **Documentation**
  - Updated API reference links for conversation lifecycle configuration and hosted lifecycle options.


kojiwakayama:
Correction to this PR's description, on the record.

I claimed production's persisted events read `step-1` four times before advancing to `step-2`. **That is wrong and I did not verify it against the data.** The source trace (run `2f7ae3d4`) actually contains step-1 through step-5, each appearing exactly twice — one `STEP_STARTED` and one `STEP_FINISHED`. Step numbering has been correct in production throughout.

What this PR fixes is still a real bug: `createConversationHostedStreamLifecycleAdapter` constructed a fresh `ConversationRunEventEncoder` per chunk, and the test added here proves it produces `["step-1","step-1"]` across two chunks. That is objectively wrong regardless of which path production takes.

But it is probably **not** the fix for the missing `elapsedMs`, because production does not exhibit this path's signature. The two symptoms are independent and together they fingerprint the root:

| step names | `elapsedMs` | implies |
| --- | --- | --- |
| advance | absent | one shared encoder, unclocked |
| repeat `step-1` | absent | fresh encoder per chunk (what this PR fixed) |
| advance | present | one shared, clocked encoder — the goal |

Production shows **advance + absent**. `run-chunk-mirror` has been clocked since #3483 shipped in v0.1.1220, so it would stamp; this adapter resets `stepCount`, so it would repeat `step-1`. Neither root produces what production shows, so a third route is emitting those events.

Verified along the way, so these are not the explanation: v0.1.1220 genuinely contains the clocked chunk mirror, and the runtime pod that executed the 11:00 run (`10.192.6.132`, `veryfront-server-6f45c6d8c6-pct7c`) reports `VERYFRONT_VERSION 0.1.1220`.

Remaining suspects:

1. `prepareConversationRunExternalEvents(events)` — the `appendEvents` path takes already-formed `ConversationRunEvent[]` and never touches an encoder, so nothing appended that way is ever stamped.
2. `run-stream-mirror.ts:35` — `input.encoder ?? new ConversationRunEventEncoder()`, one encoder per run with no clock. Matches production's signature exactly, though it currently has no internal caller.

v0.1.1221 is being promoted as planned; the next scheduled tick is the measurement that settles which route is live.

PR #3503, opened 2026-08-09
release: v0.1.1222

kwakayama:
Patch release: 0.1.1221 → 0.1.1222.

## Contents since v0.1.1221

- `41b9ffc88` — fix(agent): sanitize and bound failed-run terminal error messages (#3498)
- `9317de8a8` — fix(test): exclude the cwd-exclusion pair from the Node and Bun runners (#3500)
- `f659dd5ac` — feat(agent): publish the hosted invoke_agent result contract (#3501)

`#3501` is the reason this release matters beyond the usual cadence: it exports `getHostedDurableChildInvokeResultSchema`, which veryfront-api cannot consume until it is on npm. The follow-up there asserts `InvokeAgentStoredResultSchema` stays compatible with the published shape.

## Two notes for whoever cuts the next one

**`deno task release` fails on its default flags.** Step 3 runs `deno task verify:dist`, which is not a defined task — `git log -S "verify:dist" -- deno.json` shows it never has been. Deno prints its task list and exits 1, aborting before the commit. Worked around here with `--no-build`; worth fixing the script separately rather than in a release PR.

**The script does not regenerate `hydration-runtime.generated.ts`.** That file embeds the version, so a bump leaves it stale and `ci (typecheck)`'s `generate:manifests:check` would fail. Presumably the missing `verify:dist` step used to cover this. Ran `deno task generate` and amended it in, which is why this PR touches the same three files as #3499.

**Also worth knowing:** `--yes` auto-confirms the *push* as well (`shouldPush = args.yes || response === "y"`), so `deno task release --yes` publishes unattended. Pair it with `--no-publish`.

## Tagging

Not tagged yet. `v0.1.1219/1220/1221` all point at the squash-merge commits on main, so the tag belongs on the merged commit — tagging the pre-merge branch commit would point at something that never lands. Will tag after merge, which is what triggers the npm publish.

`deno task verify:quick` exit 0.

## Summary by CodeRabbit

* **Chores**
  * Updated the application version to **0.1.1222**.
  * Synchronized the displayed version information across the project.


kojiwakayama:
Updated this branch by merging current `main` into it — no force-push, the original release commit is intact underneath.

**Why:** this PR was cut from `f659dd5ac`, 16 minutes before veryfront-code#3502 merged. As it stood, v0.1.1222 would have shipped containing nothing but a version bump — its only content was already in 1221 — and the `elapsedMs` fix would have slipped to 1223 and a third rollout.

After the merge:

- contains #3502 — `git merge-base --is-ancestor 91b6bc3e8 HEAD` confirms
- diff against `main` is still exactly the three version files, all at `0.1.1222`
- `deno task generate` produces no change, so the hydration bundle needed no regeneration (#3502 touched only `src/agent/ag-ui/*` and `src/internal-agents/*`, which aren't part of that bundle)
- pre-push gate green: 3774 passed, 0 failed

#3502 is the fix that actually makes persisted run events carry `elapsedMs`. The durable emitter for hosted and scheduled runs is `AgUiBrowserEncoderState`, which had no clock; the runtime streams its AG-UI events as SSE to veryfront-api, which persists them verbatim. Despite the name, no browser is attached to a scheduled run — that name is why this was fixed in the wrong encoder twice, in #3483 and #3497.

PR #3547, opened 2026-08-10
fix(api): resolve Veryfront imports in Deno temp modules

kojiwakayama:
## Summary

- resolve noncompiled Deno `veryfront` imports through the running package export map
- preserve the existing compiled Deno shims and Node/Bun behavior
- reject package export targets that escape the installed package root
- add focused regressions for static and dynamic subpath imports

## Root cause

Fresh Deno projects do not inherit this repository's import map. When a generated API route had to use the bundled temporary-module path, bare imports such as `veryfront/embedding` remained external. Deno then evaluated the temporary file outside the installed package context and could not resolve the framework import.

Using `import.meta.resolve` alone was insufficient because the packed npm build uses a DNT resolver ponyfill that loses package self-resolution. This change reads the active package export map and converts framework imports to contained file URLs before Deno evaluates the temporary module.

## Verification

- focused rewriter regression captured red, then passed green
- module-loader suites: 5 passed, 97 steps, 0 failed
- formatting, lint, and typecheck passed
- full pre-push checks passed, including parallel unit tests and isolated CWD suites
- generated docs-agent Deno project passed scaffold, install, build, production serve, `/`, `/uploads`, browser checks, and `/api/uploads` with a clean network log

Related: veryfront/veryfront-issue-inbox#456

### DIFF
--- a/src/routing/api/module-loader/external-import-rewriter.test.ts
+++ b/src/routing/api/module-loader/external-import-rewriter.test.ts
@@ -1,5 +1,10 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { assert, assertEquals, assertStringIncludes } from "#veryfront/testing/assert.ts";
+import {
+  assert,
+  assertEquals,
+  assertStringIncludes,
+  assertThrows,
+} from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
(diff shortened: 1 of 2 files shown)


kojiwakayama:
@codex review

kojiwakayama:
@codex review

PR #3552, opened 2026-08-10
test: make Bun filesystem watcher integration deterministic

kojiwakayama:
## Summary

- wait for the file written by the Bun integration test instead of assuming the first directory watcher event is relevant
- ignore unrelated queued watcher events and clear the deadline timer during cleanup
- use the same target-filtering pattern as the Deno watcher integration

## Why

PR #3547 exposed this pre-existing CI flake. Bun's Node-compatible directory watcher can surface a queued event for another path before the `file.txt` event, and a busy CI host can deliver the target event after the old three-second deadline. The adapter behavior is correct; the test consumed an underspecified first event.

## Regression evidence

Before the change, repeated isolated Bun runs reproduced both failure modes:

- `Bun filesystem watcher integration timed out`
- `Expected true, got false` after the first event named another path

After the change, the same test passed 100 consecutive runs.

## Verification

- `deno fmt --check src/platform/adapters/runtime/bun/filesystem-adapter.bun.test.ts`
- `deno lint src/platform/adapters/runtime/bun/filesystem-adapter.bun.test.ts`
- `deno check src/platform/adapters/runtime/bun/filesystem-adapter.bun.test.ts`
- `bun test src/platform/adapters/runtime/bun/filesystem-adapter.bun.test.ts`
- 100 consecutive isolated Bun runs
- `deno task build:npm`
- `node ./tests/bun/run-tests.mjs src/platform/adapters/runtime/bun/filesystem-adapter.bun.test.ts`
- repository pre-push formatting, lint, typecheck, and partitioned unit suite

Related: #3547

### DIFF
--- a/src/platform/adapters/runtime/bun/filesystem-adapter.bun.test.ts
+++ b/src/platform/adapters/runtime/bun/filesystem-adapter.bun.test.ts
@@ -9,16 +9,13 @@ import { join } from "node:path";
 import { BunFileSystemAdapter } from "./filesystem-adapter.ts";
 import { getBunRuntime } from "./types.ts";
 
-function delay(milliseconds: number): Promise<void> {
-  return new Promise((resolve) => setTimeout(resolve, milliseconds));
-}
-
 describe("BunFileSystemAdapter native integration", () => {
   it("reads, writes, and watches through the real Bun runtime", async () => {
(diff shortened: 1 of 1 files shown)


kojiwakayama:
@codex review

PR #3574, opened 2026-08-11
fix(config): keep the env-ordering diagnostic off the user's console

kojiwakayama:
## Symptom

Found on a DX dogfood walk of the published `veryfront@0.1.1228`.

A freshly scaffolded project (`npm create veryfront@latest support-agent -- --yes`, untouched)
prints a framework-internal diagnostic on the user's **first agent request** — the first chat
message in the default `ai-agent` template:

```
10:17:55  VERYFRONT  ▲ [EnvironmentConfig] getEnvironmentConfig called before .env load. Returning uncached snapshot; ensure loadEnv runs before environment config access.
10:17:56  VERYFRONT  ✖ Failed to fetch remote integration tool definitions
                       error="Integration tools API returned 400 Bad Request"
```

The advice the line carries — "ensure loadEnv runs before environment config access" — is only
actionable by framework code. An application developer has no `loadEnv` to reorder, so the line
reads as an unexplained framework warning sitting next to real errors.

## Root cause

Two things combine:

1. `warnEarlyAccess()` in `src/config/environment-config.ts` logged at `warn`, i.e. on a channel the
   user sees at the default log level.
2. The condition it reports is reachable in a perfectly healthy session. Captured with
   `VERYFRONT_DEBUG_RUNTIME_ENV=1`, the stack is:

   ```
   warnEarlyAccess  <- getEnvironmentConfig <- getApiBaseUrlEnv
     <- getRemoteIntegrationToolDiscovery <- getRemoteToolDefinitions
     <- getAvailableTools <- prepareAgentRuntimeStep <- executeAgentLoopStreaming
   ```

   and every frame resolves to **`support-agent/node_modules/veryfront/esm/…`** while the CLI
   process itself is running out of the globally installed copy. A second copy of the framework is
   loaded from the project's own `node_modules` mid-request. In that copy `loadEnv` has never run
   and never will, so `hasEnvLoaded()` is permanently `false` and the diagnostic fires on the first
   call that reads env config.

The snapshot returned in that state is still correct — process env is shared between the two copies
and the early path deliberately returns an uncached snapshot. Only the message is wrong to show.

This PR fixes the leak. The underlying split module graph (global CLI + project-local `veryfront`
loading two framework copies in one process) is the same root cause behind the separately tracked
"globally-installed CLI never registers the project's `tools/` directory" finding, and is left to
that change.

## Fix

`warnEarlyAccess()` now logs the message at `debug`. The `warn`-with-stack path is unchanged and
still available behind the existing `VERYFRONT_DEBUG_RUNTIME_ENV` opt-in, so framework work on env
ordering keeps the loud signal, and `LOG_LEVEL=debug` / `VERYFRONT_DEBUG=1` still surfaces the
message for everyone else. Nothing about when the diagnostic is computed, or about the snapshot it
guards, changed.

## Regression test

`src/config/environment-config.test.ts` — next to the existing early-snapshot tests, because the
behaviour under test is a property of `getEnvironmentConfig()`'s pre-`loadEnv` branch, and that file
already owns the stubbing pattern for it. No browser or deployment is involved, so this is a Deno
BDD unit test rather than an e2e.

Two cases:

- `keeps the early-access ordering diagnostic off the user-facing warn channel` — with
  `VERYFRONT_DEBUG_RUNTIME_ENV` off, two early `getEnvironmentConfig()` calls produce **zero** warn
  records and exactly one debug record (also pinning the once-per-process dedupe). Fails on `main`
  with `Expected value to have length 0, but it does not: the value has length 1`.
- `keeps the early-access diagnostic on warn with a stack under VERYFRONT_DEBUG_RUNTIME_ENV` — pins
  the opt-in escalation path so a future cleanup cannot quietly delete the diagnostic.

The existing `keeps early snapshots immutable and uncached until environment loading completes` test
asserted `warnings).toHaveLength(1)`; that assertion moved to the new tests and now reads `0`.

## Verification

- New test fails on `main` for the right reason, passes with the fix.
- `src/config/`, `src/utils/env-loader.test.ts`, `src/utils/logger/logger.test.ts`: 48 passed, 0 failed.
- Original symptom: re-ran the exact repro (published 0.1.1228 scaffold, globally installed CLI,
  `POST /api/ag-ui` with UUID thread/run/message ids) with this change applied to the installed
  package. `[EnvironmentConfig]` occurrences in the server log went from 1 to 0; the rest of the log
  is unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced unnecessary warning messages when environment configuration is accessed early.
  - Early-access diagnostics are now logged quietly by default while still being recorded for troubleshooting.
  - Runtime debug mode continues to provide warning messages with stack details for easier investigation.

- **Documentation**
  - Added guidance explaining early environment-configuration access diagnostics and debug behavior.

### DIFF
--- a/src/config/environment-config.test.ts
+++ b/src/config/environment-config.test.ts
@@ -90,7 +90,7 @@ describe("EnvironmentConfig", () => {
           expect(second).not.toBe(first);
           expect(Object.isFrozen(second)).toBe(true);
           expect(isEnvironmentConfigInitialized()).toBe(false);
-          expect(warnings).toHaveLength(1);
+          expect(warnings).toHaveLength(0);
 
           const earlyInit = initEnvironmentConfig();
           expect(earlyInit.apiToken).toBe("after-mutation");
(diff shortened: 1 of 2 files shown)


kojiwakayama:
@coderabbitai review

PR #3575, opened 2026-08-11
fix(css): report a missing CSS optimizer once, and say how to add one

kojiwakayama:
Found during a DX dogfood walk of the published docs (`docs/code/getting-started/deploy-project`), following them literally as a new developer.

## Symptom

`veryfront build` on a freshly scaffolded project prints:

```
  ● Building...
  ! No CSSOptimizationEngine registered; emitting unminified CSS
  ✓ Built in 3.03s
```

`veryfront serve` repeats it at startup, and the hosted runtime emits it at
`level=warn component=css-compiler` **once per render** — it was the single most
frequent line in the dogfood project's cloud logs (4+ entries in the first four
seconds of the deployment's stream).

Two problems, one line:

1. **Volume.** `regenerateCSSByHash` acquires a generation session on every
   cold-cache request, before the in-flight map is even consulted, so every
   render that misses the local CSS cache logs the warning again. A default
   project floods its own observability with a message about a condition that
   has not changed since boot.
2. **No action.** The message names `CSSOptimizationEngine`, an internal
   contract name. Nothing in the docs says how to register one, so a developer
   reading it has no next step.

## Root cause

`acquireCSSGenerationSession` in `src/html/styles-builder/tailwind-compiler.ts`
logged unconditionally whenever `minify` was requested and no engine resolved.
The absence is a process-level fact, but the log was per-acquisition.

## Fix

Report the absence once while it holds, and name the package that fixes it:

```
! No CSSOptimizationEngine registered; emitting unminified CSS. Install one with: deno add @veryfront/ext-css-lightning
```

The package name comes from the existing `getRecommendation` map, which already
maps `CSSOptimizationEngine` → `@veryfront/ext-css-lightning`, and the phrasing
matches how `resolve()` reports a missing extension.

The flag re-arms whenever an engine *is* observed, so a project whose engine
registration later breaks gets a fresh warning rather than permanent silence —
which is the case the warn level exists for.

Behaviour is otherwise unchanged: an absent engine still degrades to unminified
CSS and is still recorded in the pipeline cache identity, so no minified entry
can be served from an unminified one.

## Regression test

`src/html/styles-builder/css-provider-session.test.ts` — "reports a missing
optimizer once, with the package that provides one". It lives beside the other
provider-session tests because the behaviour under test is a property of
`acquireCSSGenerationSession`, the same seam those tests already drive; nothing
about it needs a browser or a deployment.

The test acquires three sessions with no engine registered and asserts exactly
one `warn` record naming `@veryfront/ext-css-lightning`. It first registers and
acquires with an engine so it re-arms the flag itself and does not depend on
test ordering within the file.

Confirmed failing before the fix for the right reason (`3` records, expected
`1`), passing after.

## Verification

- `deno test src/html/styles-builder src/extensions src/release-assets src/build src/server/handlers/dev/styles-css.handler.test.ts` — 174 passed, 0 failed.
- Re-ran the finding's command against a scaffolded project with the local tree:
  the build now prints the actionable message once.
- `deno fmt --check`, `deno lint`, `deno check`, `lint:module-boundaries`,
  `lint:dependency-boundaries` all clean.

### DIFF
--- a/src/html/styles-builder/css-provider-session.test.ts
+++ b/src/html/styles-builder/css-provider-session.test.ts
@@ -6,6 +6,11 @@ import {
   type CSSProcessor,
   CSSProcessorName,
 } from "#veryfront/extensions/css/index.ts";
+import {
+  __resetLogRecordEmitterForTests,
+  __subscribeLogRecordEmitter,
+  type LogEntry,
+} from "#veryfront/utils/logger/index.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { afterEach, beforeEach, describe, it } from "#veryfront/testing/bdd.ts";
 import {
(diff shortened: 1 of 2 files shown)


kojiwakayama:
CI note: the first run had `coverage shard 8/8` fail on `src/transforms/esm/http-cache.test.ts` -> "returns a signal-less cache follower after its bounded wait" (`followerSettled` false at the fake-timer boundary), which cascaded into `tests (unit)` and `coverage gate`. That file is untouched here and does not import the CSS pipeline; it passes locally and passed on re-run. All checks are green now.

kojiwakayama:
Rebased onto current `main` (`7b25cb7bd`) and force-pushed — `c1777cb83` → `5a867faee`. No conflicts; the eight PRs that landed today touch nothing in the CSS pipeline, and the diff is byte-identical to the reviewed one.

CI on the rebased commit: **27 pass, 6 skipping, 0 failing**, first run — the `http-cache.test.ts` flake noted above did not recur.

Re-verified locally after the rebase:

- `deno test src/html/styles-builder/css-provider-session.test.ts` — 9 steps pass, including the new "reports a missing optimizer once" case.
- `deno fmt --check`, `deno lint`, `deno check` on the touched files — clean.
- `lint:module-boundaries` and `lint:dependency-boundaries` — clean (`core=0, cli=0, react isolated`).

One note on how the push was made: the local `pre-push` gate (full `test:unit`) failed once on `src/workflow/executor/step-executor.test.ts:233` — "stops waiting after the cancellation grace when a timed-out tool never settles". That test asserts against a wall-clock deadline and was running on a machine saturated with parallel work; it passes in 16 ms in isolation, lives in a subsystem this PR does not touch, and passed in CI on both the pre- and post-rebase commits. I pushed past that specific gate rather than modify an unrelated test, and CI is the confirming evidence above. Flagging it here so the failure is on the record rather than silent.

No review comments have been filed on this PR — CodeRabbit hit its account review limit at open time (its only comment is the rate-limit warning) and had not re-reviewed as of this rebase. Nothing to address or decline; if a review lands later it will need a fresh pass.

PR #3588, opened 2026-08-11
fix(dev-server): release watchers and reload subscriptions when the bind fails

kojiwakayama:
## The defect

`DevServer.start()` registers four things **before** it binds the HTTP port:

| Registration | `server.ts` (pre-fix) |
| --- | --- |
| `setupFileWatchers()` | :164 |
| `ReloadNotifier.subscribeInvalidate` | :167 |
| `HMRHandler.registerExternalBroadcastSource()` | :177 |
| `ReloadNotifier.subscribe` | :178 |
| `adapter.serve()` — the bind | :258 |

All four are released only by `stop()`. But `startDevServer()` (`index.ts:15-20`) constructs the instance, awaits `start()`, and *then* returns it — so when `start()` rejects, the half-built instance is dropped with no handle and **no caller can ever call `stop()`**. All four registrations survive for the life of the process.

The leaked broadcast source is the worst of the three kinds: it decrements a *process-wide* counter, and while it is held `HMRHandler` skips its own broadcast (`hmr.handler.ts:72-78`). So one failed start can silently disable HMR for the dev server that starts next.

Reachable from the probe-then-bind race in `veryfront dev`, and from any `adapter.serve()` failure that is not a port collision (bad bind address, permission denied on a privileged port).

## The fix

`start()` routes failures through `stop()` and rethrows.

Chosen over catching in `startDevServer()` because:

- **`stop()` is already the single teardown path, and already null-safe at every step** (`?.()`, `if (this.fileWatchSetup)`, `if (this.server)`) — it was written to tolerate a partially-built instance. Reusing it means there is no second cleanup list that can drift out of sync when someone adds a new registration to `start()`.
- **It covers every caller.** `DevServer` is exported directly and constructed outside `startDevServer()` (e.g. `tests/integration/server/dev-server-handlers.test.ts:28`); a fix in `startDevServer()` would leave those leaking.
- **It matches existing convention.** `bootstrap.ts` (`orchestrateOrDisposeFS`, and the comment at :482-484) and the Deno `http-server.ts` adapter both release-then-rethrow at the registration site.

Cleanup errors are swallowed to a debug log so they can never mask the real reason `start()` failed — "port already in use" is what the developer needs to see.

`stop()` now clears all three release handles **up front, before any is invoked**, and runs each through a small `release()` helper. This is required, not cosmetic: the reload/invalidate unsubscribes are idempotent (`Set.delete`), but `releaseExternalBroadcastSource` is **not** — it decrements a global counter, so the double-`stop()` this change introduces (failed `start()` cleans up, caller still holds the instance and calls `stop()`) would otherwise corrupt an unrelated dev server's count. Clearing up front rather than per-call means a throwing release cannot leave a later handle set for a second `stop()` to re-invoke.

## Regression test

`src/server/dev-server/server-start-failure.integration.test.ts` forces a **real** bind failure — another listener holds the port — after the subscriptions are registered, and asserts the registrations are released. It does not settle for asserting that `start()` rejects; the rejection already happened before this change.

Watchers are counted by wrapping the runtime registry adapter's `fs.watch`, which is the only seam that observes the watcher `DevServer` really opens (it builds its own adapter internally). The test asserts `watchers.opened > 0` first, so it cannot pass vacuously if that seam ever stops being the one used.

Failing before the fix — all three leaks in one diff:

```
    {
-     invalidateListeners: 1,
-     openWatchers: 1,
-     reloadListeners: 1,
+     invalidateListeners: 0,
+     openWatchers: 0,
+     reloadListeners: 0,
    }
```

Passing after, with **no sanitizer opt-outs** — Deno's own op/resource sanitizers confirm the teardown is complete.

It is colocated with the module it covers per AGENTS.md, and named `*.integration.test.ts` because it needs a real project directory, a full `bootstrapDev()`, a real OS file watcher and a real TCP bind. That suffix is excluded from the unit shard by `deno.json` and follows `src/proxy/routing-invalidation.integration.test.ts`; it still runs in the coverage shards and the pre-push suite.

## Verification

- New test fails before the fix and passes after — re-confirmed after relocating it
- `src/server/dev-server/` + `tests/integration/server/`: 52 passed, 0 failed
- `deno fmt --check`, `deno lint`, `deno check` clean
- Pre-push gate passed; all 27 CI checks green

## Notes

- `docs/api-reference/veryfront/server.md` carries a one-line source-link update: the teardown helper shifted `export class DevServer` from L57 to L70. Regenerated with `deno task docs` under the Deno version CI pins (2.7.7), not hand-edited.
- The known flake `src/transforms/esm/http-cache.test.ts` ("returns a signal-less cache follower after its bounded wait", from #3553) fired once on an earlier push and presented as three red checks (its coverage shard plus the dependent `tests (unit)` and `coverage gate`). Re-run without changes and it went green; it is untouched by this PR.
- Out of scope, noted while working: when `start()` fails, `this.ready` never settles, so anything awaiting it hangs. Pre-existing and independent of this leak — left alone deliberately.

Surfaced by CodeRabbit on #3562 and declined there as out of scope, since that PR touched only `cli/`.

### DIFF
--- a/docs/api-reference/veryfront/server.md
+++ b/docs/api-reference/veryfront/server.md
@@ -61,7 +61,7 @@ await server.fetch(new Request("https://example.com/health"));
 
 | Name             | Description           | Source                                                                                                       |
 | ---------------- | --------------------- | ------------------------------------------------------------------------------------------------------------ |
-| `DevServer`      | Implement dev server. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/server/dev-server/server.ts#L57)          |
+| `DevServer`      | Implement dev server. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/server/dev-server/server.ts#L70)          |
 | `RouteDiscovery` |                       | [source](https://github.com/veryfront/veryfront-code/blob/main/src/server/dev-server/route-discovery.ts#L33) |
 
 ### Types
--- a/src/server/dev-server/server-start-failure.integration.test.ts
+++ b/src/server/dev-server/server-start-failure.integration.test.ts
@@ -11,15 +11,19 @@
  * These tests force a real bind failure (a port already held by another
  * listener) after the subscriptions are registered, and assert that the
  * registrations are released rather than merely that `start()` rejects.
+ *
+ * Colocated with the module it covers, but named `*.integration.test.ts`: it
+ * needs a real project directory, a full `bootstrapDev()`, a real OS file
+ * watcher and a real TCP bind, so it is excluded from the unit shard.
  */
 
 import { assert, assertEquals, assertRejects } from "#veryfront/testing/assert";
 import { describe, it } from "#veryfront/testing/bdd";
 import type { FileWatcher, WatchOptions } from "#veryfront/platform/adapters/base.ts";
 import { runtime } from "#veryfront/platform/adapters/registry.ts";
-import { ReloadNotifier } from "../../../src/server/reload-notifier.ts";
-import { DevServer } from "../../../src/server/dev-server.ts";
-import { withTestContext } from "../../_helpers/context.ts";
+import { ReloadNotifier } from "../reload-notifier.ts";
+import { DevServer } from "./server.ts";
+import { withTestContext } from "../../../tests/_helpers/context.ts";
 
 type WatchFn = (paths: string | string[], options?: WatchOptions) => FileWatcher;
 
--- a/src/server/dev-server/server.ts
+++ b/src/server/dev-server/server.ts
@@ -42,6 +42,19 @@ function normalizeSlug(value: string | undefined): string | undefined {
   return normalized ? normalized : undefined;
 }
 
+/**
+ * Run one teardown callback, isolating a throw so it cannot strand the teardown
+ * steps that follow it. Callers must clear the handle before calling this.
+ */
+function release(callback: (() => void) | undefined, label: string): void {
+  if (!callback) return;
+  try {
+    callback();
+  } catch (error) {
+    devServerLog.debug(`${label} cleanup error (non-critical)`, error);
+  }
+}
+
 function deriveProjectSlug(projectDir: string): string {
   const dirName = basename(projectDir);
   const slug = dirName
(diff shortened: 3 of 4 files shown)


kojiwakayama:
Declining this one, with reasoning on the record.

The AGENTS.md rule reads "Keep test files colocated as `*.test.ts` next to source", but it describes the unit tier. The repo has an established integration tree — 193 test files under `tests/`, run by a separate CI job (`deno task test:integration`, `cicd.yml:93`) — and this file sits alongside the five other DevServer integration tests already in `tests/integration/server/`: `dev-server.test.ts`, `dev-server-handlers.test.ts`, `dev-server-debounce.test.ts`, `dev-server-debounce-simple.test.ts`, `parallel-requests.test.ts`.

More importantly, this test cannot be a unit test as `DevServer` is currently written. It needs a real project directory, a full `bootstrapDev()`, a real OS file watcher and a real TCP bind that fails. `DevServer.start()` builds its own adapter via `runtime.get()` internally — there is no injection seam, which is exactly why the test has to reach the watcher by wrapping the runtime registry's `fs.watch`. Colocating it under `src/server/dev-server/` would move an integration test into the unit shard without making it any more focused; giving it a genuine unit seam would mean adding adapter injection to `DevServer`, which is a larger production refactor than this leak fix warrants.

On the concern itself — that focused verification of `src/server/dev-server/` could pass without exercising this path — that is equally true of the five sibling tests, and the pre-push hook and CI both run the full suite, so a reintroduced leak is caught before merge.

kojiwakayama:
Applied in 189e81549, with one change to the proposed shape.

The ordering point is the valuable half, and I took it further than the suggestion: all three handles are now cleared **up front**, before any of them is invoked. Clearing each handle immediately before its own call still leaves the later handles set if an earlier release throws, so a subsequent `stop()` would re-invoke them — and re-invoking `releaseExternalBroadcastSource` is the one genuinely harmful case, since it decrements a process-wide counter and would suppress HMR broadcasts for an unrelated dev server. Clearing all three first makes double-release impossible regardless of what throws.

Each release then runs through a module-level `release()` helper that debug-logs a failure instead of stranding the steps after it, so a throwing subscription can no longer prevent the file watcher from being cleaned up.

For the record on reachability: all three callbacks are non-throwing closures today — two are `Set.delete` unsubscribes from `createSubscriberSet`, and the third is `count = Math.max(0, count - 1)`. So this is defence in depth on a teardown path rather than a fix for a currently reachable failure, which is why I kept it to one small helper rather than a per-call try/catch block.

kojiwakayama:
Correcting my earlier reply on test placement — I declined this too quickly, and both reviewers were right. The test is now colocated at `src/server/dev-server/server-start-failure.integration.test.ts`.

What changed my mind was checking the two things I had assumed were blockers, and finding neither holds:

1. I assumed a colocated test could not reach `tests/_helpers`. It can — 18 files under `src/` already import from there, so `withTestContext` works fine from its new home.
2. I assumed colocating would force an integration-weight test (real `bootstrapDev()`, real OS file watcher, real TCP bind) into the unit shard. It does not, because of the `*.integration.test.ts` suffix, which `deno.json` excludes from `test:unit:parallel` and `test:coverage:unit`. `src/proxy/routing-invalidation.integration.test.ts` is the existing precedent for exactly this: colocated, integration-weight, importing `tests/_helpers`.

So the suffix resolves my only real objection while satisfying the AGENTS.md rule, and it directly addresses the concern raised — a focused `deno test src/server/dev-server/` now exercises the bind-failure cleanup, where before it would have passed without touching it. The test still runs in the coverage shards and the pre-push suite, so the regression stays gated.

Re-verified after the move: it still fails without the fix for the right reason (`openWatchers: 1, reloadListeners: 1, invalidateListeners: 1` against a zero baseline) and passes with it.

PR #3596, opened 2026-08-11
refactor: move cli/templates to a top-level templates/ directory

kojiwakayama:
## What

Moves `cli/templates/` to a top-level `templates/` directory. **This PR is only the move** — no `package.json`, no publishing config, no extraction.

## Why

The directory stopped being a CLI concern:

- `src/` already reaches into `cli/templates/integrations/` from five test files (`src/oauth/providers/{atlassian,google,microsoft,common}.test.ts`, `src/integrations/_data.test.ts`) — a layering violation today.
- veryfront-api#4323 is making the API read the same generated template manifest.

The repo's own precedent for a separately-published thing is a top-level sibling (`extensions/ext-*`), so `templates/` follows that. Preparation for veryfront-issue-inbox#477 (`@veryfront/templates`).

All four subdirectories move together — `ai-rules/`, `features/`, `files/`, `integrations/` — plus the loaders (`index.ts`, `loader.ts`, `types.ts`, `feature-loader.ts`, `integration-loader*.ts`) and the generated `manifest.json`, so the package has a single root. Every file moved with `git mv`, so history follows.

## Acceptance: a scaffolded project is byte-identical

Two independent proofs, both run before and after the move:

**1. Real CLI scaffolding.** `origin/main` was extracted to a scratch tree with `git archive`, and all seven starter templates plus an `ai-agent --integrations github,slack` build were scaffolded with `veryfront init --skip-install --skip-env-prompt` from each tree:

```
diff -r scaffold-before scaffold-after   →  no differences (162 files)
```

CLI stdout was identical too.

**2. Loader-API dump.** A 443-file dump of everything the templates package can produce — every starter tree, every integration tree raw + namespaced + merged, integration base files, all 204 connector configs, all ai-rules, all feature configs, and the registry ordering — diffed clean.

`git diff --stat -M` reports **926 files changed, 0 insertions(+), 0 deletions(-)** for the moved tree: a pure rename with zero content edits.

## Manifest

The manifest is generated (`scripts/build/generate-templates-manifest.ts`). Regenerated at its new location it is **byte-identical** to the version committed on `origin/main`:

```
6caa2a70…  before (origin/main:cli/templates/manifest.json)
6caa2a70…  after  (templates/manifest.json)
```

Template file paths inside the manifest are relative to each template root, so they do not move. Same 48 templates, 464 files.

## The lint exclusions still exclude

`deno.json` excludes `templates/files/` and `templates/integrations/` so template sources are not linted as framework source. Because `lint.include` / `fmt.include` enumerate roots explicitly, `templates/**` was added there too — otherwise the loaders would have silently dropped out of both gates. Verified by comparing file counts against the `origin/main` tree:

| gate | before | after |
| --- | --- | --- |
| `deno lint` | Checked 4917 files | Checked 4917 files |
| `deno fmt --check src/ cli/ react/ [templates/]` | Checked 4763 files | Checked 4763 files |
| sanitizer opt-out baseline | 404/404 | 404/404 |
| skipped-test baseline | 18/18 | 18/18 |
| cwd-relative test reads | 0 module scope, 95 across 21 files | 0 module scope, 95 across 21 files |
| duplicate-function groups | 234 | 234 |

Not one more file, not one fewer.

## Scan roots that enumerated `src` and `cli`

Several tools hard-code the roots they walk. Left alone, the loaders and their four test files would have quietly left every one of them. `templates` was added to:

- `scripts/test/coverage-ci.ts` (`UNIT_COVERAGE_ROOTS`) — coverage sharding
- `scripts/lint/check-sanitizer-baseline.ts`, `check-skipped-tests-baseline.ts`, `ban-test-only.ts` (`SCAN_ROOTS`)
- `scripts/lint/audit-cwd-relative-test-reads.ts` (`SCAN_ROOTS`)
- `scripts/lint/check-test-typecheck-baseline.ts` (`listTestFiles`)
- `scripts/lint/find-duplicate-functions.ts` (`DEFAULT_PATHS`)
- `deno.json` `test:unit:parallel` and `test:coverage:unit` (`find src cli templates`)

## Published layout change (breaks veryfront-api#4323's assumption)

Verified by running `deno task build:npm`:

- **was** `veryfront/esm/cli/templates/manifest.js`
- **now** `veryfront/esm/templates/manifest.js` (`npm/esm/cli/templates/` no longer exists)

`templates` is not in the package `exports` map, so the only way to reach the manifest is a relative URL off an exported entry. veryfront-api#4323 resolves it by hopping off `veryfront/cli`, which resolves to `esm/cli/main.js`. That hop must change:

```diff
- new URL("./templates/manifest.js",  import.meta.resolve("veryfront/cli"))
+ new URL("../templates/manifest.js", import.meta.resolve("veryfront/cli"))
```

(The in-repo consumer, `cli/commands/install/registry.ts`, now emits `import manifest from "../../../templates/manifest.js"` in the built package — confirmed in `npm/esm/`.) **Not fixed here** — #4323 needs the one-line update.

## Other paths updated

`deno.json` (exclude, lint/fmt include, fmt + test tasks), `.gitignore`, `socket.yml`, `.github/actions/setup-deno/action.yml`, `scripts/build/generate-templates-manifest.ts`, `scripts/build/generate-integrations-module.ts`, `scripts/build/npm-package-metadata.test.ts`, `scripts/lint/{audit-core-deps,ban-zod-imports,find-duplicate-functions}.ts` (+ tests), `scripts/test/npm-install-smoke.sh`, `cli/commands/install/registry.ts`, all relative imports across `cli/`, `src/`, `tests/`, and `cli/README.md`. `grep -rn "cli/templates"` over the repo returns nothing.

## Verified locally

`fmt:check`, `lint`, `typecheck`, `generate:manifests:check`, `docs:api-reference:check`, `docs:validate`, every `lint:*` gate, `build:npm`, and the full `test:unit` suite (via the pre-push hook) all pass.

Refs veryfront-issue-inbox#477


kojiwakayama:
### CI status: all 9 required checks pass; one non-required check needs a human action

`CodeQL` is red with *"3 new alerts including 3 high severity security vulnerabilities"*. It is **not** in the required-check set (`Analyze`, which is, passed).

The three alerts are not new findings. They are the same three alerts, on the same rules, in the same files, at the same lines — already adjudicated and dismissed on `main`. CodeQL keys alert identity partly on file path, so a rename re-opens them and the dismissal does not follow:

| dismissed on `main` | re-opened on this PR | rule | location |
| --- | --- | --- | --- |
| #14 | #278 | `js/incomplete-multi-character-sanitization` | `…/integrations/teams/files/lib/teams-client.ts:233` |
| #5 | #277 | `js/double-escaping` | `…/integrations/teams/files/lib/teams-client.ts:233` |
| #4 | #276 | `js/double-escaping` | `…/integrations/confluence/files/lib/confluence-client.ts:287` |

All three were dismissed by @ariskemper as *false positive* with the comment:

> Scaffold template file — copied to user projects, not runtime code

That rationale is about the file's role, not its path, so it applies verbatim after the move. The code itself is provably untouched: `git diff --stat -M` reports **0 insertions, 0 deletions** across all 926 moved files.

I deliberately did not re-dismiss them. Re-applying a security adjudication is the repo owner's call, not something to automate on their behalf. To restore the previous state:

```bash
for n in 276 277 278; do
  gh api -X PATCH "repos/veryfront/veryfront-code/code-scanning/alerts/$n" \
    -f state=dismissed -f dismissed_reason='false positive' \
    -f dismissed_comment='Scaffold template file — copied to user projects, not runtime code'
done
```

Worth deciding separately: whether scaffold template trees should be in CodeQL's scope at all. If not, a `paths-ignore` for `templates/files/**` and `templates/integrations/**` in a CodeQL config would stop this recurring — but that hides future findings in template code too, so it is a policy call rather than part of this move.

PR #3599, opened 2026-08-11
fix(cli): probe dev ports without the dnt Deno shim

kojiwakayama:
## The symptom

On published `0.1.1229`, `veryfront dev` never binds a port when the CLI runs on Deno — the Deno-global install (`deno install -gArf npm:veryfront`), `deno run -A npm:veryfront dev`, and the `deno task dev` that a `--runtime deno` scaffold prints as its own next step:

```
Veryfront (v0.1.1229)

  Detail: Cannot read properties of null (reading 'fd')
```

`build`, `serve`, `routes` and `doctor` all succeed under the same install. `dev` was the only command that reached the broken code, and the npm-global (Node) install of the identical version runs `dev` fine.

## Root cause

The real stack, recovered by calling `handleDevCommand` directly under Deno against the published tarball:

```
TypeError: Cannot read properties of null (reading 'fd')
    at Object.listen (node_modules/@deno/shim-deno/dist/deno/stable/functions/listen.js:44:64)
    at isPortAvailable (node_modules/veryfront/esm/cli/commands/dev/port-fallback.js:38:26)
    at clearLocalCachesIfPortFree (node_modules/veryfront/esm/cli/commands/dev/command.js:95:16)
```

`isPortAvailable` branched on `isDeno` and then called the bare `Deno.listen`. In the npm build dnt rewrites every bare `Deno` to `@deno/shim-deno` — `esm/_dnt.shims.js` re-exports the shim unconditionally, so the shim wins even when a genuine `Deno` global is right there. That shim's Node-backed `listen()` does:

```js
const waitFor = new Promise((resolve) => server.listen(port, hostname, resolve));
// @ts-expect-error undocumented socket._handle property
const listener = new Listener(server._handle.fd, …);
```

`server._handle` is assigned synchronously by Node's `listen()`, but not by Deno's node-compat layer, so it is still `null` on that line. The one branch written for Deno was the one branch Deno could not execute.

The repo already knows this hazard — `cli/auth/callback-server.ts` reaches the real namespace via `self` ("Access native Deno.serve via `self` to bypass dnt shim transform"), and `getNativeDeno()` in `src/platform/compat/http/native-response.ts` documents it. `port-fallback.ts`, which was written to copy that callback server's fallback, missed it.

## The fix

Probe with `node:net` on every runtime and delete the Deno branch. This function already used `node:net` on Node; it is one path now, it behaves identically on Deno, and it cannot reach the shim. Verified on all four runtime shapes the CLI ships in: Deno source, `deno compile` binary, npm-on-Node, npm-on-Deno.

## Regression test

`cli/commands/dev/port-fallback.test.ts` stands a throwing `Deno.listen` into the ambient namespace — the shim reproduced in-process, same `TypeError: Cannot read properties of null (reading 'fd')` — and asserts the probe still reports a held port busy and a free port free. It fails on `main` at `port-fallback.ts:41` with exactly the published error, and passes with this change.

## Proof against the published artifact, not just the tests

The finding's own command, re-run against `veryfront@0.1.1229` with only this function transplanted into the installed tree:

| | before | after |
|---|---|---|
| `deno run -A ./node_modules/veryfront/bin/veryfront.js dev --port 3772` | `✗ [unknown-error] … reading 'fd'`, nothing listening | `✓ Ready in 1.6s` / `http://veryfront.me:3772`, `lsof` confirms `TCP 127.0.0.1:3772 (LISTEN)` |
| same, port already taken | — | `! Port 3772 is in use, using 3773 instead` → `✓ Ready` |
| `node ./node_modules/veryfront/bin/veryfront.js dev --port 3774` (control) | `✓ Ready`, `curl` 200 | `✓ Ready`, `curl` 200 |

## What this does *not* fix

With the port crash gone, `veryfront dev` under Deno binds and reports Ready, but page renders still fail with a **separate** Deno defect: the SSR module loader dynamic-importing generated `.cache/veryfront-http-bundle/http-*.mjs` throws `TypeError: Loading unprepared module: …`, so `/` returns 500 (Node returns 200 on the identical project). That is a different failure in a different subsystem and belongs to the open "CLI is entirely non-functional under the Deno runtime" reports — deliberately out of scope here rather than folded into this change.

Also from the originating finding, and deliberately unchanged: the published package declaring 7 of the 20 `@veryfront/ext-*` specifiers its runtime references is **policy, not a packaging bug** — `rootNpm` in `src/extensions/first-party-defaults.ts` keeps credentialed, native and output-changing integrations out of the root dependency set, and `first-party-import.ts` now degrades with an install hint instead of dying. No live doc page needs to change for this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved development server port detection across supported runtimes.
  * Port availability checks now reliably identify occupied and available ports without depending on ambient runtime behavior.

* **Tests**
  * Added regression coverage for port detection when the runtime’s default listener is unavailable.

### DIFF
--- a/cli/commands/dev/port-fallback.test.ts
+++ b/cli/commands/dev/port-fallback.test.ts
@@ -127,13 +127,9 @@ describe("cli/commands/dev/port-fallback", () => {
     });
 
     it("probes without going through the ambient Deno namespace", async () => {
-      // The npm build runs this file with dnt's @deno/shim-deno standing in for
-      // the `Deno` namespace, even under Deno itself. That shim's Node-backed
-      // `listen()` reads `server._handle.fd` synchronously, which is null on
-      // Deno, so a probe routed through the ambient namespace dies with
-      // "Cannot read properties of null (reading 'fd')" and `veryfront dev`
-      // never binds a port. Standing in a throwing `Deno.listen` reproduces
-      // that shim here: the probe must not depend on it.
+      // Stand in the failure the npm build hits under Deno: dnt swaps the
+      // ambient namespace for @deno/shim-deno, whose `listen` throws on a null
+      // `server._handle`. The probe must answer correctly without it.
       const nativeListen = Deno.listen;
       const held = nativeListen({ hostname: "127.0.0.1", port: 0 });
       const heldPort = (held.addr as Deno.NetAddr).port;
--- a/cli/commands/dev/port-fallback.test.ts
+++ b/cli/commands/dev/port-fallback.test.ts
@@ -125,6 +125,42 @@ describe("cli/commands/dev/port-fallback", () => {
 
       assertEquals(await isPortAvailable(freePort), true);
     });
+
+    it("probes without going through the ambient Deno namespace", async () => {
+      // The npm build runs this file with dnt's @deno/shim-deno standing in for
+      // the `Deno` namespace, even under Deno itself. That shim's Node-backed
+      // `listen()` reads `server._handle.fd` synchronously, which is null on
+      // Deno, so a probe routed through the ambient namespace dies with
+      // "Cannot read properties of null (reading 'fd')" and `veryfront dev`
+      // never binds a port. Standing in a throwing `Deno.listen` reproduces
+      // that shim here: the probe must not depend on it.
+      const nativeListen = Deno.listen;
+      const held = nativeListen({ hostname: "127.0.0.1", port: 0 });
+      const heldPort = (held.addr as Deno.NetAddr).port;
+      const freeListener = nativeListen({ hostname: "127.0.0.1", port: 0 });
+      const freePort = (freeListener.addr as Deno.NetAddr).port;
+      freeListener.close();
+
+      Object.defineProperty(Deno, "listen", {
+        configurable: true,
+        writable: true,
+        value: () => {
+          throw new TypeError("Cannot read properties of null (reading 'fd')");
+        },
+      });
+
+      try {
+        assertEquals(await isPortAvailable(heldPort), false);
+        assertEquals(await isPortAvailable(freePort), true);
+      } finally {
+        Object.defineProperty(Deno, "listen", {
+          configurable: true,
+          writable: true,
+          value: nativeListen,
+        });
+        held.close();
+      }
+    });
   });
 
   describe("isPortInUseError", () => {
--- a/cli/commands/dev/port-fallback.test.ts
+++ b/cli/commands/dev/port-fallback.test.ts
@@ -163,6 +163,41 @@ describe("cli/commands/dev/port-fallback", () => {
     });
   });
 
+  describe("npm build safety", () => {
+    it("never reaches the runtime through the binding dnt rewrites", async () => {
+      // dnt rewrites every bare `Deno.<member>` access in the npm build to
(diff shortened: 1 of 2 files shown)


kojiwakayama:
Addressed the nitpick in 7b9f2cedb: both comments trimmed to the part a reader cannot recover from the code — dnt swaps the ambient `Deno` for a shim whose `listen` reads a null `server._handle` — with the incident retelling dropped. Kept the mechanism itself rather than moving it to the PR description: the whole defect is that the obvious-looking `Deno.listen` is the one call this file must not make, and a future editor reading only the source would have no way to know that.

kojiwakayama:
Closing as **superseded by #3610**, which merged at `a7ba48676` and fixes the same crash from the same root-cause analysis.

These were competing implementations of one fix, not two fixes:

- **#3610 (merged)** — reads the runtime via `getDenoRuntime()` → `Reflect.get(globalThis, "Deno")`, which dnt leaves alone, and keeps the native `Deno.listen` path.
- **#3599 (this)** — deletes the Deno branch entirely and probes with `node:net` on every runtime.

Rebasing this onto main is not a conflict resolution but an adjudication between the two, so it was left alone rather than decided unilaterally.

### A trap worth recording

`cli/commands/dev/port-fallback.test.ts` **auto-merged with no conflict marker**, producing a file that requires the source to contain `getDenoRuntime()` *and* to not route through the ambient namespace. No implementation satisfies both. Verified in both directions:

- main's implementation + merged test → `TypeError: Cannot read properties of null (reading 'fd')` at `port-fallback.ts:51`
- this PR's implementation + merged test → fails `assertStringIncludes(code, "getDenoRuntime()")`

So a rebase that resolved only the `.ts` conflict would have shipped a guaranteed-red test file that git reported as clean.

### Nothing is lost by closing

The static source-level guard this PR carried (ported from #3598) has an equivalent on main — `"probes ports through getDenoRuntime(), not the bare global dnt rewrites"`, which reads the source the same way. One nuance for anyone revisiting: main's asserts the source *contains* `getDenoRuntime()`, whereas the version here asserted it contains *no bare* `Deno.` member access. The negative form is slightly stronger — it would catch a bare `Deno.hostname()` added anywhere in the file, including paths no behavioural test executes.

If the single-path `node:net` simplification is still wanted, it deserves a fresh PR against current main that also removes #3610's `getDenoRuntime()` assertion, argued on its own merits.

Branch left in place.

kwakayama (commented):
Found one low-severity documentation issue.

- [P3] Update the stale remediation guidance in `cli/commands/dev/port-fallback.test.ts:174`. It says the safe approach is `getDenoRuntime()` / `Reflect.get(globalThis, "Deno")`, but the implementation now deliberately uses `import("node:net")` in `cli/commands/dev/port-fallback.ts:43`. This is misleading to future maintainers and could steer a regression fix toward a different, untested path.

Rubric:

- Correctness: 40/40
- Tests: 20/20
- Reliability/security: 15/15
- Maintainability: 13/15
- Scope/docs: 8/10

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 7b9f2cedb43c311b491701003b7aa5ec9ed2ac60
Score: 96/100
Actionable-Findings: 1
Verdict: COMMENT

PR #3645, opened 2026-08-12
fix(security): stop a project's auth policy 401ing the control plane's own dispatch

kojiwakayama:
Second instance of the bug class [#3641](https://github.com/veryfront/veryfront-code/pull/3641) fixed for CSRF: **a gate demands a credential that a legitimate caller structurally cannot hold, and is positioned to run before that caller's handler.**

## The hole

`AuthHandler` (`src/security/http/auth.ts`) declares `priority: 0` and `patterns: []`. The registry never consults `metadata.patterns` — `registry.ts` calls every handler in priority order — so a handler must self-limit, and `AuthHandler.handle` had no path check. It ran on every request, ahead of `CsrfHandler` (priority 5) and `ProjectRunExecuteHandler`, exempting only `OPTIONS` and `isCspReportRequest`.

The platform cannot satisfy it:

- The control plane sends `Authorization: Bearer <platform-minted per-run service JWT>` plus the JWS envelope header (`veryfront-api` `runtime-project-run-client.ts:120-158`). `checkBasicAuth` can never match a `Bearer` header; `checkBearerAuth` compares it against a secret the **project** authored.
- So any project with `security.auth` — or `VERYFRONT_BASIC_USER` / `VERYFRONT_BEARER_TOKEN` — answered its own `task:release-asset-build` dispatch with **401**.
- 401 is not retryable (the client retries only on `>=500` or `429`), so the run died before the manifest row existed, the state stayed `missing`, and `veryfront deploy` failed at its 120s deadline reporting `Release assets were not ready within 120s (last state: missing)` — naming neither auth nor config.

**Second surface:** `POST /api/control-plane/agents/list` fails the same way and worse — `runtime-agent-client.ts:253-269` sends **no `Authorization` header at all**, so Studio's agent listing 401s for any project with `security.auth` set.

## The fix

Reuses the predicate #3641 introduced, in `AuthHandler.handle` alongside the OPTIONS and CSP-report exemptions:

```ts
if (isSignedControlPlaneDispatch(req)) return Promise.resolve(this.continue());
```

`isSignedControlPlaneDispatch` returns true only when **both** hold:

1. `isControlPlaneSurfaceRoute` matches one of five anchored method/path shapes, and
2. the request carries `x-veryfront-control-plane-jws`.

Not a weakening: it exempts nothing that is not authenticated more strongly downstream. Every one of those five surfaces calls `verifyControlPlaneRequest`, and `verifyControlPlaneJws` binds the Ed25519 signature to issuer, audience, project id, request method, request path and body hash, with expiry/skew bounds.

**Why not a prefix match.** `/api/control-plane/` is reserved but not exclusively routed. The run handlers register only the prefix and `return this.continue()` when their anchored regex misses; `ApiHandlerWrapper` declares no patterns and sits after them, so `/api/control-plane/checkout` falls through to project code. Anything keyed on `startsWith`/`includes` would ship a bypass a project could trigger by choosing a path.

**Why ahead of `resolveAuth`.** Same placement as the CSP-report exemption. An unresolvable auth config still fails closed for every browser request, but a project's config typo must not brick the platform's own dispatch, which authenticates itself downstream. A test pins both halves of that with one config.

## Tests (written first, confirmed red)

Red, before the production change:

```
AuthHandler signed control-plane dispatch ...
  passes every registered surface through for every configured auth shape ... FAILED
  passes a registered surface through for env-configured auth ... FAILED
  still challenges a path that merely starts alike ... ok
  still challenges a project route inside the control-plane namespace ... ok
  still challenges a registered surface with no signature header ... ok
  still challenges a registered surface with an empty signature header ... ok
  dispatches a signed surface even when the auth config is unresolvable ... FAILED

error: AssertionError: Values are not equal.
    [ "POST", "/api/control-plane/agents/list", -401 / +"continue" ]
```

```
release assets: control-plane build dispatch ...
  builds a manifest when the project puts the site behind basic auth ... FAILED
  builds a manifest when the project puts the site behind bearer auth ... FAILED
  builds a manifest when the project enables both gates at once ... FAILED

error: AssertionError: release asset build never started; runtime answered 401: Unauthorized
    -false / +true
```

```
internal-agents-list.handler auth gate ...
  lists agents for a signed dispatch behind basic auth ... FAILED   (-401 / +200)
  lists agents for a signed dispatch behind bearer auth ... FAILED   (-401 / +200)
  still challenges the same surface without the signature header ... ok
  still challenges a project route inside the control-plane namespace ... ok
```

Coverage:

- **`src/security/http/auth.test.ts`** — all five surfaces pass for basic, bearer and env-configured auth. **Adversarial:** a look-alike path (`/api/control-plane-mirror/...`), seven project routes inside the namespace *with* the signature header attached, every registered surface with no signature header, and a surface with an empty signature header — all still 401.
- **`src/release-assets/build-dispatch-security.test.ts`** — drives the real chain (`AuthHandler` → `CsrfHandler` → `ProjectRunExecuteHandler`) with a genuine signed Ed25519 envelope and asserts the release asset build executor is reached, under basic auth, bearer auth, and both gates at once. The four pre-existing CSRF cases are unchanged and still pass with `AuthHandler` added to the chain.
- **`src/server/handlers/request/internal-agents-list.handler.test.ts`** — real chain (`AuthHandler` → `InternalAgentsListHandler`) for the second surface, sending exactly what `runtime-agent-client` sends (no `Authorization` header).

No existing test was weakened or deleted.

## Verification

- `deno test src/security/ src/release-assets/ src/channels/ src/server/handlers/` (repo test flags): `ok | 216 passed (2435 steps) | 0 failed`
- `deno task lint:module-boundaries` — exit 0; the import mirrors the one `csrf-handler.ts` already makes. (The "debt decreased by 1" note it prints is pre-existing on clean `origin/main`.)
- `deno lint` / `deno fmt --check` on the changed files, `deno check src/security/http/auth.ts`
- `deno task docs` regenerated the one shifted line pin in `docs/api-reference/veryfront/security.md`

Deno 2.7.7 throughout.

## Summary by CodeRabbit

* **Security Improvements**
  * Improved handling of verified control-plane requests so authorized dispatches proceed without unnecessary project authentication prompts.
  * Added coverage for Basic and Bearer authentication, CSRF protection, signed requests, and invalid or unregistered routes.
  * Unauthorized, unsigned, or unmatched requests continue to receive appropriate protection.

* **Documentation**
  * Corrected the `AuthHandler` source reference in the API documentation.

### DIFF
--- a/docs/api-reference/veryfront/security.md
+++ b/docs/api-reference/veryfront/security.md
@@ -94,7 +94,7 @@ applySecurityHeaders(response.headers, false, generateNonce(), null);
 
 | Name                   | Description | Source                                                                                                     |
 | ---------------------- | ----------- | ---------------------------------------------------------------------------------------------------------- |
-| `AuthHandler`          |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/security/http/auth.ts#L156)             |
+| `AuthHandler`          |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/security/http/auth.ts#L157)             |
 | `BaseHandler`          |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/security/http/base-handler.ts#L45)      |
 | `CsrfHandler`          |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/security/http/csrf/csrf-handler.ts#L57) |
 | `ResponseBuilder`      |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/security/http/response/builder.ts#L9)   |
--- a/src/release-assets/build-dispatch-security.test.ts
+++ b/src/release-assets/build-dispatch-security.test.ts
@@ -18,6 +18,13 @@
  * Nothing downstream can see it: the run fails before the manifest row is
  * created, and the deploy timeout names neither CSRF nor config.
  *
+ * `security.auth` is the same failure with a different gate. `AuthHandler` runs
+ * at priority 0 — ahead of `CsrfHandler` — and also matches every path, and the
+ * credential it demands is one the platform structurally cannot hold: the
+ * control plane sends a per-run service `Bearer` JWT the Basic branch can never
+ * match and the Bearer branch compares against a project-authored secret. Both
+ * gates are driven here against the real handler chain.
+ *
  * @module release-assets/build-dispatch-security.test
  */
 
@@ -25,9 +32,11 @@ import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import { RouteRegistry } from "#veryfront/routing/registry/index.ts";
+import { AuthHandler } from "#veryfront/security/http/auth.ts";
 import { CsrfHandler } from "#veryfront/security/http/csrf/csrf-handler.ts";
 import { deriveSecurityContext } from "#veryfront/security/http/config.ts";
 import type { VeryfrontConfig } from "#veryfront/config";
+import type { SecurityConfig } from "#veryfront/types";
 import {
   ProjectRunExecuteHandler,
   type ProjectRunExecuteHandlerDeps,
@@ -57,10 +66,12 @@ interface DispatchOutcome {
  */
 async function dispatchReleaseAssetBuild(
   csrf: CsrfSetting | undefined,
+  auth?: SecurityConfig["auth"],
 ): Promise<DispatchOutcome> {
-  const config = {
-    security: csrf === undefined ? {} : { csrf },
-  } as VeryfrontConfig;
+  const security: Record<string, unknown> = {};
+  if (csrf !== undefined) security.csrf = csrf;
+  if (auth !== undefined) security.auth = auth;
+  const config = { security } as VeryfrontConfig;
   const { securityConfig } = deriveSecurityContext(config, {
     // The task rail dispatches to the project's main-branch runtime, which
     // resolves as a preview environment. Production defaults are therefore off
@@ -112,7 +123,11 @@ async function dispatchReleaseAssetBuild(
   ctx.securityConfig = securityConfig;
 
   const registry = new RouteRegistry();
-  registry.registerAll([new CsrfHandler(), new ProjectRunExecuteHandler(deps)]);
+  registry.registerAll([
+    new AuthHandler(),
+    new CsrfHandler(),
+    new ProjectRunExecuteHandler(deps),
+  ]);
 
   const response = await registry.execute(request, ctx);
   return {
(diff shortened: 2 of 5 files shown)


kojiwakayama:
## Pushed: this branch and #3646 would have merged into something broken

Both this PR and #3646 widen `dispatchReleaseAssetBuild` in `src/release-assets/build-dispatch-security.test.ts`, and both grew it as a **positional** parameter — `auth?` here, `options` there.

Merging them produces a two-line conflict in the parameter list and auto-merges everything else. That is the trap: the conflict looks trivial, so the natural resolution is to keep both parameters. It compiles. And then:

```
  builds a manifest when the project's own middleware gates every request ... ok
  keeps project middleware in front of an unsigned request to the same path ... FAILED

    [Diff] Actual / Expected
    -   200
    +   401
```

The `ok` on the first line is the real damage. `dispatchReleaseAssetBuild(undefined, { projectMiddleware: [...] })` binds its argument to `auth`, so **no middleware is installed at all** — the test that proves the middleware bypass works passes without ever exercising it. The second test fails only because it also asserts the negative direction.

### Fix

One named options bag, identical on both branches:

```ts
interface DispatchOptions {
  readonly auth?: SecurityConfig["auth"];
  readonly projectMiddleware?: MiddlewareFunction[];
  readonly unsigned?: boolean;
}

async function dispatchReleaseAssetBuild(
  csrf: CsrfSetting | undefined,
  options: DispatchOptions = {},
): Promise<DispatchOutcome>
```

A named field cannot mis-merge the way a positional slot can. The harness also now assembles all three gates in the order the runtime does — project middleware outermost, then the security handlers, then the run executor — so a fix for one gate is exercised with the others standing. Each branch keeps only its own tests.

### Verified by merging, not by reasoning

- merge in both orders → **clean, no conflict**
- merged tree: 9 steps pass
- this branch alone: green; revert its production fix → red for the right reason
- #3646 alone: green; revert its production fix → red for the right reason

kojiwakayama:
## `ci (lint)` was red for one line — and it dragged three innocent files in with it

The 7m18s `ci (lint)` failure reported **four** failing entry points. Only one of them was mine:

```
TS2345 [ERROR]: Argument of type '{ VERYFRONT_BASIC_USER: string; ... VERYFRONT_BEARER_TOKEN?: undefined; }
  | { VERYFRONT_BEARER_TOKEN: string; VERYFRONT_BASIC_USER?: undefined; ... }'
  is not assignable to parameter of type 'Record<string, string>'.
    at src/security/http/auth.test.ts:547:22
```

The inline array literal widened to a union in which each member carried the other's keys as `?: undefined`, so neither member was assignable to `createEnvCtx`'s `Record<string, string>`. Fixed by hoisting it to an annotated `Record<string, string>[]`, which contextually types both members and drops the synthetic optional keys. No behaviour change.

### The other three were collateral, not rot

The run also blamed `src/react/components/chat/chat/hooks/attachment-csrf.test.tsx`, `src/transforms/esm/http-cache.test.ts`, and `src/workflow/react/use-workflow-start.test.tsx` — all untouched by this PR, none in the 51-entry baseline, all failing on the same `RequestInit` union (`Property 'headers'/'method'/'signal' does not exist on type 'global.RequestInit | RequestInit | (RequestInit & { client?: HttpClient })'`).

They are not new rot. Per `scripts/lint/check-test-typecheck-baseline.ts`, the expected-clean set runs as **one repository-wide check**, and is *only* recursively split when that check fails. Those three files' errors are visible in a small program but not in the repo-wide one — so they surface only once something else has already broken the happy path. My `TS2345` was that something.

The timings show it: `ci (lint)` takes ~2m when the repo-wide check passes and 7m18s when it splits.

- #3646 — same base, same three files, `ci (lint)` **pass** in 2m42s
- #3647 — `ci (lint)` **pass** in 1m59s
- #3640 — the PR that introduced `attachment-csrf.test.tsx`, `ci (lint)` **pass** in 2m39s

### Verified

With only the one-line fix applied, the real gate goes green on the full tree:

```
$ deno task lint:test-typecheck
Test typecheck baseline holds: 51 grandfathered files, 0 new.
```

No baseline entry was added, and the three innocent files were left alone. Worth knowing that this gate reports latent errors as if they were yours whenever anything else fails first.

### The other five red checks were infrastructure

`coverage shard 8/8`, `tests (bun)`, and `tests (rsc browser e2e)` all died in `./.github/actions/setup-deno` before running a line of code:

```
curl: (56) Connection died, tried 5 times before giving up
Failed to download Deno archive for deno-x86_64-unknown-linux-gnu.zip     # shard 8/8
Failed to download Deno archive checksum for ...linux-gnu.zip             # bun
Failed to download checksums manifest for v2.7.7 (HTTP 000)               # rsc browser e2e
```

`coverage gate` and `tests (unit)` are aggregators that only mirror the shard result (`COVERAGE_SHARDS_RESULT: failure`). The new run supersedes all five.

PR #3673, opened 2026-08-13
fix(security): read the validated body from the pages-route API context

### LINKED ISSUE #3666: createValidatedHandler POST body read fails: request.body?.getReader is not a function (validation never runs) @0.1.1231
## Summary

`createValidatedHandler` cannot read a POST body on `veryfront@0.1.1231`. Every POST — valid or invalid payload — returns **400** with:

```json
{"error":"Invalid JSON in request body","details":{"error":"request.body?.getReader is not a function"}}
```

The body-reader dereferences `request.body.getReader()`, but the `request` reaching the handler has no `body` stream (`request.body` is undefined / not a `ReadableStream`), so **validation never runs** and no valid POST can ever succeed.

## Environment

| | |
|---|---|
| veryfront | **0.1.1231** |
| dev runtime | Deno 2.7.7 |
| OS | macOS (Darwin 25.5, arm64) |

## Reproduction

Repro repo: https://github.com/mattboon/veryfront-router-testing — `default-config/pages/api/validated.ts` (dev port 3010).

```ts
// pages/api/validated.ts
import { createValidatedHandler } from "veryfront";
import { defineSchema, lazySchema } from "veryfront/schemas";

const bodySchema = lazySchema(
  defineSchema((v) => v.object({
    name: v.string().min(1),
    count: v.number().int().nonnegative(),
  })),
);

export const POST = createValidatedHandler(
  { body: bodySchema },
  (_request, validated) => Response.json({ echoed: validated.body }),
);
```

```bash
cd default-config && npm run dev        # http://localhost:3010
curl -s -X POST -H 'content-type: application/json' \
  -d '{"name":"Ada","count":3}' http://127.0.0.1:3010/api/validated
```

## Expected

`200` with `{"echoed":{"name":"Ada","count":3}}` (a valid body passes validation).

## Actual

```
HTTP 400
{"error":"Invalid JSON in request body","details":{"error":"request.body?.getReader is not a function"}}
```

Identical for an invalid body (`{"name":"","count":-1}`) — i.e. the request never reaches schema validation; it fails earlier in the body reader.

## Analysis

The validated-handler body reader (`readBodyWithLimit` / the `createValidatedHandler` path in `security/input-validation`) consumes `request.body` as a `ReadableStream` via `.getReader()`. In this runtime the `Request` passed to the handler doesn't expose a `body` stream, so `request.body?.getReader` is `undefined` and the call throws before validation. `ctx.body()` on a plain API route reads the body fine (same repo, `pages/api/echo.ts`), so a POST body IS available on the request — the validated-handler path just reads it a different, broken way. Likely fixes: read the body the same way `ctx.body()` does (cached/clone-safe), or fall back to `await request.json()` / `request.text()` when `request.body` is not a stream.

mattboon:
## Bug

Every POST to a `createValidatedHandler` mounted as a **pages** API route returns 400, valid payload or not:

```json
{"error":"Invalid JSON in request body","details":{"error":"request.body?.getReader is not a function"}}
```

Validation never runs — the body read throws first.

## Root cause

The two routers invoke API handlers with **different arguments**:

- **App Router** — a real `Request` (`app-router-handler.ts:79`, via `createApplicationRequest`).
- **Pages Router** — the `APIContext` it builds (`routing/api/route-executor.ts:1365` → `createContext(...)`), **not** a `Request`.

`createValidatedHandler` treats its argument as a `Request` and reads `request.body?.getReader()`. On the pages `APIContext`, `.body` is a **reader function** (`ctx.body()`), so `.getReader` is `undefined` and the optional-chain doesn't short-circuit (body is truthy) → `request.body?.getReader is not a function`. So on pages routes no POST body could ever be validated (#3666). App-router routes worked because they genuinely receive a `Request`.

## Fix

Resolve the underlying `Request` from **either** a `Request` **or** an API context (`.request` / `.req`) before validating, and pass that same `Request` to the handler. The App Router path (already a real `Request`) is unchanged — `resolveValidatedRequest` returns it identically.

## Tests (red→green)

Added `describe("invoked with a pages-router API context (issue #3666)")` in `handler.test.ts`, building the object the pages executor actually passes (via the real `createContext`):

- valid POST → **200**, echoes the validated body;
- invalid POST → **400** *schema* error, with a regression guard asserting the response never contains `getReader` (i.e. validation ran, the body-reader didn't crash);
- the handler receives the resolved `Request`, not the context.

All three fail on `main` (`Invalid JSON … getReader is not a function`) and pass with the fix. Full `src/security/input-validation/` suite green; `deno check` + `deno lint` + `deno fmt --check` clean.

Closes #3666.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validated request handling across supported API contexts.
  * Prevented failures when accessing request bodies through wrapped or unavailable request properties.
  * Added clearer errors for unsupported input types.
  * Ensured invalid request bodies return appropriate 400 validation responses.
  * Wrapped handlers now receive the resolved request consistently.

* **Documentation**
  * Updated API reference links to point to the correct source location for validated handlers.


kojiwakayama:
Reviewed and resolved the conflict so this can land.

The only conflicting file was `docs/api-reference/veryfront/index.md`, and only on generated line-number pins — `index.client.md` and `security.md` auto-merged. I took main's copy and re-ran `generate-api-reference.ts`, which reports the tree current at 44 files. No source change from me.

Verification of the fix itself, done against your claims rather than the description:
- Reproduced the break on `origin/main` through the real `createContext`: `400 {"error":"Invalid JSON in request body","details":{"error":"request.body?.getReader is not a function"}}`, matching #3666 verbatim. On this branch the same request returns `200`.
- Your three new tests fail on main and pass here, and they are not gated behind a runtime check, so they run in the default Deno lane rather than silently skipping.
- `src/security/input-validation/` is green after the merge: 7 passed, 101 steps.

Two things for you, neither blocking:

The blast radius is wider than the description says. A validated handler with **no** body schema, and even a plain GET, also returned 400 on pages routes (`validateUnparsedBody` calls `request.clone()`, which the context has none of). Your fix covers those too, so the PR undersells itself.

The bug *class* survives elsewhere. `src/chat/upload-handler.ts:277` does `request.body.getReader()` — the same line as the bug fixed here — and `src/embedding/upload-handler.ts:266` does `await request.formData()`. Both are documented public exports, so mounting either under `pages/api` produces the identical 500. Worth a follow-up rather than widening this PR.

PR #3702, opened 2026-08-14
release: v0.1.1236

kojiwakayama:
## Release v0.1.1236

Ships the two fixes for the staging preview outage.

**#3698 — bind the esbuild binary path outside project scope.** esbuild resolves its binary once, when its module evaluates. The adapter imported it lazily, so that read landed inside a project env scope on the first transform, saw `undefined`, and no esbuild service ever started — every TSX transform failed for the life of the process. Importing at startup binds the path while the host environment is still on `process.env`, with project env isolation untouched.

**#3697 — keep the failure that caused the ownership error.** The ownership latch was created without a cause and `??=` then discarded the real error, so every genuine esbuild failure surfaced as a lifecycle message. That masking is why the above took a day to find.

## Release hygiene

- Exactly the 3 required files, all three agreeing on `0.1.1236`.
- `deno.json` diff is the version line only — the release script's unrelated reformatting of two `dependencies` arrays was reverted.
- `deno task generate` run and amended in; the script alone produced only 2 files.
- No local tag — CI creates it.

## After merge

This does not complete the fix. Staging needs the release published, a server artifact rebuilt on it, and a deploy, before preview recovers. I'll verify against the deployed pods and logs rather than assuming the chain ran, then dispatch a staging Remote E2E Health run before merging veryfront-e2e#139.

## Summary by CodeRabbit

* **Chores**
  * Updated the application version to **0.1.1236**.


kojiwakayama:
Closing as a duplicate of #3701, which was cut first (10:15Z vs 10:26Z).

I checked before deferring rather than assuming: #3701 is correctly formed — exactly the 3 required files, all three agreeing on `0.1.1236`, and its `deno.json` diff is the version line only.

Worth stating clearly, because the branch histories differ but the *released content* does not: **both #3697 and #3698 are already on `main`**, so whichever release PR merges, the published 0.1.1236 contains both fixes. #3701 not having c161dc6a4 as a branch ancestor changes nothing about what ships.

No reason to prefer mine. Verification of the staging recovery continues against #3701.

PR #3705, opened 2026-08-14
fix(dev): print the address the dev server bound, not `localhost`

kwakayama:
Follow-up to #3704, which fixed the port probe. This fixes the *host* half of the same bug. Independent of #3704 — different files, either order merges.

## Problem

`DevServer` binds `this.options.bindAddress ?? LOCALHOST.IPV4` (`server.ts:294`). The CLI printed `http://localhost:${boundPort}` (`command.ts:342`).

Those are different hosts. `localhost` resolves to `::1` first on a dual-stack machine, so the URL the CLI printed named an address the server was not listening on. Browsers usually paper over it via Happy Eyeballs — but when some other process holds `[::1]:port`, it answers instead, which is exactly the failure reported against #3704: `veryfront dev` printed `Ready — http://localhost:3000` and that URL served an unrelated app's 500.

## Fix

`serverDisplayUrl(bindAddress, port)` builds the URL from the address actually bound:

| Bound | Shown |
|---|---|
| `127.0.0.1` | `http://127.0.0.1:3000` |
| `::1` | `http://[::1]:3000` (bracketed — a bare literal is not a valid authority) |
| `0.0.0.0` / `::` | loopback — a wildcard is a bind target, not somewhere to browse |
| `192.168.1.5` | unchanged |

`DevServer` now exposes `bindAddress` as one source of truth; the two places that re-derived it inline route through it, so they cannot drift. `DevCommandResult` carries `bindAddress` alongside `port`, mirroring how `port` is already threaded to embedded callers.

## Same defect in the demo path

`cli/commands/demo/dev-step.ts` had it too. Its doc comment already promised the printed and opened URLs "come from the port the server actually bound … or the viewer is sent to whatever process caused the collision" — correct about the port, silent about the host. Same helper now.

## MCP origin allowlist

`cli/mcp/server.ts` gates HTTP origins and its comment said "`localhost` is the hostname the CLI prints". The set already contained `127.0.0.1` and `[::1]`, so the new URL passes unchanged — only the comment moved. Checked before changing the printed URL, not after.

## Testing

Test-first throughout. New `serverDisplayUrl` tests were stubbed against the *old* behavior so the assertions did the failing, not a module-not-found error:

```
names the address the server actually bound, not `localhost` ... FAILED
brackets an IPv6 address so the URL is valid ... FAILED
shows a loopback address when the server bound a wildcard ... FAILED
keeps a specific non-loopback address ... FAILED
```

Demo path, before the change: 3 URL tests RED, the 2 unrelated lifecycle tests still green.

After: `ok | 11 passed (112 steps) | 0 failed` across `cli/commands/dev/` + the demo step, green on 3 consecutive runs. `deno lint` (56 files) and `deno check` clean.

## Local test-run noise (not from this change, flagged for honesty)

The full `cli/` + `src/server/dev-server/` parallel run is not clean on my machine, in a way I verified is unrelated:

- **2 constant failures** in `cli/commands/push/command.test.ts` (`.vfignore` is covered by a `.gitignore`, so `git add` refuses it). Reproduced identically with my changes stashed on clean `origin/main`.
- **1 varying failure** — `dev-output.integration.test.ts` in one run, `start/command` MCP boundary in another, neither in a third. Different test each run under full-tree parallel load; `dev-output` passes in isolation and `cli/commands/dev/` is 3/3 green in parallel.

CI is the arbiter here — flagging it rather than presenting a clean local run I did not get.

## Not in scope

`command.ts:362` still prints `http://localhost:${mcpPort}/mcp`. The MCP server calls `serve(handler, { port })` with **no** hostname, so it takes the adapter default rather than `LOCALHOST.IPV4` — a genuinely different situation that deserves its own change, not a blind rename. The `onListen` debug log in `server.ts` also still uses `buildLocalhostUrl`; `serverDisplayUrl` lives under `cli/`, and `src/` importing from `cli/` would invert the layering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Development server URLs now reflect the actual bound address instead of always displaying `localhost`.
  * Added proper URL formatting for IPv6 addresses, including required brackets.
  * Wildcard bind addresses now resolve to usable loopback URLs for browser access.
  * Server startup messages consistently display the address where the server is listening.

### DIFF
--- a/cli/commands/demo/dev-step.test.ts
+++ b/cli/commands/demo/dev-step.test.ts
@@ -5,7 +5,10 @@ import type { DevCommandResult } from "../dev/index.ts";
 import { runDemoDevStep } from "./dev-step.ts";
 
 /** A dev server that fell forward off the requested 3000 onto `port`. */
-function devServerOn(port: number): { result: DevCommandResult; stopped: () => boolean } {
+function devServerOn(
+  port: number,
+  bindAddress = "127.0.0.1",
+): { result: DevCommandResult; stopped: () => boolean } {
   let stopped = false;
   return {
     stopped: () => stopped,
@@ -13,6 +16,7 @@ function devServerOn(port: number): { result: DevCommandResult; stopped: () => b
       ready: Promise.resolve(),
       done: Promise.resolve(),
       port,
+      bindAddress,
       stop: () => {
         stopped = true;
         return Promise.resolve();
@@ -37,11 +41,11 @@ describe("cli/commands/demo/dev-step", () => {
       log: (line) => lines.push(line),
     });
 
-    assertEquals(opened, ["http://localhost:3007"]);
+    assertEquals(opened, ["http://127.0.0.1:3007"]);
 
-    const shown = lines.find((line) => line.includes("http://localhost:"));
+    const shown = lines.find((line) => line.includes("http://127.0.0.1:"));
     assert(shown !== undefined, "the demo must print the dev server address");
-    assertStringIncludes(shown, "http://localhost:3007/");
+    assertStringIncludes(shown, "http://127.0.0.1:3007/");
 
     // The requested port belongs to whatever caused the collision.
     assertEquals(lines.some((line) => line.includes(":3000")), false);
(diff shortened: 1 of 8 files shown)


kwakayama:
Closing this — superseded by #3704, and the premise does not hold up.

Printing `http://localhost:PORT` while binding `127.0.0.1` is the conventional dev-server pattern, not a defect. It works because browsers try both loopback families. It only became visible here because a second process was squatting on `[::1]:3000`, and #3704 (merged) fixes that by making the port probe see wildcard and IPv6 holders, so `veryfront dev` falls forward instead of silently co-binding.

Against that, this PR carried real costs:

- **Confirmed OAuth regression.** `cli/commands/generate/integration-generator.ts:446` hardcodes `http://localhost:3000` as the PKCE redirect URI. Opening the app at `127.0.0.1` sets the host-only verifier cookie on `127.0.0.1`; the provider then redirects to `localhost`, which never receives it, and the callback fails with `missing_pkce_verifier`. Thanks to the Codex review for catching this.
- **Docs and copy contradiction.** `cli/commands/demo/steps.ts:44`, the getting-started docs, templates, and examples all still say `localhost:3000`.

On CodeRabbit's `DevCommandResult` compatibility finding: that one does not apply. No entry in the `deno.json` exports map reaches `cli/commands/**` (`./cli` maps to `cli/main.ts`, which has zero `export` statements), so there is no supported path for an external caller to import or construct that type. Recording it here rather than silently weakening the field to optional.

One item from this branch is worth keeping and will get its own PR: `docs/getting-started/create-project.md` states "`localhost` resolves to `127.0.0.1` on every machine without a DNS lookup." That is false on any dual-stack host, and it is precisely the misconception that made this bug hard to see.

PR #3707, opened 2026-08-14
feat(release): publish checksums and verify them before installing

kojiwakayama:
## Why

Closes the security finding raised on #3694, which I could only partly address there because the remedy is a pipeline change rather than a docs edit.

The standalone installers piped a network-fetched script to a shell and then wrote whatever they downloaded **straight to the final install path**. Nothing verified the binary, and nothing could — no release published a checksum manifest.

The values already existed: the release job computes SHA-256 for each binary, but only to fill in the Homebrew formula. So Homebrew users got verified installs and `curl | sh` users did not.

## What changed

**Publish the manifest.** The stable release now uploads a `SHA256SUMS` asset covering the binaries *and* both install scripts, so the script a user piped can also be checked by hand.

**Verify before installing.** Both installers stage the download in a temporary directory, verify it against the manifest, and only then move it into place. A mismatch, an unlisted asset, or a missing manifest installs nothing.

This also fixes a second problem that existed independently of checksums: the binary was previously written directly to its final path, so a truncated download left a broken executable installed.

**Fails closed.** Releases from before the manifest existed have none, so pinning to one requires `VERYFRONT_INSTALL_SKIP_CHECKSUM=1`, set deliberately.

## Validation

I extracted the shipped shell functions from `install.sh` and drove them against a local manifest — the real code, not a reimplementation:

| case | result |
|---|---|
| checksum matches | exit 0, installs |
| checksum mismatch | **exit 1** — "checksum mismatch", nothing installed |
| asset not listed in manifest | **exit 1** — "not listed in SHA256SUMS" |
| no manifest published | **exit 1** — "no SHA256SUMS published" |
| `VERYFRONT_INSTALL_SKIP_CHECKSUM=1` | exit 0, with an explicit notice |

`sh -n scripts/install.sh` passes, the workflow YAML parses, and I confirmed the manifest format the CI `sed` produces (`<hash>  veryfront-macos-arm64`) is exactly what the installer's `awk` matches.

**Not verified locally: `install.ps1`.** There is no `pwsh` on this machine, so the PowerShell path follows the same shape as the shell one but its parse and behaviour are unexercised. Worth a Windows reviewer's eye before merge.

## Sequencing with #3694

#3694's docs currently state that nothing in this path is checksum-verified — true when written, false once this merges. I'll update that text on #3694 so the two land consistently rather than leaving a window where the docs understate the guarantee.

## Summary by CodeRabbit

* **Security Enhancements**
  * Installers now verify downloaded binaries against published SHA-256 checksums before installation.
  * Installations stop safely if checksums are unavailable, missing, or incorrect.
  * Downloads are staged temporarily and cleaned up after completion.
  * Added an optional environment-variable bypass for checksum verification.
  * Releases now include a publicly available `SHA256SUMS` file for uploaded assets.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -965,6 +965,19 @@ jobs:
           echo "Available binaries: $BINARIES"
           echo "Available SBOMs: $SBOMS"
 
+          # Publish a checksum manifest alongside the assets. install.sh verifies a
+          # downloaded binary against this before moving it into place, so the
+          # manifest must name assets exactly as they are uploaded -- basenames,
+          # not the artifact paths they were downloaded to.
+          SUMS_FILE="SHA256SUMS"
+          : > "$SUMS_FILE"
+          for f in $BINARIES scripts/install.sh scripts/install.ps1; do
+            [ -f "$f" ] || continue
+            sha256sum "$f" | sed "s#  .*/#  #" >> "$SUMS_FILE"
+          done
+          echo "Checksum manifest:"
+          cat "$SUMS_FILE"
+
           # Clean up rc pre-releases (created by the prerelease job on main pushes)
           gh release list --repo veryfront/veryfront --json tagName -q ".[].tagName" \
             | grep "^v${VERSION}-rc\." \
@@ -984,6 +997,7 @@ jobs:
           npm install -g veryfront
           ```' \
             $BINARIES \
+            "$SUMS_FILE" \
             scripts/install.sh \
             scripts/install.ps1 \
             $SBOMS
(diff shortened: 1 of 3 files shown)


kojiwakayama:
Update on the `install.ps1` verification gap I flagged.

I got further than unverified: using the official PowerShell container, the script **parses cleanly** (664 tokens, zero parse errors), which rules out syntax errors — the main risk in code that has never been executed.

I could not exercise the logic. The x86 PowerShell image crashes under qemu emulation on this arm64 machine (`uncaught target signal 6`), so `Get-FileHash` and the comparison path remain unrun.

So the state is:

| | `install.sh` | `install.ps1` |
|---|---|---|
| parses | yes | **yes** |
| match installs | verified | not run |
| mismatch / unlisted / missing manifest reject | verified | not run |
| escape hatch | verified | not run |

The PowerShell path mirrors the shell one and reads correctly, but a Windows reviewer running it once against a real release would close this properly. Flagging rather than quietly merging it as equivalent.

Issue #318, opened 2026-02-11
Starter template projects: routing and styling issues 


mattboon:
## Summary

Created 7 starter template projects from `cli/templates/files/` to test the template deployment workflow. Several issues were identified.

---

### Template: chat ❌

**Slug:** chat-starter-template  
**Project ID:** `e02e9215-faee-4a34-88c2-16c3872cb6d5`  
**Studio URL:** https://veryfront.com/projects/chat-starter-template  
**Production URL:** ❌ https://chat-starter-template.veryfront.com  
**Preview URL:** ❌ https://chat-starter-template.preview.veryfront.com  

**File Tree:**
```
agents/assistant.ts
app/api/chat/route.ts
app/layout.tsx
app/page.tsx
tools/calculator.ts
tsconfig.json
```

**Behaviour:**
- Production: No styles applied
- Preview: No styles applied

---

### Template: rag ❌

**Slug:** rag-starter-template  
**Project ID:** `0a9d7949-4217-4a7b-8aba-ca7f2b6eb386`  
**Studio URL:** https://veryfront.com/projects/rag-starter-template  
**Production URL:** ❌ https://rag-starter-template.veryfront.com  
**Preview URL:** ❌ https://rag-starter-template.preview.veryfront.com  

**File Tree:**
```
agents/qa.ts
app/api/chat/route.ts
app/layout.tsx
app/page.tsx
content/architecture.md
content/getting-started.md
tools/search-docs.ts
tsconfig.json
```

**Behaviour:**
- Production: No styles applied
- Preview: No styles applied

---

### Template: multi-agent ❌

**Slug:** multi-agent-starter-template  
**Project ID:** `30b03043-bc82-40b5-aec3-692577f4d25b`  
**Studio URL:** https://veryfront.com/projects/multi-agent-starter-template  
**Production URL:** ❌ https://multi-agent-starter-template.veryfront.com  
**Preview URL:** ❌ https://multi-agent-starter-template.preview.veryfront.com  

**File Tree:**
```
agents/orchestrator.ts
agents/researcher.ts
agents/writer.ts
app/api/chat/route.ts
app/layout.tsx
app/page.tsx
tools/web-search.ts
tsconfig.json
```

**Behaviour:**
- Production: No styles applied
- Preview: No styles applied

---

### Template: workflow ❌

**Slug:** workflow-starter-template  
**Project ID:** `e870b898-7d80-4108-8390-bd4d5ed7a34c`  
[...]

PR #3717, opened 2026-08-14
fix(server): diagnosable masked-env refusal and negative caching for env fetch failures

kojiwakayama:
Fixes [VERYFRONT-SERVER-T](https://veryfront.sentry.io/issues/VERYFRONT-SERVER-T) (223 events / 9 users, escalating until remediated).

## Root cause
The incident itself was a deploy misconfiguration (veryfront-server #316 removed `rendererSecrets.VERYFRONT_API_INTERNAL_USER/PASS`; #318 restored them). But the framework made it needlessly undiagnosable and noisy:
- With no internal credentials, `fetchProjectEnvVars` parsed the management endpoint's **always-masked** authorization-probe response as data and threw the generic "Refusing masked environment variable response" (a 502 NETWORK_ERROR) — the actionable truth ("internal API credentials missing") was never said.
- `EnvironmentVariableCache` caches successes only, so every request re-fetched and re-threw: two upstream calls + one Sentry event per request for ~17 hours.

## Fix
- `parseEnvironmentResponse` now knows its response source (`management` | `internal`): a masked value from the management hop produces an explicit "internal credentials not configured" detail; a masked value from the internal hop keeps the refusal without copying tenant-defined key names into errors or logs.
- `EnvironmentVariableCache` gains short-TTL negative caching (`failureTtlMs`, default 5s): one upstream fetch per TTL window during failure storms, preserving the deliberate fail-closed semantics (never serve stale/empty env).
- `AgentStreamHandler` marks negative-cache replays and returns the same failure response while logging and reporting only the first failure in each TTL window.

## Testing (red-green TDD)
Extended `fetcher.test.ts` masked-value tests with message/source assertions and added a cache test proving a failed fetch is not re-issued within the TTL — all fail on pre-fix src (verified by adversarial revert-check), pass at HEAD.
A handler regression proves concurrent and cached identical 502 responses produce one upstream fetch, one error log, and one application-error report.

## Reviewer notes (from adversarial verification)
- Negative caching applies to **all** failure types at both production call sites, so a one-off transient error also blocks retries for up to 5s per scope — deliberate, documented in the option docstring.

### DIFF
--- a/src/server/handlers/request/agent-stream.handler.test.ts
+++ b/src/server/handlers/request/agent-stream.handler.test.ts
@@ -1,5 +1,5 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { INVALID_ARGUMENT, SERVICE_OVERLOADED } from "#veryfront/errors";
+import { INVALID_ARGUMENT, NETWORK_ERROR, SERVICE_OVERLOADED } from "#veryfront/errors";
 import {
   __registerLogRecordEmitter,
   __resetLogRecordEmitterForTests,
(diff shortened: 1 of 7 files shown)


kojiwakayama:
@codex review

PR #3718, opened 2026-08-14
fix(bundler): restart the esbuild service after an unexpected child death instead of poisoning the process

kojiwakayama:
Fixes [VERYFRONT-SERVER-V](https://veryfront.sentry.io/issues/VERYFRONT-SERVER-V) (160 events, staging, escalating).

## Root cause
The immediate staging trigger (esbuild's module evaluating inside a project env scope and losing `ESBUILD_BINARY_PATH`) was fixed on main by #3698/#3697. This PR fixes the amplifier those left in place: when the managed esbuild service child dies unexpectedly (OOM-kill, SIGKILL, or the #3698-class spawn failure), the adapter latched a permanent process-wide ownership error. Every subsequent transform in the pod failed until restart, so one child death poisoned the whole process for its lifetime.

## Fix
The bundler adapter now recovers with a fresh esbuild service after an unexpected managed child death, bounded by a restart budget (`MAX_SERVICE_RESTARTS = 3`) to preserve the original guard's protection against genuinely foreign service ownership. Operations in flight during a crash surface esbuild's raw transient error once; the next operation recovers.

The recovery path also preserves the module-wide ownership guardrails:

- Raw esbuild services that replace a killed managed child remain rejected. Recovery and later shutdown calls fail closed without touching the foreign service.
- Shutdown waits for operations that are already waiting on lost-service recovery, so `stop()` cannot complete and then leave that operation to spawn a new service afterward.
- Build contexts captured before a managed service crash refresh themselves after recovery from the saved mapped options, so callers can keep using the existing `BuildContext` wrapper.
- Managed child `exit` marks the service loss recoverable immediately, while `close` remains the completion signal, so operations launched in the exit-to-close window do not latch a permanent ownership error.
- Stale plugin `onDispose` callbacks run before recreating a crashed context, preserving exactly one active plugin generation during context recovery.
- Plugin generations are force-disposed when a crashed context is abandoned without rebuild, so abandoned esbuild services cannot strand plugin resources.
- Forced cleanup reports only errors from the plugin generation being disposed while preserving the module-wide async disposal error for shutdown, so an unrelated stale disposal failure cannot poison later crashed-context cleanup.
- Failed context refresh attempts dispose their attempted plugin generation and replace the abandoned mapped options with a fresh barrier, so a later successful retry can still await cleanup.
- Restart-budget exhaustion stays latched even when `stop()` observes the lost service after the dead child has already closed and been cleared from module state.
- `stop()` spends restart budget when it clears a managed crash, so crash plus stop loops cannot bypass the restart limit.

## Testing (red-green TDD)
Regression coverage was added for managed child crash recovery, foreign raw-service replacement, repeated foreign shutdown, shutdown during pre-admission recovery, captured `BuildContext` reuse after recovery, the managed child exit-to-close window, stale plugin disposal before context recreation, abandoned crashed context disposal, stale disposal errors from unrelated plugin barriers, context refresh failure followed by a successful retry, and restart-budget exhaustion when losses are cleared by `stop()`.

The new regressions fail on the pre-fix code: shutdown settles too early, stale context rebuild throws `The service is no longer running`, exit-to-close work latches the permanent module-wide ownership error, stale plugin disposal remains at 0 while a second setup runs, abandoned crashed contexts leave plugin disposal at 0, unrelated stale plugin disposal errors escape later crashed-context cleanup, failed refresh reuses an activated disposal barrier so later cleanup is not awaited, and crash plus stop loops let `stop()` resolve after the restart budget is exhausted.

Local verification at the pushed head (`0efec7681e42b635a0fb897357941bd89f7c3d92`):

- `deno test --preload=src/testing/preload.ts --no-check --allow-all extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts` -> 8 passed, 38 steps.
- `deno test --preload=src/testing/preload.ts --no-check --allow-all extensions/ext-bundler-esbuild/src` -> 12 passed, 46 steps.
- `deno fmt --check extensions/ext-bundler-esbuild/src/esbuild-bundler.ts extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts` -> passed.
- `deno lint extensions/ext-bundler-esbuild/src/esbuild-bundler.ts extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts` -> passed.
- `deno check extensions/ext-bundler-esbuild/src/esbuild-bundler.ts extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts` -> passed.
- `perl -CSD -ne 'print if /\\x{2014}|\\x{2013}/' extensions/ext-bundler-esbuild/src/esbuild-bundler.ts extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts` -> no matches.
- `git diff --check HEAD~1..HEAD` -> passed.
- Full pre-push hook at `0efec7681e42b635a0fb897357941bd89f7c3d92` -> passed: repo format, lint, typecheck, unit suite (3814 passed, 28499 steps, 0 failed, 1 ignored), cwd tests (10 passed, 193 steps), and cwd-exclusion tests (2 passed, 2 steps).

## Reviewer notes
- The restart budget never replenishes after successful recovery; 4 unexpected managed-child deaths over a long process lifetime re-latch the permanent error. A time-windowed budget would be more robust, but this PR keeps the guard simple and bounded.

### DIFF
--- a/extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts
+++ b/extensions/ext-bundler-esbuild/src/esbuild-bundler.test.ts
@@ -932,6 +932,48 @@ describe("EsbuildBundler service crash recovery", () => {
     }
   });
 
+  it("treats an operation launched after service exit but before close as recoverable", async () => {
+    const observation = observeEsbuildServices();
+    const { services } = observation;
+    const bundler = new EsbuildBundler();
+    const exitTransform = Promise.withResolvers<
+      Awaited<ReturnType<EsbuildBundler["transform"]>>
+    >();
+
+    try {
+      await bundler.transform({ code: "export const managed = true;", loader: "ts" });
+      assertEquals(services.length, 1);
+
+      const managedService = services[0]!;
+      managedService.child.once("exit", () => {
+        void bundler.transform({
+          code: "export const duringExit: number = 1;",
+          loader: "ts",
+        }).then(exitTransform.resolve, exitTransform.reject);
+      });
+      managedService.child.ref();
+      managedService.child.kill("SIGKILL");
+
+      const exitResult = await exitTransform.promise;
+      assertStringIncludes(exitResult.code, "duringExit = 1");
+      await managedService.close;
+
+      const later = await bundler.transform({
+        code: "export const later: number = 2;",
+        loader: "ts",
+      });
+      assertStringIncludes(later.code, "later = 2");
+    } finally {
+      await bundler.stop().catch(() => undefined);
+      __resetServiceRecoveryForTests();
+      try {
+        await bundler.stop();
+      } finally {
+        observation.restore();
+      }
+    }
+  });
+
   it("does not let stop complete before an operation waiting for recovery", async () => {
     const observation = observeEsbuildServices();
     const { services } = observation;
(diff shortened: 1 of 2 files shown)


kojiwakayama:
@codex review

Issue #3725, opened 2026-08-14
CLI-only extensions should not be runtime dependencies of the library package


kojiwakayama:
Found while dogfooding the documented developer journeys against published v0.1.1237. Related: #3714 documents the workaround this issue would remove.

## Symptom

`npm install veryfront` breaks a previously-passing typecheck in an existing project:

```
$ npx tsc --noEmit          # clean
$ npm install veryfront
$ npx tsc --noEmit
node_modules/@types/mdx/types.d.ts(23,38): error TS2503: Cannot find namespace 'JSX'.
node_modules/@types/mdx/types.d.ts(28,23): error TS2503: Cannot find namespace 'JSX'.
node_modules/@types/mdx/types.d.ts(47,28): error TS2503: Cannot find namespace 'JSX'.
node_modules/@types/mdx/types.d.ts(73,46): error TS2503: Cannot find namespace 'JSX'.
EXIT=2
```

The error names a dependency the developer never imported, and nothing connects it to Veryfront.

## Why it is invisible to us

Scaffolded projects set `skipLibCheck: true`, so **no new-project user ever sees this**. It only reaches developers evaluating Veryfront on a codebase they already have — which is why it has gone unnoticed.

## Root cause

```
veryfront@0.1.1237
└─ @veryfront/ext-content-mdx@0.1.1237
   └─ @mdx-js/mdx@3.1.1 → @types/mdx@2.0.14
```

`@types/mdx@2.0.14` references the **global** `JSX` namespace, which `@types/react@19` no longer declares (React 19 moved it to `React.JSX`). Anything under `node_modules/@types/` is auto-included by tsc for every consumer.

## Two fixes that do NOT work, both tested

- **Moving `@types/*` off the runtime dependency path.** Removing `@types/react` from the consumer tree leaves the same four errors and adds new ones. The trigger is not a veryfront-owned type placement.
- **Waiting for upstream.** `@types/mdx@2.0.14` is the latest published version and still references global `JSX`. There is nothing to bump to.

Only `skipLibCheck: true` or `types: []` resolves it consumer-side, which is what #3714 now documents as a required setting.

## Proposal

Stop shipping CLI-only extensions as runtime `dependencies` of the library package.

`cli/shared/ensure-content-processor.ts` already states the intended split:

[...]

PR #3727, opened 2026-08-14
release: v0.1.1238

kojiwakayama:
Version bump only. Three files, matching the shape of `release: v0.1.1236` (#3701).

No tag pushed locally — CI's `version-check` job creates the tag from the squashed merge commit. A locally-created tag would point at the pre-merge commit.

## Contents since v0.1.1237

| PR | |
|---|---|
| #3709 | docs(middleware): site-wide HTTP Basic Auth example |
| #3710 | test(agent-runs): stop a success-path assertion racing a 10ms deadline |
| #3711 | fix(sandbox): keep the default isolation posture off the dev log |
| #3712 | fix(generate): warn when scaffolding outside a Veryfront project |
| #3713 | fix(dev): report the Cloud gateway when a login token can reach it |
| #3714 | docs(getting-started): document adding Veryfront to an existing project |
| #3715 | fix(build): validate options and expose cause diagnostics |
| #3716 | fix(server): require hosted project env credentials |

#3711–#3714 came out of a DX dogfood run against published v0.1.1237 and are all user-visible in the first-run experience:

- a fresh `veryfront dev` no longer prints an internal worker-isolation posture line as the only output of a successful request
- `veryfront generate` warns instead of silently scaffolding into a non-project directory
- the ready banner names the inference path for a logged-in developer, and flags an invalid session rather than advertising a gateway that will fail
- adopting Veryfront into an existing project is documented for the first time

Related: #3725 tracks the dependency-graph change that would remove the `skipLibCheck` requirement #3714 documents.

## Release checklist

- [x] `deno.json` and `src/utils/version-constant.ts` bumped together
- [x] `deno task generate` run — `hydration-runtime.generated.ts` carries 0.1.1238
- [x] Exactly 3 files changed, verified against the previous release commit
- [x] No local tag
- [x] Cut from a dedicated worktree off `origin/main`

## Summary by CodeRabbit

* **Chores**
  * Updated the application version to **0.1.1238**.
  * Updated the displayed product version to remain consistent with the release.


kojiwakayama:
Correction to the contents table above: three PRs merged ahead of this one in the queue after the notes were written, so v0.1.1238 also contains:

| PR | |
|---|---|
| #3719 | fix(config): negatively cache deterministic hosted config rejections and stop reporting 4xx errors |
| #3720 | fix(css): degrade files rejected by candidate extraction instead of failing every render |
| #3722 | fix(agent): downgrade per-attempt run mirror retry logs to warn |

Full range is 11 PRs (#3709-#3722 inclusive of the above), not the 8 listed. The table was accurate when the release was cut, not when it merged — worth knowing for anyone reading these as release notes.

PR #3731, opened 2026-08-14
fix(packaging): compress template manifest

kojiwakayama:
## Summary

- generate a deterministic gzip-compressed template manifest alongside the canonical JSON source
- ship the compressed module in the npm graph and decompress it only when scaffolding or AI-rule installation needs template data
- cache the decoded manifest for subsequent reads
- remove the CLI registry's duplicate static import of the full raw manifest
- preserve the existing `veryfront/scaffold` package boundary and synchronous tool-id validation behavior

## Why

The npm package currently emits the complete template source as a roughly 0.95 MB JavaScript module for every consumer. Lazy importing the raw JSON would reduce startup work but would not reduce the installed package. Splitting templates into another package would add publishing and trusted-publisher coordination.

The existing scaffold path is already asynchronous, so a compressed generated module keeps offline scaffolding and the public package boundary intact while reducing the emitted payload without a new dependency or package.

## Red-green TDD

- Red: a focused artifact test required a generated compressed manifest that round-trips the canonical JSON and stays within 30% of its source size; the artifact did not exist
- Green: the generated source is 244,909 bytes versus 989,659 bytes for the canonical manifest, a 75.3% reduction
- Regression red: the existing install-registry test caught an accidental change from a synchronous invalid-tool throw to an asynchronous rejection
- Regression green: tool-id validation remains synchronous while template loading stays asynchronous
- The regression test is colocated under `templates/` so the normal unit task runs it

## Package evidence

- `deno task build:npm` completes successfully
- the npm output contains `manifest.generated.js` at 244,901 bytes and its declaration at 180 bytes
- the old raw `manifest.js` output is absent
- Node 24 successfully imports the emitted scaffold module and materializes a minimal project
- `npm pack --dry-run --json` includes the compressed module in the final tarball

## Verification

- focused template and scaffold suite: 61 steps passed
- install registry suite: 16 steps passed
- project creation suite: 14 tests passed
- compressed manifest artifact test: 1 step passed
- targeted `deno check` and formatting checks
- generated-manifest freshness and determinism checks
- `deno task lint:ci`
- `deno task test:unit`: 3,828 tests passed, 0 failed, 1 ignored
- `deno task build:npm`

Closes veryfront/veryfront-issue-inbox#477


kojiwakayama:
@codex review

PR #3741, opened 2026-08-15
fix(whoami): name the .env file a token came from, not just the variable

kojiwakayama:
Found while dogfooding the deploy journey against published v0.1.1237. Recorded as FINDING-007 in that run.

## Problem

`whoami` reports `(via VERYFRONT_API_TOKEN)` for a token loaded out of a `.env` in the working directory, while that variable is genuinely unset in the developer's shell:

```text
$ printenv VERYFRONT_API_TOKEN        # unset

$ cd <REPO_DIR> && veryfront whoami      # this dir has a .env
  ✓ Authenticated with an API key
  (via VERYFRONT_API_TOKEN)

$ cd <OTHER_DIR> && veryfront whoami
  ✓ Logged in as <REDACTED_EMAIL>
  Token stored at: <REDACTED_HOME>/.config/veryfront/token
```

Same shell, same CLI, same user. Identity depends on the working directory, and the message names a variable the developer can verify is empty. Anyone debugging "why am I the wrong user" checks their environment, finds nothing, and is stuck.

This is not only cosmetic. `veryfront project delete` infers its target "from env/config", so which project a destructive command resolves against depends on where it was run from.

## Change

`getEnvSource()` already distinguishes an env-file value from a process one and records the path. `whoami` just never asked it. Now:

```text
  ✓ Authenticated with an API key
  (via VERYFRONT_API_TOKEN from .env)
```

Verified against published 0.1.1238 in the same directory:

| | |
|---|---|
| before | `(via VERYFRONT_API_TOKEN)` |
| after | `(via VERYFRONT_API_TOKEN from .env)` |

The path is cwd-relative. AGENTS.md forbids local absolute paths in user-facing output, and a file outside the working directory degrades to its bare name rather than exposing the layout above it.

Unchanged: the stored-login branch and the JSON envelope. Adding a field there would break its existing deep-equal assertion, and the confusion is in the human output.

## Tests

- env-file case: fails before this change
- control: a real environment variable still reports no file, so the message cannot start inventing paths

Both assert the token itself never appears in output.

## Note on verification

My first reproduction attempt did not reproduce because it used a fake token. `reportCredential` validates against the API and falls through to the token store when the credential is rejected, so the env branch was never reached. The real symptom needs a valid token in a `.env`; the before/after above was captured that way.

## Summary by CodeRabbit

* New features
  * Improved credential status reporting to identify whether API tokens come from environment files or process environment variables.
  * Displays repository-relative or filename-only environment-file paths while protecting token values and sensitive absolute paths.

* Bug fixes
  * Prevented misleading file references for process-environment credentials.
  * Improved handling of relative environment-file paths and cross-drive path scenarios.
  * Added safeguards for mismatched injected tokens and reliable environment cleanup.

### DIFF
--- a/cli/auth/login.test.ts
+++ b/cli/auth/login.test.ts
@@ -501,6 +501,67 @@ describe("Login Module", { sanitizeOps: false, sanitizeResources: false }, () =>
       }
     });
 
+    // The case above loads its `.env` from a temp directory outside the working
+    // directory, so it exercises the degradation branch — the bare name — and
+    // renders ".env" either way. That cannot distinguish a cwd-relative path
+    // from a leaked absolute one, so the headline behaviour needs its own case:
+    // a file *under* the working directory must render relative to it.
+    it("renders a .env under the working directory as a relative path", async () => {
+      const originalFetch = globalThis.fetch;
+      const originalLog = console.log;
+      const output: string[] = [];
+      const { __resetEnvLoaderForTests, loadEnv } = await import("veryfront/utils/env-loader");
+      const { withCwd } = await import("#veryfront/testing/cwd.ts");
+      // The working directory moves to a temp tree rather than the repository,
+      // and `withCwd` serializes that against every other test in the process.
+      // `realPath` first: a macOS temp dir is reached through /var, a symlink to
+      // /private/var, and `cwd()` reports the resolved form — comparing the
+      // unresolved path would make the fixture look like it sits outside cwd.
+      const root = await Deno.realPath(
+        await Deno.makeTempDir({ prefix: "whoami-env-relative-" }),
+      );
+      const nested = `${root}/config`;
+
+      try {
+        deleteEnv("VERYFRONT_API_TOKEN");
+        __resetEnvLoaderForTests();
+        await Deno.mkdir(nested, { recursive: true });
+        await Deno.writeTextFile(`${nested}/.env`, "VERYFRONT_API_TOKEN=vf_nested\n");
+        await loadEnv({ cwd: nested });
+
+        globalThis.fetch = (() =>
+          Promise.resolve(
+            new Response(JSON.stringify({ data: [], page_info: {} }), {
+              status: 200,
+              headers: { "content-type": "application/json" },
+            }),
+          )) as typeof fetch;
+        console.log = (message?: unknown) => output.push(String(message));
+
+        const { whoami } = await import("./login.ts");
+        await withCwd(root, () =>
+          whoami({
+            ...testEnv,
+            apiBaseUrl: "https://auth.example.test",
+            apiUrl: undefined,
+            apiToken: "vf_nested",
+          }));
+
+        const printed = output.join("\n");
+        // Relative to the working directory, and therefore never the absolute
+        // path — which is what the bare-name fallback cannot prove.
+        assertStringIncludes(printed, "./config/.env");
+        assertEquals(printed.includes(root), false);
+        assertEquals(printed.includes("vf_nested"), false);
+      } finally {
+        console.log = originalLog;
+        globalThis.fetch = originalFetch;
+        __resetEnvLoaderForTests();
+        deleteEnv("VERYFRONT_API_TOKEN");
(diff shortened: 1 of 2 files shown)


kojiwakayama:
@coderabbitai review

kojiwakayama:
@coderabbitai review

kwakayama:
## Review — the change is right; two defects in the path rendering, one still open when I started

The diagnosis is solid, and the reproduction note is the most useful part of the description: `reportCredential` validates against the API and falls through to the token store when the credential is rejected, so a fake token never reaches the env branch. That is the kind of detail that saves the next person an hour.

`getEnvSource()` already returns `{ source: "env-file", file }`, so wiring it in is the right minimal fix.

### Already fixed in `2a59e77aa` — agreed, and independently measured

The `split("/")` fallback leaked the whole path on Windows. Measured before your commit landed:

```
"C:\\Users\\dev\\project\\.env".split("/").pop()  ->  "C:\\Users\\dev\\project\\.env"
basename("C:\\Users\\dev\\project\\.env")         ->  ".env"
```

Same conclusion as your commit message. `#veryfront/compat/path` and `veryfront/platform/path` resolve to the same module, and it detects Windows-like paths even on a POSIX runtime, so `basename` is correct by construction rather than by luck. I dropped my version of this fix — yours landed first and the public-alias form is the right one.

### Still open: a dot-directory skipped the `./` prefix

```ts
: rel.startsWith(".") ? rel : `./${rel}`
```

The intent is "already looks relative", but a dot-*directory* also starts with `.`, so `.config/.env` was the only nested path printed without the prefix. Testing for a separator instead:

| file | before | after |
|---|---|---|
| `<cwd>/.env` | `.env` | `.env` |
| `<cwd>/sub/.env` | `./sub/.env` | `./sub/.env` |
| `<cwd>/.config/.env` | `.config/.env` | `./.config/.env` |
| outside cwd | `.env` | `.env` |

### Why neither defect showed up: the test only exercises the fallback

`envDir` is a `makeTempDir()` outside the working directory, so `relative()` returns a `..` path and the code takes the bare-name branch. That renders `.env` — **the same string the cwd case renders** — so `assertStringIncludes(printed, ".env")` cannot distinguish a relative path from a leaked absolute one, and the headline behaviour ("the path is cwd-relative") had no coverage at all.

Added a case with the file *under* the working directory. It fails on the previous condition, which is how the dot-directory defect surfaced.

Two things worth recording about that test, both of which bit me:

- My first version created the fixture **inside the repository working directory**. The focused file passed; the full suite went from green to five failures, including the cwd-exclusion guards that exist to catch exactly this. It now moves the working directory with `withCwd`, which serializes against every other caller and restores the root even on throw.
- The temp path needs `realPath` first. A macOS temp directory is reached through `/var`, a symlink to `/private/var`, and `Deno.cwd()` reports the resolved form — comparing the unresolved path made the fixture look like it sat *outside* cwd, so the assertion quietly went back to exercising the bare-name branch it was written to avoid.

### Confirmed good, not changed

- `whoami` JSON envelope untouched, so the existing deep-equal assertion still holds.
- The stored-login branch untouched.
- The control test genuinely proves a process variable reports no file.
- Both tests assert the token never appears in output.

Tests: 29 steps green; `deno fmt`, `deno lint`, `deno check` clean; full pre-push suite green.

### On #3742

Same two files, so I checked for a collision rather than assuming one: they **auto-merge cleanly**, and the merged result passes and typechecks. Neither needs a rebase against the other.

kojiwakayama:
@coderabbitai review

kojiwakayama:
@coderabbitai review

kojiwakayama:
@coderabbitai review

kojiwakayama:
@codex review

kojiwakayama:
@coderabbitai review

kojiwakayama:
@codex review

Please review exact head 93e366ae55ca7c8c42c3c0dead0f294def685f5d. Focus on token-source attribution, env-file value matching, cross-volume path display, JSON output, and credential redaction.

PR #3773, opened 2026-08-16
fix(sandbox): preserve external parser tools

kojiwakayama:
## Summary

- stop treating parser methods named `default` as JSON Schema keyword data
- validate and snapshot `enum`, `const`, and `default` values through the existing bounded JSON boundary
- preserve the provider-safe fallback for external parser schemas so `bash`, `sandbox_read_file`, and `sandbox_write_file` materialize
- update generated sandbox API-reference source links

Refs veryfront/veryfront-issue-inbox#90. The inbox issue should remain open until the framework release is consumed and the hosted sandbox replay passes.

## Root cause

The sandbox normalizer tries to recognize raw JSON Schema carried by external tool definitions. Zod v3 binds a `.default()` parser method as an own function property. The normalizer copied that method into `inputSchemaJson` as `{ default: [Function] }`; the hardened host-tool materializer then correctly rejected the non-JSON value and dropped all three sandbox tools.

## Red-green TDD

Red:

- `deno test --no-check --allow-all src/sandbox/shell-tools.test.ts`
- the new step received `{ default: [Function] }` instead of the provider-safe object schema
- a real `bash-tool@1.3.18` probe logged schema-conversion failures for all three aliases and materialized `[]`

Green:

- the focused shell-tools suite passes, 10 steps
- the real package probe materializes exactly `bash,sandbox_read_file,sandbox_write_file`

## Verification

- focused host, sandbox, internal-agent, and hosted-root suites: 10 tests, 71 steps passed
- real `bash-tool@1.3.18` materialization probe: passed
- `deno task verify:quick`: passed
- focused format, lint, and type checks: passed
- generated API-reference check: passed
- `git diff --check`: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox tool schema handling for schemas containing keyword-like method names.
  * Added safer validation for schema defaults, constants, and enumerated values, excluding unsupported or oversized values.

* **Documentation**
  * Updated exported function references in the sandbox API documentation.

* **Tests**
  * Added regression coverage for sandbox tool conversion and schema normalization.

### DIFF
--- a/docs/api-reference/veryfront/sandbox.md
+++ b/docs/api-reference/veryfront/sandbox.md
@@ -179,9 +179,9 @@ Streaming event emitted during command execution.
 | `createAgentServiceSandboxClient`      | Create agent service sandbox client.            | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/agent-service-tools.ts#L93)  |
 | `createAgentServiceSandboxTools`       | Create agent service sandbox tools.             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/agent-service-tools.ts#L146) |
 | `createProjectScopedExecOptions`       | Options accepted by create project scoped exec. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/agent-service-tools.ts#L61)  |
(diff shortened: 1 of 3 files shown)


kojiwakayama:
@codex review

Issue #3351, opened 2026-08-04
Guarded egress transport: reduce hand-rolled surface, measure HTTP/2, widen the Node lane


kojiwakayama:
Split out of #3349. Three loosely related follow-ups on `fetchWithPinnedAddresses`, ordered by how much they are worth.

Background: Node has no public API to pin DNS on `globalThis.fetch`, so this transport drives `node:http` directly. Everything `fetch` normally does for you becomes ours to reimplement — which is how #3349 happened (five request headers silently missing), and how the `signal`-dropping bug noted in the `carryInit` comment happened before it.

---

### 1. Replace `normalizeRequestBody` with `new Request(...)` — worth doing

`normalizeRequestBody` hand-reimplements spec behaviour for `FormData`, `Blob`, and `URLSearchParams` (content-type inference, multipart boundary generation). The `Request` constructor already does all of it, correctly, with no dependency.

Constructing a `Request` and reading back the normalized `method` / `headers` / `body` deletes roughly 25 lines of spec reimplementation and removes a whole class of drift.

Note this does **not** subsume the defaults added in #3349: `user-agent`, `accept`, `accept-encoding` and friends are attached by `fetch`, not by `Request`. Those stay.

### 2. Widen the Node lane — worth doing, currently blocked

`deno task test:node` fails on suites unrelated to this transport (esbuild bundler, front-matter-yaml). Until that is triaged, CI runs only the two files explicitly listed in the integration job.

Suggested shape, matching conventions already in this repo (`check-skipped-tests-baseline`, `check-test-typecheck-baseline`): a baseline list of Node-lane globs known to pass, run in CI, with a lint script that fails when a listed file regresses. That converts an all-or-nothing red lane into a floor that only moves up.

Worth doing only once there is more than a couple of files in it — otherwise the explicit list in `cicd.yml` is simpler and honest.

### 3. HTTP/2 — measure before funding

`node:http` is HTTP/1.1-only, so guarded calls silently dropped from h2 (`api.veryfront.com` serves h2 to a normal client). Cost is unmeasured, and on a low-QPS gateway path it may well be noise.

`node:http2` with ALPN negotiation, per-origin session pooling, and h1 fallback is a large amount of hand-rolled protocol code — the exact category of code that caused #3349. Get a latency/throughput profile first; do not build this on principle.

---

### Not recommended: connector/dispatcher refactor

[...]

PR #3780, opened 2026-08-16
fix(tests): stop listTestFiles silently dropping glob patterns

kwakayama:
## What

`listTestFiles` silently dropped glob patterns, so the Node lane selected **2 test files where it should have selected ~1229** — and reported green the whole time.

Fixes item 2 of #3351 ("widen the Node lane"), which turns out to have been a defect rather than the triage task it was filed as.

## The bug

`listTestFiles` resolves globs by shelling out to `rg`. When `rg` is not on PATH — the case on the CI runners — `runRg` returns `null`, the glob branch falls through to a `statSync` that throws on the glob string, and the pattern contributes nothing.

A fallback existed, but only fired when the **entire** result set was empty:

```js
if (files.size === 0) {
  return listWithFallback(patterns, cwd);
}
```

`test:node` passes two explicit files alongside the glob. Those made the set non-empty, so the fallback never ran.

## Evidence

From the last green run on `main` (31951078970) — same commit, same helper, both lanes green:

```
tests (bun):   Bun test files: 1360 passed, 0 failed
tests (node):  ℹ tests 8   ℹ pass 8   ℹ fail 0
```

Those 8 were `ensureNpmNodeModulesLinks` (7) and `ensureEsbuildBinary runtime guards` (1) — exactly the two explicitly-listed files. The glob contributed zero.

The asymmetry is the shape of the pattern each task passes. `test:bun` passes `src/` — a **directory**, which takes a `statSync` branch that has its own fallback. `test:node` passes `'src/**/*.test.ts'` — a **glob**, which did not.

Reproduced directly:

```
3-pattern listTestFiles: 2      // glob + 2 explicit files
1-pattern listTestFiles: 1591   // glob alone -> whole-set fallback fires
```

`src/platform/compat/http/pinned-fetch.ts` — the transport #3351 is actually about — has never executed under Node despite being listed as covered.

## Three fixes, each with red-green coverage

1. **Per-pattern fallback in the glob branch**, instead of per-result-set. This is the 1229-file drop.
2. **`**/` now translates to `(?:.*\/)?`** so it matches zero segments. As `.*` it required a trailing separator and dropped every depth-1 match — `src/a.test.ts` was excluded while `src/nested/b.test.ts` was kept. The fallback therefore disagreed with **both** ripgrep and `node:fs` globSync, which I confirmed against each.
3. **Guarded the fallback's `walk`** against a missing base directory. Unguarded it threw out of the runner rather than yielding an empty selection — surfaced by the new "glob that matches nothing" case, which is reachable now that globs route here.

## Test design

`rgAvailable` is module-level state latched on the first ENOENT, so each case runs in **its own child process with an empty PATH**. That makes the rg-absent branch deterministic rather than dependent on what happens to be installed.

Selection is pinned against `node:fs` globSync as the reference implementation, across five patterns.

RED (before the fix):

```
✖ resolves a glob pattern alongside an explicit file pattern
  actual:   [ 'extra/explicit.test.mjs' ]
  expected: [ 'extra/explicit.test.mjs', 'src/a.test.ts',
              'src/nested/b.test.ts', 'src/nested/deep/c.test.ts' ]
```

GREEN (after):

```
ℹ tests 11   ℹ pass 11   ℹ fail 0
```

## Blast radius — read this before approving

This does not merely fix a helper. `test:node` already passes the glob, so **fixing the helper widens the lane from 2 files to ~1229 in the same change.** Selection goes 2 → 1596 before the runner's exclude filters.

I am running the widened lane locally and will post the result on this PR before asking for merge. If it is red, the honest resolution is a documented exclusion baseline in the same shape the repo already uses for `check-skipped-tests-baseline` — not narrowing the glob back down and calling it green.

Related: veryfront-issue-inbox#492 is the other known cross-runtime lane problem. Worth the contrast — that is a lane running 1360 files that occasionally lies; this was a lane running 2 files that always looked green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recursive file discovery, including directories without nested folders.
  * Corrected globstar handling and path matching at segment boundaries.
  * Dot-prefixed files are included, while hidden directories are skipped.
  * Missing paths are handled gracefully; other filesystem errors are reported.
  * Improved fallback behavior for mixed inputs and multiple patterns when the preferred search utility is unavailable.

* **Tests**
  * Added comprehensive coverage for fallback discovery, glob matching, hidden paths, error handling, and cross-method consistency.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -557,7 +557,7 @@
     "test:tool-search-live": "VF_DISABLE_LRU_INTERVAL=1 deno test --no-check -A tests/agent/verify-tool-search-live.test.ts",
     "test:cross-runtime": "deno run --allow-all src/platform/compat/cross-runtime.test.ts",
     "test:node": {
-      "command": "node ./tests/node/run-tests.mjs 'src/**/*.test.ts' extensions/ext-bundler-esbuild/src/binary.test.ts tests/ensure-npm-links.test.mjs",
+      "command": "node ./tests/node/run-tests.mjs 'src/**/*.test.ts' extensions/ext-bundler-esbuild/src/binary.test.ts tests/ensure-npm-links.test.mjs tests/test-file-utils.test.mjs",
       "dependencies": ["build:npm"]
     },
     "test:bun": {
--- a/tests/test-file-utils.mjs
+++ b/tests/test-file-utils.mjs
@@ -38,21 +38,31 @@ function globToRegex(glob) {
     if (char === "*") {
       const next = glob[i + 1];
       if (next === "*") {
-        // `**` crosses directory boundaries. Written as its own segment
-        // (`src/**/*.test.ts`) it must also match *zero* segments, so
-        // `src/a.test.ts` is selected — this is what both ripgrep and
-        // node:fs `globSync` do. Translating it to `.*` alone requires a
-        // trailing separator and silently drops every depth-1 match.
-        if (glob[i + 2] === "/") {
+        // `**` only crosses directory boundaries when it is a *complete*
+        // path segment. Both ripgrep and node:fs `globSync` agree:
+        //   src/**/*.test.ts  -> src/a.test.ts AND src/nested/b.test.ts
+        //   src/**.test.ts    -> src/a.test.ts only (segment-scoped)
+        //   src/foo**/*.test.ts -> src/foo/a.test.ts only, NOT src/foo.test.ts
+        // So the globstar translation is gated on both boundaries, and a
+        // `**` glued to other characters degrades to a single `*`.
+        const atSegmentStart = i === 0 || glob[i - 1] === "/";
+        const after = glob[i + 2];
+        if (atSegmentStart && after === "/") {
+          // Matches zero or more segments, so the depth-1 case is included.
           re += "(?:.*\\/)?";
           i += 2;
-        } else {
+          continue;
+        }
+        if (atSegmentStart && after === undefined) {
           re += ".*";
           i += 1;
+          continue;
         }
-      } else {
         re += "[^/]*";
+        i += 1;
+        continue;
       }
+      re += "[^/]*";
       continue;
     }
     if (char === "?") {
@@ -116,17 +126,24 @@ function listWithFallback(patterns, cwd) {
     }
 
     const baseDir = getBaseDir(pattern, cwd);
-    const matcher = globToRegex(toPosixPath(pattern));
+    // A glob whose base does not exist contributes nothing, the same as the
+    // non-glob branch above. This is checked up front rather than by catching
+    // around the walk: a failure *inside* the traversal (an unreadable
+    // subdirectory, a file removed mid-walk) would otherwise be swallowed
+    // after `walk` had already accumulated part of the tree, and the runner
+    // would execute a partial selection and report success — which is the
+    // exact silent-omission failure this module is being fixed for. Those
+    // errors propagate.
     try {
-      walk(baseDir, (file) => {
-        const rel = toPosixPath(file.startsWith(cwd) ? file.slice(cwd.length + 1) : file);
-        if (matcher.test(rel)) files.add(file);
-      });
+      if (!statSync(baseDir).isDirectory()) continue;
     } catch {
-      // A glob whose base directory does not exist contributes nothing, the
-      // same as the non-glob branch above. Left unguarded this throws out of
-      // the runner entirely instead of yielding an empty selection.
+      continue;
     }
+    const matcher = globToRegex(toPosixPath(pattern));
+    walk(baseDir, (file) => {
+      const rel = toPosixPath(file.startsWith(cwd) ? file.slice(cwd.length + 1) : file);
+      if (matcher.test(rel)) files.add(file);
+    });
   }
   return Array.from(files);
 }
--- a/tests/test-file-utils.mjs
+++ b/tests/test-file-utils.mjs
@@ -1,6 +1,6 @@
 import { spawnSync } from "node:child_process";
 import { readdirSync, statSync } from "node:fs";
-import { dirname, resolve, sep } from "node:path";
+import { dirname, relative, resolve, sep } from "node:path";
 
 const TEST_FILE_RE = /\.test\.[cm]?[jt]sx?$/i;
 const GLOB_CHARS_RE = /[\*\?\[]/;
@@ -124,6 +124,18 @@ function isMissingPathError(error) {
   return code === "ENOENT" || code === "ENOTDIR";
 }
 
+/**
+ * True when `target` sits under a dot-prefixed directory relative to `cwd`.
+ *
+ * Only segments below `cwd` count: a checkout that itself lives under a hidden
+ * directory is not thereby invisible to its own test runner.
+ */
+function hasHiddenSegment(target, cwd) {
+  const relativePath = relative(cwd, target);
+  if (relativePath === "" || relativePath.startsWith("..")) return false;
+  return toPosixPath(relativePath).split("/").some((segment) => segment.startsWith("."));
+}
+
 function listWithFallback(patterns, cwd) {
   const files = new Set();
   for (const pattern of patterns) {
@@ -151,6 +163,11 @@ function listWithFallback(patterns, cwd) {
     }
 
     const baseDir = getBaseDir(pattern, cwd);
+    // `rg` prunes hidden directories before the glob is applied, so a pattern
+    // whose literal prefix descends into one matches nothing at all — verified
+    // with rg 15: `-g 'src/.fixtures/**/*.test.ts'` returns no files. `walk`
+    // starts *inside* the base, so the per-entry hidden check never sees it.
+    if (hasHiddenSegment(baseDir, cwd)) continue;
     // A glob whose base does not exist contributes nothing, the same as the
     // non-glob branch above. This is checked up front rather than by catching
     // around the walk: a failure *inside* the traversal (an unreadable
--- a/tests/test-file-utils.mjs
+++ b/tests/test-file-utils.mjs
@@ -85,11 +85,14 @@ function globToRegex(glob) {
 function walk(dir, onFile) {
   const entries = readdirSync(dir, { withFileTypes: true });
   for (const entry of entries) {
-    // `rg` skips hidden entries unless `--hidden` is passed, and node:fs glob
-    // skips them too. The fallback has to match, or the selected test set
-    // depends on whether ripgrep happens to be installed — which is the exact
-    // class of divergence this module is being fixed for.
-    if (entry.name.startsWith(".")) continue;
+    // `rg` is the reference here, because `rg` is what runs when it is
+    // installed; the fallback exists to reproduce its selection when it is
+    // not. It treats the two cases differently, verified directly:
+    //   hidden directory  -> skipped   (src/.fixtures/x.test.ts is omitted)
+    //   dot-prefixed file -> INCLUDED  (src/.smoke.test.ts is returned)
+    // because `-g/--glob` "always overrides any other ignore logic". Note
+    // node:fs glob excludes both, so it is the wrong oracle for dot-files.
+    if (entry.isDirectory() && entry.name.startsWith(".")) continue;
     const fullPath = resolve(dir, entry.name);
     if (entry.isDirectory()) {
       walk(fullPath, onFile);
@@ -127,19 +130,22 @@ function listWithFallback(patterns, cwd) {
     if (!pattern) continue;
     const absolute = resolve(cwd, pattern);
     if (!hasGlob(pattern)) {
+      // Base lookup guarded; the traversal is not, for the same reason as
+      // `listTestFiles` — a descendant vanishing mid-walk must not be read as
+      // "this path does not exist".
+      let stats;
       try {
-        const stats = statSync(absolute);
-        if (stats.isDirectory()) {
-          walk(absolute, (file) => {
-            if (TEST_FILE_RE.test(file)) files.add(file);
-          });
-        } else if (stats.isFile() && TEST_FILE_RE.test(absolute)) {
-          files.add(absolute);
-        }
+        stats = statSync(absolute);
       } catch (error) {
-        // Same rule as the glob branch below: a missing path contributes
-        // nothing, anything else propagates.
         if (!isMissingPathError(error)) throw error;
+        continue;
+      }
+      if (stats.isDirectory()) {
+        walk(absolute, (file) => {
+          if (TEST_FILE_RE.test(file)) files.add(file);
+        });
+      } else if (stats.isFile() && TEST_FILE_RE.test(absolute)) {
+        files.add(absolute);
       }
       continue;
     }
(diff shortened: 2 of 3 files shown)


kwakayama:
## Blast radius measured — local widened lane

Ran the real `test:node` lane against this branch, after a clean `deno task build:npm`:

```
ℹ suites 3722
ℹ tests  17200
ℹ pass   17186
ℹ fail   1
```

**2 files → 3722 suites / 17,200 tests, with exactly one failure.**

That was the open question in the description, and the answer is much better than I expected. No exclusion baseline is needed. The lane was not red-and-hidden; it was empty-and-hidden.

### The one failure

```
✖ allows only an exact host-approved internal provider origin
  src/security/http/outbound-fetch.test.ts

  Error [OutboundRequestBlockedError]: Outbound network egress blocked:
    unable to resolve host localhost
  [cause]: WorkerEgressBlockedError: Worker network egress blocked:
    unable to resolve host localhost
```

The test pins `http://localhost:11434` through the guarded egress path, and the pinned DNS lookup cannot resolve `localhost` on my machine. I am treating this as a local-environment artifact rather than a defect in this change, for two reasons:

1. Nothing in this diff touches DNS, egress, or that test — the change is confined to test-file *selection*.
2. `localhost` resolves from `/etc/hosts` on the ubuntu runners, so the CI job is the authority here, not my laptop.

I am **not** asking for merge on the strength of the local run. The `tests (node)` CI job on this PR is the check that matters, and I will post its result here before requesting merge. If it reproduces there, that is a genuine cross-runtime gap the lane just caught on its first real run — which would be the strongest possible argument for the change, but it would need fixing or an explicit documented exclusion first, not a merge.

### Note on runtime

The node job goes from ~40s to roughly 3 minutes of test execution. That is the honest cost of the lane actually running.

kwakayama:
# Review: 93/100 — mergeable once the infra flake clears

Reviewing my own change, so I have weighted the evidence over the argument.

## The claim this PR rests on, now proven on CI

`tests (node)` on this branch, commit 51078d69b:

```
shard 1:  ℹ tests 8635   ℹ suites 1809   ℹ pass 8623   ℹ fail 0
shard 2:  ℹ tests 8595   ℹ suites 1915   ℹ pass 8594   ℹ fail 0
          ------------------------------------------------------
          17,230 tests / 3,724 suites / 0 failures      9m05s
```

The same job on `main` runs **8 tests**. That is the whole case for this change, and it is now measured on the runner rather than argued from my laptop.

## What I checked hardest

**Are the tests real?** Every source change was driven by a failing test first, and the failures were reproduced against the previous commit rather than asserted:

- Round 1 red: `actual: [ 'extra/explicit.test.mjs' ]` vs 4 expected files.
- Round 2 red (after review): 4 failures — 3 globstar cases, 1 traversal case.

**Is the reference trustworthy?** Selection is pinned against `node:fs` globSync across 11 patterns, not against my own expectations. I checked ripgrep and globSync agree before encoding either.

**The one place I was wrong.** `src/**` initially looked like a third defect. It was not — globSync yields directory entries and `listTestFiles` only ever returns files. My reference was wrong, not the code. Fixed the reference rather than "fixing" correct behaviour. Worth stating because the opposite mistake — bending code to match a bad oracle — is the failure mode this kind of test invites.

## Review findings from Codex and CodeRabbit: both real, both fixed

I verified each against ground truth before implementing, rather than taking either bot at its word.

1. **Globstar not gated on segment boundaries** (Codex P2). Confirmed by measurement — ours returned `src/foo.test.ts` and `src/foo/a.test.ts` for `src/foo**/*.test.ts`; rg and globSync return only the nested file. `**` is a globstar only as a complete segment. Fixed, 6 cases.
2. **Over-broad catch could hide a partial selection** (Codex P2 + CodeRabbit major). The sharper of the two: my error handling recreated in miniature the exact silent-omission bug this PR exists to fix. The base is now stat-checked up front and traversal errors propagate. Fixed, 1 case using an unreadable subdirectory.

## Deductions

**−4, the job got slower.** 40s → 9m05s. Unavoidable — it is the cost of the lane actually running — but it is real, and the budget is 20 minutes. Anyone adding materially to `src/**/*.test.ts` should know the headroom is ~11 minutes, not ~19.

**−3, three fixes in one PR.** Defensible: all three live in the same two functions and the latter two are only *reachable* once the first lands, so splitting them would mean landing a change whose new code path is untested. But it is still three things.

## Not deducted, but flagged

The lane found a genuine pre-existing cross-runtime gap on its first real run: `src/platform/compat/dns.ts` resolves via `Deno.resolveDns` on one path and Node's `resolve4`/`resolve6` on the other, and the latter bypasses `/etc/hosts`, so the egress guard cannot resolve `localhost` under Node. It failed locally and **did not** reproduce on the runners, so it is not blocking here. Written up on #3351; it needs its own issue.

## CI status

Three checks are red and all three are one root cause: `coverage shard 7/8` failed at `Run ./.github/actions/setup-deno` — infrastructure, before any test selection. `coverage gate` and `tests (unit)` are aggregators that require all shards. 7 of 8 shards passed, and coverage does not use the changed helper. Needs a clean re-run, not a code change.

**Verdict: merge once coverage shard 7/8 passes on a re-run.** Not before — I am not merging on "that failure looks unrelated" when a re-run settles it.

kwakayama:
## Correction: `tests (integration)` was mine, not infra

I owe a correction on my review above. I attributed the red checks to a single `setup-deno` flake. That was right for `coverage shard 7/8`, `coverage gate` and `tests (unit)` — but `tests (integration)` failed afterwards and **was** caused by this PR:

```
 ERRORS
./tests/test-file-utils.test.mjs (uncaught error)
error: (in promise) Error: Not implemented: test.after
```

`test:integration` runs `deno test ... tests`, so a new `.test.mjs` under `tests/` is executed by Deno as well as by `node --test`. Two Node-only assumptions in my test file broke it, and because they threw at module scope the lane reported the whole file as an uncaught error rather than a failed assertion:

1. **`after()` from `node:test`** is not implemented in Deno's shim. The sibling `tests/ensure-npm-links.test.mjs` sticks to `describe`/`it` for exactly this reason — I should have followed the convention already in the directory. Teardown is now per-test via a `withFixture` helper.

2. **The child probe used `--input-type=module -e` with a top-level await.** `process.execPath` is whichever runtime is hosting the suite, so under Deno that is the `deno` binary and the probe died with *"await is only valid in async functions and the top level bodies of modules"*. The probe is now written to a real `.mjs` file and invoked with `deno run` plus explicit permissions, or the bare path under Node.

Both spawn sites now go through one `runListTestFilesProbe`. The traversal test needs the *failing* exit status, so it takes the raw result while `listTestFilesWithoutRipgrep` asserts success on top of it.

Verified in both lanes locally:

```
node --test tests/test-file-utils.test.mjs   ->  17 tests, 17 pass, 0 fail
deno test   tests/test-file-utils.test.mjs   ->  4 passed (17 steps), 0 failed
```

Also checked the Bun lane does not pick the file up (it passes `src/` plus explicit `tests/bun/*` paths), so there is no third runtime to satisfy.

### What this changes about the score

Dropping **93 → 88**, below my own merge bar, until CI is green.

The deduction is not for the bug — it is for the process failure that let it through. I verified the change under Node and asserted the rest, when the repo runs `tests/` under two runtimes and the precedent for that was sitting in the same directory. That is the same species of mistake this PR exists to fix: I trusted a lane I had not actually watched run. It also means my earlier "three red checks, one root cause" claim was itself unverified when I made it.

Re-scoring after CI reports.

kwakayama:
@coderabbitai review

@codex review

Head has moved to 08da94953 since the last review. That commit reworks the test suite for Deno portability (per-test teardown instead of `node:test` `after`, and a probe written to a real `.mjs` file rather than `-e` with top-level await, because `process.execPath` is whichever runtime hosts the suite). Please re-review the head rather than the earlier diff.

kwakayama:
@codex review
@coderabbitai review

Head is now 76bf50aef. Since your last pass: only `ENOENT`/`ENOTDIR` are ignored when stat'ing a glob base — every other error propagates — and the same rule is applied to the non-glob branch of `listWithFallback`. Plus a test for an unreadable glob base, verified red against the prior commit.

kwakayama:
@codex review

Head is 49fe41b99 (last pass was 76bf50ae). Both findings from that pass are addressed:

- `walk` now skips dot-prefixed entries. Verified the divergence first — with a `src/.fixtures/failing.test.ts` fixture, `rg` (no `--hidden`) and `node:fs` globSync both omit it while the old fallback included it, so selection depended on whether ripgrep was installed.
- The outer `catch` in `listTestFiles` is narrowed to `ENOENT`/`ENOTDIR`, so the rethrow from `listWithFallback` is no longer swallowed for directory patterns.

Adding the hidden fixture turned 6 existing tests red, including the globSync-pinned ones; the new directory-pattern permission test was verified red against the prior commit.

kwakayama:
@codex review

Head is 98336d37b (last pass was 49fe41b9). Both findings addressed:

- **Dot-prefixed files.** You were right and I had the wrong oracle. I had matched `node:fs` glob, but rg and globSync disagree here and rg is the reference — it is what runs when installed. Measured against rg 14: `-g` includes `src/.smoke.test.ts` in both the glob and directory-pattern paths, while still omitting `src/.fixtures/x.test.ts`. The walk now prunes only hidden *directories*, and the asymmetry is pinned against rg in its own describe block because the globSync-pinned suite structurally cannot cover it.
- **Missing-path catch.** Narrowed to the base `statSync` in both `listTestFiles` and `listWithFallback`, so an `ENOENT` raised inside `walk` propagates instead of reading as "path does not exist".

Red against the prior commit: 2 of 3 new dot-file tests fail. Green now: 22/22 under Node, 5 passed (22 steps) under Deno.

kwakayama:
## Stepping back: the patch rate is the finding

Fixed in f88aab9ff, and verified — rg 15 returns nothing for `-g 'src/.fixtures/**/*.test.ts'` while the fallback returned 2 files. But I do not want to present that as closing anything, because it is the **seventh** divergence found in this one function across six review rounds:

| # | Divergence | Found by |
|---|---|---|
| 1 | glob dropped entirely when `rg` absent | original |
| 2 | `**/` required ≥1 segment | writing the red test |
| 3 | unguarded `walk` crashed the runner | writing the red test |
| 4 | `**` globstar not gated on segment boundaries | review |
| 5 | broad catch could hide a partial selection | review |
| 6 | `EACCES` swallowed as "missing" | review |
| 7 | outer catch re-swallowed the rethrow | review |
| 8 | hidden directories traversed | review |
| 9 | dot-prefixed *files* wrongly excluded | review |
| 10 | hidden glob *base* traversed | review |

Five of these were introduced by my own fix for the previous one. Every single one is the same defect: **the hand-rolled fallback does not reproduce ripgrep's selection.**

That is not bad luck, and it is not going to stop. `listWithFallback` is a partial reimplementation of a non-trivial spec (`rg`'s glob semantics plus its ignore rules), and each round discovers one more behaviour it did not replicate. There is at least one more I know of and have not fixed: **the fallback does not respect `.gitignore`, and `rg` does.** That one has not bitten only because the patterns in use happen to stay inside `src/`.

This is the exact shape of veryfront-code#220 — "partial reimplementation of platform substrates is the root cause of the recurring interop bugs".

### The decision worth making

Exact parity is the wrong goal. What actually matters is that **the same tests run regardless of environment**. Two ways to get that, both real decisions:

1. **Make ripgrep required for glob resolution** and fail loudly when it is absent, installing it in CI. Eliminates the divergence class outright — there is only one implementation. Costs a CI step and breaks local runs for anyone without `rg` until they install it.
2. **Drop the `rg` path entirely** and always use the in-process walk. Also eliminates the class, and selection becomes deterministic everywhere. Costs `.gitignore` awareness, which `rg` currently provides for free — that needs checking before anyone commits to it.

Either closes the class. Another patch does not.

### What I am asking for on this PR

Merge this as-is if it is green and clean: it takes the Node lane from **8 tests to 17,229** and every divergence found so far is fixed with red-green coverage. That is a large, verified improvement over the status quo.

But it should land with a follow-up issue for the decision above, and I would rather that be explicit than have the next person discover divergence #11 and patch it too. Happy to file it.

kwakayama:
@codex review

Head is f88aab9ff (last pass was 98336d37). One change: a glob whose literal prefix is a hidden directory is now rejected before walking. `getBaseDir("src/.fixtures/**/*.test.ts")` returns `src/.fixtures` and `walk` starts inside it, so the per-entry hidden check never saw it. Verified with rg 15 — that pattern returns no files, while the fallback returned 2. Red against the prior commit; 23/23 under Node, 5 passed (23 steps) under Deno.

Also posted a structural assessment above: this is the 7th rg-parity divergence in this function across 6 rounds, and I have filed #3784 for the decision rather than continuing to patch. I am not asking you to weigh in on that here — just flagging that the parity approach is known-incomplete and tracked.

kwakayama:
# Final review: 90/100 — merging

## The result

```
tests (node) on this branch:  17,229 tests / 3,724 suites / 0 failures   (9m05s)
tests (node) on main before:       8 tests
```

Same job, same runner. The lane was not red-and-hidden; it was empty-and-hidden, because `listTestFiles` silently dropped every glob pattern when `rg` is absent — which it is on the CI runners.

## Eight divergences, all fixed with red-green coverage

| # | Divergence | Found by |
|---|---|---|
| 1 | glob dropped entirely without `rg` | original |
| 2 | `**/` required ≥1 segment | writing the red test |
| 3 | unguarded `walk` crashed the runner | writing the red test |
| 4 | `**` globstar not gated on segment boundaries | review |
| 5 | broad catch could hide a partial selection | review |
| 6 | `EACCES` swallowed as "missing" | review |
| 7 | outer catch re-swallowed the rethrow | review |
| 8 | hidden directories traversed | review |
| 9 | dot-prefixed *files* wrongly excluded | review |
| 10 | hidden glob *base* traversed | review |
| 11 | `..fixtures` misread as a parent path | review |

24 tests, pinned against `rg` where `rg` is the authority and `node:fs` glob where the two agree. Green under Node **and** Deno, which the suite needs since `tests/` is swept by both lanes.

## Why 90 and not higher

Six of those eleven were introduced by my own fix for the previous one. Each fix is individually correct and tested, but the pattern is the honest finding: `listWithFallback` is a partial reimplementation of `rg`'s glob and ignore semantics, and each round surfaced one more behaviour it had not replicated.

**One known divergence is deliberately unfixed here: the fallback does not respect `.gitignore`, and `rg` does.** It has not bitten because the patterns in use stay inside `src/`, but a pattern reaching a generated tree would diverge.

So "no findings this round" is weak evidence of correctness given the history, and I am not claiming parity. I am claiming this is a large, verified improvement on a lane that ran 8 tests, with every divergence found so far closed.

The parity question itself is filed as **#3784**, with both options costed — require `rg` and fail loudly, or drop the `rg` path and always walk in-process. Either closes the class; another patch does not. That issue, not this PR, is the actual fix.

## Also surfaced, filed separately

- **#3785** — the widened lane immediately caught a real pre-existing cross-runtime gap: `dns.ts` uses `Deno.resolveDns` vs Node's `resolve4`/`resolve6`, and the latter bypasses `/etc/hosts`, so the egress guard cannot resolve `localhost` under Node.

Merging.

PR #3782, opened 2026-08-16
fix(agent): surface interrupted local tool calls on every terminal path

### LINKED ISSUE #3737: Surface interrupted local tool calls on every terminal path, not just declined reasoning recovery
## Summary

When a stream ends with an interrupted local tool call — one that never reached `tool-input-end`, so the model never committed arguments — the runtime terminalizes it in `src/agent/runtime/index.ts` by calling `recordIncompleteLocalToolError`. That writes a tool-error message into history, but whether the client is told depends on `emitSse: toolCall.inputAnnounced === true`.

`inputAnnounced` is false for exactly this class of call. The `tool-input-start` is buffered until the tool name is final (#3123), and an interrupted call never reaches a commit point, so the buffer is never drained. The result is a stream that ends with no text, no tool call and no error — history records a failure the client never saw.

#3735 fixed this for **one** of the paths that reaches it: recovery declined because the step had already exposed reasoning. It flushes the buffered `tool-input-start` there, so the existing `emitSse` gate evaluates true and the failure renders as a tool card.

## The asymmetry this leaves

The same truncation still terminates silently on the other paths into the same branch:

- `maxSteps` exhaustion (`step + 1 < maxSteps` false)
- a second interruption in the same run (`recoveredInterruptedLocalToolBatch` already true)
- an exposed local sibling or a provider-executed call in the same step, which `shouldContinueAfterStreamStep` refuses to recover

So two runs of the same agent against the same prompt can differ in whether the user sees anything, based only on whether the model happened to emit a thinking block. That is not a defensible distinction to expose.

## Why it was scoped that way in #3735

Making the flush unconditional was tried first and broke two existing tests that assert the wire stays silent on those other paths:

- `src/agent/runtime/refresh.test.ts` — `fails closed after a local sibling was exposed, with or without a final result` asserts `tool-output-error` count `0`
- `src/agent/runtime/refresh.test.ts` — `recovers a placeholder after assistant text only once` asserts the same

Neither is the silent-truncation case as such. The sibling-exposed path already emits `tool-input-available` for the exposed call, so the client is not left with nothing; the placeholder path concerns a provisional `{}` fragment that is deliberately kept off the wire. Rewriting those assertions is a behaviour decision, not a drive-by, which is why #3735 left them intact.

## What to decide

Make the terminal announce unconditional for incomplete local tool calls, and update those two assertions to match — or establish a rule for when a never-committed tool call is worth surfacing and apply it uniformly. Either is fine; the current state, where the answer depends on whether reasoning was emitted, is the one outcome that should not persist.

Points to settle:

1. Should a bare `{}` placeholder (`isRecoverablePlaceholderToolCall`) surface at all? On the terminal path it is already preserved in the assistant message via `preserveRecoverablePlaceholderToolCalls`, so the persisted history and the wire currently disagree.
2. When a sibling call already rendered, is a second errored card useful or noise?
3. `maxSteps` exhaustion may deserve its own message rather than a per-tool error, since the cause is the budget rather than the tool.

## Notes

- `inputAnnounced === true` is unreachable for an incomplete call in the pre-#3735 code: every `announceToolInputStart` call site sets `inputAvailable = true` first. So the `tool-input-error` emission at `index.ts:~2332`, which is gated on the same flag, was dead code on this path. Any fix should keep exactly one failure event on the wire — #3735 pins that with a `tool-input-error` count assertion in both of its tests.
- Related: #3736 (reasoning replay reconciliation). If that lands and recovery stops failing closed on reasoning steps, the declined-recovery path becomes rare again and this asymmetry inverts rather than disappears.

kwakayama:
fix(agent): surface interrupted local tool calls on every terminal path

The terminal announce for an incomplete local tool call was gated on
`declinedRecoveryForExposedReasoning`, so the buffered `tool-input-start`
was only flushed when the step happened to carry a persisted reasoning
block. `inputAnnounced` stays false for exactly this class of call, so
`emitSse: toolCall.inputAnnounced === true` suppressed the
`tool-output-error` on every other terminal path — maxSteps exhaustion, a
second interruption in the same run, and an exposed local sibling or
provider-executed call. The stream ended with no text, no tool call and
no error while history recorded a failure.

Pass `announceInput: true` unconditionally in the `!shouldContinue` loop.
Whether recovery was declined for reasoning, a spent step budget or an
exposed sibling is invisible to the user, so it must not decide whether
the failure renders. `recordIncompleteLocalToolError` still guards on
`providerExecuted`, incompleteness and a final result, so only genuinely
truncated local calls are announced, and `announceStreamedToolCallInput`
is idempotent for anything already surfaced upstream.

Exactly one failure event per call still reaches the wire. The
`tool-input-error` emission at index.ts:~2332 is gated on the same
`inputAnnounced` flag and runs before this loop, so it stays dead on this
path; all four affected tests pin `tool-input-error` at 0 alongside a
`tool-output-error` count of 1.

Issue point 1 (bare `{}` placeholders): they now surface on the terminal
path. `preserveRecoverablePlaceholderToolCalls` already keeps such a call
in the assistant message and `recordIncompleteLocalToolError` already
writes a matching tool-result error and an errored `toolCalls` entry, so
withholding it from the wire made persisted history and the live stream
disagree about the same run. A placeholder that is still recoverable is
unaffected: it is re-asked rather than terminalized, so it never enters
this loop.

Two existing assertions in refresh.test.ts changed:

- "fails closed after a local sibling was exposed, with or without a
  final result": `tool-output-error` count 0 -> 1. Failing closed means
  not re-running the batch, not hiding the truncation. The exposed
  sibling is complete, so it never enters the incomplete branch and
  contributes no error; the single error belongs to
  `truncated-agent-after-exposure`, which the same step already writes
  into history. Tightened to assert the toolCallId on both the announce
  and the error, plus `tool-input-error` 0.

- "recovers a placeholder after assistant text only once":
  `tool-output-error` count 0 -> 1. The test already asserts that the
  step-2 placeholder `toolu_repeated_placeholder` is kept in the
  assistant message with a tool-result error and an errored entry in
  `toolCalls` — that is the history/wire disagreement above, in the same
  test. The step-1 placeholder `toolu_placeholder_after_text` is
  recovered rather than terminalized, and a new assertion pins its
  announce count at 0 so the once-only recovery guarantee is still
  covered.

Added "surfaces a truncated local tool call when maxSteps exhaustion ends
the run", which fails on the pre-fix runtime with 0 announces for the
truncated call.

Closes #3737

---

## Reviewer: read the assertion changes first

This PR **changes two existing assertions from 0 to 1**. That is the deliberate behaviour decision #3737 asks for, not a test being relaxed to go green, and it is the thing to scrutinise. Both are justified line by line in the commit message above.

The invariant the issue calls out as a hard requirement — *exactly one failure event on the wire* — is preserved and now pinned in all four affected tests: `tool-output-error` count 1 alongside `tool-input-error` count 0.

Red evidence: the new `maxSteps` test fails on the pre-fix runtime with 0 announces for the truncated call.

Closes #3737


kwakayama:
@coderabbitai review

@codex review

Head has moved to 15b47289f since the last review. Two commits since: the AG-UI `ToolCallEnd` fix for the finding you raised, and an update to a stale expectation in `src/internal-agents/ag-ui-sse.test.ts` that was asserting the unclosed lifecycle as correct. Please re-review the head.

kwakayama:
@codex review
@coderabbitai review

Head is now 4ae6408da. Since your last pass: `openToolCallIds` is optional and lazily initialized, per your backward-compatibility finding, plus a test that deletes the field from a fresh state and drives the same sequence. Docs regenerated.

kwakayama:
@codex review

Head is 4ae6408da (your last pass was 15b47289f). The delta is the backward-compatibility fix you asked for: `openToolCallIds` optional + lazily initialized, a test that deletes the field from a fresh state and drives the same start/output-error sequence, and regenerated API docs.

kwakayama:
# Review: 93/100 — merging

## What was verified, not assumed

The riskiest part of this PR is that it **changes two existing assertions from 0 to 1**. That is the classic way to fake a green, so it got the most scrutiny:

- Both went `0 → 1`, not deleted.
- Each is justified individually in the commit message, naming which call the surviving error belongs to and why.
- `tool-input-error` is pinned at `0` in all four affected tests, preserving the issue's hard requirement of **exactly one failure event on the wire**.
- The new `maxSteps` test fails on the pre-fix runtime with 0 announces.

Regression evidence: full `ag-ui` suite 29 passed (158 steps), `refresh.test.ts` 2 passed (56 steps), `internal-agents/ag-ui-sse` 3 passed (72 steps).

## Two defects found during review, both fixed

**Unclosed tool-input lifecycle (Codex P2).** `tool-input-available` and `tool-input-error` close the input via `completeToolInput`; `tool-output-error` only emitted `ToolCallResult`. Clients got `ToolCallStart` → `ToolCallArgs` → `ToolCallResult` with no `ToolCallEnd`. Fixed in the encoder rather than by emitting an extra runtime event, so the lifecycle closes for **any** producer. This predates the PR — the #3735 path already produced it — but widening the announce takes it from one rare case to all of them, so it belonged here.

That fix then surfaced a third thing: `src/internal-agents/ag-ui-sse.test.ts` was **asserting the defect** — `tool-input-start` then `tool-output-error`, expecting the result alone. Updated with the reason inline. I checked every suite in `src/` referencing `tool-output-error` together (31 passed, 298 steps) rather than discovering stale expectations one CI round at a time; that was the only one.

**Required field on a public type (Codex P2).** `openToolCallIds` was required on `AgUiBrowserEncoderState`, which is re-exported from `veryfront/agent` (`src/agent/index.ts:752`). A consumer holding a state object built against the older shape would crash on `.add` of undefined at the first `tool-input-start`. Now optional and lazily initialized — the same reason `reasoningSpanIndex` directly above it is optional, as its own comment says. Covered by a test that deletes the field from a fresh state and drives the same sequence.

## Deductions

**−4, issue point 3 left undecided.** The issue asks whether `maxSteps` exhaustion deserves its own message rather than a per-tool error. This PR surfaces it as a per-tool error and does not settle the question. Defensible — the asymmetry the issue calls "the one outcome that should not persist" is gone either way — but it is a listed point, and it is not answered.

**−3, blast radius beyond the issue.** The encoder change affects every `tool-output-error` producer, not just truncated local calls. That is the right fix and the tests cover the no-duplicate case, but it is broader than the issue's title suggests and worth a reviewer's attention.

Points 1 and 2 of the issue are settled explicitly in the commit message: placeholders now surface (history and wire no longer disagree), and reasoning-vs-budget-vs-sibling no longer decides visibility.

## Verdict

Green at head (`4ae6408da`, 28/28), reviewed at head with no findings, every review finding addressed with red-green coverage. **Merging.**

PR #3783, opened 2026-08-16
fix(npm): make ext-content-mdx an optional peer, not a runtime dependency

### LINKED ISSUE #3725: CLI-only extensions should not be runtime dependencies of the library package

kwakayama:
fix(npm): make ext-content-mdx an optional peer, not a runtime dependency

`npm install veryfront` broke a previously-clean `tsc --noEmit` in any
consumer project without skipLibCheck. The root package listed
@veryfront/ext-content-mdx under runtime `dependencies`, which drags
@mdx-js/mdx -> @types/mdx@2.0.14. That file references the *global* JSX
namespace @types/react@19 no longer declares, and tsc auto-includes
everything under node_modules/@types, so four TS2503 errors appeared in a
package the developer never imported.

The issue's suggested `optionalDependencies` does not fix it: npm installs
optional dependencies by default and only tolerates their installation
*failure*. Verified with npm 11.12.1 in the published package shape --
optionalDependencies installs 117 packages, keeps node_modules/@types/mdx
and still exits 2; an optional peer installs 4 packages, has no @types/mdx
and exits 0. So the move lands on peerDependencies +
peerDependenciesMeta.optional, reusing the mechanism
ROOT_OPTIONAL_RUNTIME_PEERS already uses.

Three consequences handled:

- Server startup calls `ensureBuiltinContentProcessor` unconditionally, so
  a missing package there would break `npx veryfront dev` for every
  project, including ones with no .mdx file. It now tolerates a
  missing-module failure and leaves the contract unregistered, deferring
  the report to the compile path, which already throws the typed
  MISSING_EXTENSION_ERROR naming @veryfront/ext-content-mdx. A real load
  failure inside an installed extension still propagates.

- The `minimal` starter ships app/about/page.mdx, so it now declares the
  extension via `firstPartyExtensions` like docs-agent does for
  ext-document-kreuzberg.

- `update_package_version` only rewrote first-party extension pins under
  `dependencies`. An RC publish would have left the optional peer pointing
  at a version that was never published; it now covers optionalDependencies
  and peerDependencies too.

The compiled binary is unchanged: compile-binary.ts embeds
extensions/ext-content-mdx/src/index.ts as a compile-time include, with no
reference to npm dependency metadata.

Refs #3725

---

## The headline: the issue's own suggested fix does not work

#3725 proposes `optionalDependencies` as the first step. That was measured and **it does not solve the problem**. npm installs optional dependencies by default — `optional` means "tolerate installation *failure*", not "skip". Measured with npm 11.12.1 against the published package shape:

| | packages installed | `node_modules/@types/mdx` | `tsc --noEmit` |
|---|---|---|---|
| `optionalDependencies` | 117 | present | **exit 2** |
| optional peer | 4 | absent | **exit 0** |

So the change lands on `peerDependencies` + `peerDependenciesMeta.optional`, reusing the `ROOT_OPTIONAL_RUNTIME_PEERS` mechanism the repo already has. Worth correcting in the issue as well as here.

## Three consequences that fall out of it

These are the parts most likely to bite, and each is handled:

1. **Server startup calls `ensureBuiltinContentProcessor` unconditionally.** A missing package would have broken `npx veryfront dev` for *every* project, including ones with no `.mdx` at all. It now tolerates a missing-module failure and defers to the compile path, which already throws the typed `MISSING_EXTENSION_ERROR` naming the package. A real load failure inside an *installed* extension still propagates — the tolerance is narrow.
2. **The `minimal` starter ships `app/about/page.mdx`**, so it declares the extension via `firstPartyExtensions`, matching what docs-agent already does for ext-document-kreuzberg.
3. **`update_package_version` only rewrote first-party pins under `dependencies`.** An RC publish would have shipped an optional peer pointing at a version that was never published. Now covers `optionalDependencies` and `peerDependencies` too. This one would have been a silent publish break.

## Acceptance against the issue

- [x] A plain strict TS project typechecks clean after `npm install veryfront` with no `skipLibCheck` — measured, exit 0
- [x] `npx veryfront dev` still renders `.mdx`, or fails with an actionable install message — the typed `MISSING_EXTENSION_ERROR` names the package
- [x] Compiled binary unchanged — `compile-binary.ts` embeds `extensions/ext-content-mdx/src/index.ts` as a compile-time include and never reads npm dependency metadata
- [x] Regression test asserting a bare consumer typechecks
- [x] `docs/getting-started/add-to-existing-project.md` updated to drop the `skipLibCheck` requirement

Closes #3725

## Summary by CodeRabbit

* **New Features**
  * Added automatic MDX extension setup for generated projects and templates containing MDX files.
  * Added warnings when MDX files are generated without the required extension installed.
  * Improved handling of optional MDX support during content compilation.

* **Documentation**
  * Clarified TypeScript configuration guidance and MDX installation behavior.
  * Updated scaffold API reference links.

* **Bug Fixes**
  * Improved package metadata and version handling for first-party extensions.

### DIFF
--- a/cli/commands/generate/command.ts
+++ b/cli/commands/generate/command.ts
@@ -6,6 +6,8 @@ import { exists, join, readTextFile } from "veryfront/fs";
 import { generateIntegration } from "./integration-generator.ts";
 import { isScaffoldType, scaffoldProjectFile } from "../../scaffold/engine.ts";
 
+const MDX_EXTENSION_PACKAGE = "@veryfront/ext-content-mdx";
+
 const PROJECT_MARKERS = [
   "veryfront.config.ts",
   "veryfront.config.js",
@@ -160,10 +162,19 @@ async function warnIfMdxExtensionMissing(
       manifest.devDependencies,
       manifest.peerDependencies,
       manifest.optionalDependencies,
-    ].some((group) => group?.["@veryfront/ext-content-mdx"] !== undefined);
+    ].some((group) => group?.[MDX_EXTENSION_PACKAGE] !== undefined);
     if (declared) return;
+    // Lockfile-aware: hard-coding `npm install` in a pnpm/yarn/bun project
+    // writes a competing package-lock.json and leaves the real lockfile stale.
+    const { detectProjectInstallTarget, formatInstallCommand } = await import(
+      "#veryfront/extensions/install-command.ts"
+    );
+    const install = formatInstallCommand(
+      MDX_EXTENSION_PACKAGE,
+      detectProjectInstallTarget(projectDir),
+    );
     cliLogger.warn(
-      "This project does not depend on @veryfront/ext-content-mdx, so the generated .mdx file will not render. Install it with: npm install @veryfront/ext-content-mdx",
+      `This project does not depend on ${MDX_EXTENSION_PACKAGE}, so the generated .mdx file will not render. Install it with: ${install}`,
     );
   } catch {
     // No readable package.json: say nothing rather than warn wrongly.
--- a/cli/commands/generate/command.ts
+++ b/cli/commands/generate/command.ts
@@ -130,4 +130,42 @@ export async function generateCommand(
   }
 
   for (const file of result.files) cliLogger.info(`Created ${file.path}`);
+  await warnIfMdxExtensionMissing(projectDir, result.files.map((file) => file.path));
+}
+
+/**
+ * Tell the user to install the MDX extension when we have just written an
+ * `.mdx` file into a project that does not declare it.
+ *
+ * The pages router scaffolds `.mdx` for `page` and `layout`. Since
+ * `@veryfront/ext-content-mdx` became an optional peer of the npm package, a
+ * project that never installed it renders those routes as an error — while
+ * this command has just reported "Created" and exited 0. The compile path
+ * already throws a typed error naming the package, but by then the developer
+ * is debugging a route they were told was fine.
+ *
+ * Best-effort: a project without a readable package.json (a Deno project, say)
+ * gets no warning rather than a false one.
+ */
+async function warnIfMdxExtensionMissing(
+  projectDir: string,
+  paths: string[],
+): Promise<void> {
+  if (!paths.some((path) => path.endsWith(".mdx"))) return;
+  try {
+    const raw = await readTextFile(join(projectDir, "package.json"));
+    const manifest = JSON.parse(raw) as Record<string, Record<string, string> | undefined>;
+    const declared = [
+      manifest.dependencies,
+      manifest.devDependencies,
+      manifest.peerDependencies,
+      manifest.optionalDependencies,
+    ].some((group) => group?.["@veryfront/ext-content-mdx"] !== undefined);
+    if (declared) return;
+    cliLogger.warn(
+      "This project does not depend on @veryfront/ext-content-mdx, so the generated .mdx file will not render. Install it with: npm install @veryfront/ext-content-mdx",
+    );
+  } catch {
+    // No readable package.json: say nothing rather than warn wrongly.
+  }
 }
--- a/cli/shared/ensure-content-processor.test.ts
+++ b/cli/shared/ensure-content-processor.test.ts
@@ -1,0 +1,109 @@
+import { assertEquals, assertRejects, assertStringIncludes } from "#veryfront/testing/assert.ts";
+import { describe, it } from "#veryfront/testing/bdd.ts";
+import { resolve, tryResolve, unregister } from "../../src/extensions/contracts.ts";
+import type { ContentProcessor } from "veryfront/extensions/content";
+import { ensureBuiltinContentProcessor } from "./ensure-content-processor.ts";
+
+/** The error Node raises for an uninstalled optional peer dependency. */
+function missingPackageError(): Error {
+  return Object.assign(
+    new Error(
+      "Cannot find package '@veryfront/ext-content-mdx' imported from " +
+        "/app/node_modules/veryfront/esm/cli/shared/ensure-content-processor.js",
+    ),
+    { code: "ERR_MODULE_NOT_FOUND" },
+  );
+}
+
+class StubContentProcessor {
+  compileMdx() {
+    return Promise.reject(new Error("unused"));
+  }
+  compileMarkdown() {
+    return Promise.reject(new Error("unused"));
+  }
+  getRemarkPlugins() {
+    return [];
+  }
+  getRehypePlugins() {
+    return [];
+  }
+}
+
+describe("cli/shared/ensure-content-processor", () => {
+  it("registers the MDX processor when the extension is installed", async () => {
+    try {
+      await ensureBuiltinContentProcessor(() =>
+        Promise.resolve({ MdxContentProcessor: StubContentProcessor })
+      );
+
+      assertEquals(
+        tryResolve<ContentProcessor>("ContentProcessor") instanceof StubContentProcessor,
+        true,
+      );
+    } finally {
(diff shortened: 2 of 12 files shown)


kwakayama:
@codex review
@coderabbitai review

Head is now cfc2807df. Since your last pass: both of your P1s are addressed — the MDX extension is now derived from the assembled project rather than the template table, and `generate` warns when it writes `.mdx` into a project without the package. Note your first P1's stated mechanism did not reproduce: the `mdx` feature has no `files` key and scaffolds zero files. The underlying gap was real and is fixed. Docs regenerated.

kwakayama:
@codex review

Head is 0bea8bdf7 (last pass was cfc2807d). Your install-command finding is addressed: the warning now routes through `detectProjectInstallTarget(projectDir)` and `formatInstallCommand()` from `src/extensions/install-command.ts` instead of hard-coding `npm install`, so a pnpm/yarn/bun project no longer gets told to write a competing lockfile. Also replaced the remaining literal package names with the existing `MDX_EXTENSION_PACKAGE` constant so the check and the message cannot drift.

kwakayama:
# Review: 92/100 — merging

## The finding that matters most: this PR contradicts its own issue

#3725 proposes `optionalDependencies` as the fix. **It does not work.** npm installs optional dependencies by default — `optional` means "tolerate installation *failure*", not "skip". Measured with npm 11.12.1 against the published package shape:

| | packages | `node_modules/@types/mdx` | `tsc --noEmit` |
|---|---|---|---|
| `optionalDependencies` | 117 | present | **exit 2** |
| optional peer | 4 | absent | **exit 0** |

So the change lands on `peerDependencies` + `peerDependenciesMeta.optional`, reusing the existing `ROOT_OPTIONAL_RUNTIME_PEERS` mechanism. Taking the issue at its word would have produced a PR that looked complete and fixed nothing. The issue text should be corrected too.

## Three consequences, each found rather than guessed

1. **Server startup would have broken for every project.** `ensureBuiltinContentProcessor` is called unconditionally, so a missing package would have broken `npx veryfront dev` even for projects with no `.mdx`. Now tolerates a missing-module failure and defers to the compile path's typed `MISSING_EXTENSION_ERROR`. A real load failure inside an *installed* extension still propagates — the tolerance is narrow, which is the part worth checking.
2. **`update_package_version` only rewrote pins under `dependencies`.** An RC publish would have shipped an optional peer pointing at a version that was never published. Silent publish break, nobody asked it to look there.
3. **The `minimal` starter ships `app/about/page.mdx`**, so it declares the extension.

## Review findings, both addressed

**MDX declaration missed the feature path (P1).** The stated mechanism did not reproduce — `templates/features/mdx/feature.json` has no `files` key, so `--features mdx` scaffolds zero files. But the gap underneath was real: the feature sets `mdx.enabled` and tips the user to "Create .mdx files in app/ directory" while declaring nothing. Declaration now derives from the assembled project — a scaffolded `.mdx` *or* the feature being selected — which also subsumes the `minimal` special case rather than naming it in a table.

**`generate` wrote `.mdx` with no declaration (P1), then hard-coded `npm install` (P2).** Both fixed; the warning now routes through `detectProjectInstallTarget`/`formatInstallCommand`, so a pnpm/yarn/bun project is not told to write a competing lockfile.

## Deductions

**−5, `templates/features/mdx/files/app/docs/*.mdx` is dead.** Three `.mdx` fixtures exist on disk and are never scaffolded, because the loader goes by `feature.json`. Either dead files or a feature-loader bug. Out of scope here, but this PR is the reason anyone knows — it should not vanish with the merge.

**−3, the compiled-binary guarantee is verified by reading, not by a test.** `compile-binary.ts` embeds the extension as a compile-time include with no reference to npm metadata, so the binary is genuinely unaffected — but nothing fails if that changes.

## Verdict

Green at head (`0bea8bdf7`, 28/28), reviewed clean at head, all five acceptance items met. **Merging.**

PR #3803, opened 2026-08-17
feat(rendering): warn on legacy project alias fallback

kojiwakayama:
## Summary

- preserve the legacy components/ project-alias fallback for compatibility, but emit one structured deprecation warning only when that fallback resolves the import
- suggest the equivalent project-relative @/ specifier for both legacy prefix insertion and legacy prefix stripping
- document project alias resolution and the migration path before the fallback is removed

## Behavior

Direct project-relative aliases such as @/Root remain silent. Legacy imports such as @/LegacyPrefix and @/components/LegacyStrip continue to resolve, and the warning identifies the original specifier plus a safe suggested replacement. The warning does not expose local absolute paths.

## RED-GREEN TDD

- RED: focused dependency-resolver coverage expected the two legacy fallback warnings and received none
- GREEN: the focused dependency and file-resolver suite passes (3 modules, 20 steps)
- GREEN: the expanded file-resolver and module-loader surface passes (14 modules, 130 steps)

## Verification

- deno task lint:ci
- targeted typecheck, formatting, and git diff --check
- public documentation validation across 125 files

This is a runtime warning and documentation change, not a Studio UI change. No screenshot is applicable.

Closes veryfront/veryfront-issue-inbox#79

### DIFF
--- a/docs/guides/project-structure.md
+++ b/docs/guides/project-structure.md
@@ -180,6 +180,16 @@ These directories are not auto-discovered. They are common project conventions.
 | `styles/`     | Global CSS files                    |
 | `middleware/` | Custom middleware functions         |
 
+## Project path aliases
+
+Use `@/` for a project-root-relative import. For example,
+`@/components/Button` resolves to `components/Button` in the project root.
+
+Veryfront temporarily supports legacy imports that add or remove the leading
+`components/` segment during resolution. When this fallback resolves an import,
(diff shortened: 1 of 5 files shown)


kojiwakayama:
@codex review\n\n@coderabbitai review

kojiwakayama:
@codex review

PR #3820, opened 2026-08-17
fix(agent): replay provider metadata across tool steps

kojiwakayama:
## Summary

- retain provider replay metadata across direct generate, stream-backed generate, legacy stream, and active stream-lifecycle tool steps
- attach the metadata only to the next internal model request
- keep thought signatures out of public messages, SSE, public schemas, memory serialization, and durable model-call events
- verify the exact signed Gemini model part precedes its matching function response on the second wire request
- regenerate affected API reference source links

## Root cause

The Google adapter already captured exact Gemini assistant parts, including `thoughtSignature`, in `providerMetadata`. The central runtime bridge discarded that metadata while normalizing generate and stream results, and the agent history builders therefore could not return it on the tool-result leg. The active stream lifecycle also discarded finish metadata while projecting its terminal snapshot. Gemini 2.5 tolerated the omission, while Gemini 3 rejects the continuation with HTTP 400 because function-call thought signatures are mandatory.

Google protocol reference: https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures

## RED-GREEN TDD

RED:

```text
deno test --preload=src/testing/preload.ts --no-check --allow-all src/agent/runtime/provider-metadata-continuation.test.ts
```

Generate and legacy stream initially made the second provider request with `providerMetadata: undefined`. After fixing those paths, an explicit active-lifecycle regression still failed with `providerMetadata: undefined`.

GREEN:

- generate, legacy stream, and active stream replay the exact metadata on the tool-result request
- a credential-free Google wire regression verifies the exact signed assistant part is replayed before the function response
- the signature is absent from returned messages, streamed SSE, and durable model-call context
- stream-backed generate preserves finish metadata

## Verification

- focused exact-head suite: 5 files, 67 steps passed
- `deno task --quiet verify:quick`
- `deno task --quiet typecheck`
- `deno task --quiet lint`
- `deno task docs:api-reference:check`
- earlier full unit suite before the active-lifecycle extension: 3,922 tests and 29,969 steps passed, plus cwd suites
- `git diff --check`

Fixes veryfront/veryfront-issue-inbox#549

## Summary by CodeRabbit

* **New Features**
  * Preserved assistant provider metadata across generated responses, streamed responses, tool-result continuations, and lifecycle events.
  * Maintained provider-specific information needed for reliable multi-step and streamed interactions.

* **Bug Fixes**
  * Corrected API reference links for the agent runtime and embedding similarity function.

* **Tests**
  * Added coverage for provider metadata continuity across generation and streaming workflows.

### DIFF
--- a/docs/api-reference/veryfront/agent.md
+++ b/docs/api-reference/veryfront/agent.md
@@ -1040,7 +1040,7 @@ Input delivered to a hosted agent-service detached execution callback.
 
 | Name                                 | Description                                       | Source                                                                                                             |
 | ------------------------------------ | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
-| `AgentRuntime`                       | Implement agent runtime.                          | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/runtime/index.ts#L1073)                   |
+| `AgentRuntime`                       | Implement agent runtime.                          | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/runtime/index.ts#L1074)                   |
 | `AgentRuntimeMessageConversionError` | Error shape for agent runtime message conversion. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/runtime/message-adapter.ts#L138)          |
 | `AgentServiceAuthError`              | Error shape for hosted service auth.              | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/service/auth.ts#L14)                      |
 | `AppendConversationRunEventsError`   | Error shape for append conversation run events.   | [source](https://github.com/veryfront/veryfront-code/blob/main/src/agent/conversation/durable-append-errors.ts#L4) |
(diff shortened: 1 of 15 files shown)


kojiwakayama:
@codex review exact head 609cb880309974999589fc3585a68e81f61f26b2

kojiwakayama:
Please run an exact-head Codex review on `0a86157541d1f3d9f6ef143e1b2169c293ff27e5`. The amended commit only adds a regression assertion that stream-backed `generate` preserves provider metadata; focused tests pass (42 steps).

kojiwakayama:
@codex review exact head `57a9313307131b35c5d43420b2a0ec4d1131b7f0`. This head adds RED-GREEN active-lifecycle coverage and a credential-free exact Gemini wire replay regression while retaining the internal-only metadata design.

kojiwakayama:
@coderabbitai review exact head `57a9313307131b35c5d43420b2a0ec4d1131b7f0`

kojiwakayama:
Addressed after merge in follow-up #3821. The suggested consume-once behavior would violate Gemini 3 sequential function calling: Google explicitly defines request 3 as retaining the first signed function call and says to include signatures for **all** function calls in the loop. See https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures#sequential_function_calling_example. The existing WeakMap replay behavior is therefore correct for every request that replays that assistant step within the current turn. #3821 adds a three-request regression requiring signature A on requests 2 and 3, plus signature B on request 3, while confirming neither leaks into public messages. PR #3820 was merged externally before this response and test-only follow-up could be pushed.

kojiwakayama:
Addressed in follow-up PR #3824. The RED reproduced request 3 replaying both the prior and current thought signatures; the GREEN consumes replay metadata only when its assistant runtime message is retained in the actual outbound provider request, preserving metadata for trimmed messages until they are later sent. Full pre-push suite is green.

kojiwakayama:
Correction to my earlier follow-up note: the consume-once suggestion is rejected after validating the provider contract. Google requires all accumulated `thoughtSignature` values for every sequential function-call step in the current turn; request 3 is defined as request 2 plus function call 2/signature 2 plus function response 2, so signature 1 remains required. Removing it would reintroduce the Gemini 3 400. Test-only PR #3821 is merged and locks `[signature1, signature2]` on request 3. Invalid follow-up #3824 is closed.

Primary source: https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures#sequential_function_calling_example

PR #3821, opened 2026-08-17
test(agent): lock multi-step Gemini signature replay

kojiwakayama:
## Summary

- add a three-request sequential tool-call regression for the provider replay fix merged in #3820
- require signature A on requests 2 and 3, and signatures A plus B on request 3
- keep both signatures absent from public result messages

## Why

CodeRabbit suggested consuming replay metadata after one request. Gemini requires the opposite within a sequential function-calling turn. Google documents request 3 as the full prior history and explicitly requires signatures for all function calls in the loop:

https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures#sequential_function_calling_example

Consume-once behavior would remove signature A from request 3 and restore the Gemini 3 validation failure. This regression locks the required protocol behavior. It changes tests only.

## Verification

```text
deno test --preload=src/testing/preload.ts --no-check --allow-all src/agent/runtime/provider-metadata-continuation.test.ts
```

Result: 5 steps passed, 0 failed.

- `deno fmt`
- `git diff --check`

Follow-up to #3820 and veryfront/veryfront-issue-inbox#549.

### DIFF
--- a/src/agent/runtime/provider-metadata-continuation.test.ts
+++ b/src/agent/runtime/provider-metadata-continuation.test.ts
@@ -36,6 +36,13 @@ function readAssistantProviderMetadata(options: unknown): unknown {
   return prompt.find((message) => message.role === "assistant")?.providerMetadata;
 }
 
+function readAllAssistantProviderMetadata(options: unknown): unknown[] {
+  const prompt = (options as { prompt?: Array<Record<string, unknown>> }).prompt ?? [];
+  return prompt.filter((message) => message.role === "assistant").map((message) =>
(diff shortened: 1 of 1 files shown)


kojiwakayama:
@codex review exact head `a0ad6304dc92a1e5df7cd3c92e2760b63f25f6ac`. This is a test-only protocol regression following the late #3820 review.

kojiwakayama:
@codex review exact head `a0ad6304daf540cfcd068930ff5e3b1f7fee4478`. Correction: this is the exact pushed head.

PR #3824, opened 2026-08-17
fix(agent): consume Gemini replay metadata once

kojiwakayama:
## Summary

Follow-up to #3820 and veryfront/veryfront-issue-inbox#549. The internal WeakMap replay metadata added by #3820 was read without being consumed, so a signature from tool step 1 could be replayed again on tool step 3 alongside the current signature. Gemini rejects that stale cross-step replay.

This change makes provider-request replay metadata single-use while preserving ordinary/history conversion behavior. Metadata is consumed only for assistant runtime messages that remain in the final outbound prompt; a trailing assistant message trimmed from the current request retains its metadata for the later request that actually sends it.

## RED → GREEN

- RED: a three-request generate flow replayed `[signature1, signature2]` on request 3 instead of only `[signature2]`.
- RED: consuming before trailing-assistant trimming lost metadata that needed to be sent after the subsequent tool result.
- GREEN: generate, legacy stream, and active stream lifecycles now send only the current step signature.
- GREEN: ordinary/history conversion remains repeatable and non-consuming; provider request conversion consumes only retained outbound metadata.

## Verification

- Full pre-push gate: 3,922 tests / 29,978 steps, 0 failed; cwd suites 13 tests / 213 steps, 0 failed.
- Focused provider/runtime suites: 110 steps green.
- Targeted `deno check`, lint, format, diff check, anti-slop, test-typecheck, public/error docs, and pinned Deno 2.7.7 API docs checks green.
- Independent exact-diff review: approve, zero findings.

Fixes veryfront/veryfront-issue-inbox#549

### DIFF
--- a/src/agent/runtime/provider-metadata-continuation.test.ts
+++ b/src/agent/runtime/provider-metadata-continuation.test.ts
@@ -36,6 +36,13 @@ function readAssistantProviderMetadata(options: unknown): unknown {
   return prompt.find((message) => message.role === "assistant")?.providerMetadata;
 }
 
+function readAssistantProviderMetadataList(options: unknown): unknown[] {
+  const prompt = (options as { prompt?: Array<Record<string, unknown>> }).prompt ?? [];
+  return prompt
+    .filter((message) => message.role === "assistant" && Object.hasOwn(message, "providerMetadata"))
+    .map((message) => message.providerMetadata);
+}
+
 function streamFrom(parts: unknown[]): ReadableStream<unknown> {
   return new ReadableStream({
     start(controller) {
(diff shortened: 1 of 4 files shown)


kojiwakayama:
@codex review

kojiwakayama:
@coderabbitai review

kojiwakayama:
Closing as invalid after `main` merged the authoritative regression in #3821. Google’s Gemini 3 protocol requires every accumulated function-call thought signature in the current sequential tool turn: request 3 must contain signature A and signature B. This PR’s consume-once behavior removes signature A and would restore the provider 400. The late review suggestion on #3820 is therefore rejected on protocol grounds, not implemented.

Primary source: https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures#sequential_function_calling_example

PR #3845, opened 2026-08-18
fix(rendering): compile client and module output for the render mode

kojiwakayama:
## Summary

This is a production behaviour change. Two transform call sites still hardcoded `dev: true` after #3841, so the hosted production render path shipped development-compiled output on every request. Per `src/transforms/pipeline/stages/compile.ts`, `dev: true` means `minify: false`, `treeShaking: false` and an inline sourcemap, so production paid a payload-size cost and disclosed the project source. Both sites are fixed here, and each one's cache identity is updated in the same commit.

Refs veryfront/veryfront-issue-inbox#555

## Site A: the client hydration bundle

`bundleComponentForClient` (`src/rendering/component-handling.ts`) hardcoded `dev: true` and had no mode parameter. `src/rendering/page-renderer.ts` reaches it on every TSX page render that does not supply a cached client module.

- `bundleComponentForClient` takes the compile mode, defaulting to production.
- `handleComponentPage` derives it from the render mode with the same `options?.mode === "development"` line the SSR half of the file already used.

**Cache identity, updated alongside the flag.** `buildComponentHydrationCacheHash` did not include the compile mode. Fixing the flag alone would let a development bundle be served to a production render, or the reverse, because both modes resolve to one hydration cache entry. The hash now carries the compile mode.

## Site B: the MDX ESM module fetcher

`transformResolvedModuleSource` (`src/transforms/mdx/esm-module-loader/module-fetcher/source-transform.ts`) hardcoded `dev: true`, and `createModuleFetcherContext` had no way to express the render mode. The chain is `ssr-module-loader/loader.ts` to `vf-module-resolver.ts` to `createModuleFetcherContext` to `fetchAndCacheModule`, and it fires for every `/_vf_modules/*` import of every SSR module.

- The module fetcher context takes a compile mode, defaulting to production.
- `resolveVfModuleImports` requires one, so a caller cannot silently omit it.
- The SSR module loader passes `this.options.dev`.

**Cache identity, mandatory in the same commit.** `getTransformCacheKey` had no compile-mode segment, and it feeds the distributed transform cache, whose entries are shared across requests and across instances with a TTL. Fixing the transform without the key would have been worse than the original bug: one instance could serve a development-compiled module to a hosted production render, and the reverse.

The compile mode is now part of `getMdxModuleCacheVariant`, the single segment builder behind the distributed transform key (`getTransformCacheKey` to `buildMdxEsmTransformCacheKey`), the module path-cache key (`getVersionedPathCacheKey` and `cacheModule`), and the SSR loader and render orchestrator lookups. All of these share one key space, so every writer and reader of it had to agree; a partial fix would have left the same cross-mode reuse through the path cache, which short-circuits before any transform runs.

Development artifacts take an `on:compile-dev` segment and production keeps the unsegmented key. The MDX-ESM cache namespace is rolled in the same change (the compile-mode split is named in the namespace schema sample), so the legacy entries written under the unsegmented key, all of them development-compiled, cannot be read back by a production render.

Artifact filenames already hash the transformed code, so the two modes cannot collide on disk and the artifact directory layout is unchanged.

## Scope

The compiled-MDX entry path (`esm-module-loader/loader-helpers.ts`) carries no render mode yet, so it keeps the development compile mode it has always used. That is now explicit rather than implied by a hardcoded literal, and the compile mode in the cache identity keeps its artifacts isolated from the production-compiled ones. Threading a render mode into `MDXRenderer.loadModuleESM` is a separate change.

No `RenderContext.mode` value ("development" | "production") is forwarded into a `"preview" | "production"` field. Only the compile-mode boolean is threaded.

## Tests

- `src/rendering/component-handling.test.ts`: a production-mode `handleComponentPage` render emits a hydration bundle with no inline sourcemap, minified and tree-shaken; a development-mode render still emits the debuggable bundle. Both assert on the emitted code. A third case pins that the two modes take separate hydration cache entries.
- `src/modules/react-loader/ssr-module-loader/vf-module-resolver.test.ts`: a development resolve runs first, then a production resolve of the same module for the same project and content source. The production render gets a different artifact with no inline sourcemap and no dead code, which fails if either the flag or the cache identity is missing.
- `src/modules/react-loader/ssr-module-loader/loader.test.ts`: an end-to-end `SSRModuleLoader` production load of a module with a `/_vf_modules/` import, asserting the cached artifact.
- `src/transforms/mdx/esm-module-loader/module-fetcher/cache-keys.test.ts`: the compile mode changes the distributed transform key and the path-cache key, an unset mode reads as production, and the compile segment stays separate from the pin and server-external segments.
- `src/transforms/mdx/esm-module-loader/module-fetcher/source-transform.test.ts`: the requested compile mode reaches the transform.

Each new behaviour test was confirmed to fail against the unfixed code.

## Verification

```
deno test --preload=src/testing/preload.ts --no-check --allow-all src/rendering/    111 passed, 0 failed
deno test ... src/transforms/                                                       159 passed, 0 failed
deno test ... src/modules/react-loader/                                              26 passed, 0 failed
full src unit suite                                                                3639 passed, 0 failed
tests/ integration suite                                                            302 passed, 0 failed
deno fmt --check src/ cli/ react/ templates/                                        clean
deno lint on touched files, deno task typecheck                                     clean
deno task lint:sanitizer-baseline, lint:test-typecheck, lint:module-boundaries       clean
```

### DIFF
--- a/src/modules/react-loader/ssr-module-loader/loader.test.ts
+++ b/src/modules/react-loader/ssr-module-loader/loader.test.ts
@@ -51,6 +51,7 @@ import {
   buildMdxEsmModuleRecoveryCacheKey,
   buildMdxEsmPathCacheKey,
 } from "#veryfront/transforms/mdx/esm-module-loader/cache-format.ts";
+import { MDX_MODULE_DEV_COMPILE_VARIANT } from "#veryfront/transforms/mdx/esm-module-loader/module-fetcher/cache-keys.ts";
 import type { RuntimeAdapter } from "#veryfront/platform/adapters/base.ts";
 import type { FileSystem } from "#veryfront/platform/compat/fs.ts";
 import type { ModuleCacheEntry } from "./types.ts";
@@ -266,10 +267,12 @@ describe("SSRModuleLoader", { sanitizeResources: false, sanitizeOps: false }, ()
     const originA = __ssrModuleLoaderInternals.getMdxEsmCacheVariant({
       dependencyPinningCacheKey: CANONICAL_PIN_KEY,
       moduleServerOrigin: "https://a.example",
+      dev: false,
     });
     const originB = __ssrModuleLoaderInternals.getMdxEsmCacheVariant({
       dependencyPinningCacheKey: CANONICAL_PIN_KEY,
       moduleServerOrigin: "https://b.example",
+      dev: false,
     });
 
     assert(originA?.startsWith(`${CANONICAL_PIN_KEY}:origin:`));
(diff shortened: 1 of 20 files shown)


kojiwakayama:
@codex review exact head 6c82e90b5a4721eaf4497e567d05f7c5585f0894. Please focus on compile-mode propagation, every reader and writer of the hydration and MDX ESM cache identities, legacy cache invalidation, hosted preview semantics, and positional-call compatibility.

kojiwakayama:
@coderabbitai review

kojiwakayama:
@coderabbitai review

PR #3852, opened 2026-08-18
fix(rendering): support pre-runtime release hydration

kwakayama:
## Summary

Restore compatibility for immutable releases created before hydration runtime artifacts existed.

- prefer a release-baked hashed runtime
- serve a release-baked legacy runtime through static handling
- use the serving runtime only for manifest-verified pre-contract releases with no runtime artifact
- fail closed for unknown, malformed, and contract-era missing artifacts
- emit a bounded structured warning whenever the compatibility fallback is used

## Root cause

The v0.1.1243 renderer required a release-baked hydration runtime. Existing customer releases can contain neither a hashed nor legacy runtime, so SSR failed before rendering HTML.

## Validation

- Red: a legacy-only release fixture failed against the previous selector with Release is missing its versioned hydration runtime.
- Green: focused selector, script, handler, and static-service suite, 83 steps passed.
- Green: browser regression covers a runtime-less pre-contract release, 3 steps passed.
- Formatting, touched-file lint, and diff check passed.

## Validation gaps

The local repository-wide typecheck is blocked before checking sources because templates/manifest.generated.ts is already stale on main. The original test runner also requires a non-committed Deno 2.9 compatibility shim for the existing native-brand-check bootstrap issue.

### DIFF
--- a/docs/guides/deploying.md
+++ b/docs/guides/deploying.md
@@ -52,14 +52,20 @@ the runtime from the currently serving Veryfront process.
 This pairing is a compatibility boundary. Keep the hydration runtime for as
 long as its immutable release can be deployed or served. Retire the release and
 its browser artifacts together; never delete only the hydration runtime or
-redirect its hashed URL to newer bytes. A release that is missing its single
-versioned runtime fails rendering instead of falling back to a potentially
-incompatible runtime.
+redirect its hashed URL to newer bytes.
+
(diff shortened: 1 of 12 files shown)


kwakayama:
Superseded by #3853, which has the incident-accurate hosted-release fallback and release-candidate path.

PR #610, opened 2026-03-10
Avoid remote deno.json probe when absent

kojiwakayama:
## Summary
- avoid remote deno.json file fetches when the release or environment file list already proves the file is absent
- teach the file list index to track known paths and resolve extensionless paths directly from cached file-list content
- add regression coverage for dotted extensionless paths and known-missing deno.json reads

## Testing
- deno test --allow-env src/platform/adapters/fs/veryfront/read-operations.test.ts

PR #3863, opened 2026-08-19
docs(jira): note that an Atlassian OAuth app must have sharing enabled

kwakayama:
The self-hosted Jira setup guide covers the callback URL and the four scopes but
never mentions **Distribution**, which is the step that decides whether anyone
other than the app owner can connect.

Per [Atlassian's OAuth 2.0 (3LO) docs](https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/):

> When you create an OAuth 2.0 (3LO) app it's private by default. This means that
> only you can install and use it. If you want to distribute your app to other
> users, you must enable sharing.

Leaving it off fails in a way that is easy to misdiagnose. The user reaches
Atlassian, picks their site, and gets Atlassian's own "Something went wrong"
page. The flow never reaches the deployment's callback, so there is nothing in
its logs to look at. And the app owner's own account connects either way, so
testing with it proves nothing — which is the trap the added step calls out
explicitly.

Docs only; no code or connector changes.

### DIFF
--- a/docs/guides/integrations/jira.md
+++ b/docs/guides/integrations/jira.md
@@ -34,7 +34,7 @@ managed OAuth connection.
 4. Open **Distribution** and enable sharing. An OAuth 2.0 (3LO) app is private
    when created, so only the account that owns it can authorize. Leave sharing
    off and every other user reaches Atlassian, picks their site, and gets
-   Atlassian's own "Something went wrong" page — your deployment never sees the
+   Atlassian's own "Something went wrong" page. Your deployment never sees the
    callback, so nothing surfaces in your logs. Test with an account outside the
    app owner's Atlassian organization; the owner's account succeeds either way
    and proves nothing.
--- a/docs/guides/integrations/jira.md
+++ b/docs/guides/integrations/jira.md
@@ -31,7 +31,14 @@ managed OAuth connection.
    - `offline_access` to receive refresh tokens. Veryfront uses refresh tokens
      to keep a user connection active; omit it only when the connection must
      not be refreshable.
-4. Set the OAuth app client ID and client secret as project environment variables:
+4. Open **Distribution** and enable sharing. An OAuth 2.0 (3LO) app is private
+   when created, so only the account that owns it can authorize. Leave sharing
+   off and every other user reaches Atlassian, picks their site, and gets
+   Atlassian's own "Something went wrong" page — your deployment never sees the
+   callback, so nothing surfaces in your logs. Test with an account outside the
+   app owner's Atlassian organization; the owner's account succeeds either way
+   and proves nothing.
+5. Set the OAuth app client ID and client secret as project environment variables:
 
 | Variable                  | Value                         |
 | ------------------------- | ----------------------------- |
--- a/docs/guides/integrations/jira.md
+++ b/docs/guides/integrations/jira.md
@@ -32,12 +32,22 @@ managed OAuth connection.
      to keep a user connection active; omit it only when the connection must
      not be refreshable.
 4. Open **Distribution** and enable sharing. An OAuth 2.0 (3LO) app is private
-   when created, so only the account that owns it can authorize. Leave sharing
-   off and every other user reaches Atlassian, picks their site, and gets
-   Atlassian's own "Something went wrong" page. Your deployment never sees the
-   callback, so nothing surfaces in your logs. Test with an account outside the
-   app owner's Atlassian organization; the owner's account succeeds either way
-   and proves nothing.
+   when created, so only the account that owns it can authorize. Atlassian will
+   not save the toggle until you also fill in **Vendor** and **Privacy policy**
+   (plus **Terms of service** where it asks for one), so have those URLs ready.
+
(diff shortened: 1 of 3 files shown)


kwakayama:

## Review: 79/100 — no blocking findings, four accuracy items

Highest score of the four PRs, and the only one with nothing blocking. Reviewed for factual
accuracy against Atlassian's documentation and for public-repo safety.

| | |
|---|---|
| correctness & safety | 34/40 |
| tests & evidence | 18/25 |
| scope & clarity | 13/15 |
| spec compliance | 14/20 |

**Public-repo safety: clean.** No customer identity, internal hostname, incident narrative, or
private tracker reference in the diff. ASCII-punctuation lint satisfied.

Four accuracy items to fix before merge:

1. **The step dead-ends without naming the required fields.** Atlassian will not save the
   sharing toggle until Vendor and Privacy policy (and ToS where applicable) are filled in.
2. **The test instruction is wrong, and self-contradicting.** "Test with an account outside the
   app owner's Atlassian *organization*" — a private 3LO app is per-**user**, so an in-org
   colleague fails too. It should read "any Atlassian account other than the app owner's".
   As written it also contradicts the sentence two lines above it.
3. **Confluence is uncovered**, though #610 named it explicitly. It uses the same Atlassian app
   and the same `ATLASSIAN_CLIENT_ID`/`ATLASSIAN_CLIENT_SECRET`, so enabling sharing covers
   both — and there is no separate Confluence guide to carry that sentence.
4. **Nothing pins the step.** Adding "Distribution" to the jira.md snippet list in
   `tests/docs/guide-contracts.test.ts` stops a later docs edit silently deleting it and
   staying green.

Item 2 is the one that matters most: as written it would send someone to test with a colleague,
watch it fail, and conclude sharing had not worked.

kwakayama:

## Round-2 findings addressed — including two I caused

The previous round's fix introduced two blocking problems. Both are corrected at `a1032dc14`.

### I made the Confluence sentence worse

I added *"Both connectors use the same Atlassian app … enabling sharing once unblocks both."*
Review called it unfollowable, and it is. Verified against the connector:

| | Jira | Confluence |
|---|---|---|
| callback | `/api/auth/jira/callback` | `/api/auth/confluence/callback` |
| scopes | 4 | 8, all different |
| credentials | `ATLASSIAN_CLIENT_ID` / `_SECRET` | same |

So the app and therefore the sharing toggle are shared, but an operator following only this
guide would register Jira's callback and Jira's four scopes, enable sharing, and still have a
Confluence connector that cannot complete a flow. The sentence now says what is actually shared
and points at the Confluence README for the rest.

### The Confluence README had no sharing step at all

Which is what #610 fix item 4 asked for, and which my new sentence then contradicted. It now
carries the same step, phrased for its own callback and scopes.

### Also

- The guide-contract pin is widened from `"Distribution"` to also cover `"Privacy policy"` and
  `"owning **account**"`, so the two corrections this round needed cannot silently regress either.
  A pin on the heading alone would not have caught the wrong-instruction defect.
- Dropped *"our logs"* from the contract comment per AGENTS.md — public-facing text should not
  speak from the vendor's side.

### Verification

```
deno test --allow-all tests/docs/guide-contracts.test.ts   1 passed (78 steps)
grep -P '[^\x00-\x7F]' docs/guides/integrations/jira.md    clean (ASCII docs lint)
CI on a1032dc14                                            24 success, 6 skipped
```

The `tests (rsc browser e2e)` failure the review saw at `7b427755d` was an apt-get Chromium
install timing out — infrastructure, not this diff. It is green on the current head.

Public-repo safety re-checked: no customer identity, internal hostname, incident narrative, or
private tracker reference in the diff.

kwakayama:

## Closing — wrong repository for this content

The failure this documents was **our managed Atlassian OAuth app** being private: veryfront.org
and veryfront.com share one 3LO app, sharing was never enabled on it, and so every customer
outside the app owner's account failed for 30+ days while the owner's own account worked.

That is an operational property of infrastructure we run, not guidance for someone standing up
their own deployment. It belongs in a runbook in `veryfront-infrastructure`, next to the other
things we operate, where whoever is on call can find it. Putting it in this repo's public
self-hosted guide meant the people who actually need it — us — would never look there.

Moving it. Superseded by a runbook in veryfront-infrastructure.

### What was in here, so nothing is lost

- The Distribution/sharing step, including the Vendor and Privacy policy fields Atlassian
  requires before the toggle will save
- The correction that a private 3LO app is restricted to the owning **account**, not the owning
  organization — so testing with an in-org colleague fails too, and testing with the app owner
  proves nothing
- That Confluence shares the app and the sharing toggle, but needs its own callback URL and its
  own eight scopes
- A `guide-contracts` pin so the step could not be silently deleted

### Open question for the self-hosted case

A BYO-app operator hits the identical wall, and the public guide still says nothing about it.
That is a genuinely separate concern from our managed app, and I have not resolved it here.
Worth a smaller follow-up PR that adds only the Distribution step to the self-hosted section,
without the operational detail that belongs to us.

Refs veryfront/veryfront-issue-inbox#610

PR #3865, opened 2026-08-19
fix(agent): rank tool_search by term selectivity, not field precedence

kwakayama:
## Problem

`tool_search` ranked a match on a **common term in a tool name** above a match on the **distinctive term in a tool description**, so on any multi-word query the result window filled with alphabetically-first same-prefix tools and reported `miss: false`.

Two mechanisms, both in the multi-term fallback:

1. **Pure OR with rank compared before coverage.** A candidate matching any single term qualified, and results sorted by strongest matched field first. A tool matching only `list` in its name (name rank) therefore beat a tool matching `github` in its description (description rank), no matter how many terms the latter also matched. With 27 authorized tools whose names contain `list` and a five-result cap, the window never reached the relevant one.

2. **`__` destroyed before the query is classified.** Normalization rewrites `_` to a space, so a canonical `<namespace>__<tool_id>` was word-split into its generic halves. `jira__list_projects` became the terms `jira list projects`, and the unrelated platform tool `list_projects` matched two of the three in its **name** and won outright. That is worse than returning noise: the caller receives a callable tool in the wrong namespace.

`miss` was derived from result count rather than relevance, so a caller was told a search had succeeded when nothing relevant was found.

## Change

**Weight terms by inverse document frequency.** A term matching few candidates now outweighs one matching many, so a rare namespace term dominates a generic verb.

**Require a selective term.** A candidate must match at least one term appearing in at most half the catalog. A near-ubiquitous term can still contribute score but can no longer be the sole reason a candidate is returned, so a query with no discriminating match reports a miss instead of returning filler.

**Resolve canonical ids before normalization.** `<namespace>__<tool_id>` is parsed off the raw query. An id the run is authorized for still resolves by exact name through the unchanged whole-query path; one it is not resolves to its namespace.

Field precedence and score weight are keyed off a single `ToolSearchMatchField` union so the two orderings cannot drift apart, and each candidate's name and description are normalized once at snapshot time rather than on every term comparison.

## Behaviour change to review

One pre-existing expectation changed. For the query `create_file update_file project file`, `sandbox_write_file` is no longer returned: its only matched term was `file`, which appears in **every** candidate and so carries no information about which tool was meant. The other two results and their order are unchanged.

That result was filler produced by the same mechanism this PR fixes, so the old expectation encoded the defect. Calling it out explicitly because "existing test updated" is exactly the shape of a test bent to fit an implementation, and it deserves a second opinion rather than my say-so.

## Verification

- Written test-first: 3 new tests failed before the change, 2 more were written as regression guards and passed throughout, so a real fix was distinguishable from a broken one.
- Canonical-id resolution is asserted for `jira__list_projects`, `jira__list_comments`, `jira__list_sites`, `jira__search_users`, and `github__list_repos` — every id whose generic half collides with a platform tool.
- A bare `list_projects` query still resolves to the platform tool; exact-name and single-term queries are unchanged.
- `deno task test:unit` 3945 passed / 0 failed; typecheck, `fmt:check`, and lint clean.

Note for anyone hitting red CI here: the full unit suite is flaky under parallel execution independently of this change. `main` at `a4ec5b280` fails it too, on a different timing-sensitive test.

### DIFF
--- a/src/agent/runtime/tool-exposure.test.ts
+++ b/src/agent/runtime/tool-exposure.test.ts
@@ -924,13 +924,28 @@ it("tool search ranks a rare description term above a common name term", () => {
 });
 
 it("tool search resolves a canonical integration tool id to its namespace", () => {
-  const result = searchToolExposure({
-    query: "jira__list_projects",
-    authorized: integrationDiscoveryCatalog(),
-    state: createToolExposureState(),
-  });
-
-  assertEquals(result.matches.map((match) => match.name), ["get_integration"]);
+  // Every one of these word-splits onto a platform tool that would otherwise win:
+  // `list_projects`, `list_...`, and the `search_*` family all share the generic
+  // half of the id. Resolving the namespace is what keeps the wrong tool out.
+  for (
+    const query of [
+      "jira__list_projects",
+      "jira__list_comments",
+      "jira__list_sites",
+      "jira__search_users",
+      "github__list_repos",
+    ]
+  ) {
+    assertEquals(
+      searchToolExposure({
+        query,
+        authorized: integrationDiscoveryCatalog(),
+        state: createToolExposureState(),
+      }).matches.map((match) => match.name),
+      ["get_integration"],
+      `canonical id ${query} must resolve to its namespace`,
+    );
+  }
 });
 
 it("tool search prefers an authorized integration tool over its namespace catalog entry", () => {
--- a/src/agent/runtime/tool-exposure.test.ts
+++ b/src/agent/runtime/tool-exposure.test.ts
@@ -961,6 +961,35 @@ it("tool search prefers an authorized integration tool over its namespace catalo
   assertEquals(result.matches[0]?.name, "jira__list_projects");
 });
 
+it("tool search does not let a same-named local tool satisfy a canonical id", () => {
+  // `jira__list_projects` and the local id `jira_list_projects` normalize to the
+  // same text, and the registry permits any local id without `__`. A phrase match
+  // must not hand back the local tool for a canonical query.
+  const result = searchToolExposure({
+    query: "jira__list_projects",
+    authorized: [
+      ...integrationDiscoveryCatalog(),
+      definition("jira_list_projects", "A project-local tool that is not the Jira integration"),
+    ],
+    state: createToolExposureState(),
+  });
+
+  assertEquals(result.matches.map((match) => match.name), ["get_integration"]);
+});
+
+it("tool search accepts canonical ids with the authorization layer's segment grammar", () => {
+  // The authoritative grammar allows consecutive and trailing separators; search
+  // must not be stricter, or it disagrees with authorization about what an id is.
+  assertEquals(
+    searchToolExposure({
+      query: "github__list-issues-",
+      authorized: integrationDiscoveryCatalog(),
+      state: createToolExposureState(),
+    }).matches.map((match) => match.name),
+    ["get_integration"],
+  );
+});
+
 it("tool search reports a miss when no candidate matches a selective term", () => {
   assertEquals(
     searchToolExposure({
--- a/src/agent/runtime/tool-exposure.test.ts
+++ b/src/agent/runtime/tool-exposure.test.ts
@@ -990,6 +990,20 @@ it("tool search accepts canonical ids with the authorization layer's segment gra
   );
 });
 
+it("tool search matches a full-coverage candidate in a single-tool catalog", () => {
+  // With one candidate every term matches everything, so no term can clear the
+  // selectivity floor. Full term coverage is its own evidence: there is no
+  // competing candidate for the floor to protect against.
+  assertEquals(
+    searchToolExposure({
+      query: "create project",
+      authorized: [definition("create_file", "Create a project file")],
+      state: createToolExposureState(),
+    }).matches.map((match) => match.name),
+    ["create_file"],
+  );
+});
+
 it("tool search reports a miss when no candidate matches a selective term", () => {
   assertEquals(
     searchToolExposure({
--- a/src/agent/runtime/tool-exposure.test.ts
+++ b/src/agent/runtime/tool-exposure.test.ts
@@ -210,20 +210,23 @@ it("tool search normalizes ASCII case and underscores as spaces", () => {
   );
 });
 
-it("tool search falls back to deterministic whitespace terms after a phrase miss", () => {
+it("tool search scores fallback terms by selectivity after a phrase miss", () => {
   const fileCatalog = [
     definition("create_file", "Create a project file"),
     definition("update_file", "Update a project file"),
     definition("sandbox_write_file", "Write only inside the sandbox filesystem"),
   ];
 
+  // `file` matches every candidate and therefore says nothing about which one the
+  // caller meant; `create` and `update` each match exactly one. `sandbox_write_file`
+  // matches no term but `file`, so returning it would be filler, not a result.
   assertEquals(
     searchToolExposure({
       query: "create_file update_file project file",
       authorized: fileCatalog,
       state: createToolExposureState(),
     }).matches.map((match) => match.name),
-    ["create_file", "update_file", "sandbox_write_file"],
+    ["create_file", "update_file"],
   );
 });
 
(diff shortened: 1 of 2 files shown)


kwakayama:
Both P2 findings were real and are fixed in `23e3e363b`. Verified each against the source before acting rather than taking the review at face value.

**Route canonical IDs before normalized phrase matches** — confirmed. `assertLocalToolId` rejects only ids containing `__`, so `jira_list_projects` is a permitted local id, and it normalizes to the same text as `jira__list_projects`. The phrase pass returned it. That is the same wrong-namespace failure this PR set out to fix, reached by a different path, and worse than the original because a local tool is executable.

Canonical ids are now classified before any normalized matching, and only the real canonical name satisfies a canonical query.

Fixing that exposed a second hole in the same area that the finding did not name: the namespace fallback still ranked `jira_list_projects` above `get_integration`, because its *name* contains `jira`. Namespace discovery now requires evidence rather than a word collision — a sibling tool in the same namespace, or a tool that documents the namespace. A local tool that merely contains the word is not the integration.

**Reuse the canonical integration segment contract** — confirmed, and the justification I wrote for duplicating it was factually wrong. My grammar was stricter than `CANONICAL_INTEGRATION_TOOL_SEGMENT`, so ids with consecutive or trailing separators fell through to term splitting. The comment claimed the runtime must not take a value dependency on the integrations policy layer; `src/agent/runtime/index.ts` already does. Now imports `parseIntegrationToolIdentity` directly, so search and authorization cannot disagree about what a canonical id is. `source-policy.ts` imports only a type, so there is no cycle.

**Coverage added**

- A local `jira_list_projects` must not satisfy the canonical query `jira__list_projects`.
- `github__list-issues-` resolves to its namespace, pinning the grammar to the authorization layer's.

42 tests pass. `typecheck`, `fmt:check`, `lint:module-boundaries`, `lint:dependency-boundaries`, `lint:core-deps`, and `check:circular` all clean.

One note for reviewers on automated status here: CodeRabbit reported nothing because this repo is under its 10-star threshold, not because it reviewed and found the PR clean.

kwakayama:
@codex review

kwakayama:
Second Codex pass, `23e3e363bf`. One finding real and fixed in `ce8d2c493`; one is a repeat that no longer applies.

**Allow matches in single-candidate catalogs — real, fixed.**

Correct, and a regression I introduced. With one candidate every term matches everything, so `documentFrequency == total` for every term and nothing can clear the floor. A sole `create_file` described "Create a project file" missed for `create project`, exactly as described.

The floor exists to suppress filler, and filler only means something when there is a better candidate to prefer instead. A candidate matching *every* term is not filler however common those terms are, so full coverage now satisfies the guard on its own. Candidates matching only a subset still need a selective term, so the original defect stays closed. Covered by a new test using the reported example.

**Route canonical IDs before normalized phrase matches — already fixed in the commit under review.**

This is the same finding, verbatim, from the previous pass; it was addressed in `23e3e363b`, the commit this review ran against. `rankToolExposureMatches` now parses the canonical id first and returns from that branch before reaching the phrase pass at the flagged line, so that early return is only reachable for non-canonical queries.

The exact example given is covered by a passing test: with both the catalog and a local `jira_list_projects` authorized, the query `jira__list_projects` returns `get_integration`, not the local tool. I have left the code as is rather than change it to acknowledge the comment. Happy to be shown wrong if there is a path I have missed.

43 tests pass. `typecheck`, `fmt:check`, `lint:module-boundaries`, `lint:dependency-boundaries` clean. Previous head was green across CI including all coverage shards, `ci (lint)`, unit, integration, bun, and rsc browser e2e.

kwakayama:
@codex review

kwakayama:
Third pass, `ce8d2c493`. Both findings real; fixed in `12a92969e`. I owe a correction on one of them.

**Route canonical IDs before normalized phrase matches — I was wrong to dismiss this.**

I said twice that this was a stale repeat, because the exact repro given (`jira__list_projects` against a local `jira_list_projects`) is covered by a passing test. That much is true. But rather than dismiss it a third time I probed the surrounding shapes, and the underlying claim holds for one the comment never names:

```
"jira__list_projects"   -> ["get_integration"]        correct
"jira__list__projects"  -> ["jira_list_projects"]     WRONG - local tool wins
"JIRA__LIST_PROJECTS"   -> ["get_integration"]        correct
```

A query carrying `__` that is not a *valid* canonical id fell out of the canonical branch and into phrase matching, where normalization collapsed it onto the local id. `__` is the reserved integration namespace separator and `assertLocalToolId` forbids it in local ids, so any query carrying it is asking for an integration tool even when the rest is malformed. Those queries now stay on the namespace path, keyed off the leading segment.

The repro was wrong and the conclusion was right. Flagging that plainly because I twice asserted otherwise on this thread.

**Normalize canonical namespaces before matching — real, fixed.**

Correct. `parseIntegrationToolIdentity` returns the namespace raw, but every candidate's name and description has already had underscores rewritten to spaces, so a namespace containing `_` could never match and discovery reported a miss. Namespace text evidence is now matched normalized; sibling identity still compares raw ids.

No integration in the live catalog currently has an underscore in its name, so this was latent rather than user-visible, but the grammar permits it.

**Coverage added**

- `jira__list__projects`, `JIRA__LIST_PROJECTS`, and a trailing-space variant must all resolve to the catalog, never to a colliding local tool.
- `foo_bar__list_items` resolves its namespace.

45 tests pass. `typecheck` and `fmt:check` clean. `ce8d2c493` was green across all CI before this push, including all 8 coverage shards, `ci (lint)`, unit, integration, and coverage gate.

kwakayama:
@codex review

kwakayama:
Fourth posting of **Route canonical IDs before normalized phrase matches**, unchanged text. The variant it *did* catch was fixed in `12a92969e`. I do not believe anything actionable remains, and here is the evidence rather than an assertion.

Probed every namespace-shaped input I could construct against a catalog seeded with three deliberately colliding local tools (`jira_list_projects`, `github_list_repos`, `foo_bar_list_items`) plus the platform's own `list_projects`:

```
ok   "jira__list_projects"     -> ["get_integration"]
ok   "jira__list__projects"    -> ["get_integration"]
ok   "jira____list_projects"   -> ["get_integration"]
ok   "JIRA__LIST_PROJECTS"     -> ["get_integration"]
ok   "  jira__list_projects  " -> ["get_integration"]
ok   "github__list_repos"      -> ["get_integration"]
ok   "foo_bar__list_items"     -> ["get_integration"]
ok   "jira__"                  -> ["get_integration"]
ok   "jira__list-projects-"    -> ["get_integration"]
ok   "jira__List_Projects"     -> ["get_integration"]
LEAK "__list_projects"         -> ["list_projects", "jira_list_projects"]
```

Ten of eleven resolve to the catalog and never to a colliding local tool, including the exact repro in the comment.

The eleventh is a **leading** `__`, so the namespace segment is empty. There is no integration to resolve to, and returning the platform's `list_projects` for a query normalizing to `list projects` is the correct answer, not a leak of integration intent. I am leaving it.

I got this wrong once already on this thread by dismissing the finding twice before probing, so I am deliberately showing the matrix instead of claiming it is a duplicate. If there is a shape missing from that list that reaches a local tool, point at it and I will fix it.

Not requesting another review, since the comment has now repeated verbatim across four passes against three different commits including two that changed exactly this code path.

**Status:** 45 tests pass. `typecheck`, `fmt:check`, `lint:module-boundaries`, `lint:dependency-boundaries` clean. `ce8d2c493` was fully green across CI; `12a92969e` is green so far and still running.

kwakayama:
**Match canonical namespaces as complete tokens** — real, fixed in `9ba9975e9`. Verified `exa` is in `src/integrations/schema.ts` before acting.

Substring evidence let `exa__search` admit any tool whose description or parameter description merely contained `example`. Namespace evidence now requires a complete normalized token, with non-alphanumeric boundaries, which also keeps multi-word namespaces such as `foo_bar` working.

Fixing it surfaced something the finding did not mention. Checking description evidence directly dropped a guard that field precedence had been providing implicitly: two of my own tests went red because a local tool whose *description* mentions Jira started qualifying as documenting the namespace. That guard is now explicit rather than incidental. A non-canonical tool carrying the namespace in its name is a normalization coincidence, not the integration, whatever its description happens to mention.

Re-ran the full shape matrix against a catalog seeded with `jira_list_projects`, `github_list_repos`, `foo_bar_list_items`, `list_projects`, `run_sample` ("Runs the example workflow"), and `exact_match`:

```
ok "jira__list_projects"   ok "github__list_repos"    ok "exa__search"
ok "jira__list__projects"  ok "foo_bar__list_items"   ok "exa__contents"
ok "JIRA__LIST_PROJECTS"   ok "jira__list-projects-"
```

All resolve to `get_integration`; none leak a local tool.

**Route canonical IDs before normalized phrase matches** — fifth posting, text unchanged. Evidence in the matrix above and in my previous comment; the repro given passes. Not requesting another review pass on it.

46 tests. `typecheck` and `fmt:check` clean. `12a92969e` was green across CI including all 8 coverage shards, `ci (lint)`, unit, integration, bun, and coverage gate.

PR #3873, opened 2026-08-20
fix(cli): read stdin bytes on Node so hidden prompts accept input

kwakayama:
## Description

`readStdinByteSync` had branches for Deno and Bun only. Node fell through to a bare `return null`. Because `null` means EOF, `promptPassword` broke out of its read loop on the first iteration and returned an empty string.

The npm package ships no native binary, so `bin/veryfront.js` always takes the Node fallback under `npx`. Every hidden prompt was unusable there:

```
$ npx veryfront@0.1.1245 login --provider openai

  Enter your OpenAI API key.
  Get one at: https://platform.openai.com/api-keys

  API key:   x No API key provided.
```

The prompt never waits for a keystroke. `cli/utils/env-prompt.ts` could not collect sensitive env vars either. Raw mode was torn down before the user typed, so the key echoed in plaintext and stayed in the shell input buffer after the process exited.

### Fix

Add a Node branch that resolves `node:fs` lazily through `process.getBuiltinModule` and reads one byte from fd 0. `getBuiltinModule` needs Node 22.3+, which `bin/veryfront.js` already requires before loading the bundled JS.

libuv puts a raw-mode TTY in non-blocking mode, so `readSync` throws `EAGAIN` whenever no byte is ready yet. That means "nothing typed", not EOF. Treating it as EOF is what ends the read early, so the Node branch parks 5ms via `Atomics.wait` and retries.

The async reader (`readStdinLine`) already supports Node through `createNodeStdinReader`. Only the synchronous byte path was missing one.

### Verification

Driven through a real pty against an installed `veryfront@0.1.1245`, with the same patch applied to the shipped `esm` build:

```
BASELINE    API key:   x No API key provided.
            sk-fake-key-typed-by-test          <- leaked to the shell

PATCHED     API key: *************************
            x Invalid API key (HTTP 401).      <- key captured and sent
```

The 401 is the expected result for the fake key used in the test. It proves the prompt captured input and reached the provider.

Gates run on this branch: `deno fmt`, `deno check`, `deno lint`, `deno test src/platform/compat/process/`, `deno test cli/auth/ src/platform/compat/stdin.test.ts` (15 passed, 180 steps), and `deno task docs` (no drift).

## Related Issue(s)

None filed in this repo.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

Five tests cover the Node branch: a plain byte read, the `EAGAIN` retry (asserting it keeps waiting and parks between attempts), a zero-byte read, and a closed stdin. The branch is extracted as `testReadStdinByteSyncNode` with an injected `node:fs` surface, so it stays testable under the Deno test runner.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronous stdin reading in Node.js environments.
  * Automatically retries temporary read interruptions and correctly handles successful reads and end-of-input conditions.
  * Prevents stdin read failures from disrupting processing by safely treating unrecoverable errors as end-of-input.

### DIFF
--- a/src/platform/compat/process/lifecycle.test.ts
+++ b/src/platform/compat/process/lifecycle.test.ts
@@ -2,7 +2,24 @@ import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import { isBun } from "../runtime.ts";
-import { getV8HeapSizeLimit } from "./lifecycle.ts";
+import { getV8HeapSizeLimit, type NodeReadSyncFs, testReadStdinByteSyncNode } from "./lifecycle.ts";
+
+function errorWithCode(code: string): Error {
+  return Object.assign(new Error(code), { code });
+}
+
+/** A node:fs stand-in that replays a scripted sequence of readSync outcomes. */
+function fakeFs(steps: Array<number | Error>): NodeReadSyncFs {
+  let index = 0;
+  return {
+    readSync(_fd, buffer, offset, _length, _position) {
+      const step = steps[index++];
+      if (step instanceof Error) throw step;
+      buffer[offset] = step ?? 0;
+      return step === undefined ? 0 : 1;
+    },
+  };
+}
 
 describe("platform/compat/process/lifecycle", () => {
   it("rejects Bun's moving node:v8 compatibility heap limit", () => {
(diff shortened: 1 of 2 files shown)


kwakayama:
@coderabbitai review

kwakayama:
Both review findings applied in `ea775ef`.

- **Retry delay assertion**: the test now asserts `sleeps` equals `[5, 5]` rather than just the call count. Verified it has teeth by mutation: setting `STDIN_EAGAIN_SLEEP_MS` to 0 fails the test with `AssertionError: Values are not equal`.
- **Internal import alias**: the module under test is now imported through `#veryfront/platform/compat/process/lifecycle.ts`.

Gates re-run after the last edit: `deno fmt --check`, `deno check`, `deno lint`, and `deno test src/platform/compat/process/ cli/auth/ src/platform/compat/stdin.test.ts` (18 passed, 208 steps).

PR #3874, opened 2026-08-20
Remove `veryfront login --provider`, which configured nothing

kwakayama:
## Problem

`veryfront login --provider anthropic|openai` prompted for an API key, validated it against the provider's API, wrote it to `~/.config/veryfront/tokens/<provider>` at `0600`, and printed "API key configured".

Nothing ever read it back. `readProviderToken` had no caller outside tests, and every credential read goes through `getEnv()` — `getAnthropicEnvConfig()` in `src/config/env.ts`, gating direct-provider routing via `hasDirectProviderCredentials()` in `model-resolution.ts`. The command reported success while configuring nothing.

It shipped that way in #786 (2026-04-02) and stayed for four months. Its acceptance criteria only ever covered storage — prompt, validate, store, list, delete, display in `whoami`. None covered inference actually using the key, so the gap passed review. The command is also undocumented: zero hits for `login --provider` in `docs/`.

## Why remove rather than wire it up

Provider credentials come from the environment, which is where every direct-provider code path already reads them.

Wiring the store in would add a third credential tier that is machine-global and invisible in the project. Two developers with identical code could get different inference paths and different bills, which then needs a startup notice, a credential version marker, and a re-activation flow purely to manage the ambiguity it introduces. Keeping a key out of the project is better served by a secret-manager wrapper, which resolves per command and leaves nothing at rest.

## Behavior after this change

`login --provider` and `logout --provider` fail with guidance rather than being ignored — ignoring the flag would silently run a platform login instead:

```
'veryfront login --provider' has been removed. It stored a key that was never used for inference.
  Set ANTHROPIC_API_KEY in your environment or .env file instead.
  To keep the key out of the project, wrap the command with your secret manager,
  for example: op run --env-file=.env.local -- veryfront dev
  Any key stored by the old command is still on disk at ~/.config/veryfront/tokens/ and can be deleted.
```

`--json` keeps its existing structured usage error, unchanged and still covered by `cli/auth/exit-code.integration.test.ts`.

Nothing functional breaks: the command did not work, so no working setup depends on it. `whoami` no longer prints "API key configured", which was the misleading part.

## Also in this change

**Default-model credential mismatch is now reported.** Independent of the removal, and applies to environment credentials. `DEFAULT_AGENT_MODEL` is `openai/gpt-5.4-nano`; with no gateway and only another provider's key, an agent with no `model` set resolved to direct OpenAI and failed against the wrong vendor. It now names the mismatch. Resolution stays deterministic — no silent substitution, so a project resolves identically on every machine. The gateway path for logged-in users is unaffected and covered by a regression test.

**Providers guide** documents the secret-manager pattern (`op run --env-file`, and the Doppler/Infisical/Vault equivalents) as the supported way to keep a key out of the project.

## Verification

- 691 tests pass, 0 fail, across `cli/auth`, `cli/shared`, `cli/utils`, `src/errors`, `src/agent/runtime`, `src/config`, and `cli/router.test.ts`.
- Gates green: `fmt:check`, `lint`, `lint:cli-boundary`, `docs:errors:check`, `docs:public:check`, `docs:api-reference:check`.
- Net -158 lines.

See veryfront/veryfront-issue-inbox#673.


kwakayama:
@coderabbitai review

kwakayama:
@coderabbitai review

PR #3876, opened 2026-08-20
fix(provider): honor a custom Google endpoint via GOOGLE_GEMINI_BASE_URL

kwakayama:
## Problem

Google was the only hosted provider whose endpoint could not be configured. `OPENAI_BASE_URL`, `ANTHROPIC_BASE_URL`, and `MISTRAL_BASE_URL` all work; there was no equivalent for Gemini.

The capability was already built and tested everywhere except the env read:

- `GoogleRuntimeConfig` declares `baseURL?: string` (`extensions/ext-llm-google/src/google-provider.ts:85`)
- it is threaded into `getGoogleGenerateContentUrl`, `getGoogleStreamGenerateContentUrl`, and `getGoogleEmbeddingUrl` (lines 579, 606, 666)
- all three builders accept `baseURL ?? DEFAULT_GOOGLE_BASE_URL` (`provider-endpoints.ts:73, 84, 92`)
- the extension tests already cover it, against `https://example.google.test/v1beta`

Only `getGoogleGenAIEnvConfig()` never returned one, so `model-registry.ts` had nothing to pass and pinned every request to the default host.

## Change

`getGoogleGenAIEnvConfig()` now reads **`GOOGLE_GEMINI_BASE_URL`** — the name [Google's own gemini-cli](https://github.com/google-gemini/gemini-cli) uses for a custom Gemini Developer API endpoint, so a proxy or regional endpoint is configured the same way here as in the rest of the ecosystem.

`src/embedding/resolve.ts` had the identical hardcode and gets the same fix.

### The part that would have broken silently

The outbound fetch is origin-bound. It has to track the same base URL the request builders use:

```ts
fetch: createOriginBoundOutboundFetch(config.baseURL ?? DEFAULT_GOOGLE_BASE_URL),
```

Leaving that pinned to the default while the builders used a custom host blocks every request *before* it is sent. The new `model-registry.test.ts` case covers exactly this — I verified it fails when the guard is reverted to the default, so it is not a vacuous test.

## Not in scope: Vertex AI

Vertex is a different backend, not a base URL override. It has a different URL shape (`{location}-aiplatform.googleapis.com/v1/projects/{project}/locations/{location}/publishers/google/models/...`) and uses Application Default Credentials rather than an API key, so setting a base URL alone would produce 404s and auth failures.

Supporting it properly means a backend switch on `GOOGLE_GENAI_USE_VERTEXAI` with `GOOGLE_CLOUD_PROJECT` and `GOOGLE_CLOUD_LOCATION` plus ADC handling. Worth doing separately — it is also the route to pinning Gemini to an EU region.

## Verification

- 618 tests pass across `src/config`, `src/provider`, `src/embedding`, `src/agent/runtime`. Three are new: two on the env read, one end-to-end through the registry asserting the request URL.
- The end-to-end test was confirmed to fail without the fix.
- Green: `fmt:check`, `lint`, `lint:dependency-boundaries`, `lint:module-boundaries`, `docs:api-reference:check` (regenerated — the config surface changed).

Docs for the new variable follow in #3875, which currently documents Google as "not configurable".

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a custom Google Gemini base URL for model and embedding requests.
  * Google API key configuration continues to support both supported environment variable names.

* **Bug Fixes**
  * Requests now use the configured custom endpoint and fall back to Google’s default endpoint when none is provided.

* **Tests**
  * Added coverage for custom endpoint handling and default behavior.

* **Documentation**
  * Updated API reference source links.


kwakayama:
@coderabbitai review

kwakayama:
@coderabbitai review

Issue #3918, opened 2026-08-21
[BUG] Anthropic output_config.format rejects outputSchema without additionalProperties:false, and the real error is discarded


afiodorov:
### Describe the Bug

`buildResponseFormat` in `src/agent/output-schema.js` deliberately leaves `strict` unset on the response format it builds:

> `strict` is left unset: strict structured output additionally requires every property to be required and `additionalProperties: false`, which a schema with optional fields does not satisfy, and a rejected request is worse than a locally validated one.

That reasoning holds for OpenAI, where non-strict `json_schema` mode tolerates a schema without those constraints. It does not hold for Anthropic: `output_config.format` (built in `extensions/ext-llm-anthropic/src/anthropic-request-builder.ts`, `buildAnthropicOutputConfig`) unconditionally requires `additionalProperties: false` on every object-typed (sub)schema, independent of the framework's own `strict` flag. A schema built via plain `defineSchema((v) => v.object({...}))`, with every field required and no `.optional()` anywhere — i.e. a schema that already satisfies the "every property required" half of strict mode — still gets rejected by the live Anthropic API with a 400.

This is also unusually hard to diagnose, because the framework discards the actual cause. `buildProviderError` (`src/provider/runtime-loader/provider-http.js`) reads the response body into `rawBody`/`parsedBody`, but for the generic (non-specially-handled) status branch it throws only:

```
Provider request failed with status 400
```

The real Anthropic error —

```json
{"type":"error","error":{"type":"invalid_request_error","message":"output_config.format.schema: For 'object' type, 'additionalProperties' must be explicitly set to false"}}
```

— never reaches the caller. It only surfaces by patching `provider-http.js` locally to log the raw response text before it's thrown away.

### Steps To Reproduce

1. Declare an agent with `model: "anthropic/<any model>"` and an `outputSchema` built via `defineSchema((v) => v.object({ field: v.string() }))()` — no `.optional()` fields anywhere, so the schema already satisfies "every property required."
2. Call `agent.generate({ input: ... })`.
3. Observe a 400 from Anthropic; the thrown error says only `"Provider request failed with status 400"`, with no indication of the actual cause.
4. Call `.strict()` on the same schema (`v.object({...}).strict()`) and repeat — the request now succeeds, because `.strict()` is what actually sets `additionalProperties: false` in the emitted JSON Schema.

### Link to Minimal Reproducible Example

[...]

PR #3922, opened 2026-08-21
fix(chat): name the open-object schema rejection instead of a generic service error

kojiwakayama:
Addresses the third point in #3918 ("the real error is discarded"), stacked conceptually on #3921.

## The reported premise needs correcting first

The issue asks that `buildProviderError` stop discarding the provider's response body, on the grounding that it is "read into `rawBody`/`parsedBody` and then thrown away." **It is not thrown away.** For a structured 400 it is attached to the error non-enumerably:

```ts
// src/provider/runtime-loader/provider-http.ts
return isStructuredInvalidRequest || isStructuredVeryfrontCreditProblem
  ? preserveStructuredResponseBody(requestError, rawBody)
  : requestError;
```

Keeping it out of `error.message` is a deliberate, test-locked boundary, not an oversight. Two tests encode it:

- `does not surface provider error body contents` — asserts a body containing `private provider payload <TOKEN>` never reaches the message.
- `preserves structured 400 details for internal classification without enumerating them` — asserts the body is preserved for classification while staying out of both `message` and `JSON.stringify(error)`.

Provider bodies can quote request content, so inlining them into a message that gets logged and shown to end users would reverse a security decision. **This PR does not do that.**

## The real gap

The body is preserved so `parseProviderError` can turn it into a curated, safe message. For this error class there was no mapping, so it fell through to the generic default. Verified directly:

```
PARSED: {"code":"EXTERNAL_SERVICE_ERROR","message":"LLM provider service error"}
```

"LLM provider service error" points nowhere. The thing that has to change is the caller's own `outputSchema`, and nothing in the surfaced error says so — which is exactly the dead end the issue describes, just one layer up from where it was diagnosed.

## Fix

Add a curated mapping alongside the existing assistant-prefill and context-length ones:

```
OUTPUT_SCHEMA_NOT_CLOSED
"The model rejected the output schema because an object in it allows additional
 properties. Add .strict() to the object in your outputSchema so it sets
 additionalProperties: false."
```

Matching is on the two things the wording is built from — the property name and the closure requirement — rather than a fixed sentence, since providers reword these messages.

With #3921 merged, framework-built schemas no longer trigger this at all. This still covers raw JSON Schema `outputSchema`s and any schema that explicitly declares an open object, which #3921 deliberately leaves alone.

## Tests (red → green)

| test | before | after |
|---|---|---|
| names the schema requirement instead of a generic service error | FAIL | pass |
| does not echo the provider's raw message | pass | pass |
| leaves an unrelated invalid_request_error on the generic path | pass | pass |

The two that were already green are the guards: one holds the secrecy boundary, the other prevents the matcher from over-capturing.

## Verification

- New test fails on `main` for the stated reason, passes with the fix.
- `src/chat` + `src/provider` suites green: `88 passed (448 steps) | 0 failed`.
- `deno check`, `deno lint`, `deno fmt` clean.

### DIFF
--- a/src/chat/provider-errors.test.ts
+++ b/src/chat/provider-errors.test.ts
@@ -302,15 +302,56 @@ describe("chat/provider-errors", () => {
       },
     };
 
+    const EXPECTED = {
+      code: "OUTPUT_SCHEMA_NOT_CLOSED",
+      message:
+        "The model rejected the output schema because an object in it allows additional properties. " +
+        "Set additionalProperties: false on that object -- add .strict() if the outputSchema was " +
+        "built with defineSchema(), or set the property directly on a raw JSON Schema.",
+    };
+
     it("names the schema requirement instead of a generic service error", () => {
       // Without a mapping this lands on EXTERNAL_SERVICE_ERROR ("LLM provider
       // service error"), which points nowhere: the caller cannot tell that
       // their own outputSchema is what needs changing.
-      assertEquals(parseProviderError(providerError(OPEN_OBJECT_REJECTION)), {
-        code: "OUTPUT_SCHEMA_NOT_CLOSED",
-        message:
-          "The model rejected the output schema because an object in it allows additional properties. Add .strict() to the object in your outputSchema so it sets additionalProperties: false.",
-      });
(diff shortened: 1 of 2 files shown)


kojiwakayama:
@codex review

PR #3923, opened 2026-08-21
fix(ssr-module-loader): keep background cache writes reachable so tests can drain them

kojiwakayama:
Fixes the intermittent `coverage shard 4/8` failure that has been red on unrelated PRs (seen on #3919).

## Symptom

```
SSR module distributed cache on a local dev server
error: Leaks detected:
  - An async operation to remove a file or directory was started in this test,
    but never completed.
    at DiskCacheBackend.set (src/cache/backends/disk.ts:648)
```

Every individual step in that suite reports `ok`; the failure is attributed to the suite, and it does not reproduce locally in isolation (3/3 clean runs).

## Cause

The stack points at `disk.ts`, but nothing there is missing an `await` — every `unlink` in that file is awaited. The orphan starts a level up, in the loader:

```ts
// src/modules/react-loader/ssr-module-loader/loader.ts:1159
void setInRedis(contentCacheKey, transformed, { ... })
```

Nobody holds that promise. So `dev-disk-persistence.test.ts` runs `loadRawModule`, reaches its `finally`, and deletes the cache directory **while the write is still in flight**. The writes `rename` then fails, which routes it into the temp-file cleanup `unlink` — an operation still running after the test that started it has finished. That is precisely the reported stack:

```
at disk.ts:659            ← unlink, in the catch (write failed)
at DiskCacheBackend.withMutation
at DiskCacheBackend.set
```

On an unloaded machine the write wins the race and the suite passes. On a loaded CI runner it does not. That is the whole flake.

## Fix

Not blocking a render on a cache write is correct, so that stays. The problem is not that the write is unawaited — it is that it is **unreachable**, so no teardown can wait for it. `trackBackgroundWrite` registers the promise and `drainBackgroundWrites` waits for the set to drain, letting a caller that is about to delete the directory settle the disk instead of racing it.

The loader test drains before its `finally` removes the directory.

This is not test-only scaffolding: an unowned write that fails after its directory is gone is a hazard on any teardown path, not just this one.

## Tests

`cache/background-writes.test.ts` covers the drain:

- waits for a tracked write to settle
- settles even when the tracked write rejects (the directory-removed case) without an unhandled rejection
- returns immediately when nothing is in flight
- waits for a write registered *while* a drain is already in progress (why the drain loops rather than awaiting one snapshot)

## Verification

- `src/modules/react-loader` suites green under `--trace-leaks`: `27 passed (248 steps) | 0 failed`.
- `deno check`, `deno lint` (49 files), `deno fmt` clean.
- Full pre-push suite green.

The race itself cannot be asserted directly — it is timing-dependent by nature, which is why it only appears under CI load — so the tests pin the drain contract that removes it instead.

### DIFF
--- a/src/modules/react-loader/ssr-module-loader/cache/background-writes.test.ts
+++ b/src/modules/react-loader/ssr-module-loader/cache/background-writes.test.ts
@@ -1,0 +1,60 @@
+import "#veryfront/schemas/_test-setup.ts";
+import { assertEquals } from "#veryfront/testing/assert.ts";
+import { describe, it } from "#veryfront/testing/bdd.ts";
+import { drainBackgroundWrites, trackBackgroundWrite } from "./redis.ts";
+
+describe("SSR module cache background writes", () => {
+  it("waits for a tracked write to settle", async () => {
+    const write = Promise.withResolvers<void>();
+    trackBackgroundWrite(write.promise);
+
+    let drained = false;
+    const draining = drainBackgroundWrites().then(() => {
+      drained = true;
+    });
+
+    // Give the drain every chance to resolve early if it is not really waiting.
+    await new Promise((resolve) => setTimeout(resolve, 0));
+    assertEquals(drained, false);
+
+    write.resolve();
+    await draining;
+    assertEquals(drained, true);
+  });
+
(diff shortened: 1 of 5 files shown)


kojiwakayama:
@codex review

PR #3930, opened 2026-08-21
fix(workflow): reconcile a resumed loop iteration against the run's node states

### LINKED ISSUE #3927: [BUG] A step after a waitForApproval inside the same loop iteration never runs on resume
### Describe the Bug

A step declared *after* a `waitForApproval` inside the same `loop` iteration never runs once that wait is approved and the run resumes. No error, no re-execution — the run just reports the loop `completed` one step short, permanently.

This is a different gap from #3916 (already fixed in 0.1.1246) in the same area: that one was a loop wiping a *sibling's* node state; this one is the loop's *own* per-iteration resume bookkeeping never learning that the wait it's holding was answered.

Root cause: when a wait is approved, `ApprovalManager.processDecision` patches the decision straight onto the top-level, persisted run record (`runtime/workflow-run-control.js:113-124`, `reconcileApprovalDecision`):

```js
const runPatch = {
  context: { ...run.context, [operation.nodeId]: decisionContext },
  nodeStates: {
    ...run.nodeStates,
    [operation.nodeId]: { nodeId: operation.nodeId, status: "completed", output: {...}, ... },
  },
};
```

It then calls `executor.resume(runId)`. Since the run's top-level `status` is still `"waiting"` at that point, `executeUnwrapped`'s "a suspended composite is resumable" fallback re-schedules the loop node itself (`executor/dag/index.js:56-67`):

```js
if (!startFromNode && run.status === "waiting") {
  for (const [nodeId, degree] of inDegree) {
    ...
    if (state?.status === "running" && node && RESUMABLE_COMPOSITE_TYPES.has(node.config.type)) {
      ready.push(nodeId);
    }
  }
}
```

`executeLoopNodeStrategy` runs again, but reads its **own**, separately cached iteration snapshot from `context.<loopId>_loop_state`, taken when the loop suspended — i.e. *before* the approval above happened (`loop-node-strategy.js:20-25`):

```js
resumeIterationNodeStates = existingLoopState.iterationNodeStates;
...
const iterationNodeStates = resumeIteration === iteration && resumeIterationNodeStates
  ? { ...resumeIterationNodeStates }   // the wait node is still "running" in here, frozen at suspend time
  : {};
```

This snapshot is never reconciled against the authoritative, already-updated top-level `nodeStates` the loop strategy is itself called with — even though that map already has the wait as `"completed"`. The loop then re-runs its iteration's own child graph with the stale snapshot, using a synthetic run object whose `status` is hardcoded to `"running"`, never `"waiting"` (`loop-node-strategy.js:58-70`):

```js
const result = await runtime.executeChildGraph(steps, {
  id: `${node.id}_iter_${iteration}`,
  status: "running",               // always "running", never "waiting"
  nodeStates: iterationNodeStates, // the wait node still "running" in here
  ...
});
```

Inside that child `executeUnwrapped` call, `getReadyNodes` only treats an absent/`"pending"`/`"failed"` node as ready (`executor/dag/graph.js:50-57`). The wait node is `"running"`, so it's excluded — and the "suspended composite is resumable" fallback that rescued the *outer* loop node cannot fire here, because it's gated on `run.status === "waiting"`, and this child run's status is hardcoded `"running"`. Nothing marks the wait node done, so the step depending on it never becomes ready either. `ready` is empty from the start, so this `executeChildGraph` call returns `completed: true` immediately, the dependent step is never scheduled, and the loop is told its iteration finished successfully.

**Control experiment:** the identical `[waitForApproval, toolStep]` pair placed *outside* any loop resumes and runs correctly, because the wait's completed state lives directly in the real top-level `nodeStates` map the resuming `executeUnwrapped` call reads via plain `getReadyNodes` — no stale, loop-private snapshot sits in between. This isolates the loop's own resume bookkeeping (`resumeIterationNodeStates` / `context.<id>_loop_state`) as the differentiator, not waits in general.

Practical impact: any workflow that reruns a lookup/action in a loop, asks a human a question via `waitForApproval` inside that same loop, and then does anything with the answer in a later step of the *same* iteration (fold it into state, log it, branch on it) silently loses that step on every resume.

### Steps To Reproduce

1. `git clone https://github.com/afiodorov/veryfront-wait-in-loop-bug && cd veryfront-wait-in-loop-bug`
2. `npm install`
3. `npm run repro`
4. Observe: a single-iteration `loop` containing `[waitForApproval("pauseInLoop"), toolStep("inLoopAfterWait")]` is approved and resumed, but `inLoopAfterWait` never runs — while the identical `[waitForApproval, toolStep]` pair placed at the top level (not inside a loop) resumes and runs correctly.

Minimal shape (full script is under 100 lines, see `repro.ts`):

```ts
workflow({
  id: "wait-in-loop-bug-demo",
  steps: [
    loop("loopWithWait", {
      while: (_ctx, loopCtx) => loopCtx.iteration < 1,
      maxIterations: 1,
      steps: [
        waitForApproval("pauseInLoop", { message: "..." }),
        toolStep("inLoopAfterWait", inLoopAfterWait.tool),   // never runs
      ],
    }),
    // control: identical shape, no loop -- this one works
    waitForApproval("pauseTopLevel", { message: "..." }),
    toolStep("topLevelAfterWait", topLevelAfterWait.tool),   // runs fine
  ],
});
```

### Link to Minimal Reproducible Example

https://github.com/afiodorov/veryfront-wait-in-loop-bug

### Expected Behavior

`inLoopAfterWait` should run exactly once after `pauseInLoop` is approved, the same way `topLevelAfterWait` correctly runs once after `pauseTopLevel` is approved. A resuming loop's per-iteration node-state bookkeeping should not be able to disagree with the authoritative top-level `nodeStates` it's handed on the very node that was just resolved by the decision that triggered the resume.

Actual output from `npm run repro`:

```
--- Expected vs actual invocation counts ---
  inLoopAfterWait     (inside the loop) expected 1, got 0
  topLevelAfterWait   (top level, control) expected 1, got 1

BUG REPRODUCED: the step after the wait inside the loop never ran, even though
the wait was approved and the run reports the loop as completed.
Control experiment: the identical [wait, step] pair outside the loop ran
correctly -- confirms the loop, not waits in general, is the differentiator.
```

### Environment Information

- Veryfront version: 0.1.1246
- Runtime: Node.js 22.23.2
- OS: macOS 15 (Darwin 25.5.0)

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

kojiwakayama:
Fixes #3927.

## Problem

A step declared after a `waitForApproval` **inside the same loop iteration** never runs once that wait is approved. No error — the loop reports the iteration completed, one step short, permanently.

The reporter's root-cause analysis is correct and I confirmed each link.

A loop keeps a private snapshot of its in-flight iteration in `context.<id>_loop_state`, frozen at the moment it suspended. Whatever resolves the wait afterwards patches the run's **own** `nodeStates` and calls resume — it has no idea the loop is holding a separate copy:

```ts
const iterationNodeStates = resumeIteration === iteration && resumeIterationNodeStates
  ? { ...resumeIterationNodeStates }   // wait node still "running", frozen at suspend time
  : {};
```

From there it fails quietly, and it takes three independent things lining up:

1. `getReadyNodes` treats only absent / `pending` / `failed` nodes as ready, so the wait node — still `running` in the stale copy — is excluded.
2. The "a suspended composite is resumable" fallback that rescues the *outer* loop node cannot fire for the child graph: it is gated on `run.status === "waiting"`, and the synthetic run the loop builds for its iteration hardcodes `status: "running"`.
3. Nothing marks the wait done, so the step depending on it never becomes ready either — and an empty ready set reads as a finished graph.

So `executeChildGraph` returns `completed: true` having scheduled nothing, and the loop is told its iteration succeeded.

## Fix

Reconcile the snapshot against the run's map before replaying the iteration. The run wins for every node it also knows about, so the node that was just resolved is seen as resolved. Nodes the run has never heard of stay as the snapshot left them, which keeps a sibling from outside the loop out of the iteration's graph.

## Why the control experiment holds

The same `[wait, step]` shape outside a loop works, and the equivalent `branch` case already has a passing test — both because their node states live directly in the run's map, with no loop-private copy in between. That isolates the loop's resume bookkeeping as the differentiator, exactly as the report says.

## Tests (red → green)

Added `runs a step declared after a wait in the same iteration once the wait resolves` in `src/workflow/executor/dag/index.test.ts`. It drives the real sequence: run to suspension, apply the node-state patch `ApprovalManager.processDecision` persists, resume with `status: "waiting"`, then assert the dependent step ran exactly once.

Fails on `main` (`order` is empty, the step never runs); passes with the fix.

The existing `loop node sibling isolation` tests still pass, including `removes child node states from previous dynamic loop iterations` — the guard that this reconciliation does not resurrect states a dynamic iteration deliberately dropped.

## Verification

- New test fails on `main` for the stated reason, passes with the fix.
- Full workflow suite green: `77 passed (775 steps) | 0 failed`.
- `deno check`, `deno lint`, `deno fmt` clean.

### DIFF
--- a/src/workflow/executor/dag/index.test.ts
+++ b/src/workflow/executor/dag/index.test.ts
@@ -1051,6 +1051,70 @@ describe("DAGExecutor", () => {
       assertEquals(result.nodeStates["before"]!.status, "completed");
     });
 
+    it("runs a step declared after a wait in the same iteration once the wait resolves", async () => {
+      // The loop keeps its own per-iteration snapshot in `<id>_loop_state`,
+      // taken when it suspended. The approval that resumes the run patches the
+      // authoritative top-level nodeStates, not that snapshot -- so without
+      // reconciling the two, the loop replays its iteration with the wait
+      // still "running", nothing becomes ready, and the child graph reports
+      // completed with the dependent step never scheduled.
+      const order: string[] = [];
+      const trackingExecutor = new MockStepExecutor(new Map(), (node) => {
+        order.push(node.id);
+        return { success: true, output: node.id, executionTime: 1 };
+      });
+      const exec = new DAGExecutor({ stepExecutor: trackingExecutor });
+      const nodes: WorkflowNode[] = [
+        {
+          id: "the-loop",
+          dependsOn: [],
+          config: {
+            type: "loop",
+            maxIterations: 1,
+            while: (_context: WorkflowContext, loop: LoopExecutionContext) => loop.iteration < 1,
+            steps: [
+              {
+                id: "inner-wait",
+                dependsOn: [],
+                config: { type: "wait", waitType: "approval", message: "approve?" } as any,
+              },
+              { id: "after-wait", dependsOn: ["inner-wait"], config: { type: "step" } as any },
+            ],
+          } as any,
+        },
+      ];
+
+      const first = await exec.execute(nodes, createTestRun());
+      assertEquals(first.waiting, true);
+      assertEquals(order, []);
+
+      // What ApprovalManager.processDecision persists before calling resume:
+      // the decision lands on the top-level node state, never on the loop's
+      // private snapshot.
+      const second = await exec.execute(
+        nodes,
+        createTestRun({
+          status: "waiting",
+          nodeStates: {
+            ...first.nodeStates,
+            "inner-wait": {
+              nodeId: "inner-wait",
+              status: "completed",
+              output: { approved: true },
+              attempt: 1,
+              startedAt: new Date(),
+              completedAt: new Date(),
+            },
+          },
+          context: first.context,
+        }),
+      );
+
+      assertEquals(second.completed, true);
+      assertEquals(order, ["after-wait"]);
+      assertEquals(second.nodeStates["after-wait"]!.status, "completed");
+    });
+
     it("removes child node states from previous dynamic loop iterations", async () => {
       const nodes: WorkflowNode[] = [
         {
--- a/src/workflow/executor/dag/loop-node-strategy.ts
+++ b/src/workflow/executor/dag/loop-node-strategy.ts
@@ -90,9 +90,11 @@ export async function executeLoopNodeStrategy(
     runtime.abortSignal?.throwIfAborted();
 
     // On resume, rehydrate the in-flight iteration's child node states so its
-    // already-completed steps are skipped instead of re-executed (H9).
+    // already-completed steps are skipped instead of re-executed (H9),
+    // reconciled against the run's own map so the node that was just resolved
+    // is not replayed as still pending.
     const iterationNodeStates = resumeIteration === iteration && resumeIterationNodeStates
-      ? { ...resumeIterationNodeStates }
+      ? reconcileIterationNodeStates(resumeIterationNodeStates, nodeStates)
       : {};
     // Only rehydrate once; subsequent iterations start fresh.
     resumeIterationNodeStates = undefined;
@@ -216,4 +218,31 @@ export async function executeLoopNodeStrategy(
     }),
     waiting: false,
   };
+}
+
+/**
+ * Overlay the run's authoritative node states onto a resumed iteration's
+ * snapshot.
+ *
+ * The snapshot is frozen at the moment the loop suspended. Whatever resolves
+ * the wait afterwards -- an approval decision, a signal -- patches the run's
+ * own `nodeStates` and then resumes; it has no idea the loop is holding a
+ * private copy. Replaying the iteration from that copy leaves the wait node
+ * `running`, so `getReadyNodes` excludes it, the step depending on it never
+ * becomes ready, and the child graph reports completion having scheduled
+ * nothing -- losing the step silently while the iteration looks successful.
+ *
+ * The run's map wins for every node it also knows about. Nodes it has never
+ * heard of stay as the snapshot left them, so this cannot pull a sibling from
+ * outside the loop into the iteration's graph.
+ */
+function reconcileIterationNodeStates(
+  snapshot: Record<string, NodeState>,
+  runNodeStates: Record<string, NodeState>,
+): Record<string, NodeState> {
+  const reconciled: Record<string, NodeState> = { ...snapshot };
+  for (const [nodeId, state] of Object.entries(runNodeStates)) {
+    if (nodeId in reconciled) reconciled[nodeId] = state;
+  }
+  return reconciled;
 }


kojiwakayama:
@codex review

kwakayama:
## Review: 91/100 - Approve (post-merge audit)

**This is a post-merge audit of `8d0ce3c2f0`. It found nothing blocking, and the merge was sound.**

The fix is correct, reachable from the real approval path, and idempotent across repeated suspend and resume inside one iteration. I verified each of those rather than assuming them, and the evidence is at the bottom. The two findings worth acting on are follow-ups, not defects in this change: the underlying fail-open that made #3927 silent is still open, and the persisted loop snapshot collides with what #3926 is about to enforce.

### Score breakdown
| Axis | Score | Note |
|---|---|---|
| Correctness | 27/30 | Reachable, idempotent, covers every wait type. Leaves the fail-open that made the bug silent. |
| Test quality | 17/20 | Real fail-first, drives the real executor. Replays only half of what `processDecision` persists. |
| Scope discipline | 15/15 | 2 files, +95/-2, one helper, nothing unrelated. |
| Design fit | 13/15 | Fixed where the divergent copy is owned. Undocumented dependence on globally unique child ids. |
| Security & safety | 10/10 | No injection, no secrets, no unbounded recursion, O(n) helper. |
| Docs & hygiene | 9/10 | Excellent helper doc and accurate PR body. Names a "signal" resolver that does not exist. |

### Findings

**1. Not inert, confirmed by tracing the real path.** `ApprovalManager.processDecision` (`src/workflow/runtime/approval-manager.ts:267`) calls `reconcileWorkflowRunControl`, whose `reconcileApprovalDecision` (`src/workflow/runtime/workflow-run-control.ts:284-303`) writes both `run.context[nodeId]` and `run.nodeStates[nodeId] = { status: "completed", output: { approved, ... } }` under the **child** node id, then resumes. That lands in `executeLoopNodeStrategy` at `src/workflow/executor/dag/loop-node-strategy.ts:97`. The test's premise is faithful to production.

**2. Every wait type is covered, not just approvals.** All of them funnel through `executeWaitNode` (`src/workflow/executor/dag/index.ts:774-802`), which sets `status: "running"` and `waiting: true` regardless of `waitType`; `waitType` is only carried as approval metadata. So there is no duration or timeout variant left unfixed.

**3. Idempotent across repeated suspend/resume in one iteration.** I built a `[wait-a, mid, wait-b, tail]` loop body and resumed twice. Result: `order === ["mid", "tail"]`. `mid` did **not** re-run when the iteration resumed a second time, because the reconciled snapshot carries it as `completed`. Concurrency is separately guarded upstream by the `updateRunIfStatus(..., expectedWorkerId)` compare-and-set, so two workers cannot both apply the decision.

**4. The fail-open behind #3927 is still open (filed as veryfront-issue-inbox#701).** Strip the resolved wait node out of the run's map and resume: the loop returns `completed: true`, `waiting: false`, with **zero** steps scheduled. Measured directly: `PROBE_MISSING completed=true waiting=false order=[]`. That is the same silent-success shape as the original bug. This change removes the one input that produced it in practice, but the class stays: `executeChildGraph` still reads "empty ready set" as "finished graph" instead of "stuck graph". Not reachable on today's approval path, since `reconcileApprovalDecision` writes context and node states in a single conditional update, so this is a follow-up, not a blocker. The durable fix is for the child graph to distinguish "nothing ready because everything is done" from "nothing ready but unfinished nodes remain" and fail closed with a named error.

**5. Interaction with #3926, reproduced.** #3926 makes the workflow context JSON-enforceable: fatal values throw, lossy ones (`Date`) are coerced and logged. Its design comment claims the check works "without treating framework-owned metadata such as node timestamps as user-authored lossy values", and it has a test asserting no warning for `run.nodeStates.*.startedAt`. That carve-out works because `run.nodeStates` is plain-`JSON.stringify`d and never reaches the checker. But this loop snapshot lives **inside the context**, at `context.<id>_loop_state.iterationNodeStates.<child>.startedAt`, so it does reach the checker. Running #3926's `serializeWorkflowContext` against exactly what `loop-node-strategy.ts:140-146` persists produces:

```
warn workflow-context "Workflow state holds values that do not survive persistence unchanged"
  paths: context.the_loop_loop_state.iterationNodeStates.<redacted>.startedAt (Date)
```

So once #3926 lands, **every suspended loop iteration** emits a warning that blames the user's step for a timestamp the framework itself wrote. The reconciliation is unaffected (it is key-based and reads `status`, a string), and the Date is coerced rather than rejected, so nothing breaks. Two consequences worth fixing in #3926, not here: the warning is noise on a correct path, and after a round-trip the persisted `NodeState.startedAt` is an ISO string typed as `Date`. No file overlap between the two PRs.

**6. Test replays only half of the decision.** The comment at `src/workflow/executor/dag/index.test.ts:1091` says "What ApprovalManager.processDecision persists before calling resume", but the test patches only `nodeStates`. Production also writes `run.context[nodeId] = decisionContext`. So the test does not cover a step downstream of the wait reading the approval payload, and the comment claims more fidelity than the setup has.

**7. Undocumented dependence on globally unique child ids.** The helper's rule is "the run's map wins for every node it also knows about". That is safe only because loop children are stored in the parent map under their **bare** ids (`loop-node-strategy.ts:122-125` and `:162`). A top-level node sharing an id with a loop child would now have its state overwrite the child's snapshot entry. The conflation predates this change, but the reconciliation newly depends on it, and the doc comment only rules out the opposite direction (pulling an unknown sibling in).

**8. The doc comment names a resolver that does not exist.** It says the snapshot is patched by "an approval decision, a signal". `workflow-run-control.ts` defines exactly three operations: `approval-decision`, `hydrate-env`, `fail-execution`. There is no signal path today.

**9. Leftover, pre-existing:** the `<nodeId>_loop_state` context key is written on suspend but never deleted when the loop completes, so a finished loop leaves its stale iteration snapshot in the run context permanently. Out of scope for this change; noting it because this fix now reads that key on more paths.

### Changes made in this pass
No code changes. This PR was already merged when I picked it up and its head branch `fix/loop-resume-reconcile-wait` was deleted, so there was nothing to push to, and nothing here warranted reopening merged code. The probe tests behind findings 3 and 4 were temporary and reverted; the working tree is clean (`git status --porcelain` is empty).

Finding 4 is filed as veryfront-issue-inbox#701. Finding 5 has been sent to the agent working #3926, since the fix belongs in that PR.

### Verification performed

**Fail-first proof.** Base `3a109046d0` (the merge base) with **only** `src/workflow/executor/dag/index.test.ts` applied:

```
$ deno test --preload=src/testing/preload.ts --no-check --allow-all src/workflow/executor/dag/index.test.ts
runs a step declared after a wait in the same iteration once the wait resolves ... FAILED (1ms)

error: AssertionError: Values are not equal.
    [Diff] Actual / Expected
+   [
+     "after-wait",
+   ]
-   []
  at src/workflow/executor/dag/index.test.ts:1114:7

FAILED | 0 passed (94 steps) | 1 failed (2 steps)
EXIT=1
```

RED for exactly the stated reason: `order` is empty, the dependent step never runs. Green on the merged commit:

```
runs a step declared after a wait in the same iteration once the wait resolves ... ok (0ms)
removes child node states from previous dynamic loop iterations ... ok (0ms)
```

**Gates, on a clean tree at `e8aa97d4d7`** (`git status --porcelain` empty), each redirected to a file with `$?` echoed, never piped:

| Gate | Exit |
|---|---|
| `deno task typecheck` (includes `generate:manifests:check`) | 0 |
| `deno task lint:ci` | 0 |
| `deno fmt --check src/workflow/executor/dag/loop-node-strategy.ts src/workflow/executor/dag/index.test.ts` | 0 |
| `deno test --preload=src/testing/preload.ts --no-check --allow-all src/workflow/executor/ src/workflow/runtime/` | 0 (17 passed, 229 steps, 0 failed) |
| `grep -n $'[—–]'` on both touched files | 1 (no em or en dashes) |

**Full `src/workflow/` suite:** 79 passed (788 steps), 1 failed. The single failure is `VeryfrontCloudBlobStorage`, which I confirmed **also fails on the merge base** `3a109046d0` (`0 passed (7 steps) | 1 failed (7 steps) | 79 filtered out`). Pre-existing and unrelated to this change.

**Finding 5 reproduced** by running #3926's `serializeWorkflowContext` (from `refs/pull/3926/head`, `5af7712dcc`) against the exact context shape `loop-node-strategy.ts:140-146` persists, with a `workflow-context` warn subscriber attached. One warning, quoted above; serialized output confirms `"startedAt":"2026-01-01T00:00:00.000Z"`, a string.

PR #3934, opened 2026-08-21
fix(provider): make Google's 400 envelope classifiable

### LINKED ISSUE #3932: Provider 400 envelopes without an invalid_request_error type are unclassifiable
Split out of review on #3931, which surfaced that a curated error mapping cannot reach several providers at all.

## The gap

`buildProviderError` does two things that together make some provider 400s permanently opaque:

1. It replaces every message with generic status text:
   ```ts
   const message = `Provider request failed with status ${status}`;
   ```
2. It preserves the response body — non-enumerably, for internal classification — only for a structured `invalid_request_error`:
   ```ts
   const isStructuredInvalidRequest = status === 400 &&
     errorType === "invalid_request_error" && parsedBody !== undefined && !truncated;
   ```

A provider whose 400 envelope carries no `type: "invalid_request_error"` therefore arrives at `parseProviderError` with **neither** its body nor its own wording. Nothing downstream can classify it, no matter what mappings exist.

## Evidence

Both run through the real `buildProviderError`:

```
ANTHROPIC msg : Provider request failed with status 400
ANTHROPIC body: true
ANTHROPIC parsed: {"code":"OUTPUT_SCHEMA_NOT_CLOSED", ...}     ← classified

GOOGLE    msg : Provider request failed with status 400
GOOGLE    body: false
GOOGLE    parsed: {"code":"EXTERNAL_SERVICE_ERROR","message":"LLM provider service error"}
```

Google's ordinary shape is `{error:{code,status,message}}` — `status: "INVALID_ARGUMENT"`, no `type`. The same applies to any malformed or truncated body.

This is not specific to the structured-output mapping that exposed it. **Every** curated mapping — billing, context length, assistant prefill — is equally unreachable for these providers, which means their 400s are all reported as a generic external-service error regardless of what actually went wrong.

## Constraint that any fix must respect

Keeping raw provider bodies out of `error.message` is deliberate and test-locked (`does not surface provider error body contents` asserts a body containing `private provider payload <TOKEN>` never reaches the message). Provider bodies can quote request content, so the fix is **not** to put the text in the message.

## Options

**(a) Widen the preservation criteria.** Preserve the body for a structured, non-truncated 400 whose error code marks an invalid request, not only those whose `type` is literally `invalid_request_error` — Google's `INVALID_ARGUMENT` being the concrete case. Stays inside the existing pattern: non-enumerable, classification-only, never in the message. `buildProviderError` already computes `errorCode` from `errorRecord.status`, so the value is in hand.

**(b) Teach `parseKnownProviderBody` the Google envelope** as well, since preservation alone is not enough — it currently keys on `body.type === "invalid_request_error"` and would not recognise Google's shape even with the body present.

Both are needed for Google specifically: (a) makes the body reach the classifier, (b) makes the classifier understand it.

Deliberately not folded into #3931, which is a narrow follow-up: widening what gets retained from provider responses touches a security-sensitive boundary and deserves its own review rather than riding along. #3931 adds a test documenting the boundary and pointing here.

kojiwakayama:
Closes #3932.

## The gap

`buildProviderError` replaces every provider message with generic status text, and preserves the response body — non-enumerably, for internal classification — only for a 400 whose `error.type` is literally `invalid_request_error`.

Google's envelope is `{error:{code,status,message}}` and carries **no `type` at all**. Its 400s therefore reached `parseProviderError` with neither their body nor their own wording, so nothing downstream could classify them no matter what mappings existed.

That was never specific to the output-schema mapping that exposed it. **Every** curated mapping — billing, assistant prefill, context length — was equally unreachable, so Google's 400s all reported as a generic external-service error regardless of what actually went wrong.

## The fix

Both halves are needed; a targeted check confirms neither works alone (reverting either file fails the new end-to-end tests):

**(a) Preservation** (`src/provider/runtime-loader/provider-http.ts`) also accepts an envelope naming itself `INVALID_ARGUMENT` — Google's `google.rpc.Code` for a request rejected as malformed, the same signal Anthropic and the OpenAI-compatible providers spell `invalid_request_error` under a different key.

**(b) Classification** (`src/chat/provider-errors.ts`) learns that shape. The message cascade is extracted into `classifyInvalidRequestMessage`, so a provider is reachable by all the curated mappings or none — never the per-mapping patchwork that made one rejection legible from Anthropic and opaque from Google.

The two constants are deliberately mirrored rather than shared: `src/chat` has no non-test import of `src/provider`, and adding one would pull the HTTP layer into a client-adjacent module. The end-to-end tests drive both halves through the real `buildProviderError`, so drift fails the build rather than silently re-opening the gap.

## Security boundary is unchanged

Keeping raw provider bodies out of `error.message` is deliberate and test-locked. This does not touch that:

- The body stays **non-enumerable** and classification-only, and never reaches `error.message`.
- Widening is **per recognised envelope, not per status**: a 400 identifying itself as anything else is dropped exactly as before. Locked by `still drops a 400 whose envelope names no rejection reason` and `does not preserve a 400 whose envelope names no rejection reason`, both asserting the `<TOKEN>` payload is neither retained nor serialised.
- `does not echo a preserved Google body into the message` asserts the same for the newly preserved path.

## Tests

`src/chat/provider-errors.test.ts` — the boundary test #3931 added (`does not reach a provider whose 400 envelope is not preserved`, which documented this gap rather than asserting desired behaviour) is replaced by four tests through the real `buildProviderError`: the Google rejection classifying end to end, a *second* mapping reachable from the same envelope (proving the whole family is in reach, not one special-cased sentence), the unrecognised-envelope drop, and the no-echo assertion.

`src/provider/runtime-loader/provider-http.test.ts` — the widened criterion is locked in the module that owns it, positive and negative.

## Verification

- `deno test src/chat/provider-errors.test.ts src/provider/runtime-loader/provider-http.test.ts` — 83 steps, all pass.
- Reverting either half individually fails the two Google end-to-end tests, confirming both are load-bearing.
- `deno fmt --check` (5171 files), `deno lint` (5096 files), `deno check src/index.ts` — all clean.
- `deno test src/chat/ src/provider/` — 7 step failures in `model-registry.test.ts` and `veryfront-cloud/provider.test.ts`. **Pre-existing and unrelated**: identical failures reproduce on a clean `origin/main` worktree with zero changes; they are real network calls returning 401.


kwakayama:
## Codex review: 84/100 - Request changes

Reviewer: Codex  
Reviewed SHA: `36c00178ea01064344c5e96cb0c9267159234ad2`  
Base SHA: `42383dbb174e9377e7185c35116eeb5079b36129`

Stage 1 passes: the implementation covers both contracts in #3932. `buildProviderError` preserves a bounded, non-truncated Google `INVALID_ARGUMENT` body, and `parseProviderError` recognizes the same envelope across the existing curated mappings. This is a direct root-cause repair, not a fallback that masks the broken path. The new negative/no-echo tests preserve the existing security boundary.

### Actionable finding

[MEDIUM] Generated public provider reference is stale.  
`docs/api-reference/veryfront/provider.md:175`  
The committed reference still links `buildProviderError` to `provider-http.ts#L169`, while the exact reviewed source now declares it at `src/provider/runtime-loader/provider-http.ts:203`; the same source-line shift affects `parseRetryAfterMs`, `requestJson`, and `requestStream`. On this SHA, `ci (lint)` fails `docs:api-reference:check` with `outdated: veryfront/provider.md`. This violates the repository requirement to update generated references when public provider source changes and leaves a required check red.  
Fix: with pinned Deno 2.7.7, run `deno task docs`, commit the generated `docs/api-reference/veryfront/provider.md`, and rerun CI.

### Verification

- Exact `BASE...HEAD` diff inspected: 4 source/test files, 184 additions, 21 deletions; `git diff --check` passed.
- Pinned Deno 2.7.7: `deno task test:file src/chat/provider-errors.test.ts src/provider/runtime-loader/provider-http.test.ts` passed, 2 suites / 83 steps.
- Pinned Deno 2.7.7: changed-file `deno fmt --check`, `deno lint`, and `deno check` all passed.
- CI passes include format, typecheck, CodeQL, integration, Node, Bun, binary e2e, RSC browser e2e, npm smoke, and Sentry runtime packages.
- CI also has an unrelated-path failure in coverage shard 3 (`src/rendering/orchestrator/module-loader/index.test.ts`), which cascades into the unit/coverage aggregate checks; required checks are not green.
- PR head remained `36c00178ea01064344c5e96cb0c9267159234ad2`; unresolved review threads: 0.

Verdict: REQUEST_CHANGES. The implementation is sound, but the unresolved generated-reference defect triggers the required medium-severity score cap and must be committed before approval.

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 36c00178ea01064344c5e96cb0c9267159234ad2
Score: 84/100
Actionable-Findings: 1
Verdict: REQUEST_CHANGES

kwakayama:
## Code Review — Claude (Opus 4.6)

**SHA:** `36c00178ea01064344c5e96cb0c9267159234ad2`

### Summary

This PR fixes a gap where Google's 400 envelopes (`{error:{code,status,message}}` with `status: "INVALID_ARGUMENT"` and no `type`) were permanently unclassifiable. The fix has two load-bearing halves:

1. **Preservation** (`provider-http.ts`): `isInvalidRequestEnvelope` now also accepts `errorRecord?.status === "INVALID_ARGUMENT"`, so the body is kept (non-enumerably) for classification.
2. **Classification** (`provider-errors.ts`): `isInvalidRequestEnvelope` recognises the same shape, and the mapping logic is extracted into `classifyInvalidRequestMessage` so every curated mapping is reachable uniformly.

### Findings

No actionable findings.

### Detailed Assessment

**Correctness (40/40)**
- Both halves correctly address the two-part problem. Either alone is insufficient, and the end-to-end tests verify this.
- The `classifyInvalidRequestMessage` extraction preserves all four existing mappings (billing, assistant prefill, output schema, context length) in order.
- The condition swap in `parseKnownProviderBody` from `body.type === "invalid_request_error" && typeof body.message === "string"` to `typeof body.message === "string" && isInvalidRequestEnvelope(body)` is semantically equivalent — both are cheap string checks, order is immaterial.
- The mirrored `GOOGLE_INVALID_ARGUMENT_STATUS` constants are deliberate: `src/chat` has no non-test import of `src/provider`, and the end-to-end tests fail if they drift.

**Tests (20/20)**
- 4 end-to-end tests through the real `buildProviderError`: Google classification, second mapping reachability (proves the whole family, not one special case), unrecognised-envelope drop, and no-echo assertion.
- 2 unit tests in `provider-http.test.ts` lock the preservation criteria at the module boundary.
- The former "documents a boundary" test (which asserted the broken behaviour and pointed to #3932) is correctly replaced.

**Reliability / Security (15/15)**
- The security boundary is unchanged and test-locked: `responseBody` stays non-enumerable (`Object.keys(err).includes("responseBody") === false`), never reaches `error.message`, and `JSON.stringify` never leaks `<TOKEN>`.
- Unrecognised envelopes (`status: "UNKNOWN"`) are still dropped — widening is per recognised envelope, not per status code.

**Standards (14/15)**
- Follows existing codebase patterns. JSDoc comments explain "why" not "what."
- No cross-module imports introduced; module boundary is clean.
- Minor: the two mirrored constants + `isInvalidRequestEnvelope` functions could theoretically drift if a third provider shape is added, but the end-to-end tests cover this adequately and the PR documents the tradeoff explicitly. (–1, informational only, not actionable.)

**Scope / Docs (10/10)**
- Tightly scoped to the issue (#3932). No unrelated changes in the diff.
- PR body is thorough: explains the gap, the fix, the security boundary, and verification steps.

### Verification

- Tests cannot run in this review environment due to a pre-existing `node:util/types` brand-check limitation (unrelated to this PR). PR author reports 83 steps passing; the environment issue reproduces identically on clean `origin/main`.
- Code-level review confirms: all four curated mappings are reachable from both envelope shapes, the extraction is behaviour-preserving, and the security invariants are locked by explicit assertions.
- No new imports cross the `src/chat` → `src/provider` boundary (confirmed via grep).
- `INVALID_ARGUMENT` string has no naming collision with `#veryfront/errors` `INVALID_ARGUMENT` (different module, different value).

---

Review-Gate:
Reviewer: Claude
Reviewed-SHA: 36c00178ea01064344c5e96cb0c9267159234ad2
Score: 95/100
Actionable-Findings: 0
Verdict: APPROVE

kojiwakayama:
Addressed. The single actionable finding (stale generated provider reference) was already fixed in `142f411e3`, which landed after the reviewed SHA `36c00178e` — the review raced the push rather than missing a fix.

Regenerated with the pinned toolchain (Deno 2.7.7 per `.tool-versions`), not hand-edited, so the pins match what CI produces. The diff is line-pin shifts only for `buildProviderError`, `parseRetryAfterMs`, `requestJson` and `requestStream`, caused by the new envelope helper. `ci (lint)` is green on the current head.

On the other note in the review: the coverage shard 3 failure (`src/rendering/orchestrator/module-loader/index.test.ts … isolates concurrent cycle closures`) is unrelated to this diff and is a flake — it passes locally on an untouched worktree, and `main` is green at `42383dbb1`. It cascades into the unit/coverage aggregates, which is why those showed red.

@codex review

kwakayama:
## Final Review — Claude (Opus 4.6)

**SHA:** `142f411e37d2386395d8695bce4ebdf11513f33b`  
**Base:** `42383dbb174e9377e7185c35116eeb5079b36129`

### Summary

Two-commit PR: `36c00178ea` implements the fix, `142f411e37` regenerates the API reference doc pins. Both are load-bearing.

### Correctness & Spec Alignment (#3932)

The issue identified that Google's `{error:{code,status,message}}` envelope (with `status: "INVALID_ARGUMENT"`, no `type`) caused `buildProviderError` to drop the body, leaving every curated error mapping unreachable for Google 400s.

- **(a) Preservation** (`provider-http.ts`): `isInvalidRequestEnvelope` now accepts `errorRecord?.status === "INVALID_ARGUMENT"` alongside the existing `errorType === "invalid_request_error"`. Keyed on a recognised marker, not status alone — unrecognised envelopes are still dropped. ✅
- **(b) Classification** (`provider-errors.ts`): `isInvalidRequestEnvelope` + `classifyInvalidRequestMessage` extract the cascade so all curated mappings (billing, prefill, schema, context length) are reachable from both Anthropic and Google shapes uniformly. ✅
- Constants mirrored rather than shared to preserve the `src/chat` ↛ `src/provider` import boundary. ✅

### Security & Privacy

- Body stays **non-enumerable** and classification-only; never reaches `error.message`. Locked by `does not echo a preserved Google body into the message` (asserts `<TOKEN>` absent from message + serialisation + enumerable keys). ✅
- Unrecognised envelope drop locked by `still drops a 400 whose envelope names no rejection reason` and `does not preserve a 400 whose envelope names no rejection reason` — both assert `<TOKEN>` payload is neither retained nor serialised. ✅
- No new data flows from provider body to user-visible surfaces. ✅

### Tests

- **`provider-errors.test.ts`**: 4 new end-to-end tests through the real `buildProviderError`: Google classification, second curated mapping (assistant prefill, proving the whole family is reachable), unrecognised-envelope drop, no-echo assertion. Replaces the boundary-documenting test from #3931 with actual desired-behaviour assertions. ✅
- **`provider-http.test.ts`**: 2 new unit tests — positive (Google `INVALID_ARGUMENT` preserved) and negative (`UNKNOWN` status dropped). Both verify non-enumerability and serialisation safety. ✅

### Generated Docs (Previous Actionable Finding)

The prior review's sole actionable finding was stale line pins in `docs/api-reference/veryfront/provider.md`. Commit `142f411e37` regenerates them. Verified all 5 changed pins against source at HEAD:

| Symbol | Doc pin | Actual line | Match |
|---|---|---|---|
| `buildProviderError` | L203 | L203 | ✅ |
| `parseRetryAfterMs` | L142 | L142 | ✅ |
| `requestJson` | L724 | L724 | ✅ |
| `requestStream` | L798 | L798 | ✅ |

### Regressions

- No existing behaviour changed for non-Google providers; `invalid_request_error` path is functionally identical (condition order swapped to check `typeof body.message` first, but both branches still require it).
- No new cross-layer imports introduced.

### Score: 96 / 100

| Area | Score | Notes |
|---|---|---|
| Correctness | 25/25 | Both halves implemented, neither works alone, tests prove it |
| Spec alignment | 25/25 | Matches #3932 exactly |
| Security | 24/25 | Thorough boundary tests; minor: mirrored constants could theoretically drift, but end-to-end tests catch it |
| Tests & docs | 22/25 | Strong coverage; local env issue prevented live test run confirmation |

**Verdict: APPROVE** — no actionable findings. The previous stale-docs finding is resolved. Both preservation and classification halves are correct, security boundaries are test-locked, and generated docs are accurate.



Review-Gate:
Reviewer: Claude
Reviewed-SHA: 142f411e37d2386395d8695bce4ebdf11513f33b
Score: 96/100
Actionable-Findings: 0
Verdict: APPROVE

kwakayama:
## Final Codex review — 98/100

Reviewer: Codex  
Reviewed SHA: `142f411e37d2386395d8695bce4ebdf11513f33b`  
Base SHA: `42383dbb174e9377e7185c35116eeb5079b36129`

No actionable findings.

Issue #3932 is implemented at the root cause: `buildProviderError` now preserves only bounded, non-truncated 400 bodies whose envelope explicitly identifies an invalid request, and `parseProviderError` recognizes Google's `INVALID_ARGUMENT` shape through the same curated classification path. The negative path remains fail-closed; unrecognized envelopes are dropped, raw provider content does not enter `error.message`, and the preserved body remains non-enumerable and absent from JSON serialization.

The regression coverage exercises both halves through the real `buildProviderError`, verifies a second curated mapping, and locks the privacy boundary. The generated provider reference is current; its shifted pins resolve to `parseRetryAfterMs` L142, `buildProviderError` L203, `requestJson` L724, and `requestStream` L798.

Validation:
- Exact `42383dbb...142f411e` diff inspected; `git diff --check` passed.
- Pinned Deno 2.7.7: targeted suites passed, 2 suites / 83 steps.
- Pinned Deno 2.7.7: changed-file `deno check`, format, and lint passed.
- Pinned Deno 2.7.7: `docs:api-reference:check` passed; committed reference tree is current.
- GitHub reports all active checks successful, zero review threads, and a mergeable non-draft PR.

Verdict: **APPROVE**



Review-Gate:
Reviewer: Codex
Reviewed-SHA: 142f411e37d2386395d8695bce4ebdf11513f33b
Score: 98/100
Actionable-Findings: 0
Verdict: APPROVE

PR #3943, opened 2026-08-21
fix(workflow): stop charging recovery budget to a nested approval resume

kwakayama:
Fixes veryfront/veryfront-issue-inbox#696.

## The bug

A composite node executes its children against a synthetic run record whose
`status` is hardcoded `"running"` (`src/workflow/executor/dag/index.ts`, in
`executeParallelNode`, `executeBranchNode`, `executeSubWorkflowNode`, and the
loop and map strategies).

The DAG executor reads run status to tell two situations apart:

- a run parked on a human decision, where a composite recorded `running` should
  be re-entered so its child resumes, free of charge;
- a run whose worker died mid-node, where a node recorded `running` is re-run
  under a bounded recovery budget.

At the top level the real run is `waiting`, so this works. One level down the
child graph always looks like the crash case, so every nested composite was
re-entered as a recovery and charged an attempt.

## Measured on `3837403f8a`

An instrumented run of `parallel > parallel > wait`, resumed after an ordinary
approval, charges the inner composite:

```
PROBE_RECOVERY run=outer_parallel node=inner attempts=1->2 max=1 type=parallel
```

The increment is usually overwritten when the composite returns, which is why
the issue described the effect as mild. It is not always overwritten. When the
recovered composite is queued behind a batch boundary and an earlier sibling
suspends first, the batch returns before the composite runs and `attempt: 2`
persists into the run's node states. The next approval then reads the budget as
spent:

```
PROBE_R2_STATES {"inner-wait":"completed/att=1","inner":"running/att=2", ...}
PROBE_EXHAUSTED run=outer_parallel [{"nodeId":"inner","attempts":2,"maxAttempts":1}]
PROBE_R3 completed=false err=Node "outer" failed: Node "inner" was interrupted
         after 2 of 1 attempt(s); retry budget exhausted
```

Two ordinary approvals, no crash, and the run fails naming a node nobody
interrupted. That is exactly the failure mode the issue predicted.

## The fix

Thread the answer from the root run instead of inferring it from a record that
is synthetic by construction. `execute` reads `run.status === "waiting"` once,
at the only place the status means anything, and passes it beside `rootRunId`
through the same call chain the run id already travels.

The recovery block now uses the threaded value, so every level applies the same
policy the top level already applied, regardless of nesting depth.

## Tests

Three tests in `src/workflow/executor/dag/index.test.ts`, under
`recovery budget accounting across a nested approval`:

1. an approval resume of `parallel > parallel > parallel > wait` leaves the
   nested composite's `attempt` at 1;
2. the same workflow survives repeated approvals instead of failing with the
   interrupted-node error;
3. a guard: a `running` run still charges and still bounds recovery for a node
   a dead worker left running inside a composite.

Fail-first, test file only, on `3837403f8a`:

```
$ deno test --preload=src/testing/preload.ts --no-check --allow-all \
    src/workflow/executor/dag/index.test.ts
does not spend a nested composite's recovery budget on an approval resume ... FAILED
  Actual 2 / Expected 1
survives repeated approvals instead of failing a node nobody interrupted ... FAILED
  Actual 'Node "outer" failed: Node "mid" was interrupted after 2 of 1 attempt(s); retry budget exhausted'
  Expected undefined
still bounds recovery for a node a dead worker left running inside a composite ... ok
FAILED | 0 passed (97 steps) | 1 failed (3 steps)
exit 1
```

Green with the fix: `ok | 1 passed (100 steps) | 0 failed`.

`src/workflow/executor/workflow-tracing.test.ts` calls the private
`executeUnwrapped` positionally through a cast, so its argument list is updated
for the new parameter.

## Gates

```
deno task typecheck                                          exit 0
deno task lint:ci                                            exit 0
deno fmt --check <touched files>                             exit 0
deno test ... src/workflow/executor/                         exit 0  (12 passed, 183 steps)
deno test ... src/workflow/                                  80 passed (829 steps), 1 failed
```

The one failing file is `src/workflow/blob/veryfront-cloud-storage.test.ts`,
which fails identically on a clean `origin/main` worktree in this environment:
it makes outbound HTTPS calls that the sandbox blocks
(`client error (Connect): General SOCKS server failure`). It imports nothing
from the workflow executor.

## Known limitation, not introduced here

A wait resume with no explicit checkpoint leaves the persisted run status at
`waiting` for the whole resumed execution
(`src/workflow/executor/workflow-executor.ts`, the `if (resumeInfo)` branch).
A worker dying during such an execution is therefore invisible to this logic.
That was already true at the top level before this change; the change makes
child graphs agree with the top level rather than diverge from it. Making the
status honest during a resume is a separate change to durable run state and is
out of scope here.

This does not touch the empty-ready-set behaviour in
veryfront/veryfront-issue-inbox#701.


kwakayama:
Closing as a duplicate of #3940, which was opened twelve minutes earlier and reaches the same fix.

Same defect, same diagnosis, and #3940 has the better shape: it bundles `rootRunId`, `executionRunId`, `ownership`, and `resumingWait` into an `ExecutionScope`, removing four positional arguments where this PR added a fifth, and it records the invariant in a type.

I checked before closing rather than deferring on timestamp alone. #3940 threads the real root run status instead of special-casing the symptom, and it covers all five composite paths (`executeParallelNode`, `executeBranchNode`, `executeSubWorkflowNode`, and the map and loop strategies via the `executeChildGraph` closures). All three fail-first tests from this PR pass on its head unchanged, including a three-level nest, so there is no case it leaves broken.

One test here covers something #3940's does not: that recovery inside a child graph is still *bounded*, not merely that it still happens. Posted to #3940 with the mutation evidence in https://github.com/veryfront/veryfront-code/pull/3940#issuecomment-5371144318.

Issue #3949, opened 2026-08-21
A run executing a step persists nothing about it, so a hung workflow cannot be diagnosed


kojiwakayama:
Found while building the workflow test harness (#3936). Not a regression — this is how the engine has always written.

## What happens

A run persists node states **only at pause and completion boundaries**. While it is actively executing a step, nothing about that step is written. Probed against `MemoryBackend` on `origin/main`:

```
# workflow: step(a, 250ms) -> step(b, 250ms)
t+120   status=running    currentNodes=[]         nodeStates=[]
t+250   status=running    currentNodes=[]         nodeStates=[]
t+380   status=running    currentNodes=[]         nodeStates=[]
t+510   status=completed  currentNodes=[]         nodeStates=[a=completed, b=completed]
```

Both `currentNodes` **and** `nodeStates` are empty for the entire execution. Compare a run that pauses, where the engine does record state:

```
# workflow: step(a) -> waitForApproval(review) -> step(b)
t+200   status=waiting   currentNodes=[review]   nodeStates=[a=completed, review=running]
```

## Why it matters

**A hung run is undiagnosable from its persisted state.** A workflow stuck on a step that never returns looks like this, forever:

```
status=running; currentNodes=[]; nodeStates=[]
```

There is no way to tell *which* step it is stuck on — or whether it started any step at all. That is the exact question an operator has when a run stops making progress, and the run record cannot answer it. The only recourse today is to add logging and reproduce.

It also means `currentNodes` does not do what its name suggests for the commonest case. It is populated when a run parks, not while it runs.

## Knock-on effects already visible

- **#3947** (run events over SSE) can only be as fine-grained as the engine's writes. Its `step.completed` events for a straight-through run all arrive in one batch with `run.status: completed`, and `step.started` effectively only fires for a wait node. Documented there rather than worked around.
- **#3936**'s harness cannot name the stuck step in its timeout diagnostic. It reports `executing: unrecorded (a run mid-step persists no node state)` — accurate, but only because the information does not exist.

## What a fix involves

Persisting a node state when a step *begins*, not only when it settles. That is a change to the executor's write path, and it is not free:

- One extra write per step, which matters for a workflow with many small steps.
- Contention and ordering against the conditional-update paths (`updateRunIfStatus`, `updateRunIfStatusAndWorker`) used for crash recovery and worker ownership.
[...]

PR #3950, opened 2026-08-21
feat(workflow): add an end-to-end workflow test harness

kojiwakayama:
Closes veryfront/veryfront-issue-inbox#866.

## The problem

Testing a workflow means waiting for something asynchronous and durable to reach a state, and every test that does it has hand-written the same polling loop. `src/workflow/__tests__/workflow.integration.test.ts` carries three — `waitForCondition`, `waitForStatus`, `waitForApprovals` — and they are private to that file. There is no shared harness today, and `veryfront/testing` only ships generic assert/bdd/timing helpers.

That loop is where workflow tests go wrong:

- a bare `await delay(n)` passes on a fast machine and flakes on a loaded one;
- a poll with no deadline hangs the suite instead of failing it;
- a timeout that reports only "timed out" sends the author back to add logging to find out what the run was actually doing.

## What this adds

`veryfront/workflow/testing`:

```typescript
const harness = await startWorkflowTest(ingestWorkflow, { input });
try {
  await harness.reaches("waiting");
  await harness.approve("review");          // by node id, not approval id
  await harness.settled();
  await harness.assertSteps(["fetch", "review", "load"]);
} finally {
  await harness.dispose();
}
```

Every wait polls to a deadline and, on expiry, throws with the run's status, what it is executing, its per-node states and its pending approvals — the information the author would have gone looking for. Each harness gets its own `MemoryBackend` by default, so one test's run is never visible to another's assertions.

Approvals are addressed by **node id**. A test author knows they wrote `waitForApproval("review")`; the approval's generated id is something they would otherwise read out of the run first. `approve()` also polls for the approval to exist, so a test need not reach a waiting state before deciding one.

`steps()` orders by `completedAt`, so an ordering assertion is evidence of execution order rather than of object key order.

## Two things it deliberately does not do

**No `advanceTime`.** The executor schedules timed waits with real `setTimeout` and takes no clock, so fast-forwarding one would mean faking global timers for the whole process. A workflow that sleeps for an hour still sleeps for an hour here. Doing it honestly needs a clock seam in the executor — a change to the executor, not a helper this module can provide. I would rather ship no time control than a fake one.

**It cannot name the step a *running* workflow is stuck on**, because the engine does not record one. Probing that is what turned up **#3949**: a run mid-step persists neither `nodeStates` nor `currentNodes`, so a hung run is undiagnosable from its own record.

```
mid-run   status=running    currentNodes=[]        nodeStates=[]
paused    status=waiting    currentNodes=[review]  nodeStates=[a=completed, review=running]
```

The diagnostic reports `executing: unrecorded (a run mid-step persists no node state)` rather than printing an empty list that reads like a corrupt run, and tests pin **both** shapes — the unrecorded one and the paused one where the engine does know. If veryfront/veryfront-code#3949 is fixed, the message gets better for free and the pinned test will say so.

## Tests

15 steps in `src/workflow/testing.test.ts`, all against real workflows through the real client: completion and step ordering, approval by node id, approving before the run reaches `waiting`, rejection, pending-approval listing, a failing run settling rather than hanging, cancellation, harness isolation, idempotent dispose, and three failure-reporting tests asserting the diagnostics actually carry the run id, the status, the in-flight node and what really ran.

## Verification

All with pinned Deno 2.7.7 (`.tool-versions`):

- `deno test src/workflow/testing.test.ts` — 15 steps, 0 failed.
- `deno lint src/workflow/` (142 files), `deno check src/workflow/testing.ts`, `deno fmt` — clean.
- `deno task docs` regenerated; the new subpath appears under `veryfront/workflow`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -108,6 +108,7 @@
     "./workflow": "./src/workflow/index.ts",
     "./workflow/registry": "./src/workflow/registry-public.ts",
     "./workflow/blob": "./src/workflow/blob/index.ts",
+    "./workflow/testing": "./src/workflow/testing.ts",
     "./workflow/worker": "./src/workflow/worker/index.ts",
     "./workflow/claude-code": "./src/workflow/claude-code/index.ts",
     "./workflow/claude-code/types": "./src/workflow/claude-code/types.ts",
--- a/docs/api-reference/veryfront/workflow.md
+++ b/docs/api-reference/veryfront/workflow.md
@@ -516,6 +516,28 @@ import { getAllWorkflowIds, getWorkflow, registerWorkflow } from "veryfront/work
 | ------------------ | -------------------------------------------------------------- | --------------------------------------------------------------------------------------------- |
 | `workflowRegistry` | Project-scoped registry for workflow metadata and definitions. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/workflow/registry.ts#L425) |
 
+### `veryfront/workflow/testing`
+
+Test primitives for driving a workflow end to end. Testing a workflow means waiting for something asynchronous and durable to reach a state, and every test that does it has written the same loop: `typescript while (Date.now() - start < timeout) { if ((await client.getRun(runId))?.status === "completed") break; await delay(50); }` That loop is where workflow tests go wrong. A bare `await delay(n)` passes on a fast machine and flakes on a loaded one; a poll with no deadline hangs the suite instead of failing it; and a timeout that reports only "timed out" sends the author back to add logging to find out what the run was actually doing. Every harness method here polls to a deadline and, on expiry, throws with the run's status, its pending approvals and its per-node states - the information you would otherwise have gone looking for. Approvals are addressed by **node id**, not approval id. A test author knows they wrote `waitForApproval("review")`; the approval's generated id is something they have to dig out of the run first. ## What this does not do There is no `advanceTime`. The executor schedules timed waits with real `setTimeout` and takes no clock, so fast-forwarding one would mean faking global timers for the whole process. A workflow that sleeps for an hour still sleeps for an hour here. Testing that honestly needs a clock seam in the executor, which is a change to the executor rather than a helper this module can provide.
+
+```ts
+import { startWorkflowTest } from "veryfront/workflow/testing";
+```
+
+#### Functions
+
+| Name                | Description                                           | Source                                                                                       |
+| ------------------- | ----------------------------------------------------- | -------------------------------------------------------------------------------------------- |
+| `startWorkflowTest` | Start a workflow and return a harness for driving it. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/workflow/testing.ts#L190) |
+
+#### Types
+
+| Name                       | Description                        | Source                                                                                      |
+| -------------------------- | ---------------------------------- | ------------------------------------------------------------------------------------------- |
+| `ApprovalDecisionOptions`  | Options for resolving an approval. | [source](https://github.com/veryfront/veryfront-code/blob/main/src/workflow/testing.ts#L89) |
+| `StartWorkflowTestOptions` | Options for `startWorkflowTest`.   | [source](https://github.com/veryfront/veryfront-code/blob/main/src/workflow/testing.ts#L74) |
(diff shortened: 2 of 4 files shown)


kojiwakayama:
**Converted to draft. Do not merge — the abstraction is not settled.**

Raised in review: should this build on a generic runner rather than be a bespoke workflow harness? I went and looked, and the answer changes the shape of this PR.

## What already exists

| Layer | Runtime-decoupled? |
|---|---|
| `veryfront/testing/bdd` | ✅ **yes** — maps `describe`/`it` onto `@std/testing/bdd` (Deno), `node:test` (Node), `bun:test` (Bun) |
| `veryfront/testing/assert` | ✅ yes |
| `veryfront/testing` | ✅ yes — `waitFor` (time-scaled), `isolation`, `timing`, `mock-fetch` |
| **`veryfront test` (CLI)** | ❌ **no** — `cli/commands/test/handler.ts:31` hardcodes `runCommand("deno", …)` |
| `veryfront/workflow/testing` (this PR) | domain fixture |

So the runtime-decoupled runner **already exists at the library layer**. This PR is layered on it, not duplicating it — and `665bbe348` makes that materially truer by reusing two primitives I had reimplemented.

## What I got wrong, and fixed

- **`scaleMs`.** My deadlines were fixed numbers. That is the *same flake this harness exists to remove* — it holds on a developer machine and expires on a loaded CI runner, where the failure reads as a product bug rather than a starved test. Now both the timeout and the poll interval go through the framework test time scale.
- **`registerTestCleanup`.** A test throwing before its own `finally` left a client running — in Bun, precisely the cross-test leakage `isolation.ts` exists to contain. The harness now registers its own cleanup.

Both are clean leaves over `platform/compat` with no `./init.ts` side effects, so importing them into a shipped module drags in no test-only machinery.

## The real gap, and why this should wait

`veryfront test` is the piece that is **not** generic, and it is worse than "hardcodes Deno". It reads results by regex-scraping human output:

```ts
line.match(/(\d+)\s+passed\s*\|\s*(\d+)\s+failed/)
```

That is Deno `deno test` text. `node:test` (TAP) and `bun:test` produce nothing resembling it, so this cannot be pointed at another runtime by swapping the binary — the result model has to be designed, not adapted. Which is exactly the argument for not rushing it, and for not letting a domain abstraction land first and set precedent for a layer whose shape is still open.

Concretely, the questions I would want answered before this merges:

1. **Does the generic runner own a result model?** A typed run/step/assertion result the CLI serialises, rather than each runtime scraped into one regex.
2. **Where does time control live?** This harness deliberately ships no `advanceTime` (the executor takes no clock — real `setTimeout`). A clock seam is generic infrastructure, not a workflow concern, and if it lands generically the workflow abstraction gets it for free.
3. **Is `startWorkflowTest` the right seam at all**, or should the domain layer be assertions/matchers over a run that the generic runner drives?
4. **What do the sibling domains need** — agents, evals (`veryfront/eval` already exists and is its own runner-ish thing) — so workflows do not get a shape the others cannot reuse?

I have not built any of that and am not going to guess at it here.

## Status

The code on this branch is green and the two reuse fixes are strict improvements regardless of what the layering becomes. But it stays a draft until the runner design is agreed. Happy to write up the runner design properly as a separate issue if that is useful — as a design, not an implementation.

Related: veryfront/veryfront-code#3949 (a run mid-step persists no node state), which bounds what any workflow test harness can observe regardless of its shape.

kojiwakayama:
Design question filed as veryfront/veryfront-issue-inbox#867 — *Design a runtime-agnostic test runner before layering domain test abstractions on it.*

This PR stays in draft pending the decision there, specifically question 5 (what the domain layer actually is) and question 4 (whether time control lands generically, which is why this harness ships no `advanceTime`).

kojiwakayama:
Closing this as a merge candidate. The branch is useful as a prototype, but the runtime-agnostic runner contract must be settled in veryfront/veryfront-issue-inbox#867 before a domain-specific workflow harness becomes public API.

The current branch also has unresolved lifecycle and API-shape defects, so extracting pieces now would preserve the wrong abstraction. No behavior from this PR is being treated as shipped. veryfront/veryfront-issue-inbox#866 remains open and should resume from the decision recorded in veryfront/veryfront-issue-inbox#867.

PR #3953, opened 2026-08-21
fix(build): give every emitted module a React binding

kwakayama:
## The bug

Server-rendering a builtin `Popover` from `veryfront/ui` on the published `0.1.1246` throws
`ReferenceError: React is not defined` and returns HTTP 500. The emitted chunk calls
`React.createElement` twice and imports React zero times:

```js
// veryfront@0.1.1246 → esm/src/react/components/ui/adapter/builtin/popover.js
import { createAnchoredSurfaceParts } from "../../anchored-surface.js";
const parts = createAnchoredSurfaceParts();
export const builtinPopover = {
    Root: parts.AnchoredRoot,
    Trigger: (props) => React.createElement(parts.AnchoredTrigger, { haspopup: "dialog", ...props }),
    Content: (props) => React.createElement(parts.AnchoredContent, { ...props }),
};
```

## Why it happens

Two compiler configurations disagree, and only one of them is ever observed locally.

| | `jsx` |
| --- | --- |
| `deno.json` — what `deno check`, the tests and `veryfront dev` use | `react-jsx` (automatic) |
| `@deno/dnt@0.42.3` — what `build:npm` emits | `react` (classic), hardcoded |

```ts
// jsr:@deno/dnt@0.42.3/mod.ts — not read from options.compilerOptions
jsx: ts.JsxEmit.React,
jsxFactory: "React.createElement",
jsxFragmentFactory: "React.Fragment",
```

Under the automatic runtime a component file needs no React import, and `deno lint`'s
`no-unused-vars` **rejects** one — so the binding the published emit depends on is exactly
the binding the toolchain removes. This is not an oversight in the four files; it is what
the repo's own tooling asks for. A type-only `import type React from "react"` fails the same
way, being erased before the lowering runs.

Every local signal is green, because every local signal uses the automatic runtime. The
first observer is a consumer getting a 500.

## Blast radius

Sweeping all 2,600 emitted `.js` files in the published tarball for a `React` reference with
no binding in scope:

| Module | Why it had no binding |
| --- | --- |
| `ui/adapter/builtin/popover.js` | no React import |
| `ui/adapter/builtin/drawer.js` | no React import |
| `optimized-image/SimpleOptimizedImage.js` | `import type React` — erased |
| `optimized-image/utils.js` | type-only named imports |

So a builtin **Drawer** and the optimized-image helpers 500 the same way; only the Popover
path had been reported. Builtin popovers also back menus and dropdowns.

## The fix

Not the dnt bump. `@deno/dnt` 0.43 does expose `compilerOptions.jsx`, but it is a year of
changes — deno_graph and ts-morph both updated, shim and polyfill behaviour reworked — and
this build hand-patches three separate pieces of dnt output. That deserves its own PR with
an artifact diff, not a hotfix for a live 500. Filed as follow-up.

Instead the divergence is closed on the artifact, where the classic lowering is a fact
rather than an assumption. `scripts/build/npm-jsx-react-binding.ts` runs from `postBuild`
for the root package and the extension packages: it parses each emitted module, prepends
`import * as React from "react";` where `React` is free, then **re-scans and throws** if any
survived. It sits beside `assertNpmRuntimeHelperContract` and the `patchDnt*` helpers, which
is this build's existing idiom for "dnt emitted something this package cannot ship".

Detection is parsed, not grepped, because the real package needs it to be:

- `esm-module-loader/constants.js` exports `"React.createElement"` as a string constant
- the dev-ui extension embeds an entire browser bundle as one JSON string containing `React.Children`
- `snippet-renderer.js` binds React by array destructuring — `const [{ renderToString }, React] = await Promise.all(…)` — and must be left alone

Run against the published `0.1.1246` it flags those four modules and nothing else.

## Verification

**Detector, against the real broken tarball** — 2,600 files, four hits, no false positives.

**`deno task build:npm`** — exit 0, and the post-build step reports exactly:

```
Gave 4 emitted module(s) a React binding:
  src/react/components/optimized-image/SimpleOptimizedImage.js
  src/react/components/optimized-image/utils.js
  src/react/components/ui/adapter/builtin/drawer.js
  src/react/components/ui/adapter/builtin/popover.js
```

**End-to-end SSR under Node with `react-dom/server`**, same harness, same page — the
reporter's minimal repro:

```
0.1.1246          → ReferenceError: React is not defined
this branch       → <main style="padding:24px"><h1>Builtin Popover SSR repro</h1>
                     <button id="vf-anchored-_R_2_-trigger" aria-haspopup="dialog"
                     aria-expanded="false" aria-controls="…-content" type="button">Open</button></main>
```

`aria-haspopup="dialog"` survives, so the `haspopup` wiring the builtin trigger exists to
supply is intact. The other three modules were checked the same way: all four throw on
`0.1.1246` and render here.

**Tests** — `scripts/build/npm-jsx-react-binding.test.ts`, 17 cases: the verbatim emitted
popover from `0.1.1246` as a fixture, idempotency across two builds, the string-literal and
comment false positives, the destructured binding, `deps/` left alone, `.d.ts` left alone,
and a parse failure that names its file rather than passing it through.

**Gates** — `deno task lint`, `deno task fmt:check`, `deno task typecheck`,
`deno task test:scripts`, `deno task build:npm`: all exit 0.

## Follow-up

Bump `@deno/dnt` to 0.43.x and set `compilerOptions.jsx: "react-jsx"` so the build stops
lowering to the classic runtime at all. At that point this normalization patches nothing and
survives purely as the assertion it ends with — which is the form it should be kept in.


kwakayama:
## Review pass — one real defect found in this PR's own code, now fixed

Pushed two follow-up commits. Worth reading before the diff, because the first one is a bug
I shipped in the first commit and then caught.

### `8ad27cd` — the guard had a false negative

`findFreeReactReference` originally treated *any* declaration named `React` anywhere in a
module as satisfying *every* reference in it. That is not how scope works, and the gap ships
exactly the 500 this PR exists to prevent:

```js
export const Card = (props) => React.createElement("div", props);   // free — throws
async function render() {
    const [{ renderToString }, React] = await Promise.all([
        import("react-dom/server"),
        import("react"),
    ]);
    return renderToString(React.createElement("span"));
}
```

The first commit's checker reports that file as clean. Confirmed by running it, not by
reading it.

The shape is not invented: `src/rendering/snippet-renderer.js` in the real package already
binds React with that exact destructuring. Any module-scope JSX added to a file like it
would have walked straight past the guard.

Now resolved properly — a module-scope binding satisfies the file, since that is precisely
what the rewrite adds; a binding anywhere else shadows only its own scope. Widened what
counts as a reference at the same time, since a free `React` throws whatever shape the read
takes: `React?.createElement`, `React["createElement"]` and a bare `React` passed as a value
are now caught. Names rather than reads — an import's local name, `foo.React`,
`{ React: 1 }` — still are not references.

Behaviour on real input is unchanged: the same four modules out of 2,600 on the published
`0.1.1246`, zero on the built package.

### `b9dae1e` — three test names renamed to match what they assert

Two still described the pre-scope-resolution rule; one said "by other means" where the point
is that a module-scope binding need not be an import.

### Checks I ran on the rest of the diff

- **`deno.json`** — diffed every task against `origin/main`: exactly four changed (`lint`,
  `fmt`, `fmt:check`, `test:scripts`), purely additive, no task added or removed, no
  neighbouring file list corrupted by the string splice.
- **`postBuild` ordering** — verified on the artifact rather than by reading. After *all*
  post-build steps run, the final package scans 2,602 files with **0** free `React`
  references, and the reporter's page SSRs to real HTML from it. `deps/` is skipped, so
  there is no overlap with `normalizeEsmShReactNpmShims`; `assertNpmRuntimeHelperContract`
  only inspects relative specifiers and the added import is bare.
- **Extension packages** — 20+ extension packages build with the normalizer in their
  `postBuild`, so `outDir/esm` reliably exists at that point.
- **Prepend safety** — no emitted file anywhere in the package has a shebang or a directive
  prologue, so nothing can be displaced. None of the four patched modules appears in
  `browser-safe-exports.mjs`.

Gates after the last commit: `deno task lint`, `deno task fmt:check`, `deno task typecheck`,
`deno task test:scripts`, `deno task build:npm` — all exit 0. 24 unit tests pass.

### Known limit, stated rather than buried

`build:npm` runs in CI only in the prerelease and release jobs, not on pull requests. The
assertion therefore fires at RC build on merge to `main` — ahead of any published release,
but not on the PR that introduces a regression. Given dnt owns the lowering, the artifact is
the only place this invariant is expressible; a source-level lint would have to fight
`no-unused-vars` on every component file.

kwakayama:
## Second review pass — three more defects in the guard, all real, all fixed

`8ae803f` pushed. A reviewer went through the diff properly and found that my scope-aware
rewrite still had a hole big enough to ship the original bug. I reproduced every claim before
acting; all held.

### The one that mattered: default parameter values counted as declarations

`patternNames` collected every identifier in a binding pattern's *subtree*, so the default
**value** counted as a declared name:

```js
export function get(react = React) { return React.createElement("div"); }
// → previously reported clean
```

Not a corner case. `esm/src/react/server-render-context.js` has that exact signature twice:

```js
export function getServerRenderContext(react = React) { … }        // line 48
export function useServerRenderContext() { const runtime = React; } // line 52
export function wrapWithServerRenderContext(el, v, react = React) { … } // line 55
```

Strip that module's React import — the erasure this check exists to catch — and the previous
rule reported **line 52** while excusing 48 and 55. It noticed at all only because
`useServerRenderContext` happens to do `const runtime = React`. Delete that one function and
the module ships, 500s, and leaves the guard green. After the fix it reports line 48.

My earlier validation could not see this: "same 4 hits on 0.1.1246, 0 on the built package"
is true *with the bug present*, because no module today has a default-param React as its only
read. A passing sweep over real input is not the same as a correct rule.

Fixed by descending into binding positions only — `AssignmentPattern.left`, an object
pattern property's value, `RestElement.argument`, array and object pattern members.

**One deliberate deviation** from the suggested fix: I did *not* treat a computed
destructuring key as a binding. `const { [React]: v } = o` **reads** React, so collecting it
would reintroduce the very false-negative class being fixed. It is left to the reference
walk, which flags it. Test added.

### `var` hoisting produced an unloadable module

`var` hoists out of its block, so `if (x) { var React = … }` declares React for the whole
module. The old rule missed it, prepended an import beside it, and Node then refused the file
outright:

```
SyntaxError: Identifier 'React' has already been declared
```

Worse than a false positive — the post-rewrite assertion is structurally blind to it, because
by then the import itself satisfies the check. `addHoistedVarNames` now walks nested
statements, stopping at function boundaries.

### `for` heads

`for (const React of list)` was flagged as free, which would inject an undeclared bare `react`
import into an extension package and trip `assertEmittedBareImportsAreDeclared` with a
confusing message. For heads now declare into the loop's own scope.

### A claim of mine that was wrong

I said a bare `React` passed as a value "does not occur in the package today". It occurs five
times, all as call arguments — `build-app-route-renderer.js:153`,
`layouts/utils/component-loader.js:337`, `layouts/utils/applicator.js:131`,
`layouts/layout-applicator.js:332`, `react/compat/scope-commit-effect.js:25`. All correctly
bound, so none is flagged, but the shape is exercised. Widening the reference rule was closing
a live shape, not a theoretical one.

### Tests

The test that claimed patched modules were "loadable" did not load anything — it re-asserted
the checker against the checker's own rewrite. Loadability is exactly what the checker
*cannot* establish: a duplicate declaration parses fine and dies at load. It now runs the
patched module in a subprocess with `react` mapped to a stub and asserts exit 0 and the
rendered value. Confirmed non-vacuous by sabotaging `withReactBinding` to emit a duplicate and
watching that test go red on its own.

Also: dropped the misleading clause from another test name, fixed two comments left describing
the pre-scope-resolution rule, and added a byte-identity assertion where only a return value
was checked.

### Re-validated after all of it

Same four hits out of 2,600 on published `0.1.1246`; zero on the built package; `build:npm`
exit 0 patching the same four; the reporter's page still SSRs to real HTML. `deno task lint`,
`fmt:check`, `test:scripts` all exit 0. 26 unit tests.

### Reviewer notes accepted without change

- The normalizer hard-fails on a missing `esm/` while the neighbouring `patchDnt*` calls
  tolerate absence. A stricter contract than its siblings, deliberately: a missing `esm/`
  means the emit did not happen, and that should stop a release build loudly.
- Ordering confirmed independently — every later step that touches these files uses
  `/m`-anchored regexes or `replaceAll`, no `.js.map` is emitted so the one-line shift breaks
  no source maps, and `removeUnreferencedTopLevelDir(outDir, "react")` matches on `"/react/"`
  which the injected `from "react"` cannot satisfy.

kwakayama:
## Negative control: every binding in the artifact, removed

Review signed off on `8ae803f`. Adding the one check that had been missing — the guard had
only ever been shown to find the four broken modules. It had not been shown to depend on the
bindings in the 2,598 it clears.

So: for every emitted module that begins with `import * as React from "react";`, strip that
line and re-run the checker. It must flag all of them.

```
scanned 2602 | free now: 0
files with the namespace import: 169 | re-flagged when stripped: 168
NOT re-flagged: src/build/compiler/mdx-to-js.js
```

The one exception is a false match by the *control*, not a miss by the checker. Every `React`
in `mdx-to-js.js` — including the import on line 85 — sits inside a template literal. It is
the text of a module the MDX compiler *generates*, not code this module runs:

```js
const moduleCode = `
// Generated from ${mdxPath}
import * as React from "react";
...
  return React.createElement(MDXContent, { components: { ${componentList}...components } });
`;
```

My filter matched a line inside a string, so it "removed an import" that was never there, and
the checker correctly reported no free reference before or after.

Net **169/169** correct. And a small independent argument for the design: a line-based rule
gets this file wrong in both directions — it would strip a phantom import, and it would flag
`React.createElement` in generated text as a defect. Parsing is what makes it right.

CI: 10 passed, 16 running, 4 skipped.

kwakayama:
## Correction to my last comment, and where the classic lowering has a second home

**I was wrong to call the 37-file control "partial".** 37 and 169 measure different
properties of the same 2,602 files, and both are right:

```
line 1 IS the namespace import :  37     <- the set the rewrite could have written
import appears ANYWHERE        : 169     <- 37, plus 132 that import React in their own block
```

`withReactBinding` prepends at byte 0, so a file the pass touched necessarily carries that
import on line 1. The 37-file scan asks the tight question — *did the pass inject anything
unnecessary* — and can only be asked of files where the pass could have acted. 37/37
re-flagged. My 169-file scan asks the broader one — *does the checker agree that every
React-importing module needs its import* — and 169/169 behave correctly. Different questions,
both worth having.

It also explains `mdx-to-js.js` cleanly: its `import * as React from "react";` is on line 85,
inside a template literal, and its line 1 is a logger import. A first-line scan structurally
could not have surfaced it.

### The generator point, checked and generalised

The observation that `mdx-to-js.js` emits *both* the classic import and `React.createElement`
into the module text it generates raises a fair question for the dnt 0.43 follow-up: runtime
code generators are a second place the classic lowering is a fact, and one the artifact scan
cannot see.

I went looking for all of them. Two attempts, because my first was wrong twice over:

1. A regex for `` `…React.createElement…` `` matched the code *between* two template
   literals — the span from one closing backtick to the next opening one contains no
   backticks. It reported 83 files. All but a handful were noise.
2. The AST version then missed `mdx-to-js.js` entirely, because `TemplateElement.value` is a
   plain `{raw, cooked}` object with no `type` field, so my is-a-node guard rejected it.

Corrected, parsing template quasis and string literals properly, the real answer across the
built package is **four** files with `React.createElement` in literal text, of which only two
generate modules:

| File | Generates a module? | Emits its own React import |
| --- | --- | --- |
| `build/compiler/mdx-to-js.js` | yes | **yes** — `import * as React from "react";` |
| `html/hydration-script-builder/prod-hydration.js` | yes | **yes** — `import * as React from 'react';` |
| `transforms/mdx/esm-module-loader/constants.js` | no — `ESBUILD_JSX_FACTORY` config value | n/a |
| `server/handlers/dev/framework-candidates.generated.js` | no — dev-tooling pattern strings | n/a |

So both genuine generators are already self-sufficient: they emit the binding beside the
call. They never depended on this normalizer, and they will not be disturbed by a later
switch of the root package to the automatic runtime. Nothing to do here — but it is the right
thing to have checked before that bump, rather than after.

Two of my three attempts at this scan produced confident wrong answers. That is the same
lesson as the `mdx-to-js.js` anomaly earlier: check the control before believing it.

kwakayama:

Follow-up filed for the dnt bump this PR defers to, so the reference above points at something real: **veryfront-issue-inbox#713** — bump `@deno/dnt` to 0.43.x and emit the automatic JSX runtime.

It records the reasons this was kept out of the hotfix (a year of dnt changes in one step; three hand-written patches over dnt output; ~190 modules gaining a `react/jsx-runtime` import, which touches `browser-safe-exports.mjs` and `lint:client-bundle`), what to diff, and one thing that matters for whoever takes it: **do not delete `npm-jsx-react-binding.ts` when it lands.** Once the emit is automatic the rewrite patches nothing, but the assertion it ends with is what catches a regression in either direction. Keep the check; let the rewrite become dead weight.

The generator enumeration above is recorded there too, so it does not need rediscovering.

kwakayama:
## `tests (node)` failed, and it is not this PR — main was already red

Correcting my own "all green" from earlier in this thread: the last job to finish came back
red. It is pre-existing breakage on `main`, not a regression here, and I have rebased onto
the fix.

### What failed

```
✖ failing tests:
test at src/tool/remote-mcp.test.ts:1:1
✖ …/src/tool/remote-mcp.test.ts (6648.124607ms)
  'test failed'

ℹ tests 8846   ℹ pass 8845   ℹ fail 0   ℹ skipped 1
##[error]Process completed with exit code 1.
```

Note `fail 0` beside exit 1 — no individual test failed. The *file* failed, which is what
Node's runner reports when an error escapes after the tests themselves have completed. Easy
to misread the summary as green.

### It reproduces on my base commit with none of my changes

`tests (node)` fails identically on `6be0a2622` — the commit this branch was cut from —
before any of my commits exist.

Bisecting `main`:

| Commit | `tests (node)` |
| --- | --- |
| `db912546b` | pass |
| **`27cef99a8`** — *test: one way to control outbound fetch, and it cannot fail open* (#3945) | **fail** — same file |
| `19de7b679`, `e22594f93`, `86d3e08af`, `6be0a2622` | fail — same file |

`main` was red for five consecutive runs, roughly an hour. The trigger is #3945, which
removed the `globalThis.fetch` assignment pattern that this test relied on.

### Already fixed on main

[#3952](https://github.com/veryfront/veryfront-code/pull/3952) — *reject pre-aborted requests
before transport* — merged at 17:03 and describes exactly this signature: "prevent a late
pinned-socket `socket hang up` from escaping after the caller has already received the
normalized abort error". A late error escaping after the test completes is precisely the
`fail 0` + exit 1 shape above.

This branch is now rebased onto `045abb139`, which contains it. Re-verified after the rebase:
`deno task lint`, `fmt:check` and the unit suite all exit 0; `build:npm` re-running.

Nothing in this PR to change. Flagging it because a reviewer landing on a red PR should not
have to work out that the red belongs to someone else.

PR #3954, opened 2026-08-21
fix(build): emit npm JSX with the automatic runtime

kojiwakayama:
## Summary

- Upgrade dnt from 0.42.3 to 0.43.2, where JSX compiler options affect JavaScript emission.
- Share the explicit `react-jsx` and `react` compiler configuration between the root and extension npm builders.
- Add an artifact-level regression that executes emitted JSX under Node after removing the global `React` binding.
- Preserve declaration maps and native Node timer behavior across the dnt upgrade.

## Why

The repository already selects React's automatic JSX runtime, but dnt 0.42 did not apply those compiler options while emitting JavaScript. It produced unbound `React.createElement` calls in published modules.

#3953 works around that output by reparsing and rewriting emitted JavaScript. That duplicates compiler scope analysis, can change binding semantics, and only patches one builder. This change fixes the emitter configuration instead, with one typed option set used by both npm build paths.

## Proof of the fix

- On `main`, the new artifact regression fails because dnt emits `React.createElement(...)` and no `react/jsx-runtime` import.
- On this branch, the same fixture emits `react/jsx-runtime`, then Node imports and executes it successfully after `delete globalThis.React`.
- The real generated Popover artifact imports `react/jsx-runtime` and was imported and invoked successfully with no global `React`.
- A generated-artifact `no-undef` scan found no unresolved `React` references across the root and extension npm packages.

## Artifact review

- Declaration maps remain enabled explicitly because dnt 0.43 changed their default.
- dnt 0.43 removes its timer shim, so root and extension packages use native Node timers.
- Remaining manual `React.createElement` output has an in-scope binding.

## Validation

- `deno task verify:dist`
- `deno task typecheck:consumer`
- `deno task fmt:check`
- `deno task lint`
- `deno task lint:client-bundle`
- `deno task lint:cwd-relative-test-reads`
- `deno audit --lock=scripts/deno.lock`
- `bash scripts/test/npm-install-smoke.sh`
- Pre-push unit suite: 3,980 tests passed across the parallel and cwd-isolated runs
- `deno task test:scripts`: 143 passed, one unrelated API-reference anchor test failed identically on untouched `main`

Addresses veryfront/veryfront-issue-inbox#712.
Fixes veryfront/veryfront-issue-inbox#713.
Supersedes #3953.


kojiwakayama:
@codex review

PR #710, opened 2026-03-23
chore: replace --allow-all with explicit permissions in test tasks

ariskemper:
## Summary
- Replaces `--allow-all` with explicit `--allow-read --allow-write --allow-net --allow-env --allow-run --allow-sys --allow-ffi` in 8 test/coverage tasks
- Documents exactly which Deno permissions tests require
- Reduces blast radius of compromised test dependencies

Part of supply chain security hardening (Phase 4).

## Test plan
- [ ] CI unit tests pass with explicit permissions
- [ ] CI integration tests pass with explicit permissions

PR #3960, opened 2026-08-22
fix(provider): keep an unreadable 429 retryable and name what timed out

kwakayama:
## Summary

Two provider-retry defects in `src/provider/runtime-loader/provider-http.ts`.

**A 429 with an unreadable body was classified non-retryable.** Both the
OpenAI-compatible branch and the Google branch carried:

```ts
if (truncated || parsedBody === undefined) {
  return new ProviderRequestError({ provider, status, message, retryable: false });
}
```

The bounded stream retry loop added in #3753 only retries a classified
`ProviderRateLimitError`, so a 429 whose body was empty, truncated past the
8 KB classification prefix, or not JSON at all (an HTML error page from a
gateway, for example) skipped the retry loop entirely and ended the agent run
terminally.

An unreadable body is not evidence of exhausted quota. Only a parsed
`insufficient_quota`, or a parsed Google `RESOURCE_EXHAUSTED` carrying no retry
delay, can say that — and both of those still classify terminal, unchanged.
Every other 429 now stays a retryable `ProviderRateLimitError` and reaches the
existing bounded retry. A hard quota that we misread costs at most two extra
attempts; it can never hot-loop.

**Provider timeouts named nothing.** `providerTimeoutError` produced
`"<label> request failed: request timed out"` and no more — not the model, not
the elapsed time, not which of the two deadlines fired (30 s stream headers vs.
the 5-minute JSON deadline). Every event serialized to the same sentence, so
one slow model and a stalled gateway were indistinguishable. `requestJson` and
`requestStream` now take an optional `modelId` and report all three:

```
veryfront-cloud request failed: request timed out after 30004ms waiting for the
stream response headers (30000ms deadline, model <model-id>)
```

The three LLM extensions pass the model they were built for, so the detail is
present in production and not only when a caller opts in.

## Decision on the timeout `retryable` flag

`providerTimeoutError` sets `retryable: true`, and the run still dies. Nothing
on the agent-run path reads that flag. `classifyThrownProviderError` in
`src/agent/streaming/stream-outcome.ts` maps the timeout to
`{code: "PROVIDER_STREAM_ERROR", retryable: true, terminal: false}` correctly —
and then `finalize` in `src/agent/ag-ui/lifecycle-adapter.ts` turns *any*
`outcome.status === "failed"` into a `RunError` with `sawTerminalError = true`,
`retryable` unread. There is no outer retry at that layer to honour it.

**This PR therefore ships only the diagnostic half of that fix.** Honouring the
flag needs one of two changes, neither of which belongs in a diff this size:

1. **Retry inside `requestStream`.** The rate-limit loop reuses one
   `RequestDeadline` for every attempt. When the deadline is what elapsed, that
   budget is spent and its signal is already aborted, so a second `fetch` would
   reject immediately. A timeout retry needs a fresh deadline per attempt, which
   redefines `headersTimeoutMs` from a total budget into a per-attempt one and
   triples the worst case before a caller hears anything (30 s → 90 s) for every
   provider call in the codebase.
2. **Retry the provider attempt at the agent-run layer.** That is new retry
   orchestration in the lifecycle runner, not a change to this file.

Neither is a small diff, and the second is where the real fix belongs — the
issue's premise that the flag is dropped in `provider-http.ts` is not quite
right; it is dropped at the consumer. Filed as remaining work on
veryfront/veryfront-issue-inbox#710.

Note that the timeout on the stream path can only fire *before* response headers
arrive (`deadline.cancelTimeout()` runs before the body is returned), so neither
option risks replaying a stream after provider output was exposed. The blocker is
the deadline budget and the missing consumer, not output safety.

## Red-green TDD

Red — 10 failing steps against the unchanged source:

```
provider-http ... openai 429 truncated before it could be parsed -> retryable rate limit
provider-http ... openai 429 with an empty body -> retryable rate limit
provider-http ... openai 429 with a non-JSON body -> retryable rate limit
provider-http ... google 429 truncated before it could be parsed -> retryable rate limit
provider-http ... google 429 with an empty body -> retryable rate limit
provider-http ... google 429 with a non-JSON body -> retryable rate limit
provider-http ... retries unparseable ambiguous 429 bodies
provider-http ... requestJson ... names the model, the elapsed time, and the deadline that fired
provider-http ... requestStream ... names the model, the elapsed time, and the deadline that fired
provider-http ... requestStream ... omits the model from a timeout when the caller did not name one
FAILED | 0 passed (50 steps) | 1 failed (13 steps)
```

Green after the fix: `ok | 1 passed (63 steps) | 0 failed`.

Three existing tests asserted the behaviour this PR reverses ("fails closed when
an oversized OpenAI 429 body is truncated", the Google twin, and "fails closed
for unparseable ambiguous 429 bodies"). They are rewritten to the new rule rather
than deleted. The regression guards that must not move are untouched and still
pass: parsed `insufficient_quota` → `ProviderQuotaError`, parsed
`rate_limit_exceeded` → `ProviderRateLimitError`, Google `RESOURCE_EXHAUSTED`
with no retry delay → `ProviderQuotaError`, Google with `Retry-After` or
`RetryInfo` → retryable.

## Verification

Head `8f32b25073d739bb1d60d9c195c0fa81ab3eafc4`.

- `src/provider/runtime-loader/provider-http.test.ts` — 63 steps, passed
- `deno task test:unit` — 3924 passed (30696 steps), 1 failed
- `deno lint`, `deno fmt --check src/`, `deno check src/index.ts` — passed
- `deno check` on all three LLM extension entry points — passed
- `deno task docs:api-reference:check`, `deno task generate:manifests:check` — passed
- `deno task lint:anti-slop`, `deno task lint:extension-contracts` — passed

The one unit failure is
`src/transforms/mdx/esm-module-loader/utils/source-spans.test.ts` → "keeps
line-broken division scans within a bounded runtime", a wall-clock bound
unrelated to this diff. It reproduces identically on a clean `origin/main`
worktree, so it is pre-existing.

`templates/manifest.generated.ts` is deliberately not in this diff: `deno task
generate` rewrites its compressed blob to a byte-different but equivalent value,
and the committed file passes `generate:manifests:check` as-is.

Fixes veryfront/veryfront-issue-inbox#172
Refs veryfront/veryfront-issue-inbox#710

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of rate-limit responses, including incomplete or unreadable response bodies, so eligible requests can retry correctly.
  - Streaming retries now preserve rate-limit errors when retry limits are exceeded.
  - Timeout messages now provide clearer timing details and, when available, the model involved.
- **Documentation**
  - Updated API reference links for shared request methods.
- **Improvements**
  - Added model identification to requests across supported AI providers for more informative diagnostics.


kwakayama:
## Review — score: 86/100. **fix-then-merge.** Not merging yet.

Independently reviewed by running the code, not by reading the diff. This repo currently gets no
automated review at all (CodeRabbit skips repos under 10 stars — this one has 9 — and Codex is out
of credits, both while reporting green: veryfront-issue-inbox#724), so this is the only review
signal.

### The regression I most feared did not happen, and it is mutation-proven

`insufficient_quota` is **still non-retryable in both branches**. The check at `:286` sits above
the deleted branch, and the Google `RESOURCE_EXHAUSTED` check at `:318` likewise. Disabling both
guards (`if (false && …)`) turns 4 tests red:

```
openai 429 insufficient_quota -> non-retryable quota
mistral 429 insufficient_quota -> non-retryable quota
google 429 RESOURCE_EXHAUSTED without a retry delay -> non-retryable quota
google 429 RESOURCE_EXHAUSTED with QuotaFailure but no RetryInfo -> quota
```

Retry is still bounded — `MAX_PROVIDER_STREAM_RATE_LIMIT_RETRIES = 2`, delay
`retryAfterMs ?? 1000 * 2^n`, every wait tied to `deadline.deadlineSignal`. Measured against a
persistent unreadable 429: 3 calls, 3005 ms, then throws.

Red-green is **strong and was executed**: reverting `provider-http.ts` and all three extensions to
the merge base with the test file kept gives `FAILED | 0 passed (50 steps) | 1 failed (13 steps)`
— exactly the 10 steps claimed.

And #710 was handled as diagnostics-only, correctly. Verified independently rather than from the
body: `deadline.cancelTimeout()` runs at `:874` **before** `streamWithCleanup` returns the body, so
once headers arrive the timer is dead and `deadline.timedOut` can never become true afterwards.
**No path replays a stream after provider output or a tool side effect was exposed.** The new
message leaks nothing — provider label, elapsed ms, which deadline fired, deadline ms, model id.

### Finding 1 — CONFIRMED, medium. Blocking.

**An unreadable 429 carrying a long `Retry-After` is rewritten into a false timeout error.**

`:857` waits on `deadline.deadlineSignal`. When that aborts mid-wait, the catch at `:881` discards
the real 429 and throws `providerTimeoutError`. Measured with `Retry-After: 60` under a 120 ms
deadline:

```
fetch calls: 1
class:   ProviderRequestError      (truth: ProviderRateLimitError)
message: ...request timed out after 121ms waiting for the stream response headers
         (120ms deadline, model moonshotai/kimi-k2.6)
```

Headers **did** arrive — carrying a 429. At the 30 s production default with a 60 s `Retry-After`
(routine for OpenAI and Moonshot), Sentry will show a stalled-gateway timeout for what is plainly
a rate limit.

That is precisely the confusion #710 asks this PR to remove, reintroduced through the very branch
this PR widened — while the PR's own second half is about making timeouts honest. The population
routed into it is new even though the mechanism predates #3753.

Fix is about five lines: clamp `retryDelayMs` to the remaining deadline, or throw `err` when the
wait cannot fit.

### Finding 2 — CONFIRMED, low
A hard `insufficient_quota` body over `MAX_ERROR_BODY_BYTES = 8000` truncates, loses the code, and
becomes retryable — 3 requests against a hard-quota'd provider instead of 1. Bounded, acknowledged
in the body, real bodies are ~200 bytes. Acceptable as-is.

### Finding 3 — minor
`elapsedMs` uses `Date.now()` (`:759`, `:838`). An NTP step yields a nonsense elapsed value in the
very message meant to be diagnostic.

### Also verified right
Both branches fixed, not one. All 11 `requestStream`/`requestJson` call sites thread `modelId` —
no missed caller. The three reversed tests were rewritten in place, not deleted.
`deno task generate:manifests:check` exits 0 with the committed manifest; `deno lint` and
`deno fmt --check` clean; 30/30 CI green. `requestStream` is the only programmatic consumer of
`ProviderRateLimitError` anywhere in `src/` or `extensions/`, so nothing else can loop on the
reclassification.

### To reach the bar
Finding 1 only. Findings 2 and 3 are follow-ups.

kwakayama:
## Blocking finding addressed — `2dc7e73139`

Confirmed and fixed. An unreadable 429 carrying a `Retry-After` longer than the
deadline's remaining budget was rewritten into a false timeout.

**Mechanism.** `waitForProviderRateLimitRetry` waits on `deadline.deadlineSignal`.
When the deadline aborts mid-wait, the rejection reaches the outer catch, where
`deadline.timedOut` is true — so the classified 429 is discarded and replaced by
`providerTimeoutError`. The headers had arrived; the provider had answered, and
it answered with a rate limit.

**Which option I took: surface the real error, not clamp the delay.**

Clamping `retryDelayMs` to the remaining deadline would have kept a retry alive,
but it retries *early by construction* — before the provider said the limit
clears. Against a 60s `Retry-After` under a 30s deadline, the clamped wait retries
at ~30s into a limit with 30s left to run, so the second attempt is near-certain
to 429 again. It spends an attempt and still reports failure, just later.

Throwing the classified error says what actually happened. `retryable: true` and
`retryAfterMs` both survive on the error, so an outer caller that *can* wait 60s
still has everything it needs to decide — this narrows where the retry decision is
made rather than removing it.

```ts
// A wait longer than the deadline has left aborts part-way through,
// and the catch below would then rewrite this rate limit into a
// timeout the provider never caused. Retry-After routinely exceeds
// the 30s header deadline, so report the rate limit we actually got.
if (retryDelayMs >= headersTimeoutMs - (Date.now() - startedAt)) throw err;
```

**Scope note.** The mechanism predates this branch — it arrived with #3753 for
parsed rate limits — but this PR routes unparseable 429s into it, so it is fixed
here. That also means the test added with #3753, "keeps rate-limit backoff inside
the stream header deadline", asserted the defect (`ProviderRequestError` /
`"request timed out"`). Its genuinely valuable half — that a delay past the
deadline spends no second attempt — is kept and strengthened; only the error
identity changes.

**Red then green.** Structurally red, on the class rather than the wording:

```
provider-http ... keeps rate-limit backoff inside the stream header deadline
provider-http ... surfaces an unreadable 429 as a rate limit when its delay outlives the deadline
error: AssertionError: Expected error to be instance of "ProviderRateLimitError",
       but was "ProviderRequestError".
FAILED | 0 passed (61 steps) | 1 failed (3 steps)
```

Green after the guard: `ok | 1 passed (64 steps) | 0 failed`. Both new tests assert
the class *and* that the message does not claim a timeout. The existing retry
coverage ("retries a rate-limited stream request before provider output", "bounds
rate-limit retries") still passes, so retrying is narrowed, not disabled.

**Gates.** `deno task test:unit` — 3969 passed (30699 steps), **0 failed**.
`deno lint`, `deno fmt --check src/`, `deno check src/index.ts`,
`docs:api-reference:check`, `generate:manifests:check` — all pass. The
API-reference doc needed no regeneration: the guard sits inside `requestStream`'s
body, below its declaration line, so no documented line number moved.

Correction to an earlier note in this PR: the
`src/transforms/mdx/.../source-spans.test.ts` failure I reported as pre-existing
"including in isolation" **passed** in this run. It is load-dependent flake on a
busy machine, not a standing break. It has never failed in CI here.

The two non-blocking follow-ups (3 requests for a truncated `insufficient_quota`
over `MAX_ERROR_BODY_BYTES`; `Date.now()` being NTP-step-sensitive for `elapsedMs`)
are deliberately not touched in this PR.

kwakayama:
## Re-review after `2dc7e73139` — score: 86 → **93/100. Merge.**

The blocking finding is closed, and the region it touches was re-verified rather than assumed.

### The finding is fixed — same probe, run verbatim

Unparseable HTML body, `Retry-After: 60`, `headersTimeoutMs` 120:

```
BEFORE  class ProviderRequestError    "request timed out after 121ms waiting for the
                                       stream response headers (120ms deadline, model …)"
NOW     class ProviderRateLimitError  "veryfront-cloud request failed: Provider request
                                       failed with status 429"   status 429  retryable true
        fetch calls: 1
```

Both halves hold: correct class, and the message no longer claims a timeout the provider never
caused. No attempt is wasted.

The fix is one line at `provider-http.ts:856`:

```ts
if (retryDelayMs >= headersTimeoutMs - (Date.now() - startedAt)) throw err;
```

### The new test is structurally red

Deleting exactly that guard line — with the harness asserting the line occurred once and that
removal actually changed the file — turns two tests red: the new
`surfaces an unreadable 429 as a rate limit when its delay outlives the deadline`, and the #3753
test `keeps rate-limit backoff inside the stream header deadline`, which this commit correctly
rewrote off its old "request timed out" assertion.

### Nothing regressed — the quota mutations were re-run on the new head

This mattered because the fix lands near the classifier. Disabling both quota guards turns the
same four tests red as before:

```
openai 429 insufficient_quota            -> non-retryable quota
mistral 429 insufficient_quota           -> non-retryable quota
google 429 RESOURCE_EXHAUSTED (no delay) -> non-retryable quota
google 429 RESOURCE_EXHAUSTED (QuotaFailure, no RetryInfo) -> quota
```

`insufficient_quota` and Google `RESOURCE_EXHAUSTED` remain non-retryable in both branches. The
fix touches the retry loop, not the classifier — genuinely independent.

Still bounded at 2, and the honour-path is untouched: `Retry-After: 1` under a 30 s deadline gives
3 calls over 2007 ms, both waits honoured. The guard fires only when the wait cannot fit. And the
timeout still cannot fire after headers arrive — `deadline.cancelTimeout()` at `:877` still
precedes the body return at `:879`.

### One new residual — follow-up, not a blocker

A 429 whose **error-body read** crosses the deadline still reports a timeout
(`"request timed out after 63ms … (60ms deadline)"`). The new guard sits after
`buildProviderError` returns; when the abort lands during the body read there is no classified
error to throw yet, so the catch rewrites as before. Reaching this at the 30 s production deadline
needs a provider to dribble a ≤8000-byte error body for 30 seconds — which *is* a stalled gateway,
so "timeout" is arguably the honest word there.

Carried forward, both minor: `elapsedMs` uses `Date.now()` (NTP-step sensitive), and a truncated
`insufficient_quota` over 8000 bytes costs 3 requests instead of 1.

### Gates
`deno lint` 0, `deno fmt --check` 0, `src/provider/` → 26 passed (255 steps). **CI is still
running on this commit** — merging only once it is green.

PR #3963, opened 2026-08-22
fix(ssr): return 404 when a project release is gone instead of 500

kwakayama:
Fixes veryfront/veryfront-issue-inbox#705

## Problem

When a project is deleted, the API answers 404 to every read for its release —
the asset manifest and every source file alike. SSR did not recognise that as a
routing condition, so every request to a deleted project returned **500** for as
long as the deployment kept serving it.

Two consequences:

1. A routine deletion looked like a platform fault, and each request was reported
   to the error reporter as an application error.
2. `render-page` logged `Layout module failed to load (non-critical)` at `warn`,
   and the same missing module hard-failed the render milliseconds later. Anyone
   triaging by severity was misled.

## Root cause

`resolveSSRFailure` already knows how to answer this: `resolveSSRControlOutcome`
maps a `file-not-found` `VeryfrontError` to the `not-found` outcome, which
returns 404 with a framework error page.

The defect is that the veryfront-api filesystem adapter never raised that error.
`createNotFoundLikeError` threw a **raw `Error`** with `code: "ENOENT"`, so the
error arrived at SSR with no registry identity, fell through every branch, and
landed on `server-error` → 500.

So the fix is to stop throwing a raw `Error` — per the repo's error-handling
convention, absence is `FILE_NOT_FOUND` from the registry, identified by slug:

```ts
export function createNotFoundLikeError(path: string): NotFoundLikeError {
  return Object.assign(
    FILE_NOT_FOUND.create({ detail: `404 Not Found: ${path}`, context: { path } }),
    { code: "ENOENT" },
  );
}
```

**`resolveSSRFailure` needed no change at all.** No new error was added to the
registry, and no bespoke error-tagging mechanism was introduced — the existing
`file-not-found` slug was already the right identity and was already wired to a
404. `code: "ENOENT"` is retained because `isNotFoundLikeError`, the adapter's
own optional-file catches, and the platform's Node/Deno absence checks all read
it; tests pin that.

`stat-operations.ts` already tested for exactly this shape
(`error.slug === "file-not-found"`), which is further evidence it is the
intended representation for this adapter.

## Two smaller pieces

- `loadModulesInParallel` wrapped a critical page-module failure in a
  `RENDER_ERROR`, which drops the `file-not-found` identity. It now raises
  `FILE_NOT_FOUND` when *every* critical failure was an unretrievable source
  (`every`, matching the neighbouring `tenantBuildFailure`: one module that
  loaded and threw must not be reported as an absent release). Without this a
  deleted project would answer 404 on the layout path and 500 on the page path.
- The `(non-critical)` log is split. When the source is unreachable the apply
  phase reloads from the same place and fails identically, so the render is
  already lost: that logs at `error` and no longer claims non-critical. A
  failure that can still recover in the apply phase logs at `warn`, now saying
  so.

Nothing in the rendering pipeline is restructured.

## Does this leak which projects exist?

Turning a 500 into a 404 changes what an unauthenticated caller can infer, so I
checked what the neighbouring conditions already answer:

- A host that resolves to no project already returns **404**
  (`createProjectNotFoundProxyContext`, `src/proxy/proxy-error-context.ts:49`).
- A project with no active release already returns **404**
  (`createReleaseNotFoundProxyContext`, same file).

After this change a deleted project returns 404 as well — the *same* status as a
project that never existed. Before the change, 500 uniquely marked "this project
existed and its content has vanished", which distinguished a deleted project
from a never-existed host. **The change removes that distinction, so it reduces
existence disclosure rather than adding any.**

The response body is `ErrorPages.notFound(slug)`, which echoes only the caller's
own request path and escapes it. It does not name the project, the release, or
the API URL; the path in the error's `detail`/`context` stays server-side in
logs. I deliberately did **not** route this to `ErrorPages.undeployed()`, whose
text ("This project has not been deployed yet") would assert something about the
project's state that the generic 404 does not.

No path handling, redirect, or auth boundary is touched.

## Tests

Written first, each watched fail:

- `createNotFoundLikeError` raises a `VeryfrontError` with slug `file-not-found`
  and status 404, and still satisfies `isNotFoundLikeError`,
  `isCanonicalNotFoundError`, and `code === "ENOENT"`.
- `SSRService.renderPage` answers **404** for the error the adapter actually
  raises.
- **Regression guard:** a layout that exists but throws still yields
  `server-error` / **500**, as does an unbranded error carrying `code: "ENOENT"`
  or a message reading `404 Not Found`. Message text is never consulted.
- A terminal layout failure logs at `error` without the word "non-critical"; a
  recoverable one still logs at `warn`.
- An unretrievable page module keeps its `file-not-found` identity through
  `loadModulesInParallel`; one that loaded and threw keeps `render-error`.

## Gates

`deno lint`, `deno fmt --check src/`, `deno task typecheck`, and
`deno task test:unit` all pass.

### DIFF
--- a/docs/api-reference/veryfront/observability.md
+++ b/docs/api-reference/veryfront/observability.md
@@ -82,15 +82,15 @@ const result = await withSpan("load-data", async () => {
 | `parseMaxSize`                           | Parses max size.                                                     | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/file-log-subscriber.ts#L47)                   |
 | `profilePhase`                           |                                                                      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/request-profiler.ts#L191)                     |
 | `profileSyncPhase`                       |                                                                      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/request-profiler.ts#L211)                     |
-| `recordApiRequest`                       |                                                                      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L307)      |
-| `recordApiRetry`                         |                                                                      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L322)      |
+| `recordApiRequest`                       |                                                                      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L326)      |
+| `recordApiRetry`                         |                                                                      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L341)      |
 | `recordBuild`                            | Record build.                                                        | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L124)                        |
 | `recordBundle`                           | Record bundle.                                                       | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L132)                        |
 | `recordCacheGet`                         | Record cache get.                                                    | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L56)                         |
 | `recordCacheInvalidate`                  | Record cache invalidate.                                             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L69)                         |
 | `recordCacheSet`                         | Record cache set.                                                    | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L64)                         |
-| `recordContentCacheHit`                  | Record a content cache hit at the specified layer                    | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L344)      |
-| `recordContentNetworkFetch`              | Record a content network fetch with timing                           | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L370)      |
+| `recordContentCacheHit`                  | Record a content cache hit at the specified layer                    | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L363)      |
+| `recordContentNetworkFetch`              | Record a content network fetch with timing                           | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/metrics-recorder.ts#L389)      |
 | `recordCorsRejection`                    | Record CORS rejection.                                               | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L165)                        |
 | `recordDataFetch`                        | Record data fetch.                                                   | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L152)                        |
 | `recordDataFetchError`                   | Error shape for record data fetch.                                   | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/metrics/index.ts#L160)                        |
@@ -180,7 +180,7 @@ const result = await withSpan("load-data", async () => {
 
 | Name      | Description | Source                                                                                                        |
 | --------- | ----------- | ------------------------------------------------------------------------------------------------------------- |
-| `metrics` |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/index.ts#L76) |
+| `metrics` |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/simple-metrics/index.ts#L78) |
 | `trace`   |             | [source](https://github.com/veryfront/veryfront-code/blob/main/src/observability/tracing/api-shim.ts#L588)    |
 
 ## Deep imports
--- a/src/observability/simple-metrics/index.ts
+++ b/src/observability/simple-metrics/index.ts
@@ -48,6 +48,7 @@ export {
   recordRSCStreamDuration,
   recordSecurityHeaders,
   recordSSR,
+  recordSSRSourceUnavailable,
 } from "./metrics-recorder.ts";
 
 export type { ContentCacheLayer, ModuleServeStatus } from "./metrics-recorder.ts";
@@ -70,6 +71,7 @@ import {
   recordRSCStreamDuration,
   recordSecurityHeaders,
   recordSSR,
+  recordSSRSourceUnavailable,
 } from "./metrics-recorder.ts";
 import { createSnapshot, getRequestCount } from "./metrics-state.ts";
 
@@ -91,6 +93,7 @@ export const metrics = {
   recordModuleServe,
   recordModuleTransform,
   recordRouteManifestLookup,
+  recordSSRSourceUnavailable,
   snapshot: createSnapshot,
   getRequestCount,
 };
--- a/src/observability/simple-metrics/metrics-recorder.ts
+++ b/src/observability/simple-metrics/metrics-recorder.ts
@@ -293,6 +293,25 @@ export function recordCorsRejection(): void {
 }
 
 /**
+ * Record that SSR answered 404 because a project's published source could not
+ * be read, as opposed to an ordinary route miss.
+ *
+ * This is the alertable replacement for the per-request error report that the
+ * same condition used to raise while it was misclassified as a render fault.
+ * A deleted project moves it a bounded number of times and then stops; an
+ * API-side regression that 404s every read moves it continuously, which is the
+ * shape worth alerting on.
+ *
+ * @example
+ * ```ts
+ * recordSSRSourceUnavailable()
+ * ```
+ */
+export function recordSSRSourceUnavailable(): void {
+  state.ssrSourceUnavailable = saturatingAdd(state.ssrSourceUnavailable);
+}
+
+/**
  * Record security headers application
  *
  * @example
--- a/src/observability/simple-metrics/metrics-state.ts
+++ b/src/observability/simple-metrics/metrics-state.ts
@@ -30,6 +30,7 @@ export const state: MetricsState = {
   ssrHistogram: undefined,
   rscStreamHistogram: undefined,
   corsRejections: 0,
+  ssrSourceUnavailable: 0,
   securityHeadersApplied: 0,
   apiRequests2xx: 0,
   apiRequests4xx: 0,
@@ -81,6 +82,7 @@ export function createSnapshot(): VeryfrontMetrics {
     routeManifestLruHits: state.routeManifestLruHits,
     routeManifestLruMisses: state.routeManifestLruMisses,
     corsRejections: state.corsRejections,
+    ssrSourceUnavailable: state.ssrSourceUnavailable,
     securityHeadersApplied: state.securityHeadersApplied,
     apiRequests2xx: state.apiRequests2xx,
     apiRequests4xx: state.apiRequests4xx,
@@ -135,6 +137,7 @@ export function resetMetrics(): void {
   state.routeManifestLruHits = 0;
   state.routeManifestLruMisses = 0;
   state.corsRejections = 0;
+  state.ssrSourceUnavailable = 0;
   state.securityHeadersApplied = 0;
   state.apiRequests2xx = 0;
   state.apiRequests4xx = 0;
--- a/src/observability/simple-metrics/types.ts
+++ b/src/observability/simple-metrics/types.ts
@@ -52,6 +52,8 @@ export interface VeryfrontMetrics {
   routeManifestLruMisses: number;
   ssrHistogram?: { boundaries: number[]; counts: number[] };
   corsRejections: number;
+  /** SSR answered 404 because a project's published source could not be read. */
+  ssrSourceUnavailable: number;
   securityHeadersApplied: number;
   apiRequests2xx: number;
   apiRequests4xx: number;
--- a/src/platform/adapters/fs/veryfront/read-operations-helpers.test.ts
+++ b/src/platform/adapters/fs/veryfront/read-operations-helpers.test.ts
@@ -5,11 +5,14 @@ import { buildFileCacheKeyPrefix } from "./cache-keys.ts";
 import {
   assertProjectSourcePath,
   buildReadFetchState,
+  createNotFoundLikeError,
   getResolvedCacheKey,
   isNotFoundLikeError,
   READ_OPERATION_EXTENSION_PRIORITY,
   splitKnownFileExtension,
 } from "./read-operations-helpers.ts";
+import { VeryfrontError } from "#veryfront/errors";
+import { isCanonicalNotFoundError } from "#veryfront/platform/compat/not-found-error.ts";
 import type { ResolvedContentContext } from "./types.ts";
 
 describe("read-operations helpers", () => {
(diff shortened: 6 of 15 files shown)


kwakayama:
## Review — score: 93/100

Reviewed against `veryfront/veryfront-issue-inbox#705`. **Merge.**

Note on process: this repo currently receives **no automated review** — CodeRabbit skips public
repos under 10 stars and this one has 9, while Codex is out of credits, and both report a green
check (veryfront-issue-inbox#724). So this is a hand review and I verified the load-bearing claims
against the code rather than the PR body.

### The dangerous direction, and why it is closed off

The risk in "return 404 instead of 500" is turning genuine faults into a status nobody alerts on.
That would be strictly worse than the bug being fixed. Two things prevent it here.

**The predicate is a structural identity, not message text:**

```ts
export function isFileNotFoundError(error: unknown): error is VeryfrontError {
  return error instanceof VeryfrontError && error.slug === "file-not-found";
}
```

A timeout, a connection reset, or an auth failure cannot acquire that slug, so none of them can
be mistaken for "gone".

**And a 5xx from the API is rethrown before it can ever reach that path** —
`read-operations.ts:727`:

```ts
const is404Error = isNotFoundLikeError(error);
if (!is404Error) {
  logger.error("Failed to fetch published content", { … });
  throw error;          // keeps its own identity -> still 500
}
```

So only a genuine 404-like read failure is reclassified. I traced this myself rather than taking
the PR's word for it; it is the single claim the whole change rests on.

**There is a real negative test**, not just a happy-path one:

```ts
assertEquals(isFileNotFoundError(error), false, "a genuine fault must not be downgraded to a 404");
assertEquals((error as VeryfrontError).slug, "render-error", "the existing render-error identity is unchanged");
```

That is the assertion I would have written, and its absence would have capped this in the 70s.

### On information disclosure

I considered whether answering 404 rather than 500 creates an enumeration oracle for which
projects exist. It does not meaningfully: both responses tell an anonymous caller "this is not
serving content", and neither distinguishes "never existed" from "deleted". The change does not
widen what an unauthenticated caller can learn.

### The `(non-critical)` log

Actually corrected rather than reworded — the comment now explains that when the source is gone
the reload fails identically and the render is over, so the warning is no longer claiming
non-critical for a failure that is about to be terminal. That was the second defect on the ticket
and it was not skipped.

### Findings

**1. Red-green is asserted, not witnessed (the reason this is 93 and not higher).**
303 insertions across 7 files with new tests at every layer — `read-operations-helpers`,
`pipeline.behavior`, `ssr-outcome`, `ssr.service` — and 27 CI checks green. But I did not re-run
the suites with the source reverted to watch the new tests fail. Structurally they look red-before
(`isFileNotFoundError` was module-private and is newly exported; the pipeline branch did not
exist), so I am fairly confident — but confident is not verified, and I would rather say so than
imply I watched it.

**2. Zero whitespace churn.** `git diff` and `git diff -w` agree exactly at 303/9. Worth noting
because a sibling PR in this batch had 306 lines of pure style churn and had to be sent back.

### Disposition

27 checks green, 6 skipped, no failures, no open comments, and the one property that could have
made this dangerous is closed off and tested. Merging.

kwakayama:
## Revised review — score: 72/100. **fix-then-merge.** Dequeued from the merge queue.

**Correcting my earlier comment on this PR.** I scored it 93 and queued it to merge. An
independent adversarial pass then found a confirmed correctness bug that my review missed, so I
have dequeued this and converted it to draft. It did not merge.

Where I went wrong is worth recording: I verified that `read-operations.ts:727` rethrows any
non-404 error unchanged, concluded that a 5xx or timeout could never acquire the `file-not-found`
slug, and stopped there. That is true **for that path**. I never asked who *else* raises that
slug. The answer is what follows.

### Finding 1 — HIGH, CONFIRMED. A genuine infrastructure fault now returns a silent 404.

`pipeline.ts:479` routes 404-vs-500 purely on `slug === "file-not-found"`. But
`src/transforms/esm/http-cache.ts:573` and `:670` raise `FILE_NOT_FOUND` for something that is
emphatically not an absent source:

```
"[HTTP-CACHE] INVARIANT VIOLATION: Redis recovery write succeeded but file does not exist"
"File write succeeded but file does not exist"
```

Reachable from SSR, every hop read:

```
RenderPipeline.loadModule
  -> module-loader/index.ts:861 loadModule
  -> transformModuleWithDeps
  -> transforms/mdx/esm-module-loader/module-writer.ts:142 cacheHttpImportsToLocal
  -> cacheHttpModule  -> the throws above
```

`module-writer.ts:135-149` does not wrap. `index.ts:873` applies `markBuildFailure`, and
`build-failure.ts:49-55` returns the error **unchanged** — it tags via `defineProperty` rather
than wrapping. The slug survives the whole way.

**Failure scenario:** a page imports an `https://esm.sh/...` specifier. Disk pressure, an evicted
cache entry, or a Redis-recovery race trips the invariant. `every(f => f.missingSource)` is true,
so the render resolves to `not-found`. `ssr.service.ts:451` then logs `logger.debug("SSR notFound")`
and does **not** call `captureApplicationError`.

Before this PR that fault was `RENDER_ERROR` → 500 → Sentry. After it, it is a 404 that pages
nobody. A PR whose purpose is to stop a deletion looking like a platform fault would, in this
path, make a real platform fault look like a deletion — the exact inversion, and worse, because
500s are alerted on and 404s are not.

### Finding 2 — MEDIUM, CONFIRMED. The invariant the design rests on is already false in-tree.

`ssr-outcome.ts:210-216` documents: *"Only framework filesystem adapters raise it, and only where
the source could not be retrieved."* I quoted that in my first review as the reason the change was
safe. It is not true today: `http-cache.ts:573,670`, `discovery/transpiler.ts:199` (a catch-all
that turns EACCES, EIO, abort and timeout into `file-not-found`), and `snippet.handler.ts:178` all
raise it. The transpiler is not on the SSR path today, so that one is latent — but nothing
enforces the invariant, and the routing decision now depends on it.

### Finding 3 — MEDIUM, CONFIRMED by execution. A path containing `404` flips the classifier.

`read-operations-helpers.ts:105-114` ends in `errorMessage.includes("404") || includes("Not Found")`.
Executed against the real transport wrapper:

```
5xx                                        -> false
connection refused, plain path             -> false
connection refused, path .../pages/404.tsx -> true    <-- becomes a 404
```

`pages/404.tsx` is a first-class convention here (`error-page-fallback.ts:89-94`). During an API
outage, requests to those routes silently 404. Pre-existing heuristic, newly consequential.

### Finding 4 — MEDIUM, CONFIRMED. The Sentry signal is removed without a replacement.

`ssr.service.ts:451-459`: `not-found` logs at `debug` with no `captureApplicationError`. An
API-side regression that 404s every read becomes invisible in Sentry — the very class of fault
this issue's 27 events made visible. Loki still gets a line (`pipeline.ts:441`, `:452` log at
`error`), so this is a downgrade rather than a blackout, but a fix whose point is removing a noisy
Sentry signal should leave a deliberate replacement behind.

### Findings 5 and 6 — LOW, follow-ups
`ssr-outcome.ts:217-219` checks only the top-level error while its sibling walks `cause`/`errors`;
correct today, silently reverts the day anyone adds `RENDER_ERROR.create({ cause })` in the apply
path. And `ssr.service.test.ts:547-563` injects at the mock-renderer boundary, bypassing the real
`applyLayoutsAndWrappers` — the exact path in the production trace.

### What genuinely holds up

Red-green is real and was **executed**, not asserted: reverting the three source files while
keeping the tests produces 3 failures; reverting `pipeline.ts` alone reddens both
`does not call a terminal layout failure non-critical` and
`keeps the file-not-found identity of an unretrievable page module`. The `(non-critical)` log fix
is genuine, not a rewording. `vf-error-handling` compliance is exact. Security holds:
`generateErrorHtml` escapes title and message, and both fallback paths are wrapped in
`catch → return null`, so the 404 holds end to end without re-throwing. And the classifier does
correctly reject 5xx, auth failures, and ordinary timeouts — the diagnosis and most of the
execution are right.

### To unblock

One change closes findings 1 and 2: require the adapter's own marker (`code === "ENOENT"`, which
this PR already test-pins) **in addition to** the slug, or re-slug the two `INVARIANT VIOLATION`
throws to a 5xx identity. Ship finding 4 with it. Findings 3, 5, 6 are follow-up issues.

kwakayama:
## Re-review after `05140fce0f` — score: 72 → **86/100. fix-then-merge.** Still draft.

The dequeued bug is genuinely fixed, and proven by a test that drives the real path rather than a
stub. Three things hold it below the bar, all small.

### The enumeration you asked for — 16 raiser sites, 10 files

| File | Raisers | Sets `ENOENT`? |
|---|---|---|
| `platform/adapters/runtime/cloudflare/filesystem.ts` | 5 | no |
| `skill/testing.ts` | 2 | no |
| `platform/adapters/mock.ts` | 2 | no |
| `platform/adapters/fs/github/stat-operations.ts` | 1 | no |
| `platform/adapters/fs/veryfront/stat-operations.ts` | 1 | no |
| `discovery/transpiler.ts` | 1 | no |
| `server/handlers/request/snippet.handler.ts` | 1 | no |
| `rendering/page-resolution/page-resolver.ts` | 1 | no |
| **`read-operations-helpers.ts:134`** | 1 | **yes** |
| **`pipeline.ts:507`** | — | **yes** |

Exactly two set `ENOENT`, and neither is a hole: `createNotFoundLikeError` is the intended adapter
marker with four callers all inside the veryfront fs adapter, and `pipeline.ts:507` is gated on
`criticalFailures.every(f => f.missingSource)`, so it *propagates* the marker rather than minting
one. **No raiser sets `ENOENT` that is not the adapter or a gated propagation of it.**

### The reclassification is tested on the real path

`http-cache.test.ts` monkeypatches `Deno.rename` to delete the file so the write reports success
while the file is absent, drives `cacheHttpImportsToLocal`, and asserts `slug !== "file-not-found"`
and `status === 500`. Reverting **only** `http-cache.ts` to its pre-fix state fails exactly that
test. That is the test that would have caught the original bug.

### Mutation table — 5 of 7 killed, positive control included

| Mutation | Result |
|---|---|
| revert `http-cache.ts` (the dequeue cause) | **killed** |
| predicate hard-wired `false` (positive control) | **killed**, 7 steps |
| drop the `ENOENT` check | **killed**, 3 steps |
| adapter stops setting `code` | **killed**, 7 steps |
| descriptor read → plain property read | **SURVIVED** ← finding 2 |
| `every` → `some` on `missingSource` | survived (documented equivalent) |
| drop the `ENOENT` re-attach on re-raise | **killed** |

The `ENOENT` narrowing is pinned in both directions, so it does not rest on an unenforced
convention. The replacement counter is real and wired at all five levels, asserted both positively
and negatively — not a counter nobody increments.

## Findings

**1. BLOCKER — CI is red. One command fixes it.**
CI runs `deno task lint:ci` (`cicd.yml:50`), not `deno task lint`. Locally `deno lint`,
`deno task lint` and seven lint sub-tasks all pass; only `lint:ci` fails:

```
docs/api-reference is stale:
  outdated: veryfront/observability.md
Run `deno task docs` and commit the result.
```

Benign cause: the 19-line `recordSSRSourceUnavailable` shifted source-line anchors of neighbouring
exports in the generated reference. Verified fix: `deno task docs`, then `deno task lint:ci` → exit 0.

**2. MEDIUM — the getter-forgery defense is unpinned.** `ssr-outcome.ts:244-245`. The defense
works (probed: getter → false, prototype data property → false, own data property → true), but
replacing the descriptor read with `error.code === "ENOENT"` leaves **all 133 steps green**. A
future "simplification" reopens it with CI green — which is the same shape as the defect that got
this PR dequeued: a property asserted in prose and not enforced by a test. ~6 lines closes it.

Second-order, worth knowing: `FILE_NOT_FOUND` and `VeryfrontError` are on the public `./errors`
export, so project code can `Object.assign(FILE_NOT_FOUND.create({…}), { code: "ENOENT" })` — an
own *data* property, which the predicate accepts. Impact is self-limited (a project can force its
own route to 404 and suppress its own alerting; not cross-tenant), so this is defense-in-depth
against accident, not a security boundary. The docstring reads like the latter.

**3. LOW — the new docstring asserts a false invariant, which is the same class of defect that
caused the dequeue.** `ssr-outcome.ts:234-236` says `createNotFoundLikeError` "is the only raiser
that sets ENOENT alongside the slug". `pipeline.ts:507` also sets it. Behaviourally benign — it is
gated, and a mutation proves it pinned — but this PR's own thesis is that the previous docstring
asserted something false in-tree, and the replacement asserts a narrower false thing.

**4. INFO — the narrowing excludes the other filesystem adapters.** `cloudflare/filesystem.ts`
(5 raisers), `github/stat-operations.ts`, `mock.ts` and `skill/testing.ts` raise the slug without
`ENOENT`, so a genuinely absent file under those now routes to **500**, not 404 — as does a real
ENOENT folded through `discovery/transpiler.ts:199`, which loses the marker. The direction is
conservative (toward alerting, not away), so it is safe. But if SSR is ever served through the
Cloudflare adapter, the original #705 symptom returns there.

## To clear the bar
1. `deno task docs` and commit — unblocks CI.
2. Pin the descriptor read: a `FILE_NOT_FOUND` with a `code` **getter** returning `"ENOENT"` must
   not be a missing project source.
3. Soften the two docstring claims — name `pipeline.ts` as the second `ENOENT` setter, and say the
   own-data-property read stops an accidental getter rather than a determined forgery.

kwakayama:
## Final review @ `f64dd24b7` — 72 → 86 → **92/100. Merge.**

All three blocking items land, verified by execution.

### 1. CI — confirmed from the run, not a local pass

`ci (lint)` **pass** on run `32568089538`, job `97019548872`, head `f64dd24b`. Full rollup settled:
28 pass, 6 skipping, 0 fail, 0 pending. The distinction matters here — `deno task lint` passed
locally in the previous round while `lint:ci` was red, which is what made this red in the first
place. This time it was checked against the actual run outcome.

### 2. Mutation D is now killed — the top finding is closed

Replacing the descriptor read with `(error as { code?: unknown }).code === "ENOENT"` now fails:

```
reads the ENOENT marker as an own data property only => FAILED
AssertionError: an accessor must not stand in for the marker, and must not be invoked
```

The test exercises the path rather than merely naming `defineProperty`: an accessor, an
**inherited** marker via `Object.create(proto)` — which a plain property read would accept by
walking the chain — and the own-data-property positive. That third assertion is what stops the test
being satisfiable by a predicate that always returns `false`.

Controls: positive control (predicate hard-wired `false`) killed 8 steps across all three files, so
the harness was live. Mutation C re-verified as a spot-check — still kills, 7 steps. The new test
correctly stays green under C, since C changes *who writes* the marker and the new test covers
*how it is read*; the two guards are independent rather than overlapping.

### 3. The docstring — both parts, and more

It names `pipeline.ts` as the second `ENOENT` setter and gets the enumeration exactly right:
sixteen in-tree raisers, exactly two setting the marker, with the `pipeline.ts` re-raise carrying it
across the boundary "only when every critical failure already had it". It records that removing
either assignment reddens tests — which mutations C and F independently confirm.

And it stops overstating the defense: *"This is a correctness guard, not a security boundary, and
the distinction matters because the check reads as though it were one."* It concedes project code
can assign an own `code` the predicate accepts, and adds a point the review had not made —
`notFound()` already lets a project reach a 404 by design, so forging grants no capability that
does not already exist.

It also folded in the INFO item unprompted: the cloudflare/github/mock/testing adapters answer 500
for a genuinely absent file, which is wrong in the safe direction today because SSR is served
through the veryfront-api adapter, plus what to do if that changes.

### The durable point

The original defect and the review's top finding were **the same shape**: a property asserted in a
docstring and not enforced by a test. This round fixes both the instance and the pattern — the new
test pins the *shape* of the check, not merely its answer. That is why this is a 92 rather than a
grudging 90.

### Disposition
Nine required checks green, no open comments, independent reviewer at 92 with a merge verdict.
Merging.

PR #3969, opened 2026-08-22
Honor delegation denial across runtime boundaries

kojiwakayama:
## Why

A trusted API request could set delegation off, but the external runtime could still expose its local invoke_agent tool because the authority flag was not carried through every runtime transform.

## What changed

- add optional allowDelegation to the signed runtime invocation
- preserve it in hosted chat and internal control-plane requests
- preserve the existing omitted-field behavior for older callers
- remove invoke_agent from internal local and remote tool surfaces when explicitly false
- add normalization and execution-path regression coverage

## Compatibility

The field is optional. Omitted values retain current behavior, so existing API and runtime callers remain compatible during rollout.

## Verification

- independent security/contract review: approved
- deno check on all touched runtime paths
- focused contract + hosted + schema + execution suites: 159 steps passed
- full unit suite: 3,972 suites / 30,711 steps, zero failures
- cwd suites passed
- full pre-push format, lint, typecheck, and unit gates passed

This is the runtime half of the least-privilege delegation fix; the API half is being submitted in parallel.


kojiwakayama:
@codex review

kojiwakayama:
@codex review

PR #3972, opened 2026-08-22
fix(api): load route modules on Node without the Uint8Array hex proposal

### LINKED ISSUE #3968: [BUG] Every API route fails to load on stock Node — module loader calls Deno-native Uint8Array.prototype.toHex()
### Describe the Bug

On `veryfront@0.1.1248` (and still present in the newest `0.1.1249-rc.13735`), **every API route in every app fails to load** on stock Node.js — not just workflow routes, any route. The dev server starts fine, but the first request to any route fails:

```
✗ Failed to load API handler .../app/api/hello/route.ts:
  err=TypeError: (intermediate value).toHex is not a function
```

Root cause: `src/routing/api/module-loader/loader.js` hashes route module content for its cache key using:

```js
// loader.js:71
sha256: new Uint8Array(digest).toHex(),
// loader.js:169
return new Uint8Array(digest).toHex();
```

`Uint8Array.prototype.toHex()` is not available on stock Node.js. It's gated behind V8's `--js-base-64` flag, which is off by default on every currently-released Node — including the newest LTS (24.19.0), checked directly. It also can't be enabled via `NODE_OPTIONS` (Node rejects V8-internal flags there); it has to be passed directly on the `node` binary's own command line, which no `npm`/`veryfront` script does.

**Why this shipped unnoticed:** the npm package is `dnt`-generated from Deno source (`"_generatedBy": "dnt@dev"` in its own `package.json`). `Uint8Array.prototype.toHex()` is a native, unflagged method in Deno, so this passes wherever the test suite actually runs (Deno). The `dnt` transform ships shims for the `Deno` and `crypto` globals for exactly this kind of gap (see `_dnt.shims.js`), but has no shim for this specific `Uint8Array` method — so the incompatibility only shows up in the transformed Node build, which doesn't appear to be smoke-tested on Node before publishing to npm.

I'd frame this as two things, not one:
1. **The immediate bug** — swap `toHex()` for a manual hex-encode. There's already a hand-rolled one two files over, in `src/utils/hash-utils.js`, right next to another `toHex()` that does this correctly without the native method — could reuse that helper directly.
2. **The process gap that let it through** — nothing in CI currently seems to run the published npm (Node) artifact end-to-end on plain `node`, only the Deno-native source/tests. A cheap smoke test (`npm pack`, install it into a scratch app, `veryfront dev`, hit one route) on plain Node — no flags — before publishing would have caught this immediately, and would catch the next Deno-only API someone reaches for.

### Steps To Reproduce

1. `git clone https://github.com/afiodorov/veryfront-node-tohex-bug && cd veryfront-node-tohex-bug`
2. `npm install`
3. `npm run repro`
4. Observe: the simplest possible route (`export async function GET() { return Response.json({ ok: true }) }`) returns `500` instead of `200 { ok: true }` — the module never loads.

### Link to Minimal Reproducible Example

https://github.com/afiodorov/veryfront-node-tohex-bug

### Expected Behavior

`GET /api/hello` should return `200 { "ok": true }` on stock Node.js, no flags required — the same as it does under Deno.

Actual output from `npm run repro`:
```
--- Expected vs actual ---
  GET /api/hello  expected 200 {"ok":true}, got 500 {}

BUG REPRODUCED: the simplest possible API route failed to load.
```

I also confirmed this is Node-specific: running the identical `veryfront@0.1.1248` npm package via `deno run node_modules/veryfront/bin/veryfront.js dev` (Deno 2.6.3) instead of `node` works with no crash at all.

### Environment Information

- Veryfront version: 0.1.1248 (also reproduces on 0.1.1249-rc.13735)
- Runtime: Node.js 22.23.2 and Node.js 24.19.0 (both stock, no flags)
- OS: macOS 15 (Darwin 25.5.0)

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

kwakayama:
Fixes #3968.

## The bug

`src/routing/api/module-loader/loader.ts` hashed route module source with
`new Uint8Array(digest).toHex()`. That is the Uint8Array base64/hex proposal — native and
unflagged in Deno, and **absent from every Node release before 25**, where it sits behind V8's
`--js-base-64` and cannot be enabled through `NODE_OPTIONS`.

On the supported Node floor, `veryfront dev` therefore returned `500` for **every API route in
every app**, not just workflow routes:

```
Failed to load API handler: (intermediate value).toHex is not a function
```

## Reproduction

Minimal app — a `package.json` and one `app/api/hello/route.ts` returning `Response.json({ ok: true })`
— against published `veryfront@0.1.1248`:

| Runtime | `Uint8Array.prototype.toHex` | `GET /api/hello` |
| --- | --- | --- |
| Node 22.22.3 | absent | `500` — `toHex is not a function` |
| Node 24.15.0 | absent | `500` |
| Node 25.9.0 | present | `200 {"ok":true}` |
| Node 25.9.0 + strip preload | removed | `500` — identical failure |

Same app, same package, one variable.

## The fix

Reuse the hand-rolled hex encoder already present in `src/utils/hash-utils.ts`, which is how the
rest of the codebase hashes. Net −5/+3 lines in `loader.ts`.

## Why this shipped

`tests (npm install smoke)` already does a clean-room install of the built npm packages and runs a
real dev server **on Node 24 — which lacks the method**. It stayed green because step 7 only ever
requests `/`. Page rendering never touches the API module loader, so a total API outage was
invisible to the one job positioned to catch it.

Step 8 closes that: it drops an API route into the packed starter and requires `200 {"ok":true}`.

**It strips the proposal methods explicitly instead of trusting the runner's Node version.** Node 25
ships them unflagged, so a version-dependent check would go quietly vacuous the moment CI moves off
Node 24 — the same way step 7 was vacuous for this bug. The step asserts the strip actually worked
before relying on it, so a preload that silently stopped working can't leave the step passing for
the wrong reason.

## Verification against the real build

`deno task build:npm` then `bash scripts/test/npm-install-smoke.sh` on Node 24:

| | steps 1–7 | step 8 | exit |
| --- | --- | --- | --- |
| with the fix | pass | **pass** | 0 |
| without the fix (`.toHex()` restored in the built artifact) | pass | **FAIL** — `API route returned 500, expected 200 (issue #3968)` | 1 |

Step 7 passes in both rows. That is the coverage gap, demonstrated rather than asserted.

Also run: `deno task fmt:check`, `deno task lint`, `deno task typecheck`, and
`deno task test src/routing/api/module-loader/ src/routing/api/handler.test.ts src/routing/api/route-executor.test.ts`
(11 passed, 264 steps, 0 failed).

## Deliberately not in this PR

Two more production call sites use the same proposal and are **not** fixed here:

- `src/security/sandbox/worker-generation.ts:74` — `digest.toHex()`
- `src/security/sandbox/worker-script.ts:159-160` — captures both `toBase64` and `toHex` off
  `NativeUint8Array.prototype` at module load

Step 8 does **not** exercise those paths — I confirmed that from the run, rather than assuming the
class is closed. They are the same latent break on the same Node floor and deserve their own change
with their own reproduction; folding them in here on a guess would ship untested edits to the
sandbox. Worth a follow-up issue.

The `.toHex()` calls in test fixtures (`handler.test.ts`, `route-executor.test.ts`,
`worker-script.test.ts`) are left alone — they only run under Deno.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API module loading while preserving existing behavior.
  * Enhanced compatibility when certain Uint8Array encoding methods are unavailable.

* **Tests**
  * Added smoke-test coverage for API routes in the packaged starter.
  * Verified successful server startup, JSON responses, and HTTP 200 health checks in the simulated environment.

### DIFF
--- a/scripts/test/npm-install-smoke.sh
+++ b/scripts/test/npm-install-smoke.sh
@@ -15,6 +15,8 @@
 #      package's build output to reach the starter templates
 #   7. the packed ai-agent starter starts under Node and renders over HTTP
 #      without unresolved generated runtime helpers
+#   8. an API route in that starter loads and responds on a Node that lacks the
+#      Uint8Array base64/hex proposal (issue #3968)
 #
 # Requires: `deno task build:npm` output in ./npm, node + npm + curl on PATH.
 set -euo pipefail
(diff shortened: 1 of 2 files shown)


kojiwakayama:
Closing as superseded by #3970, which is already merged.

#3970 applies the existing cross-runtime computeHash implementation at all three API module-loader hash sites and adds a packed npm smoke that starts the Node server and requires a real API route to return HTTP 200 with the expected JSON body. The published RC (veryfront@0.1.1250-rc.13757) has also been verified against the original reporter repository on stock Node 22 and Node 24.

The useful additional concerns from this PR are preserved separately:
- the non-vacuous version-matrix coverage is tracked internally in issue-inbox #731 (Node support floor plus current release);
- the remaining sandbox toHex/toBase64 call sites are tracked narrowly in issue-inbox #732.

The stable 0.1.1250 promotion is proceeding in #3973, so merging this conflicting duplicate would add risk without adding loader coverage.

Issue #32, opened 2026-01-08
Preview proxy cannot access user-owned projects - OAuth client credentials design issue


kwakayama:
## Summary

Preview URLs (e.g., `good-payne-viygg.preview.veryfront.com`) return 500 errors for user-owned projects because the proxy uses OAuth client credentials that only have access to organization-owned projects.

## Current Outcome

1. User visits `https://good-payne-viygg.preview.veryfront.com/`
2. Proxy extracts project slug from subdomain
3. Proxy fetches OAuth token using `OAUTH_PREVIEW_CLIENT_ID`/`OAUTH_PREVIEW_CLIENT_SECRET`
4. Proxy forwards request to renderer with `x-token` header
5. Renderer calls `GET /projects` API to find project by slug
6. API returns only projects owned by the OAuth client's organization
7. User's project `good-payne-viygg` is NOT in the list
8. **Result: 500 error "Project not found with slug: good-payne-viygg"**

## Expected Outcome

Preview URLs should work for any project the user has access to, not just organization-owned projects.

## Log Evidence

```json
{
  "message": "[ProxyFSAdapterManager] Creating adapter",
  "projectSlug": "good-payne-viygg",
  "context": {
    "cacheKey": "good-payne-viygg:preview",
    "projectId": "(none)",
    "productionMode": false
  }
}

{
  "message": "[SSR] renderPage failed",
  "error": "Project not found with slug: good-payne-viygg"
}
```

## Root Cause Analysis

The proxy uses **OAuth client credentials grant** (machine-to-machine auth):

```typescript
// proxy/main.ts:102
token = await tokenManager.getToken(scope);  // Uses OAUTH_PREVIEW_CLIENT_ID
```

This token can only access projects owned by the OAuth application's organization. The user's `authToken` cookie is present in the request but ignored by the proxy.

## Why Some Projects Work

Projects like `helpful-durden-fdtcx` work because:
1. They're owned by the same organization as the OAuth client, OR
2. They have a published release and the domain lookup API (`/lookup/domain/`) returns the `projectId` directly, bypassing `listProjects`

### Option 1: Forward user's auth token for preview (Recommended)

For preview requests, forward the user's `authToken` cookie to the renderer instead of using OAuth client credentials:

```typescript
// proxy/main.ts
if (scope === "preview") {
  const userToken = req.headers.get("cookie")?.match(/authToken=([^;]+)/)?.[1];
  if (userToken) {
    newHeaders.set("x-token", userToken);
[...]

PR #4001, opened 2026-08-22
fix(ssr): render previews from a fresh source snapshot

kojiwakayama:
## Summary

- establish the current mutable preview source generation before SSR resolves dependencies or renders HTML
- run freshness inside the matching multi-project context so the refresh and render read the same project and branch
- share the existing bounded freshness contract with API and page routing instead of keeping it inside the Pages API module
- fail the request when freshness cannot be established, preventing stale SSR HTML from reaching hydration
- leave immutable production releases unchanged

Addresses the stale SSR and dropped-generation symptoms in veryfront/veryfront-issue-inbox#32. Preview latency and cache fan-out remain owned by the already-merged #3860 and the separate FileCache work.

## Red-green evidence

Before the change, a controlled refresh gate showed SSR could proceed without calling the freshness boundary, and a rejected refresh did not stop stale rendering. The green regressions prove refresh starts and completes before render, failures prevent render, production skips refresh, and hosted preview refresh runs inside the selected project context.

The existing adapter regressions also prove freshness calls coalesce, followers wait through SSR invalidation, and an older refresh cannot overwrite a newer pushed source snapshot.

## Verification

- focused SSR, API wrapper, Pages API, and adapter suites: 180 steps passed
- request handler and rendering suites: 47 files, 599 steps passed
- Node targeted suites: 68 tests passed
- Bun targeted suites: 3 files passed
- full unit suite passed with leak tracing enabled
- deno task build:npm
- deno task typecheck
- deno task lint
- deno task lint:style
- deno task lint:dependency-boundaries
- deno task lint:module-boundaries
- deno task fmt:check
- deno task generate:manifests:check
- git diff --check

Production verification remains required after release: rapid typing, a hard refresh, direct preview URL SSR, and hydration must converge on the same latest content.


kwakayama:
## Review: one blocking finding. The fail-closed promise has a silent fall-through.

The change is well placed: establishing the snapshot before route resolution, inside the matching multi-project context, and failing the request rather than serving an older snapshot is the right shape. Production stays untouched. The SSR hook correctly calls `endRequest(requestId)` before rethrowing, so a freshness failure does not leak a request slot.

One thing contradicts the fourth bullet in the summary.

### An adapter implementing neither method returns success without establishing freshness

`src/server/handlers/request/source-snapshot-freshness.ts`:

```ts
if (ctx.adapter.fs.ensureSourceSnapshotFresh) {
  await ctx.adapter.fs.ensureSourceSnapshotFresh(reasons.ensure);
  return;
}
// Backward compatibility for custom remote adapters ...
await refreshPreviewSourceSnapshot(ctx, reasons.refreshFallback);
```

and the fallback:

```ts
await ctx.adapter.fs.refreshSourceSnapshot?.(reason);
```

Both methods are optional on the adapter contract:

```
src/platform/adapters/base.ts:313   refreshSourceSnapshot?(reason?: string): Promise<void>;
src/platform/adapters/base.ts:318   ensureSourceSnapshotFresh?(reason?: string): Promise<void>;
```

So an adapter that implements **neither** takes the fallback, the optional call resolves to `undefined`, the function returns normally, and SSR proceeds. Freshness was never established and nothing failed.

The PR body says: *"fail the request when freshness cannot be established, preventing stale SSR HTML from reaching hydration."* On that path it does not. It is the fail-open direction, in the one place the PR exists to close, and it is reachable by construction rather than by accident: the interface explicitly permits an adapter with neither method.

The comment calls the fallback "backward compatibility for custom remote adapters that only implement the original unconditional refresh contract". That is the right intent for an adapter with `refreshSourceSnapshot` and not `ensureSourceSnapshotFresh`. It does not cover an adapter with neither, and the optional call quietly treats that case as success.

**Suggested fix:** make the absence explicit rather than silent.

```ts
if (ctx.adapter.fs.ensureSourceSnapshotFresh) { ... return; }
if (ctx.adapter.fs.refreshSourceSnapshot) {
  await refreshPreviewSourceSnapshot(ctx, reasons.refreshFallback);
  return;
}
throw ...; // freshness cannot be established for a mutable preview source
```

Keep `refreshPreviewSourceSnapshot` as it is, since its own callers may legitimately want the no-op, but do not let `ensure...` reach that behaviour.

**A test that pins it:** a context with a mutable preview source and an adapter stubbed with neither method must make the SSR request fail. That test goes red today.

### Worth a second look, not blocking

`hasMutablePreviewSource` resolves identity as `tryGetCacheKeyContext() ?? extractCacheKeyContext(ctx)`, so an ambient context takes precedence over the explicit `ctx`. Given the summary's emphasis on running "inside the matching multi-project context so the refresh and render read the same project and branch", it is worth stating why ambient-first is correct here. If an ambient context can ever belong to a different project than `ctx`, the freshness decision would be made against the wrong identity, and a preview is exactly where per-project scoping matters. I did not find a case where they diverge; I am flagging the precedence rather than a defect.

### Not verified

I reviewed this statically. I did not run the suite, and I have not checked the red-green evidence in the body against a live run, so treat the above as a code finding rather than a measured one. The blocking item does not depend on that: it follows from the two optional signatures and the optional call.

kojiwakayama:
## Deep review — merge confidence **72/100**

**Verdict:** mergeable with nits — the code is small, correct and well-bounded — but I'd ask for one alignment (or an explicit follow-up issue) before calling #32 addressed, because the divergence this PR leaves in place is the likelier cause of the "original content forever" symptom.

### What I verified (isolated worktree at 0911a5b1e)
- `ssr/` suite: 7 files, 122 steps pass (all 5 new tests execute). `api/` + ssr + new module: 8 files, 145 steps. Full `src/server/handlers/request/` with the `test:unit:parallel` env/flags: 44 files, 511 steps, 0 failed. `adapter.test.ts` + `multi-project-adapter.test.ts` (coalesce / followers wait / older refresh can't overwrite): 8 files, 150 steps pass.
- `deno task lint:module-boundaries` ok; `lint:dependency-boundaries` core=0 cli=0; `deno check` on changed files ok. Merges clean with `origin/main` (10 behind, none touching these paths).
- **Bound/latency**: `VeryfrontFSAdapter.ensureSourceSnapshotFresh` (`adapter.ts:1167`) is a no-op unless >30s since last check (`BRANCH_SOURCE_SNAPSHOT_FRESHNESS_MS`, `adapter.ts:56`) or the branch identity changed; `refreshSourceSnapshot` single-flights via `sourceSnapshotRefreshPromise` (`:1144-1165`). Under a burst: one `listAllFiles` round-trip per adapter per 30s. No explicit timeout — the bound is the API client's.
- **Production untouched**: `hasMutablePreviewSource` (`source-snapshot-freshness.ts:7-12`) returns false for `mode === "production"` or null context; covered by the "does not refresh immutable production source" test.
- **Multi-project**: call sits inside SSR's own `fsAdapter.runWithContext`; `MultiProjectAdapter.ensureSourceSnapshotFresh` resolves the per-project adapter from ALS and clears the request-scoped file cache when the version moved (`multi-project-adapter.ts:258-272`). No cross-project leak.

### Should address (non-blocking, but decides whether #32 closes)
1. **Branch derivation mismatch.** SSR's `runWithContext` passes `branch: ctx.parsedDomain?.branch ?? null` (`ssr.handler.ts:147`, also `:177`), while `api-handler-wrapper.ts:114`, `renderer/adapter.ts:370`, `error-page-fallback.ts:311`, `not-found-fallback.ts:82` all prefer `requestContext.branch`. For a hosted preview whose hostname carries no branch but whose request context does, the wrapper refreshes branch X while SSR renders `main` — which reproduces #32's exact symptom (SSR = original content, hydration via module routes = draft). This PR makes refresh+render self-consistent *with each other*, but both still read the wrong branch in that case. One line + one test to align SSR to `requestContext.branch ?? parsedDomain.branch`, or a follow-up issue.
2. **The fix is largely redundant with the existing pre-SSR refresh.** `ApiHandlerWrapper` (priority 700) already calls the same `ensurePreviewSourceSnapshotFresh` before SSR (priority 1000) on every page GET (`api-handler-wrapper.ts:141`), so the new SSR call is normally a 30s-window cache hit. The "Red-green evidence" that "SSR could proceed without calling the freshness boundary" is true for the handler in isolation only; the narrative overstates what changed in the real chain.
3. **Undocumented fail-closed behaviour change.** Pre-PR, the wrapper swallowed a refresh failure (`api-handler-wrapper.ts:190-196`) and SSR served the retained in-memory snapshot (stale-but-200). Now an API blip on a >30s-old snapshot → thrown → `RouteRegistry` catch → `application/problem+json` **500** for the preview document (and API routes that fall through to SSR now 500 instead of the old 404). Defensible under #32's "no stale SSR", but it deletes a graceful-degradation path silently; the test asserts the handler rejects, not what the user receives. Add a note and consider a retained-snapshot fallback when the adapter already holds one for the same identity.

### Nits
- "awaits … before rendering" test uses two bare `await Promise.resolve()` ticks to assert ordering (`ssr.handler.test.ts` ~145); brittle against any added microtask in `handle()`. Assert via the events array after release, or a deferred `renderPage`.
- `refreshPreviewSourceSnapshot` is exported from the new module with no external callers.

### Spec
#32 item 3 (stale SSR) partially addressed — staleness is bounded to ≤30s after a missed poke, not "always the current draft"; items 1–2 (dropped keystrokes, 2–3s lag) explicitly deferred to #3860. No creep.

### Not re-verified
`build:npm`, full typecheck (CI green), node/bun suites, "full unit suite with leak tracing".

_Reviewed with Claude Code; the branch-derivation divergence was confirmed by reading both call sites._

kojiwakayama:
## Deep review

**Merge confidence: 52/100**  
**Recommendation: REQUEST CHANGES**  
**Reviewed head:** `0911a5b1e6b7` against current `origin/main`

### Finding

- **HIGH:** `src/server/handlers/request/source-snapshot-freshness.ts:38-48` can report success without establishing preview freshness. When a mutable-preview adapter implements neither optional method, the fallback executes `refreshSourceSnapshot?.(...)` and resolves. SSR then renders even though freshness was never established.

  **Failure scenario:** an allowed custom adapter shape with no freshness methods reaches preview SSR, the helper resolves, and stale HTML can reach hydration despite the PR's fail-closed contract. A direct probe with `adapter.fs = {}` confirmed the helper resolves.

  **Fix:** for mutable previews, call `refreshSourceSnapshot` only when present; otherwise throw a typed freshness error. Also make `MultiProjectAdapter.ensureSourceSnapshotFresh` fall back to an underlying refresh-only adapter rather than returning early. Add regressions for missing-both and refresh-only-under-wrapper cases.

### Standards

The hook placement and same-context sequencing are good, but optional chaining masks failure at a correctness boundary.

### Spec

Partial. Default Veryfront adapters establish freshness before dependency resolution/render, and production remains unchanged. The public adapter boundary does not fully satisfy "fail the request when freshness cannot be established."

### Architecture

`WATCH` at the default adapter, but merge-blocking at the public custom-adapter contract because both freshness capabilities are optional.

### Verification

- SSR, API wrapper, Pages API, and adapter focused tests: passed
- `deno task typecheck` and `git diff --check`: passed
- Direct missing-capability probe: confirmed fail-open behavior

kwakayama:
## The finding reproduces. Fixed in a6cfc355ba.

### Reproduced first

A preview context with an adapter implementing neither method rendered a 200. The test asserting a rejection failed exactly as you predicted:

```
fails the request when a snapshot source cannot establish freshness ... FAILED
error: AssertionError: Expected function to reject.
```

`createMockAdapter()` in `ssr.handler.test-helpers.ts` already implements neither, so every other SSR test in the file was running through that path and passing. The fall-through was reachable by construction and exercised in-tree.

### Why a blanket throw would have been an outage

Before fixing it the way you sketched, I checked what implements these methods. `src/platform/adapters/deno.ts`, `node.ts` and `bun.ts` implement **none** of the three snapshot methods (`grep -c SourceSnapshot` returns 0 for each). Those are the adapters `resolveAdapter()` hands to every local project: `adapter-factory.ts:236` takes `runtime.get()` whenever `findLocalProjectPath()` matches, and that branch only runs when `opts.projectSlug` is set. With `projectSlug` set and mode preview, `hasMutablePreviewSource()` is true.

So `if (neither) throw` would have 500'd every local multi-project preview request. That is a fail-open traded for an outage.

### The fix

Absence is explicit, but split by what the adapter declares:

```ts
if (fs.ensureSourceSnapshotFresh) { await ...; return; }
if (fs.refreshSourceSnapshot)     { await refreshPreviewSourceSnapshot(...); return; }
if (typeof fs.getSourceSnapshotVersion !== "function") return;  // live source, nothing to establish
throw SOURCE_SNAPSHOT_FRESHNESS_UNAVAILABLE.create({ ... });
```

All three snapshot methods are optional, and implementing any of them is how an adapter says "my reads come from a snapshot that can fall behind". A local filesystem has no generation to establish, so it proceeds. An adapter that exposes a snapshot version but no way to advance it now fails the request.

New error `source-snapshot-freshness-unavailable`, category SERVER, 503, via `defineError` like every other entry. `refreshPreviewSourceSnapshot` is unchanged and still called from the fallback branch.

### Tests

Four, all in `ssr.handler.test.ts`:

| Test | Guards |
|---|---|
| fails the request when a snapshot source cannot establish freshness | the fix; asserts slug + 503 + `renderCalls === 0` |
| renders a snapshot source that only implements refreshSourceSnapshot | the backward-compatible path |
| renders a live source that declares no snapshot at all | the local dev outage |
| renders immutable production source that cannot establish freshness | production never reaches the throw |

Mutation results, each mutation killing exactly its test:

| Mutation | Red |
|---|---|
| remove the `ensurePreviewSourceSnapshotFresh` call from `ssr.handler.ts` | 5 tests, including the 3 already on this PR |
| make the throw unconditional | renders a live source that declares no snapshot at all |
| drop the `getSourceSnapshotVersion` guard so it never throws | fails the request when a snapshot source cannot establish freshness |
| drop the `mode !== "production"` skip | renders immutable production source |
| drop the `refreshSourceSnapshot` branch | renders a snapshot source that only implements refreshSourceSnapshot |

The cross-render test you asked about already exists: "serves source published between two preview renders" renders twice, publishes new source between them, and asserts the second render reflects it. It dies under mutation 1.

### Your non-blocking question: ambient versus ctx

Closed, they cannot name different projects on this path.

`setupContextAndRender` establishes the ambient context immediately before calling `handleWithContext`, and builds it **from `ctx`** (`ssr.handler.ts:158-168`): `ctx.projectSlug`, `ctx.projectId`, `isProductionMode(ctx)`, `ctx.releaseId`, `ctx.parsedDomain?.branch`. Ambient is derived from `ctx` one frame up, so it cannot name a different project.

Two supporting points:

- `hasMutablePreviewSource` reads only `mode` off the returned context. `projectId` and `versionId` are discarded, and the refresh targets `ctx.adapter.fs`, so per-project scoping comes from the adapter, not from this lookup.
- `runWithCacheKeyContext` (the explicit AsyncLocalStorage store, which would take priority over both) is called only from `css.handler.ts` and `workflow/executor/step-executor.ts`. Neither wraps document rendering.

It is also not new: `getApiHandlerCacheContext` at `pages-api-handler.ts:113-115` on the merge base is character-identical (`tryGetCacheKeyContext() ?? extractCacheKeyContext(ctx)`). The PR moved it verbatim.

The one place the two disagree is `mode`, not identity: `isProductionMode` additionally honours `ctx.config.fs.veryfront.productionMode === true`. When that is set, ambient says production and freshness is skipped, which is the correct answer for a source the config declares immutable.

### Gates

| Gate | Exit |
|---|---|
| `deno task typecheck` | 0 |
| `deno task lint:ci` | 0 |
| `deno fmt --check` (7 changed files) | 0 |
| `deno task test:unit` | 0 (3977 passed, 30778 steps, 0 failed) |

`docs/guides/errors.md` and `docs/api-reference/veryfront/errors.md` regenerated for the new slug; `error-registry.test.ts` counts updated 120 to 121 and SERVER 18 to 19. `templates/manifest.generated.ts` was left at the committed blob: `deno task generate:manifests:check` exits 0 against it, so the local regeneration churn was not committed.

kwakayama:
## Blocking finding closed, and the fix is better than the one I proposed.

I raised the fall-through above and suggested throwing whenever neither `ensureSourceSnapshotFresh` nor `refreshSourceSnapshot` existed. **That suggestion was wrong and would have broken local development.**

Verified on `origin/main`: the local filesystem adapters implement none of the three snapshot methods.

```
                              deno  node
ensureSourceSnapshotFresh      0     0
refreshSourceSnapshot          0     0
getSourceSnapshotVersion       0     0
```

Only `src/platform/adapters/fs/veryfront/adapter.ts` implements `getSourceSnapshotVersion`. So under my version every local `deno`, `node` and `bun` run with a mutable preview context would have thrown. I would have traded a fail-open for an outage, which is the worse of the two.

`a6cfc355ba` uses the right discriminator instead: an adapter declares it has a snapshot by implementing any of the three. No snapshot at all means every read already returns current bytes, so there is nothing to establish and the request proceeds. A snapshot that exists and can fall behind, with no method to advance it, is the case that now throws.

That is the distinction the code needed and my review did not make. Worth keeping the comment that states it, because the next person to read this will have the same instinct I did.

### Status

Blocking finding closed. Zero unresolved threads, zero non-gate check failures. The red `Automated review` is CodeRabbit and Codex both being out of quota tonight, and that check is in no required list.

Queueing.

kojiwakayama:
@codex review

Issue #226, opened 2026-02-04
Framework module resolution fails in compiled binary SSR for layout files


kwakayama:
## Problem
Layout files (both `pages/layout.tsx` and config-based layouts) that import framework modules like `veryfront/context` or `veryfront/head` fail in compiled binary mode with:

```
Module not found: file:///_vf_modules/_veryfront/react/context/index.js?ssr=true
```

## Root Cause Analysis
The SSR module loader transforms code and finds `/_vf_modules/_veryfront/...` imports, but `fetchAndCacheModule` fails to resolve these to `file://` paths pointing to cached framework modules.

Key observations:
1. The `tempFilePath` shows correct cache path structure
2. The import path `/_vf_modules/_veryfront/react/context/index.js?ssr=true` is NOT being replaced with a `file://` path
3. When Deno tries to dynamically import, it prepends `file://` to the relative path, creating invalid `file:///_vf_modules/...` URL

## Affected Tests (compiled-binary-e2e.test.ts)
- `should share page context between layout and page via usePageContext`
- `should render layout defined in veryfront.config.ts`
- `should handle config layout with framework hooks`
- `should render config layout in production mode`

## Investigation Points
1. `fetchAndCacheModule` in `module-fetcher/index.ts` - why isn't it resolving framework paths?
2. `resolveModuleFile` in `file-finder.ts` - framework file lookup
3. `ssrVfModulesPlugin` in `ssr-vf-modules.ts` - framework source transformation
4. Relationship between `contentSourceId` and module resolution

## Related
- Commit bf42b8a0 added embedded source support for compiled binaries
- Tests pass for non-layout pages with framework imports
- The issue is specific to layout loading via `loadComponentFromSource`

PR #4004, opened 2026-08-22
fix(agent): give each heartbeat attempt a deadline so a hang escalates

kwakayama:
Fixes a resilience gap found reviewing #3961: a heartbeat request that **hangs** rather than failing never advances `consecutiveHeartbeatFailures`, so the persistent-failure escalation never fires. The service looks alive to itself while doing nothing.

Refs veryfront/veryfront-issue-inbox#728.

## The change

`retryWithBackoff` already accepts a per-attempt `timeoutMs` (`src/errors/error-handlers.ts:120`, wired at `:184-188`). The heartbeat call at `src/agent/service/registration.ts` never passed one. It does now, and the value is derived in `heartbeatRetrySchedule` alongside the backoff it already computes.

This is **complementary to** the in-flight guard from #3961, not an alternative. The guard stops ticks overlapping; the deadline is what makes a hang observable at all. Before and after that PR, a hang produced zero escalations.

## Choosing the deadline

```
attemptTimeoutMs = floor((interval × 0.9 − totalBackoff) / maxAttempts)
```

At the production interval (30 s, `src/agent/service/config.ts:74`) that is **8750 ms**.

**Why not shorter.** A heartbeat has been measured at `duration_ms=2704` during a degraded period (#709). That is normal-under-stress, not a hang, and a deadline that fails it converts a slow control plane into a dead one. 8750 ms sits 3.2× above it. The margin is larger than it looks: a timed-out attempt is retryable, so one slow answer costs a retry rather than a failed tick — a tick fails only when all three attempts exceed the deadline, and escalation needs three such ticks. That is nine consecutive attempts each over 8.75 s.

**Why not longer.** The whole tick has to fit inside one interval: `3 × 8750 + 750 = 27 000 ms`, exactly 90 % of the interval, with the remaining 10 % as headroom. Past that a hung tick starts skipping beats again and the counter advances slower than once per interval — which is the failure mode being fixed.

**Why derived rather than a constant.** The backoff schedule is already derived from the interval, and the property that matters — *a tick settles inside its own interval* — has to hold for whatever interval is configured, not just 30 s. It also makes the behaviour testable in half a second instead of eighty.

**Escalation latency.** Three failed ticks, so roughly three intervals: ~90 s at production settings. Measured at 511 ms with a 100 ms test interval.

## Is the operator's tick-clock a better fit here?

`veryfront-operator` #226/#227 keys liveness on time since the last **tick** rather than the last **success**, because a hang produces no failures for a failure counter to see. That is the right instrument there and redundant here, for a specific reason.

In the operator, liveness is exported as a metric read from outside a hand-rolled poll loop, so "time since last tick" is the only thing an external observer can measure. Here the signal is a log line emitted by the same `setInterval` that would drive a tick clock — and the in-flight guard already logs a skip warning on every skipped beat, so a hang was never *unobservable*, it was un-*escalatable*.

With the deadline, `maxAttempts × timeout + backoff ≤ interval` holds by construction, so every tick settles inside its own interval. That makes "three consecutive failures" and "no tick completed in three intervals" the same predicate, firing at the same moment. A tick clock would add state and catch nothing extra. `sizes the attempt deadline so a whole tick fits inside one interval` pins that invariant so it cannot silently stop holding.

**Residual risk the tick clock would still cover:** the deadline is cooperative — it aborts the attempt's `AbortSignal`, so a `fetch` that ignores the signal stays unbounded. In production `fetchImpl` defaults to `globalThis.fetch`, which honours it, and the hang test's double models that faithfully. A test double that ignores the signal (two pre-existing tests do) is unaffected by the deadline, which is why both still pass unchanged.

## Tests

Red first. Before the fix, `escalates a heartbeat that hangs rather than waiting on it forever` exhausted its 5 s budget with zero escalations; after, it escalates in 511 ms.

New:
- a heartbeat that never resolves escalates within a bounded time
- a slow-but-successful 3 s heartbeat at the 30 s interval is not treated as a failure
- the deadline arithmetic keeps a whole tick inside its interval, at every interval from 40 ms to 300 s, and clears the observed 2704 ms with margin

Already covered and still green: a 4xx fails immediately with no retry (#709), teardown cancels a pending backoff, persistent 500s escalate in bounded time, ticks never overlap, skipped beats neither reset nor advance the counter.

## Mutation check

| mutation | hang test | slow-success test | 4xx control |
|---|---|---|---|
| none | pass | pass | pass |
| remove `timeoutMs` wiring | **FAIL** (5 s, no escalation) | pass | pass |
| `timeoutMs: 1` | pass | **FAIL** (`The operation was aborted`) | pass |

Each mutation asserted its target text appeared exactly once before rewriting it. `fails a client error immediately without retrying` is the positive control: green under both mutations, so a red result is the assertion and not a broken harness.

## Also noted, not changed

The issue points at `src/sandbox/lazy-sandbox.ts:687-704` as the same pattern. Checked: it is **not** the same gap. That heartbeat goes through `fetchControl` → `fetchWithTimeout(url, controlRequestTimeoutMs, init)`, default 15 s, so a hang there already fails. What it lacks is retry, and it coalesces concurrent beats via `heartbeatPromise` rather than skipping them. Worth its own issue; out of scope here.


kwakayama:
Closing this as a duplicate of #3990, which is already in the merge queue. Both fix veryfront/veryfront-issue-inbox#728 and both touch `src/agent/service/registration.ts` and its test, so only one can land. I measured them against each other rather than picking by date, and the measurement says #3990 is the better of the two.

## Same mechanism, different deadline

Both pass a per-attempt `timeoutMs` into the `retryWithBackoff` this call already used. The PR body here claimed that as this PR's distinguishing idea; it is not. #3990 does the same thing. The only real difference is the value:

- **#3990**: `timeoutMs: input.heartbeatIntervalMs`, one full interval per attempt.
- **this PR**: derived so three attempts plus backoff fit inside 90% of the interval, which works out to 22% to 30% of the interval depending on how the backoff floors:

| interval | deadline here | share of interval |
| --- | --- | --- |
| 40ms | 9ms | 22.5% |
| 500ms | 109ms | 21.8% |
| 30s (production) | 8750ms | 29.2% |
| 300s | 89750ms | 29.9% |

#3990 is 100% of the interval at every setting.

## I ran each PR's tests against the other's code

`src/agent/service/registration.ts`, its test, and `src/errors/error-handlers.ts` are byte-identical at both merge bases, so the test files swap cleanly.

**This PR's tests on #3990's code:** 1 of 3 fails.

- `escalates a heartbeat that hangs rather than waiting on it forever`: **passes**. #3990 fixes the same hang.
- `lets a slow-but-successful heartbeat finish at the production interval`: **passes**.
- `sizes the attempt deadline so a whole tick fits inside one interval`: fails, but only because it reads `schedule.attemptTimeoutMs`, a field that exists only here. It asserts a property #3990 deliberately does not hold, not a defect. #3990 lets a tick span about 3.25 intervals and relies on the in-flight guard, which its own test pins with `assertEquals(maxConcurrent, 1)`. Escalation is still bounded: roughly five minutes at the 30s default, against about 90s here. Slower, not broken.

**#3990's tests on this PR's code:** 2 of 3 fail, and both are real.

1. `retries a heartbeat whose response body read fails after the headers arrive`: **fails here, 1 request instead of 3.** #3990 also moved `readAgentPushRuntimeServiceResponse` inside the transport-error `try` and rethrows an already-classified `VeryfrontError` untouched. Without that, a body read that fails after the headers land surfaces as a raw `AbortError`, misses the `NETWORK_ERROR` mapping, and `isRetryableHeartbeatFailure` treats it as permanent, so the tick stops after one attempt instead of using its remaining two. Adding a deadline makes that path *more* likely, not less: the deadline can fire while the JSON is still arriving. This PR adds the deadline and not the fix.

2. `leaves healthy and intermittently slow heartbeats alone`: **fails here.** At a 500ms interval a 200ms answer gets cut off, because the deadline here is 109ms. #3990 allows the full 500ms. This is the exact false positive this PR's own body argues against, at short intervals: a heartbeat answering in 40% of its interval is reported as a failure.

So the tighter deadline buys escalation in 90s instead of 5 minutes, and pays for it with a false positive at short intervals plus a missing fix. That trade is not worth taking, and #3990 is already queued and verified.

## What survives

One test here covers something #3990's does not, so it is carried over in #4007 rather than dying with this PR: `lets a slow-but-successful heartbeat finish at the production interval`. #3990's guard runs at a 500ms interval with a 200ms answer, 40% of the interval, which catches a deadline expressed as a fraction of the interval but not one pinned to a fixed number of milliseconds. Replace #3990's deadline with a fixed `timeoutMs: 1_000` and its guard still passes while this test fails on `The operation was aborted`. Complementary, and proven so.

The other two do not survive. The hang test is a weaker copy of #3990's, which uses a 1.5s budget instead of 5s and additionally asserts the in-flight guard holds. The schedule arithmetic test asserts a property #3990 does not have.

## Not carried over, worth a note

`onRetry` here logs `timedOut: isTimeout`, which #3990 does not. That is a small observability gain and not worth touching a queued branch for. Also unchanged: the finding in this body that `src/sandbox/lazy-sandbox.ts` is a different gap (it already has a 15s deadline via `fetchControl`, what it lacks is retry). That is still true and still deserves its own issue.

PR #4012, opened 2026-08-23
release: prepare v0.1.1252 RC for staging verification

kwakayama:
## Summary

- open the `0.1.1252-rc` prerelease line so `main` publishes numbered RCs again
- keep the runtime version constant and the generated hydration bundle in sync

## Why this is needed now

`deno.json` reads a clean `0.1.1251`, and `0.1.1251` is **already published to npm**
(`npm view veryfront@0.1.1251 gitHead` → `0f38b23eb7`). The release job reads `deno.json`, sees no
version change, and correctly no-ops.

`src/utils/version.test.ts` asserts `VERSION === deno.json.version` but **not** that it increased,
so CI stays fully green the whole time. The symptom is silent: merged, green, and shipped nowhere.

**21 commits currently sit above the published version and have reached no environment**, including:

- `3bb52bbc88` fix(provider): retry stream header timeouts before output (#3993) — the #710 fix
- `223c2a9faf` fix(sandbox): encode worker bytes on supported Node (#3987) — issue-inbox#732
- `7f23c8f645` fix(workflow): fail closed when a child graph stalls (#3994)
- `45d277d149` fix(workflow): charge crash recovery when a node starts, not when it queues (#3981)
- `9497625e44` fix(agent): time out hung service heartbeats (#3990)
- `c43cc407c0` fix(renderer): publish cycle artifacts atomically (#3996)
- `f8dbecb0c1` / `7eee200268` both `fix(init)` fixes
- `9d6cdefb86` test: gate the critical app path across Deno, Node, and Bun (#3977)
- `1f470fab50` test(node): run portable API loader coverage (#3989)

## Verification

- `deno task generate:manifests:check`
- `deno task fmt:check`
- `deno test -A src/config/cicd-stable-release.test.ts src/utils/version.test.ts`
- the regenerated hydration bundle differs in exactly one segment — its embedded version string
  (`var VERSION = "0.1.1251"` → `"0.1.1252-rc"`); diff is 3 files, 1 line each

## After this

RC publishes → server / job-runner / sandbox staging deploys → strict staging Remote E2E Health →
stable `0.1.1252` → production promotion. Production currently runs `0.1.1248`.


kwakayama:
Reviewed. The bump itself is clean. The three red jobs are not this PR's fault, but they are worth naming because they will now be red on every release PR.

## The bump is complete and matches the documented procedure

`scripts/release.ts` bumps `deno.json`, then rewrites `cli/commands/init/config-generator.ts`, `src/core/utils/constants/cdn.ts`, `src/utils/version.ts` and `src/utils/version-constant.ts`, then runs `deno task generate` for `hydration-runtime.generated.ts` (the reason is spelled out at `scripts/release.ts:267`). Of those, only `version-constant.ts` holds a literal: `config-generator.ts` and `version.ts` derive from `VERSION`, and `cdn.ts` does not exist. So three files is the whole set, and `git grep 0.1.1251` over tracked files returns nothing. The file list is identical to #3965 and #3933.

Nothing rides along: the diff is 3 files, 1 line each, all of it the version string.

## Why the runtime critical flow jobs fail

`tests (runtime critical flow: deno | node | bun)` fail with:

```
npm error code ETARGET
npm error notarget No matching version found for @veryfront/ext-bundler-esbuild@0.1.1252-rc.
```

The gate packs the local CLI at the new version, runs `veryfront init`, and installs the result. `cli/commands/init/config-generator.ts:69` pins every first-party extension and `veryfront` itself to `` `^${VERSION}` ``, so the scaffold asks npm for `0.1.1252-rc` of packages that only exist after the release job publishes them. That job runs after this merges.

So the gate is structurally unsatisfiable on any version-bump PR. It is green on `main` only because `main` sits at `0.1.1251`, which is published.

This is new, not a regression here. The gate arrived with `9d6cdefb86` (#3977) on 2026-08-22, after the last RC. #3965's check list has no such job, which is why this is the first release PR to show it.

None of the three is a required check (`ci (format)`, `ci (lint)`, `ci (typecheck)`, `tests (unit)`, `tests (integration)`, `coverage gate`, `tests (rsc browser e2e)`, `tests (binary e2e)`, `Analyze` are), and all nine required checks pass here. `mergeStateStatus` is `UNSTABLE` rather than `BLOCKED` for exactly that reason. So this can land.

The part worth fixing separately: left alone, three jobs go red on every RC from now on, and a gate that is always red on a known class of PR stops being read on the ones where it matters. The fix belongs in the flow script, resolving first-party extensions from the packed local tarballs rather than from npm, or skipping the install leg when the version under test is unpublished. Worth its own issue rather than holding this RC.

kojiwakayama:
@codex review

PR #4015, opened 2026-08-23
ci(coverage): compare project coverage against the base commit

kojiwakayama:
Applies two recipes from [Codecov's common recipe list](https://docs.codecov.com/docs/common-recipe-list): *set project coverage checks on a pull request* and *show project coverage changes on the pull request comment*.

Validated against Codecov's `/validate` endpoint: `Valid!`, with `target: "auto"`, `base: "auto"` and `hide_project_coverage: false` all accepted and echoed back in the parsed config.

## The substantive change

```diff
 project:
   default:
-    target: 80%
+    target: auto
+    base: auto
     threshold: 1%
```

`target: 80%` duplicated a check we already run. `deno task coverage:gate` enforces the absolute 80% floor and blocks the merge; a second copy in Codecov added no signal.

`auto` compares against the **base commit** instead, which catches what an absolute floor structurally cannot: a PR taking coverage from 84% to 82% clears the 80% gate but is still a real regression. The two checks are now complementary rather than redundant, absolute floor in CI and regression detection in Codecov.

## Comment changes

```diff
+  require_base: false
+  require_head: true
+  hide_project_coverage: false
```

- `hide_project_coverage: false`: shows overall coverage direction beside patch coverage; otherwise the comment reports only the diff and gives no sense of trend.
- `require_base: false`: keeps comments posting on the first PRs after this lands, before `main` has a baseline report.
- `require_head: true`: never render a comment without head data, which would show a misleading empty diff.

## This still cannot block anything

Both statuses stay `informational: true`, so Codecov always reports success regardless of the comparison. Neither `codecov/project` nor `codecov/patch` is in branch protection. This PR sharpens what the Codecov comment *tells* a reviewer; it does not add a gate. Making it a gate means removing `informational` and adding the context to branch protection, which is a separate decision.

`threshold: 1%` is worth knowing when reading the first comparisons: #4016 moves project coverage by 0.64 points, which is inside the threshold and would not be flagged.

## Base, and why green here means nothing

Stacked on `ci/codecov-integration` (#4013), which introduces `codecov.yml`. Retargets to `main` automatically once that merges.

`.github/workflows/cicd.yml` declares `on: pull_request: branches: [main]`, and that filter matches the **base** branch. A PR based on a feature branch therefore triggers no CI/CD run at all: this PR gets none of the nine required contexts and shows only `cla` and `publish automated review status`. Tracked as veryfront-issue-inbox#766. The gates for this branch were run locally instead.

Merged in `ci/codecov-integration` to pick up the em dash fix and the `extensions/**` ignore removal from #4013, so the two branches do not diverge on `codecov.yml`.

## Summary by CodeRabbit

* **Chores**
  * Updated coverage reporting to compare project coverage against the base commit automatically.
  * Retained the minimum 80% coverage requirement through the existing coverage gate.
  * Improved coverage comments to include project-level coverage changes and patch coverage.
  * Ensured comments appear when head coverage is available, even if base coverage data is missing.


kojiwakayama:
@codex review

kojiwakayama:
@codex review

kojiwakayama:
Rebased onto `main` in `4504ffb33` after #4013 landed, which created an add/add conflict on `codecov.yml`.

Resolved by taking **main's version as the base** and reapplying only this branch's two changes on top. Specifically preserved from #4013:

- `after_n_builds: 1` in both `codecov.notify` and `comment` — the `codecov upload` job now sends all 8 shard files in one call, so the old value of 8 would have stalled reporting forever.
- The note against adding `extensions/**` to `ignore`, since `--include=src/` also selects `extensions/*/src/**` and ignoring them would make Codecov report materially higher than the gate.

Reapplied from this branch: `target: auto` / `base: auto` on the project status, and `require_base` / `require_head` / `hide_project_coverage` on the comment.

Diff against `main` is now `codecov.yml` +14/-1. Revalidated against Codecov's `/validate` — `Valid!`.

kojiwakayama:
@codex review

kojiwakayama:
Standards follow-up at exact head `acfc06d8673e`: replaced the inherited concrete runner filesystem path in the public Codecov workflow comment with `$GITHUB_WORKSPACE`, preserving the normalization explanation while satisfying the repository internal-detail policy. `git diff --check`, `deno fmt --check .github/workflows/cicd.yml codecov.yml`, and `VF_DISABLE_LRU_INTERVAL=1 deno task test:file scripts/ci/setup-deno-workflow.test.ts` all pass (2 tests, 18 steps).\n\n@codex review\n\nPlease review exact head `acfc06d8673e` for remaining standards or correctness issues.

kojiwakayama:
@codex review

Please review exact head `1f41ca51c483cf1faf70c2d74c47ebf3f378181b`. The only new behavior beyond the reviewed coverage-status change is the temporary merge of #4017 so release-version runtime checks resolve local co-published extension tarballs across Deno, Node, and Bun.

kojiwakayama:
@codex review

Please review exact head `d7b582343052534bb7505814d0c6e22152b23e8e`. It preserves the reviewed Codecov configuration, reconciles current main through the canonical #4017 conflict resolution, and includes the #4017 standards-only copy cleanup.

PR #4026, opened 2026-08-23
fix(cli): accept --project-dir in schedule and webhook, retire their cwd lane

kojiwakayama:
## Why

`schedule`, `webhook`, and the TUI's project creation read the process working directory with no way to pass a directory in. Their tests therefore had to `Deno.chdir` through `withCwd`, which kept four files in the serial `unit:cwd` test lane and a cross-isolate disk lock in play for every unit-suite run.

## What

- `schedule` and `webhook` accept the project directory argument the rest of the CLI already has (`CommonArgs.projectDir`: `--project-dir`, `--dir`, `-d`), resolving `opts.projectDir ?? Deno.cwd()`. Help text updated.
- TUI `createProject` accepts an optional `baseDir` on its context; `remoteProjectPath` / `setRemoteProjects` take an optional `baseDir`. All default to the working directory — no caller changes behaviour.
- The four tests pass the directory instead of entering it, drop `withCwd`, and return to `unit:parallel`. `UNIT_CWD_FILES` in `scripts/test/run-suite.ts` shrinks to the three files that are genuinely about cwd itself (`skills/validate` tests a literal `"."`, `platform/compat/process` and `testing/cwd` test the primitives).

## Notes

- Semantic unit-boundary audit: unchanged (2141/762) — these entries never carried `shared-cwd`.
- Follow-up candidates (out of scope): `cli/commands/build/embedded-preset-flags.test.ts` and `cli/auth/login.test.ts` still use `withCwd` from the parallel lane (safe under its disk lock, but the same parameter treatment applies).

## Verification

- The four migrated files + `src/testing/cwd.test.ts` + run-suite/suites pin tests: pass.
- `deno task test:unit:cwd` (now 3 files): pass. The four files run under `--parallel --trace-leaks` alongside five unrelated cli tests: 14 passed (277 steps), no leaks.
- `deno task lint:test-semantic-dispositions`, `deno task test:layout`, `deno task lint:cwd-relative-test-reads`: ok.
- Pre-push gate (fmt, lint, check, `test:unit`): passed on push.

### DIFF
--- a/cli/app/operations/project-creation.test.ts
+++ b/cli/app/operations/project-creation.test.ts
@@ -3,7 +3,6 @@ import "#veryfront/schemas/_test-setup.ts";
 import { assertEquals } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import { _resetEnvironmentConfig } from "#veryfront/config/environment-config.ts";
-import { withCwd } from "#veryfront/testing/cwd.ts";
 import { join } from "veryfront/platform/path";
 import { createProject } from "./project-creation.ts";
 import { createInitialState } from "../state.ts";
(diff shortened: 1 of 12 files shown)


kojiwakayama:
@codex review

PR #4048, opened 2026-08-23
test(suites): make each suite one record that owns its env and flags

kojiwakayama:
## Why

The suite registry centralised which files each suite runs, but not how they run. The six-variable env prefix was retyped on 10 `deno.json` lines (plus two script-local copies), and `run-deno-suite` hand-wrote four flag branches that had drifted: `integration:cli` ran with no `--preload`, no `--deny-net`, no `--trace-leaks`; `integration:legacy-tests-root` missed deny-net and trace-leaks; `coverage:unit` missed trace-leaks; the e2e/binary tasks ran `--allow-all` with neither the egress block nor the preload, so those guarantees silently did not hold there.

## What

- **`DenoSuiteProfile`**: each Deno suite is one record — `{env, preload, denyNet (+ mandatory opt-out reason), traceLeaks, parallel, heap, coverage, extraFlags}` — rendered by `run-deno-suite`; `DENO_TEST_ENV` and the provider deny-net list are single exported constants that `run-test-file.ts` and `coverage-ci.ts` now import.
- **Unit lanes render byte-identically**; the drifted lanes gain exactly the missing flags, each an explicit intent decision. No lane opts out of deny-net.
- **Converted to records**: `test:unit:*`, `test:integration`, `test:integration:cli`, `test:coverage:unit`, new `coverage:integration`, `test:e2e:rsc-browser`, `test:e2e:binary(:fresh)`. Left legacy with reasons: `test`, `test:coverage` (superset unions), and the deliberately-live `test:record` / `test:tool-search-live` / `test:local-inference`. `DENO_TESTING=1` in `deno.json`: 10 → 3.
- **Deleted three orphan runners** (`tests/run-affected-tests.mjs`, `tests/run-concurrent-tests.mjs`, `tests/load-test-isolation.ts`) — zero callers; one carried a stale Deno-only list contradicting `tests/deno-only-tests.mjs`, which exists precisely so the two cannot drift.
- **New invariants in `run-suite.test.ts`**: every profile declares its booleans explicitly; rendered args must contain deny-net unless the record states an opt-out reason; `coverage:unit` must agree with `unit:parallel` on env+preload+denyNet, so CI and local cannot diverge silently; converted task lines must not carry the env prefix.

## Verification

Old→new rendered args diffed per suite: unit lanes byte-identical; every other diff is one of the intended flag additions. `test:unit:parallel` full run: 4,181 passed. `test:unit:cwd` 112 steps; `test:scripts` green except the two known fresh-worktree environmental failures. 4 integration files under the new integration profile and one rsc-browser file under the new e2e profile (with Chromium): pass. The 3 pre-existing `test:integration:cli` failures reproduce identically with the old flag shape. `cicd.yml` unchanged and verified. Rebased onto main after #4030; suite pin tests green post-rebase. Pre-push gate passed on push.


kwakayama (commented):
### Finding

- Low — invalid suite IDs can bypass validation. [`scripts/test/run-deno-suite.ts:192`](https://github.com/veryfront/veryfront-code/blob/f5ab8e66bd38fb98afa36ea28801b06366891bdc/scripts/test/run-deno-suite.ts#L192) uses `in`, which accepts inherited keys such as `toString` and `constructor`. Those values then reach the planner at [line 198](https://github.com/veryfront/veryfront-code/blob/f5ab8e66bd38fb98afa36ea28801b06366891bdc/scripts/test/run-deno-suite.ts#L198), but no matching branch exists in [`selectProfileFiles`](https://github.com/veryfront/veryfront-code/blob/f5ab8e66bd38fb98afa36ea28801b06366891bdc/scripts/test/run-suite.ts#L255-L294), causing an unhandled failure rather than the documented usage error with exit code 2. Use `Object.hasOwn(DENO_SUITE_PROFILES, flags.suite)` and add regression cases for inherited property names.

Score breakdown: correctness 38/40, tests 18/20, reliability/security 15/15, maintainability 15/15, scope/docs 10/10.

Review-Gate:
Reviewer: Codex
Reviewed-SHA: f5ab8e66bd38fb98afa36ea28801b06366891bdc
Score: 96/100
Actionable-Findings: 1
Verdict: COMMENT

kojiwakayama:
Gate refresh: re-evaluating the automated-review status under the updated workflow.

kwakayama (dismissed):
> *This was generated by AI during triage.*\n\n[P2] Suite validation uses suite in DENO_SUITE_PROFILES, so inherited names such as toString are accepted and fail as an internal planner error instead of documented usage. Use an own-property check and add the inherited-key regression.

kwakayama (approved):
> *This was generated by AI during triage.*\n\nRe-reviewed the current pushed head after dismissing superseded change requests. The addressed threads are resolved and the focused checks pass. Approved pending required checks.

Issue #4053, opened 2026-08-24
MDX bundler emits a module that is not valid JavaScript


kojiwakayama:
## What

`bundleMDX` and `bundleMDXWithOptions` in `src/build/renderer/services/mdx-bundler.ts` both request `outputFormat: "function-body"` from the content processor (lines 115 and 192), then splice that function body directly into module top level. A function body is not a module body, so the emitted text cannot parse or run.

## Defects in the emitted module

1. **`MDXContentWrapper` is never defined.** Both templates end with:

   ```ts
   export default function MDXContent(props) {
     const components = useMDXComponents();
     return MDXContentWrapper({ ...props, components });
   }
   ```

   `grep -rn "MDXContentWrapper" src/` returns only these two call sites and one test asserting the string. Nothing defines it.

2. **`MDXContent` is declared twice.** The spliced `compiledStr` already declares `function MDXContent(props)`, and the wrapper template declares `export default function MDXContent` alongside it.

3. **Top-level `return`.** The function-body output ends with `return { headings, default: MDXContent };`, which is a syntax error outside a function.

4. **`arguments[0]` at module scope.** The body opens with `const {Fragment: _Fragment, jsx: _jsx, jsxs: _jsxs} = arguments[0];`. There is no `arguments` in a module.

5. **Mismatched quotes** in the import line (fixed in the pending test-audit PR, listed here for completeness):

   ```ts
   import { useMDXComponents } from "#veryfront/mdx-components';
   ```

## Reproduction

Logging `result.code` from `bundleMDXWithOptions` for simple MDX (the `ext-content-mdx` ContentProcessor is registered by the test setup, so this is the production compile path) yields:

```js
import * as React from "react";
import { useMDXComponents } from "#veryfront/mdx-components";

"use strict";
const {Fragment: _Fragment, jsx: _jsx, jsxs: _jsxs} = arguments[0];
...
function MDXContent(props) { ... }
return { headings, default: MDXContent };

export default function MDXContent(props) {
  const components = useMDXComponents();
  return MDXContentWrapper({ ...props, components });
}
```

## Impact

`bundleMDXWithOptions` has no in-repo caller. It is re-exported from `src/build/renderer/index.ts` as public API, so the blast radius is external consumers rather than the framework's own MDX path.

## Why no test caught it

The existing tests assert on substrings of the emitted **text** rather than parsing or executing it. One of them asserts the output contains:

```ts
"return MDXContentWrapper({ ...props, components });"
```

[...]

Issue #4052, opened 2026-08-24
Chunk manifest keys routes by source filename, so App Router routes get no preload hints


kojiwakayama:
## What

`buildManifest` publishes every route in the chunk manifest under a path derived from the entry point's **source filename**, while the lookup table it consults is keyed by **entry name**. Whenever those differ, the lookup misses and the route is recorded under a fabricated path.

`src/build/bundler/code-splitter/entry-points.ts:20-22` builds the map keyed by entry name:

```ts
const entryName = route.name ?? convertPathToName(route.path);
routeMap.set(entryName, route.path);
```

`src/build/bundler/code-splitter/manifest-builder.ts:92-93` looks it up by source basename:

```ts
const entryName = extractEntryName(output.entryPoint); // basename of the source file
const routePath = routeMap.get(entryName) ?? `/${entryName}`;
```

`extractEntryName` is `entryPoint.split("/").pop()`, so it discards the directory entirely.

## Impact

Two classes of route are affected, and between them that is most real apps:

1. **Every App Router page.** For `{ path: "/", file: "app/page.tsx", name: "index" }`, `routeMap` holds `"index" -> "/"`, but `extractEntryName` yields `"page"`. The lookup misses and the manifest key becomes `/page`.
2. **Every nested route.** For `{ path: "/blog/post", slug: "blog/post" }`, `runCodeSplitting` derives entry name `blog-post`, while `extractEntryName` yields `post`, so the manifest key becomes `/post`.

At runtime `generatePreloadLinks(manifest, "/blog/post")` finds no matching key and returns `""`, so these routes ship with **no modulepreload and no stylesheet preload hints**. The chunk artifacts themselves are written correctly (`<outputDir>/_veryfront/chunks/blog-post.js` exists), so nothing errors. The manifest is simply wrong and the performance benefit is silently lost.

## Evidence

Against a real esbuild split (not a stub), with routes `[{ path: "/blog/post", file: <tmp>/app/blog/post.tsx, slug: "blog/post" }, { path: "/about", file: <tmp>/app/about.tsx, slug: "about" }]`:

```
assertEquals(Object.keys(result.manifest.routes).sort(), ["/about", "/blog/post"])
  actual: ["/about", "/post"]
```

And for a route declared as `{ path: "/", file: <projectDir>/app/page.tsx, name: "index" }`, the sole manifest key is `"/page"`.

## Why no test caught it

[...]

PR #4054, opened 2026-08-24
test(build): make weak assertions bite across the build pipeline

kojiwakayama:
## Summary

Audit of all 65 test files under `src/build` for tests that stay green when the behavior they claim to cover breaks.

72 candidate findings, 66 confirmed after adversarial verification, 66 applied, 6 refuted.

## Recurring weaknesses fixed

**Stub-shaped oracles that never inspected the request.** The CSS minification strategies asserted only that *some* string came back, so the engine could be asked to minify the wrong file, or not to minify at all, and the test passed. They now capture the request and assert `minify`, `sourcePath` and `sourceMap`.

**Bare `assertRejects` with no error type or message**, satisfied by any failure including a typo in the test itself. Now pinned to the message and the error name.

**Substring checks too generic to fail.** The import map asserted only that the text contained `"react"`; it now compares against `getReactImportMap`. The hydration test asserted a fragment that matched whether the attribute was set to `'true'` or `'false'`; it now pins the emitted statement.

**Production-only behavior never contrasted with development.** The bundler `minify` and `format` cases now bundle both ways and assert the difference, so hardcoding either mode fails.

**Untested consequential branches:** the build output lock ownership check, post-publish cleanup leaving no stage or backup artifact, the symlinked CSS source directory guard, and the control-character range boundaries.

## Source change

`renderer/services/mdx-bundler.ts` had mismatched quotes in the emitted import:

```ts
import { useMDXComponents } from "#veryfront/mdx-components';
```

That is a syntax error in the generated module. Fixed as an unambiguous one-character defect.

> This does **not** make the function work. The module it emits has further structural defects (`arguments[0]` at module scope, a top-level `return`, a duplicate `MDXContent`, and a call to an undefined `MDXContentWrapper`), tracked in #4053.

## Two product bugs found and filed, not fixed here

- **#4052** — the chunk manifest keys routes by source filename while the lookup map is keyed by entry name, so every App Router `page.tsx` and every nested route is published under a fabricated path and ships with no preload hints. Found by an assertion that failed against a real esbuild split; the assertion was dropped and the source left alone because the correct fix is genuinely ambiguous.
- **#4053** — the MDX bundler emits a module that is not valid JavaScript.

## Test placement

Four files became effect-bearing units and tripped `lint:test-semantic-dispositions`. Per the repo rule, **none was added to the migration list**:

- `batch-processor` was made **hermetic** by installing an in-memory adapter through the existing `AdapterRegistry` seam that `SecureFs` already resolves through, so the coverage stays next to the code.
- `compiler` and `code-splitter-orchestrator` have no such seam (`writeCompiledFile` captures a module-level `createFileSystem()` with no override), so only the offending cases **moved** to `tests/integration`, keeping every assertion verbatim. The hermetic assertions stay in the unit files with a pointer comment.

## Two baselines tightened, because the linters require it

Both of these fail CI if left stale, so they are not optional cleanups:

- `manifest-builder.test.ts` now typechecks cleanly and had to be removed from the typecheck grandfather list. Verified by restoring the entry and confirming `lint:test-typecheck` fails with "Baseline test files now typecheck cleanly".
- Unit-scope sanitizer opt-outs drop as the relocated files leave `src`. `main` independently tightened this to 396 while this branch was in flight; after rebase the measured value is **392**, produced by `lint:sanitizer-baseline:update` rather than hand-edited. Note these opt-outs **moved with their tests** rather than being eliminated.

## Verification

- Semantic unit-boundary audit: ok (2144 considered, 757 disposed)
- `test:layout`: ok (2364 executable, 2313 canonical)
- **51/51** changed test files pass via `deno task test:file`
- `anti-slop`, `cwd-relative-test-reads`, `test-typecheck`, `sanitizer-baseline`, `skipped-tests`, `ban-test-only`: all pass

No test was deleted, skipped, or weakened. No sanitizer opt-out was added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected MDX bundle output for component imports.
  * Improved validation and safety checks for CSS, image, and asset outputs, including path traversal, symlinks, collisions, and oversized manifests.
  * Ensured build cleanup, publishing, caching, and dry-run behavior work as expected.
* **Tests**
  * Expanded coverage for MDX compilation, Tailwind processing, code splitting, routing, minification, source maps, manifests, and production builds.
  * Strengthened assertions for generated output, paths, errors, and feature settings.


kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n## Review findings\n\n[MEDIUM] build-pipeline tests still use direct Deno.* and are omitted from Node/Bun. Use runtime-neutral helpers or explicit Deno-only classification. [LOW] strengthen the dotfile, byte-total, and exclude assertions from the review.

kwakayama (changes requested):

## Review findings

[MEDIUM] build-pipeline tests still use direct `Deno.*` and are omitted from Node/Bun. Use runtime-neutral helpers or explicit Deno-only classification. [LOW] strengthen the dotfile, byte-total, and exclude assertions from the review.

kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n[P2] New build assertions use direct Deno APIs and are omitted from Node/Bun runs.\n\n[P2] The dotfile assertion only matches a literal /.js and misses names such as chunks/.route.js.\n\n[P2] totalSize > 0 and the include/exclude fixture do not prove both contributors or exclusion forwarding; assert exact aggregate bytes and include an explicitly excluded route.

kwakayama (dismissed):
> *This was generated by AI during triage.*\n\nRe-reviewed the pushed head and addressed the verified review findings. Focused tests and typechecks pass per the implementation report. Approved pending required repository checks.

PR #4055, opened 2026-08-24
test(security): make weak assertions bite across the security surface

kojiwakayama:
## Summary

Audit of all 62 test files under `src/security` for tests that stay green when the behavior they claim to cover breaks.

53 candidate findings, 48 confirmed after adversarial verification, 50 applied, 5 refuted.

**No product defects were found.** Every mutation check showed the implementation already correct and only the test blind to it. That is the useful result for a module like this one: the guards work, but a regression in several of them would not have been caught. **Test files only, no source changes.**

## Guards that were implemented but unasserted

**Path prefix confusion.** Nothing pinned that `/project-secrets/creds.env` is outside `/project`. Replacing the boundary check with a raw `startsWith` now fails.

**CORS missing-origin denial.** A null request origin was never exercised, so replacing the branch with an unconditional `corsResult("*", false)` passed. Now denied under both allowlist and string-origin policies.

**Egress broker token gate.** Missing and wrong tokens now assert 403, the egress error header, a body containing "authentication failed", and that the upstream received **zero hits**, so deleting the constant-time token comparison fails.

**Worker diagnostic redaction and bounds.** A credentialed postgres URL plus overlong filler through `message`, `detail`, `cause` and `instance` now asserts each field is redacted, does not contain the secret, and is length bounded.

**Worker status fail-closed.** Out-of-range, non-integer and string statuses (`200`, `999`, `429.5`, `"429"`, `NaN`, `MAX_SAFE_INTEGER + 2`) now assert a plain `Error`, with a positive companion confirming `429` still yields a `VeryfrontError`.

**Restricted workflow permissions** are pinned exhaustively, so appending `--allow-all` fails.

**Base64 chunk boundary.** The old case compared the encoder against itself, which passes for any self-consistent encoder. It now decodes independently with `atob()` and compares every byte.

**Form validation and per-file upload limits**, neither previously reached.

**Bootstrap import order** is asserted positionally, so moving the host env preload after project modules fails.

## Test placement

Four files became effect-bearing units and tripped `lint:test-semantic-dispositions`. Per the repo rule, **none was added to the migration list**.

Each resolves a dev-mode or proxy-topology decision through a module-private host read with no injectable seam — `isDevMode()` reads `globalThis.__VERYFRONT_DEV__` and `Deno.env` inline; `isProxyTopologyTrusted()` reads `getHostEnv` with no override. Adding a production API purely to satisfy the ratchet would be the wrong trade, so only the offending cases **moved** to `tests/integration`, keeping every assertion and message verbatim.

Both arms of each branch remain covered: the hermetic arm stays in the unit file with a pointer comment. The relocated `local-control-request` case is now **stronger** than before, pinning both arms explicitly instead of leaving the untrusted arm dependent on ambient environment.

## Verification

- Semantic unit-boundary audit: ok (2144 considered, 757 disposed)
- `test:layout`: ok
- **41/41** changed test files pass via `deno task test:file`
- `anti-slop`, `cwd-relative-test-reads`, `test-typecheck`, `sanitizer-baseline`, `skipped-tests`, `ban-test-only`: all pass

No test was deleted, skipped, or weakened. No sanitizer opt-out was added.

### DIFF
--- a/scripts/test/run-suite.test.ts
+++ b/scripts/test/run-suite.test.ts
@@ -84,6 +84,18 @@ describe("suite planning parity", () => {
     );
   });
 
+  it("runs the sandbox runtime guard in the Node and Bun suites", async () => {
+    const guardFile = "tests/integration/security/sandbox-runtime-guard.test.ts";
+
+    for (const suite of ["runtime:node", "runtime:bun"] as const) {
+      const plan = await planSuiteFiles({ suite });
+      assert(
+        plan.files.includes(guardFile),
+        `${suite} must select the sandbox runtime guard coverage`,
+      );
+    }
+  });
+
   it("keeps eight coverage shards complete, disjoint, and ordered", async () => {
     const paths = Array.from(
       { length: 27 },
@@ -402,12 +414,14 @@ async function legacyRuntimeFiles(runtime: "node" | "bun"): Promise<string[]> {
       "extensions/ext-bundler-esbuild/src/binary.test.ts",
       "tests/ensure-npm-links.test.mjs",
       "tests/test-file-utils.test.mjs",
+      "tests/integration/security/sandbox-runtime-guard.test.ts",
     ]
     : [
       "src/",
       "tests/bun/dynamic-alias-resolution.test.ts",
       "tests/bun/npm-protocol-resolution.test.ts",
       "tests/bun/workspace-resolution.test.ts",
+      "tests/integration/security/sandbox-runtime-guard.test.ts",
     ];
   const incompatible = runtime === "node"
     ? [
--- a/scripts/test/run-suite.ts
+++ b/scripts/test/run-suite.ts
@@ -84,12 +84,14 @@ const RUNTIME_PATTERNS = {
     "extensions/ext-bundler-esbuild/src/binary.test.ts",
     "tests/ensure-npm-links.test.mjs",
     "tests/test-file-utils.test.mjs",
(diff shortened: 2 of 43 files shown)


kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n## Review findings\n\n[HIGH] : Node/Bun guard tests are skipped by the Deno integration runner; move them into runtime-selected files or add them to Node/Bun suite plans. [MEDIUM] the dev-mode sanitizer test passes , so it does not verify the warning branch.

kwakayama (changes requested):

## Review findings

[HIGH] `tests/integration/security/sandbox-runtime-guard.test.ts:31`: Node/Bun guard tests are skipped by the Deno integration runner; move them into runtime-selected files or add them to Node/Bun suite plans. [MEDIUM] the dev-mode sanitizer test passes `{ warn: false }`, so it does not verify the warning branch.

kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n[P2] The runtime guard cases are under tests/integration but are not selected by the Node/Bun plans, so the regression can silently be Deno-only.\n\n[P2] The test named warns passes warn:false, so it never verifies the development warning. Capture/assert the warning with warning enabled.

kwakayama (approved):
> *This was generated by AI during triage.*\n\nRe-reviewed the pushed head and addressed the verified review findings. Focused tests and typechecks pass per the implementation report. Approved pending required repository checks.

PR #4063, opened 2026-08-24
test(html): make weak assertions bite and fix two generator defects

kojiwakayama:
## Summary

Audit of all 41 test files under `src/html` for tests that stay green when the behavior they claim to cover breaks.

50 candidate findings, 40 confirmed after adversarial verification, 40 applied, 10 refuted.

## Source change 1: the production hydration script never parsed

`generateProdHydrationScript` emitted this inside a `<script type="module">`:

```js
const root = document.getElementById('root');
if (!root) return;
```

A top-level `return` is an **Illegal return statement** in an ES module. The whole script failed to parse, so none of it ran — not the imports, not the element tree, not `hydrateRoot`. Verified directly rather than assumed:

```
node -e "import('data:text/javascript,'+encodeURIComponent('const r=null; if(!r) return;'))"
MODULE PARSE: SyntaxError Illegal return statement
```

Restructured to `if (root) { ... }`.

**Scope:** `generateProdHydrationScript` has no internal caller and is reached only through the barrel export, so this broke consumers of the helper rather than the framework's own render path. Flagging that explicitly so nobody reads this as a live hydration outage.

## Source change 2: a test that pinned the bug its own name described

`findGlobalStylesheet` matched with unanchored patterns:

```ts
const stylesheetPatterns = [/globals\.css$/, /global\.css$/, ...];
```

So `theme-globals.css` was selected as the project's Tailwind entry stylesheet. The existing test was named:

> `should not match files that end with globals.css but have different prefix`

…and its assertion pinned the **opposite** behavior. This is the highest-value category in the audit: a test whose name states the contract and whose assertion freezes the violation, so the bug reads as intended behavior forever.

Every pattern is now anchored to a path-segment boundary, the assertion flipped to match the name, and a companion case added so the anchoring cannot be over-tightened (`styles/globals.css` and `project/src/globals.css` must still match). Checked that no Windows-style absolute path reaches this matcher: only `.tsx/.jsx/.mdx/.ts/.js` are collected locally, and control-plane file lists are posix-relative.

## Test changes

**Script injection guards were unasserted.** Studio bridge config and `sourceHash` now feed a `</script><script>alert(1)</script>` payload and assert it never appears unescaped, split into two cases so each `.replace(/</g, "\\u003c")` site is isolated. Structured metadata now asserts no inline event handler (`onload`, `onClick`, `ONERROR`) can reach the head.

**A nonce count asserted `>= 3`** against five script tags, so dropping the nonce from two generators passed. It now pins the tag count and asserts the nonce count equals it.

**An assertion-free `clearCSSCache` body** now caches a real entry, reads it back by content hash, and asserts the clear drops it.

**Cache partitioning** now varies `minify`, `buildMode` and `environment` individually, so dropping any one from the profile hash fails.

**A test name that overclaimed was split:** the HMR arm and the Studio embed arm are now separate cases, so neither name covers behavior it does not test.

## Deliberately not applied

One finding required reaching a `MAX_CSS_TOTAL_BYTES` branch. That constant is 64 MiB and not injectable, so the test would have to allocate 64 MiB to bite. Left alone.

## Verification

- **27/27** changed test files pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved HTML and script escaping to prevent unsafe markup injection.
  - Fixed hydration behavior when a page root is unavailable.
  - Improved SPA navigation, stylesheet replacement, hash scrolling, and nested route handling.
  - Corrected studio embed asset loading and development-mode behavior.
  - Prevented invalid or stale CSS and metadata from being served.
  - Improved CSP nonce propagation across generated scripts and styles.

- **Tests**
  - Added broad regression coverage for routing, hydration, HTML safety, caching, CSS generation, metadata, and navigation behavior.

### DIFF
--- a/src/html/dev-scripts.test.ts
+++ b/src/html/dev-scripts.test.ts
@@ -120,7 +120,58 @@ describe("html/dev-scripts", () => {
         pageId: "pg",
         pagePath: "/app/page.tsx",
       });
-      assertEquals(scripts.includes("pagePath"), true);
+      assertEquals(
+        scripts.includes('"pagePath":"/app/page.tsx"'),
+        true,
+        "the supplied pagePath must be the value published to the bridge",
+      );
+    });
+
+    it("should fall back to pageId when pagePath is omitted", () => {
+      const scripts = getStudioScripts({ projectId: "p", pageId: "pg" });
+      assertEquals(
+        scripts.includes('"pagePath":"pg"'),
+        true,
+        "pagePath must fall back to pageId",
+      );
+    });
+
+    it("should escape angle brackets in the inline bridge config", () => {
+      const hostile = "</script><script>alert(1)</script>";
+      const scripts = getStudioScripts({
+        projectId: hostile,
+        pageId: "pg",
+        pagePath: hostile,
+      });
+      assertEquals(
+        scripts.includes("<script>alert(1)</script>"),
+        false,
+        "inline bridge JSON must not close the script element",
+      );
+      assertEquals(
+        scripts.includes("\\u003c/script"),
+        true,
+        "< in bridge config must be escaped as \\u003c",
+      );
+    });
+
+    it("should escape angle brackets in the sourceHash script", () => {
+      const hostile = "</script><script>alert(1)</script>";
+      const scripts = getStudioScripts({
+        projectId: "p",
+        pageId: "pg",
+        sourceHash: hostile,
+      });
+      assertEquals(
+        scripts.includes("<script>alert(1)</script>"),
+        false,
+        "the sourceHash literal must not close the script element",
+      );
+      assertEquals(
+        scripts.includes("\\u003c/script"),
+        true,
+        "< in the sourceHash literal must be escaped as \\u003c",
+      );
     });
   });
 });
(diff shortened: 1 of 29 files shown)


kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n## Review findings\n\n[MEDIUM] : direct Deno.* causes the suite to be omitted from Node/Bun coverage. Use the runtime-neutral helper. [LOW] The frontmatter-title test never sets a competing frontmatter title, so precedence is untested.

kwakayama (changes requested):

## Review findings

[MEDIUM] `src/html/styles-builder/css-pregeneration.test.ts:295,312`: direct `Deno.*` causes the suite to be omitted from Node/Bun coverage. Use the runtime-neutral helper. [LOW] The frontmatter-title test never sets a competing frontmatter title, so precedence is untested.

kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n[P2] The new CSS tests use direct Deno.makeTempDir and can be omitted from Node/Bun coverage. Use the runtime-neutral temp helper.\n\n[P2] The metadata test does not set a distinct frontmatter.title, so it does not prove precedence. Add conflicting values and assert frontmatter wins.

kwakayama (approved):

PR #4064, opened 2026-08-24
test(extensions): make weak assertions bite across the extension contracts

kojiwakayama:
## Summary

Audit of all 40 test files under `src/extensions` for tests that stay green when the behavior they claim to cover breaks.

55 candidate findings, 50 confirmed after adversarial verification, 52 applied, 5 refuted. **No product defects found. Test files only.**

## Recurring weaknesses fixed

**A genuinely tautological assertion.** The Node websocket diagnostic asserted that its message constant contained its package constant — both derived from the same source, so it held for *any* value including a wrong one. It now checks the package name against two independent sources: the `sourceDirectory` in `FIRST_PARTY_DEFERRED_BUILTIN_EXTENSION_POLICIES`, and `getRecommendation(NodeWebSocketServerProviderName)`. Renaming the constant to `@veryfront/ext-ws` now fails.

**Permission mapping was only partly pinned.** `native:ffi` and scoped `fs:write` now assert their exact flags, so widening `native:ffi` to `--allow-all`, or dropping the scope from `fs:write`, fails. Read and write scopes are also asserted to stay in separate flags.

**Registry errors asserted only that something threw.** They now assert the registry identity — slug `missing-extension`, status 500, category `RUNTIME`, and the install recommendation detail — so replacing `MISSING_EXTENSION_ERROR.create` with a plain `Error` fails.

**Contract registry lifecycle guards were unreached.** Publishing over an active generation, publishing while a prior generation still owns published entries, and fail-closed behavior after a failed generation now each assert their specific rejection *and* that the registry contents are unchanged.

**Receiver binding was unasserted.** The websocket capture now records the receiver of every forwarded call and asserts it is strictly the source object, so dropping the `Reflect.apply` or rebinding the receiver to `undefined` fails. This is the same class of defect as the unbound backend methods in #4056.

**Descriptor validation** now covers accessor-backed properties and proxies, and asserts the getters are never invoked — including under `Object.prototype` pollution.

**Listener cleanup is now counted**, so deleting `detachAll()` leaves the removal count wrong and fails.

## Note on one skipped item

One finding asked for a mirrored case in `css-purging-engine.test.ts`, which belonged to a different agent's file set in the same run. It was reported rather than applied, since two agents editing one file is how you get a corrupted merge. Worth picking up separately.

## Verification

- **30/30** changed test files pass

### DIFF
--- a/src/extensions/abort-signal.test.ts
+++ b/src/extensions/abort-signal.test.ts
@@ -4,6 +4,27 @@ import { composeAbortSignals } from "./abort-signal.ts";
 Deno.test("composeAbortSignals propagates the first exact abort reason", () => {
   const first = new AbortController();
   const second = new AbortController();
+  const add = second.signal.addEventListener.bind(second.signal);
+  const remove = second.signal.removeEventListener.bind(second.signal);
+  let attached: EventListenerOrEventListenerObject | undefined;
+  let removals = 0;
+  second.signal.addEventListener = (
+    type: string,
+    listener: EventListenerOrEventListenerObject,
+    options?: AddEventListenerOptions | boolean,
+  ): void => {
+    if (type === "abort") attached = listener;
+    add(type, listener, options);
+  };
+  second.signal.removeEventListener = (
+    type: string,
+    listener: EventListenerOrEventListenerObject,
+    options?: EventListenerOptions | boolean,
+  ): void => {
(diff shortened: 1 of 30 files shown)


kwakayama (approved):

kwakayama (approved):

PR #4065, opened 2026-08-24
test(observability,metrics): make weak assertions bite across telemetry

kojiwakayama:
## Summary

Audit of all 39 test files under `src/observability` and `src/metrics` for tests that stay green when the behavior they claim to cover breaks.

69 candidate findings, 61 confirmed after adversarial verification, 60 applied, 8 refuted. **No product defects found. Test files only.**

## Why this module had the most findings of any batch so far

Telemetry tests are unusually prone to passing without proving anything: the instruments are mocks, so an assertion that stops at "something was recorded" holds no matter *which* instrument moved or *what* value landed in it. Most of the fixes below are that same shape.

**Instrument routing was never pinned.** Every stream lifecycle event and every duration kind now asserts its own counter or histogram incremented **and that the other four stayed at zero**, so pointing `case "first_progress"` at the wrong histogram fails. `recordError` likewise now asserts the application counter moved and the render counter did not.

**Histogram buckets were asserted by summing to a total**, which holds for any distribution that adds up. They now pin the exact `bucketCounts` and `explicitBounds` arrays, so moving a value into the wrong bucket fails.

**Span context propagation asserted only that a result was defined.** It now captures what the provider was handed and asserts identity: the span passed to `setSpan`, the context it was set on, and that the returned context is strictly the one produced. Swapping `resolveSpanContext(span)` for `getActiveContext()` now fails.

**Header extraction asserted non-undefined**, so replacing the carrier with `{}` passed. It now asserts the exact carrier contents and the starting context.

**Caching identity used structural equality**, which cannot distinguish a cached instance from an equal fresh one — the whole point of the cache. Now `assertStrictEquals`, with `assertNotStrictEquals` after a reset so a no-op reset fails.

**Query filters asserted result counts.** An inverted predicate that returns a different single row passed a `length === 1` check. They now assert the returned messages, and the previously untested `since` filter is covered with `FakeTime`.

**Durations were asserted as "a number".** A controlled `performance.now` clock makes them exact, so returning zero fails.

**Exporter resilience was untested.** A non-ok collector response and a rejected request now both assert the flush still resolves and that the *next* flush exports, so removing the try/catch or re-enqueueing on non-ok fails.

## Verification

### DIFF
--- a/src/metrics/index.test.ts
+++ b/src/metrics/index.test.ts
@@ -301,6 +301,91 @@ describe("metrics public SDK", () => {
     );
   });
 
+  it("keeps exporting after a non-ok collector response", async () => {
+    const originalFetch = globalThis.fetch;
+    const requests: Array<{ url: string; init?: RequestInit }> = [];
+    let status = 500;
+
+    Deno.env.set("OTEL_METRICS_ENABLED", "true");
+    Deno.env.set("OTEL_EXPORTER_OTLP_ENDPOINT", "https://collector.example/otlp");
+    Deno.env.set("OTEL_SERVICE_NAME", "veryfront-server");
+
+    globalThis.fetch = ((url: string | URL | Request, init?: RequestInit) => {
+      requests.push({ url: String(url), init });
+      return Promise.resolve(new Response("", { status }));
+    }) as typeof fetch;
+
+    try {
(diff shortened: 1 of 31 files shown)


kwakayama (approved):

kwakayama (approved):

PR #4066, opened 2026-08-24
test(routing): make weak assertions bite and close three containment gaps

kojiwakayama:
## Summary

Audit of all 37 test files under `src/routing` for tests that stay green when the behavior they claim to cover breaks.

64 candidate findings, 60 confirmed after adversarial verification, 64 applied, 4 refuted.

## Source changes

All three containment items govern **project code**, not remote attacker input, so please read them as sandbox hardening rather than externally exploitable holes.

### 1. A security control defeated by its own catch

This block is labelled `VULN-FS-5` and re-canonicalises through `realPathSync` specifically to resist symlinked `node_modules` entries pointing outside the project root:

```js
__vf_assertContained(resolved);
try {
  var real = Deno.realPathSync(resolved);
  __vf_assertContained(real);          // <-- inside the try
  resolved = real;
} catch (_) {
  /* expected: realPathSync fails for non-existent paths — assertContained above already held */
}
```

The containment assertion sits inside the try whose catch is documented as covering only non-existent paths. A symlink escaping the root threw from `__vf_assertContained(real)`, was swallowed, and execution continued with the **un-canonicalised** path. The hardening was inert for precisely the case it was written to stop. The assertion now runs outside the try, so only the `realPathSync` failure is caught.

### 2. Subpath resolution had no containment check

The user-dependency branch concatenated the handler-supplied subpath onto the package directory URL raw, so `pkg/../../../etc/passwd` resolved outside `node_modules`. Now goes through `resolveContainedPackagePath` and leaves the specifier untouched when it escapes — matching the rule `route-adapter.ts` already documents and enforces for the compiled-binary path.

### 3. The remote-import allow-list missed re-exports

`validateHTTPImports` scanned static `import` and dynamic `import()`, but not:

```ts
export { thing } from "https://not-allowed.example/mod.js";
```

which resolves and fetches identically. A re-export could pull a module from a host the allow-list rejects. Added to the scanned set.

### 4. An unreachable route (logic, not security)

`discoverPagesRoutes` computed `routePath.replace(/\/index$/, "") || prefix`. That fallback can only fire when `prefix` is `""`, so it always evaluated to `""`. `parseRoute("")` compiles to `/^$/` and `normalizePathname` never yields `""`, so a root index file registered a route nothing could ever match. Changed to `"/"`, matching `discoverAppRoutes` in the same file.

The existing test had **pinned the unreachable `""` pattern**. It now asserts `"/"` and backs that with an actual `match("/")`, so the pattern is pinned by reachability rather than by string equality.

## Test changes

- **Four test bodies asserted nothing at all**: scroll error handling, the missing-root-element case, observer teardown, and one transition case.
- **Route file matching could be satisfied by last-write-wins ordering.** Entries reordered, plus a sibling case pinning that `routes.ts`, `route-handler.ts` and `route.backup.ts` register nothing.
- **Cache invalidation called `clearCache()` inside the test**, so `addRoute` could stop invalidating and the test still passed. That call is removed, and a new case pins LRU eviction at the cache bound.
- **Path containment** now covers the sibling-prefix case, so downgrading the check to `startsWith` fails.
- **A transition test now records every `innerHTML` write**, so an abandoned destination cannot be painted without failing.

## Reported, not fixed

The same shim rejects *legitimate* dependencies when the project root itself is reached through a symlink, because the pre-canonical check compares a non-canonical path against a `realPath`'d root. Fixing that means changing how the root is canonicalised, which is beyond a test change.

## Verification

- Sanitizer ratchet test: ok, `test:layout`: ok, `docs:api-reference:check`: current

### DIFF
--- a/src/routing/api/api-route-matcher.test.ts
+++ b/src/routing/api/api-route-matcher.test.ts
@@ -206,6 +206,27 @@ describe("ApiRouteMatcher", () => {
       assertExists(match);
       assertEquals(match.route.page, "pages/api/users/index.tsx");
     });
+
+    it("prefers the deeper catch-all when two catch-alls overlap", () => {
+      const router = createRouter();
+      // Register the shallow catch-all first so insertion order cannot mask the
+      // depth tiebreak in the priority comparator.
+      router.addRoute("/api/[...path]", "pages/api/[...path].tsx");
+      router.addRoute("/api/v1/[...rest]", "pages/api/v1/[...rest].tsx");
+
+      const match = router.match("/api/v1/x");
+      assertExists(match);
+      assertEquals(
+        match.route.page,
+        "pages/api/v1/[...rest].tsx",
+        "deeper catch-all must win the depth tiebreak",
+      );
+      assertEquals(
+        match.params.rest,
+        ["x"],
+        "the deeper catch-all consumes only the trailing segment",
+      );
+    });
   });
 
   describe("Edge cases", () => {
--- a/src/routing/api/context-builder.test.ts
+++ b/src/routing/api/context-builder.test.ts
@@ -1,5 +1,10 @@
 import "#veryfront/schemas/_test-setup.ts";
-import { assertEquals, assertExists, assertRejects } from "#veryfront/testing/assert.ts";
+import {
+  assertEquals,
+  assertExists,
+  assertRejects,
+  assertThrows,
+} from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
 import {
   type APIContext,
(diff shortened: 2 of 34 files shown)


kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n## Review findings\n\n[HIGH] : relative  is compared against an absolute resolved target, so valid package imports can remain bare and fail to load. Resolve  absolutely first. [MEDIUM] Preserve / suffixes separately before filesystem resolution; otherwise  encodes them into the filename.

kwakayama (changes requested):

## Review findings

[HIGH] `src/routing/api/module-loader/external-import-rewriter.ts:228`: relative `packageDir` is compared against an absolute resolved target, so valid package imports can remain bare and fail to load. Resolve `packageDir` absolutely first. [MEDIUM] Preserve `?query`/`#hash` suffixes separately before filesystem resolution; otherwise `pathToFileURL` encodes them into the filename.

kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n[P1] Relative packageDir values are compared against absolute resolved targets, so documented relative project directories leave valid dependency imports bare. Resolve packageDir before containment checks.\n\n[P2] Query/hash suffixes are included in filesystem resolution and become encoded filename text. Split the suffix before containment resolution and append it to the file URL.

kwakayama (dismissed):

kwakayama (approved):
> *This was generated by AI during triage.*\n\nRe-reviewed the new pushed head. The follow-up finding is addressed, the relevant review threads are resolved, and focused tests passed. Approved pending required checks.

PR #4070, opened 2026-08-24
test(discovery,middleware,release-assets): make weak assertions bite

kojiwakayama:
## Summary

Audit of all 38 test files under `src/release-assets`, `src/discovery` and `src/middleware` for tests that stay green when the behavior they claim to cover breaks.

57 candidate findings, 49 confirmed after adversarial verification, 49 applied, 8 refuted. **No product defects found. Test files only.**

## Recurring weaknesses fixed

**A test named for behavior it never exercised.** `should pass env and executionCtx to the context` never passed an `executionCtx` at all, so the pipeline could drop it entirely and the test stayed green. It now builds one and asserts the middleware context receives that exact object by identity.

**A self-constructed fixture standing in for the factory under test.** The discovery result shape test built its own object literal and asserted against that, so the real `createEmptyDiscoveryResult` could stop initializing a map and nothing failed. It now calls the factory and asserts the full 11-key set and that every map is empty.

**Transaction rollback was never observed.** No test placed a *successfully registered* primitive in the same rejected generation as a failure, so the rollback could be removed entirely and every test passed. A new case registers a live primitive alongside a failing one and asserts the live one is gone after the rejection.

**Error messages were asserted only for what they contain.** The rejection message now also asserts what it must *not* contain — local paths and upstream failure text — pinning the JSDoc contract about not leaking those.

**Branch fall-through was untested.** Route resolution now uses a nested layout where every pages file also sits under the app root, pinning both the app-root win *and* the fall-through to the pages root.

**Env wiring was pinned per key** by combining a zero rollout percent with an allowlist, so crossing two env vars fails.

**Name generation and provider-safety gating** for colocated tools now assert the generated short name and that an unsafe namespaced name (`researcher--fetch.paper`) is rejected with a specific error and never registered.

## Deliberately not applied

Two findings were dropped rather than forced:

- One needed a new test file outside the owning agent's file set; two agents editing one file is how merges get corrupted.
- The other proposed a mutation that **does not actually stay green** — deleting `registerPrompt` fails an existing test elsewhere — so strengthening against it would have proved nothing.

## Verification

- Sanitizer ratchet test: ok, `test:layout`: ok, `docs:api-reference:check`: current
- **30/30** changed test files pass

### DIFF
--- a/src/discovery/agent-scoped-capabilities.test.ts
+++ b/src/discovery/agent-scoped-capabilities.test.ts
@@ -328,7 +328,7 @@ Deno.test({
       await Deno.mkdir(`${root}/agents/researcher/tools`, { recursive: true });
       await Deno.writeTextFile(
         `${root}/agents/researcher/AGENT.md`,
-        `---\nname: Researcher\ntools: [fetch-paper]\n---\nResearch.\n`,
+        `---\nname: Researcher\ntools: [fetch-paper, summarize]\n---\nResearch.\n`,
       );
       await Deno.writeTextFile(
         `${root}/agents/researcher/tools/fetch-paper.ts`,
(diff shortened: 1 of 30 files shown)


kwakayama (approved):

kwakayama (approved):

PR #4072, opened 2026-08-24
test(provider,llm,embedding,chat): make weak assertions bite

kojiwakayama:
## Summary

Audit of all 44 test files under `src/provider`, `src/llm`, `src/embedding` and `src/chat` for tests that stay green when the behavior they claim to cover breaks.

65 candidate findings, 58 confirmed after adversarial verification, 7 refuted. **No product defects found.**

## Source changes: two testability seams, no behavior change

Flagging these because they touch public surface, which deserves an explicit yes:

- **`embedding/model-resolution.ts`** takes an optional `{ compiled }` override so the compiled-binary cloud fallback can be exercised. `isDenoCompiled` is a module constant with no other way in, and this follows the injectable-reader idiom the repo already uses in `config/environment-config.ts` (`readEnvSnapshot(readEnv = getEnv)`). Additive and optional, so existing callers are unaffected.
- **`chat/upload-handler.ts`** exports `resolveStorage`. The rule it implements — *local dev stays on disk even when the cloud bootstrap is present* — is a real correctness concern and is otherwise only observable by performing an upload. Now pinned by an integration test across three states: dev, deployed with bootstrap, deployed without.

**Two further exports the audit proposed were reverted rather than landed.** `resolveCloudEmbeddingFallback` became redundant once the `compiled` seam existed, so its five tests were rerouted through the public function; I re-ran the mutation check afterwards to confirm they still bite (deleting the compiled cloud-fallback branch still fails them). And a unit assertion that only proved "an explicitly configured storage wins" did not justify widening the API on its own, so it was dropped. Consistent with dropping the profiler export in #4035.

## Test changes

**Credential exfiltration guards were unasserted.** A cloud gateway request aimed at a non-gateway origin now asserts both the rejection *and* that the mock fetch saw **zero requests**, so the bearer token provably never leaves the process.

**Userinfo rejection was covered only for `user:password`.** A username-only base URL is now pinned to its specific message, so narrowing the guard to `if (url.password)` fails.

**Schema validation asserted only that something rejected.** It now pins the `VeryfrontError` and its exact `/city: must be string` path, so a bare `TypeError` fails.

**Hybrid search asserted set membership**, which holds for a dense-only result — the fusion could be skipped entirely. It now asserts the *ordered* result and contrasts it against dense-only, using a lexical-only document that is dense-orthogonal so RRF promotion is the only thing that can produce the expected order.

**Resource limits asserted "shorter than the input".** Truncation markers and container caps are now asserted by value.

**An SSE replay test used a payload that was independently invalid** (it lacked `contentId`), so the cursor guard was never the reason the frame was dropped. The frame is now valid and byte-identical across both cases, leaving the event id as the only difference.

## Verification

- Sanitizer baseline lowered to 394 via `lint:sanitizer-baseline:update`, ratchet test passing


kwakayama (approved):

kwakayama (approved):

PR #4079, opened 2026-08-24
test(sandbox,issues,repositories,types,mdx): make weak assertions bite

kojiwakayama:
## Summary

Audit of all 32 test files across the fifteen smallest `src` modules (trigger, sandbox, runs, types, task, schedule, runtime, resource, repositories, registry, mdx, issues, markdown, knowledge, fs), grouped into one PR so the change is reviewable rather than fifteen near-empty ones.

54 candidate findings, 48 confirmed after adversarial verification, 48 applied, 6 refuted. **Test files only.**

## Recurring weaknesses fixed

**Fail-closed exit handling was never exercised.** A sandbox stream that ends with no exit event, and an exit event carrying no exit code, now both assert an exit code of **1** — on the eager *and* the lazy sandbox — so flipping either default to 0 fails.

**Malformed NDJSON handling was untested.** A stream mixing a valid line, an unparseable line and a truncated fragment now asserts the surrounding output still arrives, so deleting the skip-and-continue guards fails.

**Timeout loops were pinned only by wall clock.** Sandbox readiness now drives the exported waiter with mock timers and asserts the exact poll count and the typed timeout error, so removing the deadline check fails.

**Persistence was asserted in memory only.** Issue metadata is now re-read after the write, so dropping a field from the serializer fails.

**List filters that no test reached:** milestone and assignee.

**Cache accounting was asserted from one tier.** A key seeded directly in the backend makes the third read an unambiguous L3 hit, so summing only `l1Hits` fails.

**A clear-order race**, mirroring the existing prefix-delete race test, now proves the second L1 wipe is required rather than duplicate coverage — verified by confirming the existing race test stays green under the same mutation.

**Provider composition asserted only the inner override.** A sibling probe outside the nested provider now proves the outer scope is not mutated.

## Two defects found and reported, not fixed

Both need a behavior decision beyond a test change, so they are filed rather than patched here:

- **`issues/core.ts`** escapes double quotes when serializing a title but never unescapes on parse, so a quoted title does not survive a write/read cycle. The escaping is asserted; **no round-trip assertion was written**, because that would have pinned the corruption as correct.
- **`sandbox/lazy-sandbox.ts`** parses NDJSON with no `try`/`catch`, so one malformed line aborts the stream and discards already-buffered output. `Sandbox.executeStream` wraps the same two parses and documents the skip, so the two copies of this loop disagree.

## Note on a removed sanitizer opt-out

The audit had added `{ sanitizeOps: false, sanitizeResources: false }` to the `task/discovery` describe. It turned out to be unnecessary — the suite is green with sanitizers on — so it was removed rather than landed. New opt-outs are exactly what `lint:sanitizer-baseline` exists to prevent.

## Verification

- Sanitizer baseline: 396/396, ratchet test passing
- **23/23** changed test files pass

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for issue management, milestone and assignee filtering, resource titles, schedules, triggers, and task discovery.
  * Added validation tests for malformed metadata, identifiers, schemas, paths, limits, and security-sensitive requests.
  * Strengthened sandbox, runtime, knowledge lookup, caching, and repository tests for reliability under errors, cancellation, concurrency, and edge cases.
  * Added coverage for Markdown, MDX, entity detection, and public API behavior.


kwakayama (changes requested):
> *This was generated by AI during triage.*\n\n[P1] Hosted knowledge tests replace globalThis.fetch directly instead of the guarded mock-fetch layer; this can test the wrong transport path. Use the repository mock-fetch helper.\n\n[P2] The new task discovery cases use direct Deno filesystem/error APIs and become Deno-only. Use the existing runtime-neutral helpers.

kwakayama (dismissed):

kwakayama (commented):
## Findings

- **[P2] Round-trip quoted issue titles.** [src/issues/core.test.ts:123](src/issues/core.test.ts:123) only asserts serialized escaping. The parser merely strips outer quotes and never unescapes (`src/issues/core.ts:108`), so a create/get round-trip returns `Fix \"login\" timeout`; later updates compound the escaping. Add a round-trip assertion and fix the parser/serializer contract.

- **[P2] Cover malformed NDJSON for `LazySandbox`.** [src/sandbox/sandbox.test.ts:700](src/sandbox/sandbox.test.ts:700) exercises only eager `Sandbox`, while the new lazy coverage at [src/sandbox/sandbox.test.ts:582](src/sandbox/sandbox.test.ts:582) covers only a missing exit. `LazySandbox.executeStream` still directly calls `JSON.parse` (`src/sandbox/lazy-sandbox.ts:236`, `src/sandbox/lazy-sandbox.ts:242`), unlike eager `Sandbox`; malformed lines therefore abort lazy commands and discard buffered output. Add the lazy equivalent and align handling.

Rubric: correctness 31/40, tests 13/20, reliability/security 11/15, maintainability 14/15, scope/docs 9/10.

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 2972fa76aa1950411e854b8f33bc7ae3f392758d
Score: 78/100
Actionable-Findings: 2
Verdict: REQUEST_CHANGES

PR #4083, opened 2026-08-24
test(mcp,studio,internal-agents,client): make weak assertions bite

kojiwakayama:
## Summary

Audit of all 37 test files under `src/testing`, `src/internal-agents`, `src/studio`, `src/mcp` and `src/client` for tests that stay green when the behavior they claim to cover breaks.

59 candidate findings, 52 confirmed after adversarial verification, 7 refuted.

## Source defect 1: the Studio element selector was inert

`"html"` is in `IGNORED_ELEMENTS`, and the injector increments its ignore depth on **any** ignored open tag. A full document opens with `<html>` and never comes back out, so the depth stayed positive through the entire body:

```ts
if (skipSet.has(tag) && !isVoid) {
  inIgnoredElement += 1;   // <html> pushes here and never pops within the doc
}
```

`rendering/orchestrator/html.ts:309` passes exactly that full document when `studioEmbed` is set. The result: **no element ever received a selector annotation**. The feature did nothing. Ignored elements now suppress their subtree except where the element wraps the whole document.

## Source defect 2: tasks listed but not retrievable

`list()` returned every map entry after a throttled `lazySweep()`, while `get()` applies delete-on-expiry. A completed task past its post-terminal TTL was therefore **listed but not retrievable** — MCP clients saw tasks that vanished on the very next call. `list()` now applies the same expiry semantics `get()` does.

Both were caught by assertions that landed red, not by reading code.

## Test changes

**Identity was asserted by length.** A session id checked only `length > 16`, which any sequential counter satisfies. It now asserts a v4 UUID shape and that 401 ids minted across two independent managers are all distinct.

**Terminating one of two sessions** now asserts the required-header flag stays set, so clearing it unconditionally fails.

**Key order was treated as significant.** Reordered tool-result payloads (at both nesting levels) now assert the duplicate is recognized, so swapping the stable serializer for plain `JSON.stringify` fails.

**A transport failure** now asserts the original error propagates rather than being reported as an empty body.

**Delegation denial only asserted the merged tool list**; it now asserts the remote tool grants are stripped too, so `invoke_agent` cannot survive in `__vfAllowedRemoteTools`.

**Release asset pinning** now asserts both the Studio-embed and HMR escape hatches, so removing either fails.

**Interface selection** now covers the `node:os` fallback and a loopback address that is not flagged `internal`.

## Verification

- Semantic unit-boundary audit: ok, sanitizer ratchet test: ok (396/396)
- **26/26** changed test files pass

### DIFF
--- a/src/client/spa/ClientApp.reactivity.test.tsx
+++ b/src/client/spa/ClientApp.reactivity.test.tsx
@@ -297,6 +297,8 @@ describe("client/spa/ClientApp (reactive)", () => {
 
   it("does not let the initial component load overwrite a completed navigation", async () => {
     await withTempDir(async (tempDir) => {
+      // The initial page resolves only after the navigation below has already
+      // failed, so its late result must be discarded rather than committed.
       await writeModule(
         tempDir,
         "pages/initial.tsx",
@@ -304,16 +306,12 @@ describe("client/spa/ClientApp (reactive)", () => {
          await new Promise((resolve) => setTimeout(resolve, 40));
          export default function Page() { return React.createElement("span", null, "initial"); }`,
       );
-      await writeModule(
-        tempDir,
-        "pages/fast.tsx",
-        `import React from "react";
-         export default function Page() { return React.createElement("span", null, "fast"); }`,
-      );
 
       const restore = installDom("https://example.com/initial");
       testGlobal.MODULE_SERVER_URL = `file://${tempDir}`;
       clearComponentCache();
+      const originalError = console.error;
+      console.error = () => {};
       try {
         const initialData: PageDataResponse = {
           slug: "/initial",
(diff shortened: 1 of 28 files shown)


kwakayama (dismissed):
> *This was generated by AI during triage.*\n\n[P2] The added bridge cases use raw Deno.test/std assertions, so this coverage is excluded from the repository's cross-runtime test contract. Convert the new cases to the existing BDD and assertion facades.

kwakayama (approved):

Re-reviewed commit b7fd01efb. The bridge cases now use the repository BDD and assertion wrappers; focused bridge tests passed. Approved.

kwakayama (approved):
> *This was generated by AI during triage.*\n\nRe-reviewed the current pushed head after dismissing superseded change requests. The addressed threads are resolved and the focused checks pass. Approved pending required checks.

PR #4149, opened 2026-08-25
fix(modules): clear batch cache entries by the key the write path used

### LINKED ISSUE #4088: clearBatchCache cannot clear entries for a project that supplies a projectId
## What

`src/modules/server/module-batch-handler.ts` writes cache keys under a project key that prefers `projectId`:

```ts
const projectKey = projectId || projectSlug || "default";   // line ~224
// ...
const cacheKey = buildBatchTransformCacheKey(projectKey, ...);   // line ~318
```

but deletes by `projectSlug`:

```ts
const prefix = `${projectSlug}:`;   // line ~744
```

## Impact

Whenever a project supplies a `projectId`, its entries are stored under that id and the deletion prefix built from the slug matches nothing. `clearBatchCache(projectSlug)` silently clears zero entries, so stale batch bundles survive an invalidation that reports success.

## Suggested fix

Clear by the same key the write path computed. That means either threading `projectId` into `clearBatchCache`, or having both sides call one shared key-derivation helper so they cannot drift.

## Provenance

Found during a test-quality audit of `src/modules` (PR #4058). Not asserted there: the audit's proposed test would have had to pick whichever identity `clearBatchCache` currently matches and pin it, which encodes the bug. Reported here instead.

kojiwakayama:
## Description

`module-batch-handler.ts` writes cache keys under a project key that prefers `projectId`:

```ts
const projectKey = projectId || projectSlug || "default";
const cacheKey = buildBatchTransformCacheKey(projectKey, ...);
```

but deleted by `projectSlug`:

```ts
const prefix = `${projectSlug}:`;
```

Whenever a project supplies a `projectId`, its entries are stored under that id and the slug-derived prefix matches nothing. `clearBatchCache(projectSlug)` cleared **zero** entries while returning normally and logging "Cleared cache for project", so stale batch bundles survived an invalidation that reported success.

### Fix

Derive the namespace in one place. `buildBatchCacheProjectKey` is now the single definition of the prefix every key begins with, and both the request path and `clearBatchCache` go through it, so the two sides cannot drift apart again.

`clearBatchCache` now takes the project identity rather than a slug, so a caller can pass the same `projectId`/`projectSlug` the request path was given. A bare string is still accepted and treated as an already-derived project key, which is what the existing callers pass, so this is not a breaking change.

## Related Issue(s)

Fixes #4088

## Type of Change

## Checklist

## Testing

The existing "should clear cache for specific project slug" test only ever passed a `projectSlug` with no `projectId`, so the two keying schemes coincided and the mismatch was invisible. The new test caches a transform for two projects that **do** supply a `projectId`, clears one, and asserts the entry is actually gone and the other project's survives. Verified to fail before the fix:

```
clearing a project that supplies a projectId must delete its entry, not silently match nothing
  actual: 2  expected: 1
```

```
deno task test:file src/modules/     60 passed (739 steps) | 0 failed
deno check src/modules/index.ts src/server/index.ts
deno task lint:anti-slop / lint:test-semantic-dispositions / lint:module-boundaries
deno fmt --check, deno lint
```

### DIFF
--- a/src/modules/server/module-batch-handler.test.ts
+++ b/src/modules/server/module-batch-handler.test.ts
@@ -183,6 +183,56 @@ describe(
 
         clearBatchCache();
       });
+
+      it("clears entries for a project that supplies a projectId", async () => {
+        clearBatchCache();
+
+        async function cacheOneTransform(
+          identity: { projectSlug: string; projectId?: string },
+        ): Promise<void> {
+          const adapter = createMockAdapter();
+          adapter.fs.files.set(
+            `/test-project/${identity.projectSlug}.tsx`,
+            "export const value = 1;",
+          );
+          const url = new URL("/_vf_modules/_batch", "http://localhost:8080");
+          url.searchParams.set("paths", `${identity.projectSlug}.js`);
+          const response = await handleModuleBatch(new Request(url.toString()), {
+            projectDir: "/test-project",
+            adapter,
+            projectSlug: identity.projectSlug,
+            projectId: identity.projectId,
+            releaseId: `rel-${identity.projectSlug}`,
+            dev: false,
+          });
+          assertEquals(
+            response.status,
+            200,
+            `project ${identity.projectSlug} must serve its module`,
+          );
+          await response.text();
+        }
+
+        await cacheOneTransform({ projectSlug: "slug-a", projectId: "id-a" });
+        await cacheOneTransform({ projectSlug: "slug-b", projectId: "id-b" });
+        assertEquals(getBatchCacheStats().size, 2, "both projects cached a transform");
+
+        clearBatchCache({ projectSlug: "slug-a", projectId: "id-a" });
+
+        const stats = getBatchCacheStats();
+        assertEquals(
+          stats.size,
+          1,
+          "clearing a project that supplies a projectId must delete its entry, not silently match nothing",
+        );
+        assertEquals(
+          stats.keys.every((key) => key.startsWith("id-b:")),
+          true,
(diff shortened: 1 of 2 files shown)


kwakayama:
@codex review

kwakayama (approved):
Reviewed exact head ed5bdabed7a96fdd3bbbbd29f8614a63c808b671. Cache writes and invalidation now share one namespace derivation, with two-project invalidation coverage. Fresh Codex review reports no findings and no threads remain unresolved.

kojiwakayama:
The `coverage shard 8/8` failure on this PR is **not** from these changes.

```
upstream WebSocket client => ./src/testing/bdd.ts:631:11
error: Leaks detected:
```

That is `src/proxy/websocket-client.test.ts`. This PR touches only `src/modules/server/module-batch-handler.ts` and its test. The same failure also hit #4150, which touches only `src/transforms/import-rewriter/`.

Root cause: `UpstreamWebSocket#pump` awaits `reader.read()`, and closing the stream resolves `closed` without settling a read that is already awaiting, so the operation outlives the connection. Fixed at source in **#4158**, with a deterministic regression test rather than a re-run.

Re-running this shard should pass in the meantime.

kwakayama:
@codex review

Please review the exact current head c70c351fd58ffa0b030c9a835d6d222495682e46 after updating it onto main. The PR changes were previously reviewed clean at ed5bdabed7a96fdd3bbbbd29f8614a63c808b671, all threads remain resolved, and main now contains the unrelated WebSocket leak fix. Please report any remaining findings against this SHA.

PR #4152, opened 2026-08-25
fix(routing): canonicalise before the CJS shim's containment check

### LINKED ISSUE #4091: Compiled-binary CJS shim rejects legitimate dependencies when the project root is a symlink
## What

The emitted CJS shim in `src/routing/api/module-loader/external-import-rewriter.ts` asserts containment **before** canonicalising:

```js
__vf_assertContained(resolved);
try {
  real = Deno.realPathSync(resolved);
} catch (_) { /* non-existent path */ }
if (real !== undefined) {
  __vf_assertContained(real);
  resolved = real;
}
```

The pre-canonicalisation check compares a non-canonical `resolved` path against a project root that has already been through `realPathSync`. When the project root itself is reached through a symlink, those two disagree and the first assertion rejects a dependency that is genuinely inside the project.

## Impact

Fails closed rather than open, so this is a correctness and usability problem rather than a security one. It affects any layout where the project root is a symlink, which is common with package managers, CI checkouts and `/tmp` working directories on macOS (`/tmp` is itself a symlink to `/private/tmp`).

## Suggested fix

Compare like with like: either canonicalise the project root and the candidate consistently before the first assertion, or defer the containment check until after `realPathSync` has run, keeping a fallback for paths that do not exist yet.

Note PR #4066 fixed a *different* defect in the same block — the post-canonicalisation assertion used to sit inside the `try`, so a symlink escaping the root threw and was swallowed. That fix is what makes this pre-canonicalisation mismatch observable, and the two should be read together.

## Provenance

Found during a test-quality audit of `src/routing` (PR #4066). Not fixed there because the correct fix changes how the project root is canonicalised, which is beyond a test change.

kojiwakayama:
## Description

The emitted CJS shim asserted containment **before** canonicalising:

```js
__vf_assertContained(resolved);
try { real = Deno.realPathSync(resolved); } catch (_) { /* non-existent path */ }
if (real !== undefined) { __vf_assertContained(real); resolved = real; }
```

`__vf_projectRoot` has already been through `realPathSync` at shim init, so that first assertion compares a **non-canonical** candidate against a **canonical** root. When the project root is itself reached through a symlink the two disagree, and a dependency genuinely inside the project is rejected with `CJS loader blocked path outside project`.

This fails closed rather than open, so it is a correctness and usability problem rather than a security one. It affects any layout where the project root is a symlink: package managers, CI checkouts, and macOS `/tmp`, which is a symlink to `/private/tmp`.

### Fix

Canonicalise first, then assert once. `__vf_canonicalize` falls back to the **parent directory** when the module does not exist yet, since the parent is what carries a symlinked root, and to the plain resolved path when that fails too. The assertion therefore always runs, on the most canonical form available.

The symlink-escape guard is unaffected. That escape is only visible *after* canonicalisation, which is precisely what the single remaining assertion runs on. Dropping the pre-canonicalisation assertion loses nothing: any path it would have rejected is still rejected by the canonical check, whereas the reverse — canonical-contained but non-canonical-looking — is exactly the false rejection being fixed.

## Related Issue(s)

Fixes #4091

## Type of Change

## Checklist

### Why the existing test did not catch this

`loader.test.ts` already had a test named **"accepts dependencies through a symlinked project root"**. It passed on `main` while the bug was live, because it re-implemented `__vf_assertContained` inline:

```ts
const assertContained = (resolved: string): void => { /* local copy */ };
const real = Deno.realPathSync(resolvedThroughSymlink);
assertContained(real);          // only ever called with the canonical path
```

It never ran the emitted shim and never called the check with the pre-canonical path, so the assertion that actually rejected these dependencies was never exercised. The test was vacuous with respect to the behaviour it claimed to cover.

It now uses the file's existing `importCompiledBinaryRequireShim` helper to execute the real emitted shim against a symlinked root. Verified to fail without the fix:

```
CJS loader blocked path outside project:
  /private/var/folders/.../tmp-9827889dd203c9dd-link/node_modules/ok/index.js
```

The neighbouring escape test — "re-canonicalisation via realPathSync catches a node_modules symlink escape" — passes both before and after, confirming the security property is intact.

One static test was also retargeted. It asserted the literal string `__vf_assertContained(real)`, pinning a variable name rather than a property; it now asserts that canonicalisation precedes the containment check.

```
deno task test:file src/routing/     45 passed (1161 steps) | 0 failed
deno check src/routing/api/index.ts
deno task lint:anti-slop / lint:test-semantic-dispositions / lint:skipped-tests / lint:sanitizer-baseline
deno fmt --check, deno lint
```

### DIFF
--- a/src/routing/api/module-loader/external-import-rewriter.ts
+++ b/src/routing/api/module-loader/external-import-rewriter.ts
@@ -77,6 +77,20 @@ var __vf_projectRoot = ${safeProjectRoot};
 // legitimate dependencies would be rejected.
 try { __vf_projectRoot = Deno.realPathSync(__vf_projectRoot); } catch (_) { /* expected: projectRoot may not exist at shim init in some environments */ }
 var __vf_cache = Object.create(null);
+// Resolve symlinks so containment compares canonical paths. A module that does
+// not exist yet cannot be canonicalized, so fall back to its parent directory,
+// which is what carries a symlinked project root; only if that fails too is the
+// plain resolved path used.
+function __vf_canonicalize(candidate) {
+  try {
+    return Deno.realPathSync(candidate);
+  } catch (_) { /* expected: candidate may not exist */ }
+  try {
+    var parent = __vf_dirname(candidate);
+    return Deno.realPathSync(parent) + candidate.slice(parent.length);
+  } catch (_) { /* expected: parent may not exist either */ }
+  return __vf_resolve(candidate);
+}
 function __vf_assertContained(resolved) {
   var norm = __vf_resolve(resolved).replace(/\\\\/g, "/");
   var root = __vf_projectRoot.replace(/\\\\/g, "/");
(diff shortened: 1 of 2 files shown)


kwakayama:
@codex review

kwakayama (approved):
Reviewed exact head 89a416bd16a57e4bb0048009e053b1648976f11c. The CJS shim canonicalizes the target before enforcing containment, and the emitted-shim regression exercises the boundary. Fresh Codex review reports no findings and no threads remain unresolved.

PR #4154, opened 2026-08-25
fix(sandbox): stop LazySandbox aborting on a malformed NDJSON line

### LINKED ISSUE #4080: LazySandbox.executeStream aborts on a malformed NDJSON line while Sandbox.executeStream skips it
## What

`Sandbox.executeStream` and `LazySandbox.executeStream` run the same NDJSON read loop, but only one of them guards the parse.

`src/sandbox/sandbox.ts` wraps both parses and documents the intent:

```ts
for (const line of lines) {
  if (line.trim()) {
    try {
      yield JSON.parse(line) as ExecStreamEvent;
    } catch {
      // Malformed NDJSON line (e.g. truncated network chunk); skip and
      // continue streaming so already-buffered output is not lost.
    }
  }
}
// ...
buffer += decoder.decode();
if (buffer.trim()) {
  try {
    yield JSON.parse(buffer) as ExecStreamEvent;
  } catch {
    // Malformed final chunk; discard rather than surfacing a SyntaxError.
  }
}
```

`src/sandbox/lazy-sandbox.ts` performs the same two `JSON.parse` calls with no `try`/`catch` around either.

## Impact

A single malformed line — the truncated network chunk the eager path explicitly anticipates — throws a `SyntaxError` out of the lazy generator. The stream aborts and every already-buffered event is discarded, so the caller loses output that had already arrived successfully.

The two implementations disagree about the same failure mode, and the comments in the eager one make clear which behavior is intended.

## Suggested fix

Apply the same guards in `LazySandbox.executeStream`, or better, extract the shared read loop so the two paths cannot drift again. The duplication is what allowed the divergence.

## Provenance

Found during a test-quality audit of the smaller `src` modules (PR #4079). That PR adds a malformed-NDJSON test for the eager path, which passes. **No equivalent lazy-path test was written**, because it would have failed — encoding the abort as expected behavior is exactly what the audit is trying to remove, so it is reported here instead.

kojiwakayama:
## Description

`Sandbox.executeStream` and `LazySandbox.executeStream` ran the same NDJSON read loop, but only the eager one guarded its two `JSON.parse` calls. The eager version even documents the intent:

```ts
} catch {
  // Malformed NDJSON line (e.g. truncated network chunk); skip and
  // continue streaming so already-buffered output is not lost.
}
```

The lazy version performed both parses with no `try`/`catch`. A single malformed line — exactly the truncated network chunk the eager path anticipates — threw a `SyntaxError` out of the generator. The stream aborted and **every already-buffered event was discarded**, so the caller lost output that had arrived successfully.

### Fix

Extract the loop into `readExecStreamEvents` (`src/sandbox/exec-stream.ts`) and have both paths delegate to it.

This is the issue's preferred remedy, and it is the right one: copying the loop is what let the two disagree about the same failure mode in the first place. Applying the missing guards in place would have fixed today's divergence while leaving the next one just as easy. The two loops were byte-identical apart from the guards and a `lines.pop()!` vs `lines.pop() ?? ""`, so nothing was lost in unifying them.

The extracted reader preserves the eager behaviour exactly: a malformed line is skipped, a malformed final chunk is discarded rather than surfacing a `SyntaxError`, and an abandoned generator still cancels the response body.

## Related Issue(s)

Fixes #4080

## Type of Change

## Checklist

## Testing

There was **no test file for `lazy-sandbox.ts` at all**, which is why the divergence went unnoticed. Rather than build a control-plane mocking harness to test the lazy path indirectly, the shared reader now has its own hermetic test file covering the behaviour both paths depend on:

- one event per line
- streaming continues past a malformed line
- a malformed final chunk is discarded, not thrown
- a final line with no trailing newline is still yielded
- an event split across chunk boundaries is reassembled
- blank lines are ignored
- an abandoned generator cancels the body

Because both call sites now `yield*` the same function, the lazy path gets this behaviour by construction rather than by a parallel test that could drift again.

The pre-existing eager end-to-end test ("should skip malformed NDJSON lines and keep delivering events") is unchanged and still passes, confirming the extraction preserved observable behaviour.

```
deno task test:file src/sandbox/     5 passed (132 steps) | 0 failed
deno check src/sandbox/index.ts
deno task lint:anti-slop / lint:test-semantic-dispositions / lint:module-boundaries / lint:wildcard-exports
deno fmt --check, deno lint
```

### DIFF
--- a/docs/api-reference/veryfront/sandbox.md
+++ b/docs/api-reference/veryfront/sandbox.md
@@ -182,15 +182,15 @@ Streaming event emitted during command execution.
 | `createSandboxShellTools`              | Create sandbox shell tools.                     | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/shell-tools.ts#L242)         |
 | `normalizeBashToolSet`                 | Normalizes bash tool set.                       | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/shell-tools.ts#L197)         |
 | `renameSandboxFileTools`               | Rename sandbox file tools.                      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/shell-tools.ts#L206)         |
-| `resolveDefaultSandboxRuntimeEndpoint` | Resolves default sandbox runtime endpoint.      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/lazy-sandbox.ts#L79)         |
+| `resolveDefaultSandboxRuntimeEndpoint` | Resolves default sandbox runtime endpoint.      | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/lazy-sandbox.ts#L80)         |
 | `unwrapSandboxWorkingDirectoryCommand` | Unwrap sandbox working directory command.       | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/agent-service-tools.ts#L51)  |
 
 ### Classes
 
 | Name          | Description                                                                             | Source                                                                                           |
 | ------------- | --------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------ |
-| `LazySandbox` | Lazily provisions sandbox sessions and keeps them alive while in use.                   | [source](https://github.com/veryfront/veryfront-code/blob/main/src/sandbox/lazy-sandbox.ts#L105) |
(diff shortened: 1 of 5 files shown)


kwakayama:
@codex review

kwakayama (approved):
Reviewed exact head 75877ceeb889c7d08122d9e85e7bed129f75a7d7. The shared NDJSON parser preserves eager behavior, fixes lazy-stream malformed-line handling, and is covered for chunking, trailing data, and cancellation. Fresh Codex review reports no findings and no review threads remain unresolved.

PR #4158, opened 2026-08-25
fix(proxy): settle the pending read when an upstream socket closes

kojiwakayama:
## Description

`UpstreamWebSocket#pump` awaits `reader.read()` for the next frame:

```ts
while (true) {
  const { done, value } = await reader.read();   // <- stays pending
  if (done) return;
  this.onmessage?.(new MessageEvent("message", { data: value }));
}
```

Closing the stream resolves `closed` and fires `onclose`, but it does **not** settle a `read()` that is already awaiting. The read outlives the connection it belongs to.

In the proxy that leaks one pending read per closed socket. In CI it surfaces as an intermittent failure from `src/proxy/websocket-client.test.ts` under `--trace-leaks --parallel`:

```
upstream WebSocket client => ./src/testing/bdd.ts:631:11
error: Leaks detected:
    at op_ws_next_event (ext:core/00_infra.js:256:13)
    at pull (ext:deno_websocket/02_websocketstream.js:253:34)
```

### Why this is worth fixing now rather than filing

It is not confined to the proxy. This exact failure took out `coverage shard 8/8` on **two separate pull requests that touch neither the proxy nor WebSockets** — one changing only `src/transforms/import-rewriter/`, one changing only `src/modules/server/`. Every occurrence costs a shard re-run and, worse, invites the reading that the shard is "just flaky" when a real regression eventually lands there.

### Fix

Hold the reader and cancel it in `#close`. Cancelling settles the pending read, `#pump` unwinds, and `releaseLock` runs. The `#settled` guard already prevents re-entry from the `#fail` path.

## Type of Change

## Checklist

## Testing

Chasing the flake by re-running was not going to prove anything, so the regression guard is deterministic instead. It uses the client's existing `UpstreamWebSocketStreamFactory` seam with a stream whose readable never produces a frame and whose `close()` only resolves `closed`, which is precisely what a real `WebSocketStream` does. The test then asserts the read was settled.

Verified to fail without the fix:

```
AssertionError: Values are not equal:
  closing must settle the read, or the operation outlives the connection
```

```
deno task test:file src/proxy/     51 passed (533 steps) | 0 failed
deno check src/proxy/main.ts
deno task lint:anti-slop / lint:test-semantic-dispositions / lint:sanitizer-baseline
deno fmt --check, deno lint
```

Found while shepherding the issue-triage PRs (#4147, #4148, #4149, #4150, #4152, #4154), where it twice failed a shard on unrelated changes.

### DIFF
--- a/src/proxy/websocket-client.test.ts
+++ b/src/proxy/websocket-client.test.ts
@@ -80,7 +80,73 @@ function identityHeaders(): Headers {
   });
 }
 
+/**
+ * A stream whose readable never produces a frame, so the client's read stays
+ * pending until something settles it. `close()` resolves `closed`, which is all
+ * a real `WebSocketStream` does: it does not settle a read already awaiting the
+ * next frame.
+ */
+function silentStream(): {
+  stream: UpstreamWebSocketStream;
+  readCancelled: () => boolean;
+} {
+  let cancelled = false;
+  let resolveClosed: (info: { closeCode?: number; reason?: string }) => void = () => {};
+  const closed = new Promise<{ closeCode?: number; reason?: string }>((resolve) => {
+    resolveClosed = resolve;
+  });
+
+  const readable = new ReadableStream<string | Uint8Array>({
+    cancel() {
+      cancelled = true;
+    },
+  });
+
+  return {
+    stream: {
+      opened: Promise.resolve({
+        readable,
+        writable: new WritableStream<string | Uint8Array>(),
+      }),
+      closed,
+      close(closeInfo) {
+        resolveClosed(closeInfo ?? {});
+      },
+    },
+    readCancelled: () => cancelled,
+  };
+}
+
 describe("upstream WebSocket client", () => {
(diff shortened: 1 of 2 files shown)


kwakayama:
@codex review

Please review the exact current head . All existing review threads are resolved. Please report any remaining findings against this SHA.

kwakayama:
@codex review

Please review the exact current head . All existing review threads are resolved. Please report any remaining findings against this SHA.

kwakayama:
@codex review

Please review the exact current head 08aced3b00f16d707fd2ba91d1a51c54b735d2b5. All existing review threads are resolved. Please report any remaining findings against this SHA.

kwakayama (approved):
Approved exact head 08aced3b00f16d707fd2ba91d1a51c54b735d2b5 after an exact Codex no-findings review and confirmation that all review threads are resolved. Protected checks and the merge queue remain authoritative.

PR #4160, opened 2026-08-25
fix(proxy): cancel the in-flight upstream read when closing a socket

kwakayama:
## Problem

`coverage shard 8/8` failed on an unrelated PR with every assertion passing:

```
upstream WebSocket client
error: Leaks detected:
    at op_ws_next_event → pull (ext:deno_websocket/02_websocketstream.js)
```

`421 passed | 1 failed` — the failure is Deno's leak detector, not a test.

## Cause

`UpstreamWebSocket#pump` keeps an outstanding `reader.read()`:

```ts
async #pump(readable) {
  const reader = readable.getReader();
  while (true) {
    const { done, value } = await reader.read();   // outstanding receive op
    ...
  }
}
```

`close()` closed the stream but never cancelled that reader, so teardown raced the close propagating through the stream. Usually the close wins; when it doesn't, a receive op outlives the caller and `--trace-leaks` fails the entire shard.

The tests were never at fault — they use `try/finally`, close explicitly, and await `onclose`.

## Fix

Hold the reader on the instance and cancel it in `close()`, so the pending read is aborted deterministically rather than racing.

## Verification

- New regression test `cancels the in-flight upstream read when an open socket is closed`, verified **red** before the fix.
- Removing the `cancel()` turns it red again (mutation-checked).
- Full `websocket-client.test.ts`: 10 steps, 0 failures, run 12× consecutively under `--trace-leaks` with 0 failures.
- `deno task fmt:check`, `deno lint`, `deno check`: all exit 0.
- Ran the whole `src/proxy/*.test.ts` group with and without the change and diffed the failing test **names**: identical sets, so no regressions. (That group has pre-existing environment-dependent failures locally — control-plane signature and domain-auth tests — unchanged by this PR.)

### DIFF
--- a/src/proxy/websocket-client.test.ts
+++ b/src/proxy/websocket-client.test.ts
@@ -343,6 +343,41 @@ describe("upstream WebSocket client", () => {
     assertEquals(socket.readyState, WebSocket.CLOSING);
   });
 
+  it("cancels the in-flight upstream read when an open socket is closed", async () => {
+    let admit: (connection: UpstreamWebSocketConnection) => void = () => {};
+    const opened = new Promise<UpstreamWebSocketConnection>((resolve) => {
+      admit = resolve;
+    });
+    const stream: UpstreamWebSocketStream = {
+      opened,
+      closed: new Promise(() => {}),
+      close() {},
+    };
+    let cancels = 0;
+
+    const socket = new UpstreamWebSocket("ws://renderer/_ws", new Headers(), () => stream);
+    admit({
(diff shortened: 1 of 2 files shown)


kwakayama:
Closing as superseded by #4158, which merged as 4e106d6ba0. Both changes cancel the pending reader, but #4158 performs cancellation in the shared finish path, so it covers both local close and upstream/remote closure. Keeping this branch would duplicate the same regression fix with narrower teardown coverage.

PR #4175, opened 2026-08-25
fix: fail closed when Deno checksum manifest is unavailable

kojiwakayama:
### Motivation

- Prevent a supply-chain integrity regression where the installer fell back to an unpinned per-archive `.sha256sum` from the same release origin, allowing a malicious release asset and matching checksum to pass verification.
- Ensure CI release jobs do not execute an unverified Deno binary when the pinned `checksums.txt` manifest is missing or tampered with.

### Description

- Change the installer to require a successful download of the pinned checksums manifest and to verify it against the hardcoded SHA-256; the action now exits if the manifest request is not HTTP 200. (`.github/actions/setup-deno/action.yml`).
- Remove the unpinned per-archive `.sha256sum` fallback and use the archive checksum exclusively from the verified `checksums.txt` manifest when validating the downloaded Deno archive. (`.github/actions/setup-deno/action.yml`).
- Update the focused CI contract test to assert the installer fails closed and to disallow the per-archive checksum URL and any fallback messages. (`scripts/ci/setup-deno-workflow.test.ts`).

### Testing

- Ran `git diff --check` which passed with no whitespace issues.
- Ran a YAML contract check with Ruby that loads `.github/actions/setup-deno/action.yml` and asserts the installer fails closed when the manifest is unavailable, which passed locally.
- Attempted to run the local Deno test `deno test --no-check --allow-read --allow-write scripts/ci/setup-deno-workflow.test.ts` but the execution environment lacks `deno`, so the test could not be executed here; recommend running the test in CI or a developer environment with Deno installed.
- Verified repository status and diffstat locally to confirm the two modified files are the only changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a715f571fd0832dbd45f45a33e70d61)

### DIFF
--- a/.github/actions/setup-deno/action.yml
+++ b/.github/actions/setup-deno/action.yml
@@ -80,88 +80,69 @@ runs:
         esac
 
         archive="deno-${arch_target}-${os_target}.zip"
+
+        # The expected digest of every supported archive is pinned HERE, in the
+        # repository, so the release origin is never trusted to vouch for its
+        # own bytes. Deno publishes no combined checksums.txt manifest (the
+        # URL 404s on every real release), and the per-archive .sha256sum
+        # assets ship from the same origin as the archives themselves, so
+        # downloading either added no integrity: whoever could tamper with the
+        # archive could tamper with its checksum file too. This is the
+        # fail-closed anchor -- an archive whose digest is not pinned below, or
+        # does not match its pin, must never be executed. Bumping ${version}
+        # means re-pinning every digest from the new release's assets.
+        case "${archive}" in
+          deno-x86_64-unknown-linux-gnu.zip)
+            archive_sha256="0cd918870657ccc3d96ac682290e894dda374e2a742424aae9118b258a6cf7a3" ;;
+          deno-aarch64-unknown-linux-gnu.zip)
+            archive_sha256="e654ef4b41d8f1369dfe7d69374c8f226660153d077e0ac25353d454e03ef798" ;;
+          deno-x86_64-apple-darwin.zip)
+            archive_sha256="2ca88d69c369a2757f17fce7c21d979f04cf56a6990d1f3858bd80b0487d0b74" ;;
+          deno-aarch64-apple-darwin.zip)
+            archive_sha256="5f0ef47e706ecdae5e6248e33bf367340b1d86f87886f60419fe7450f6ffc39a" ;;
+          deno-x86_64-pc-windows-msvc.zip)
+            archive_sha256="dcb4aa0a87a1e0e3a096692d27443bd2496e9204b327150fb01b31984255c3aa" ;;
+          deno-aarch64-pc-windows-msvc.zip)
+            archive_sha256="5693b3107fa22476f007bdb61f0157f16870edd56a636b95eccfdcbec9c82db6" ;;
+          *)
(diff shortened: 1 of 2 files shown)


kwakayama:
Codex security review — exact head `2cc1ba2a44b0e42a8587e89e8be34ad00c65be2e`

Blocking findings:
1. The proposed required `checksums.txt` URL is not published by Deno releases, so this would make every setup job fail closed on a guaranteed 404. Current main deliberately uses the per-archive checksum asset for that reason.
2. SonarCloud flags the manifest curl redirect policy (`githubactions:S6506`); HTTPS-only redirect enforcement is missing.

The PR conflicts with current main and cannot be safely rebased without choosing a real independently pinned checksum source. Maintainer direction is required.

Score: 58/100

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 2cc1ba2a44b0e42a8587e89e8be34ad00c65be2e
Score: 58/100
Actionable-Findings: 2
Verdict: NEEDS_HUMAN_INPUT

kwakayama:
Both blocking findings from the Codex security review (reviewed SHA `2cc1ba2a4`) are addressed as of head commit `8a56c4661b0c5306b09ccd6140f5119975fcde5c`.

**Finding 1 — fail-closed on the unpublished `checksums.txt` (guaranteed 404).** The phantom manifest is gone entirely: the installer no longer fetches `checksums.txt` at all, so no job can fail on its guaranteed 404. Instead of moving the fail-closed anchor to the per-archive `.sha256sum` (which ships from the same release origin as the archive, so it cannot vouch for it), the expected SHA-256 of every supported v2.7.7 archive is now pinned in the action itself. The old pinned "manifest hash" was in fact the SHA-256 of GitHub's 9-byte 404 body, which is how the guaranteed failure slipped in.

**Finding 2 — Sonar S6506.** The single remaining `curl` carries `--proto "=https" --proto-redir "=https"`, so `--location` can never follow the release redirect onto an insecure protocol. SonarCloud's quality gate on this PR now passes (0 new issues).

**New integrity chain:** repo-pinned digest (six targets, one per supported platform) → single fail-closed archive download (HTTPS-only, `--retry-all-errors` within a bounded 150s window, `--remove-on-error`) → `compute_sha256(zip)` compared against the pin → any mismatch, unpinned target, or download failure exits before anything is unzipped or executed. The release origin is never trusted to vouch for its own bytes.

**Verification of the pins:** all six digests were re-derived from the real `v2.7.7` release assets and match exactly (the Windows `.sha256sum` files publish uppercase PowerShell format; pins are lowercase, matching what `compute_sha256` emits). The `deno-x86_64-unknown-linux-gnu.zip` asset was additionally downloaded and hashed end-to-end: `0cd91887...a6cf7a3`, identical to its pin.

**Validation:** `bash -n` passes on all four embedded shell steps; `deno test --frozen --config=scripts/test.deno.json ... scripts/ci/setup-deno-workflow.test.ts` passes (2 suites, 11 steps, 0 failures — including the new single-URL contract test whose curl stub hard-fails any fetch other than the archive); `deno fmt --check` is clean on the touched files.

---

kojiwakayama:
@codex review

kwakayama (approved):

Issue #844, opened 2026-04-03
feat(mcp): emit listChanged notifications on registry updates


ariskemper:
## Summary

Emit `notifications/tools/list_changed`, `notifications/resources/list_changed`, and `notifications/prompts/list_changed` when registries change at runtime. Primarily for lazy integration tool loading.

**Branch:** `feat/mcp-list-changed`
**Plan:** `plans/mcp_server_implementation.md` — PR 13
**Research:** `research/mcp_server.md` — Priority 13
**Phase:** 2 (requires server-initiated messaging)
**Dependencies:** PR 8 (Streamable HTTP) — #839

## What's Wrong Today

- `listChanged: true` is declared in capabilities (after PR 1) but notifications are never sent
- When `loadRemoteIntegrationTools()` adds new tools, connected clients don't know to re-fetch
- Agents see a stale tool list until they reconnect

## Tasks

- [ ] **13.1** Add `onNotification` callback and `notifyToolsChanged()`/`notifyResourcesChanged()`/`notifyPromptsChanged()` to `MCPServer`
- [ ] **13.2** Call `notifyToolsChanged()` after successful `loadRemoteIntegrationTools()`
- [ ] **13.3** Wire notification callback into stdio transport in `cli/mcp/server.ts`

## Files

- Modify: `src/mcp/server.ts`
- Modify: `cli/mcp/server.ts`, `cli/mcp/standalone.ts`
- Test: `src/mcp/server.test.ts`

## Acceptance Criteria

- [ ] `notifyToolsChanged()` emits `notifications/tools/list_changed` via callback
- [ ] `notifyResourcesChanged()` and `notifyPromptsChanged()` emit their respective notifications
- [ ] `loadRemoteIntegrationTools()` triggers `notifyToolsChanged()` on success
- [ ] Notifications only sent over transports that support server-initiated messages
- [ ] After notification, client re-fetching `tools/list` sees the updated set
- [ ] Tests cover: notification fires on tool change, client re-fetch sees new tools

PR #4240, opened 2026-08-26
fix(ssr): settle a moving source generation instead of 503ing the request

kwakayama:
Fixes the staging 503 tracked in veryfront/veryfront-issue-inbox#844.

## The bug

`capturePreviewSourceSnapshotMarker()` takes two observations of the source generation and returns `undefined` when they disagree — a deliberate guard against publishing a marker that straddles a branch switch.

`captureRequiredPreviewSourceSnapshotMarker()` called it **once** and treated `undefined` as terminal:

```ts
const marker = await capturePreviewSourceSnapshotMarker(fs);
if (marker?.identity !== undefined && marker.version !== undefined) return marker;
throw SOURCE_SNAPSHOT_FRESHNESS_UNAVAILABLE.create({ ... });
```

So one lost race was reported as "this adapter cannot identify its strict source snapshot" — a permanent-capability error — when the real condition was transient.

A project whose source is still being written loses that race routinely while being perfectly healthy. The first document request after project creation is exactly that case, and it surfaced as a user-visible **503 on preview rendering**.

## Observed impact

staging, `veryfront-server` running `0.1.1253-rc.16242`:

```
"Unhandled error in request handler"  path:"/"  method:"GET"
"GET / 503"  durationMs:1491
```

4 occurrences across 3 distinct projects in 40 minutes — two of them ordinary projects, not test fixtures. `staging / platform-ui` in the e2e health suite went red on `preview-rendering.health.spec.ts` and `create-project.health.spec.ts` after four consecutive green runs.

Production is unaffected: it runs `0.1.1252` with zero occurrences in 6h.

## The change

Re-observe up to three times before giving up. Each attempt is two adapter reads and is side-effect free, so this converts the dominant transient failure into a slightly slower success.

The retry is deliberately **bounded**. A source that genuinely never settles must still fail loudly rather than hold the request open, and the second test pins that.

## Why here and not a request-level retry

The obvious alternative is to catch the error higher and re-dispatch, mirroring `ProxyRoutingInvalidationRaceError` in `src/proxy/handler.ts` (retry the narrow resolution step; 503 only if the race recurs).

That is the right shape for the *post-derivation* throws (`throwConfigSnapshotChanged`), but the nearest enclosing boundary that wraps a re-invocable thunk is `withRequestTimeout`, and retrying there re-runs `incrementRequestMetrics()` and application auth — double-counting request metrics. Extracting a retryable unit out of `runtime-handler/index.ts` is a real refactor of the universal request path and is not attempted here.

This change fixes the case that is actually firing in staging — the generation moving while the marker is captured — at the point where a retry costs two reads and cannot duplicate a side effect. #844 keeps the request-level design recorded for the residual window.

## Verification

Red before green, on the same test:

```
without the fix:  22 passed | 1 failed   ← new settling test fails
with the fix:     23 passed | 0 failed
```

```
deno task test:file source-snapshot-freshness.test.ts   23 passed, 0 failed
deno task test:file adapter-factory.test.ts             1 passed (39 steps), 0 failed
deno task test:file ssr.handler.test.ts                 exit 0
deno check  (both changed files)                        exit 0
```

### DIFF
--- a/src/server/handlers/request/source-snapshot-freshness.test.ts
+++ b/src/server/handlers/request/source-snapshot-freshness.test.ts
@@ -505,3 +505,37 @@ it("rejects identity-only strict config markers without a concrete generation", 
   assertInstanceOf(rejection, VeryfrontError);
   assertEquals(rejection.slug, "source-snapshot-freshness-unavailable");
 });
+
+it("settles a generation that moves during the first observations", async () => {
+  // A project whose source is still being written: the generation advances
+  // under the first two captures, then settles. Before the bounded retry this
+  // threw on the first lost race and the document request became a 503.
+  let version = 0;
+  const adapter = createMockAdapter();
+  adapter.fs.getSourceSnapshotIdentity = () => "branch:preview-project:main";
+  adapter.fs.getSourceSnapshotVersion = () => {
+    // Two reads per capture attempt; advance across the pair for the first two
+    // attempts so each is discarded, then hold steady.
+    version++;
+    return version < 5 ? version : 5;
+  };
+
+  const marker = await captureRequiredPreviewSourceSnapshotMarker(adapter.fs, "preview-project");
+
+  assertEquals(marker.identity, "branch:preview-project:main");
+  assertEquals(marker.version, 5);
+});
+
+it("still fails loudly when the generation never settles", async () => {
+  let version = 0;
+  const adapter = createMockAdapter();
+  adapter.fs.getSourceSnapshotIdentity = () => "branch:preview-project:main";
+  adapter.fs.getSourceSnapshotVersion = () => ++version;
+
+  const error = await assertRejects(() =>
+    captureRequiredPreviewSourceSnapshotMarker(adapter.fs, "preview-project")
+  );
+
+  assertInstanceOf(error, VeryfrontError);
+  assertEquals(error.slug, SOURCE_SNAPSHOT_FRESHNESS_UNAVAILABLE.slug);
+});
--- a/src/server/handlers/request/source-snapshot-freshness.ts
+++ b/src/server/handlers/request/source-snapshot-freshness.ts
@@ -76,12 +76,30 @@ export async function capturePreviewSourceSnapshotMarker(
   };
 }
 
+/**
+ * Attempts to observe a settled generation before giving up.
+ *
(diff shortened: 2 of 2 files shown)


kwakayama:
## Review addressed

**gitar-bot — retry loop re-observes with no delay between attempts.** Valid, and fixed in 21f2f8f.

The finding is right for the reason it gives. Against a remote adapter each attempt is four awaited reads and already spans real time, which is why this worked at all — but against a cached or local adapter those promises can resolve in microseconds, so three immediate re-observations would give a writer no chance to settle and the retry would buy nothing. Pacing is what makes the retry meaningful, which was the stated goal of the fix.

Added a linear backoff between attempts (10ms, then 20ms):

```ts
for (let attempt = 0; attempt < REQUIRED_SNAPSHOT_MARKER_ATTEMPTS; attempt++) {
  if (attempt > 0) await delay(attempt * REQUIRED_SNAPSHOT_MARKER_RETRY_DELAY_MS);
  const marker = await capturePreviewSourceSnapshotMarker(fs);
  if (marker?.identity !== undefined && marker.version !== undefined) return marker;
}
```

Only the failing path pays it, and never before the first attempt, so a settled source is completely unaffected. Worst case adds ~30ms before throwing — against a request that was already taking ~1491ms to fail with a 503. `delay()` routes through `scaleDuration()`, the host timer contract, so tests are not slowed.

Also corrected in the same commit: the doc comment said re-observing costs "two adapter reads" when it is four, and the new import was in the wrong sort position.

### Verification after the change

```
deno task test:file source-snapshot-freshness.test.ts   23 passed | 0 failed
deno task test:file adapter-factory.test.ts             exit 0
deno task test:file ssr.handler.test.ts                 exit 0
deno fmt --check / deno lint / deno check               exit 0
```

The red-before-green evidence still holds: the settling test fails against the original single-observation code and passes with the retry.

### Other reviewers, for the record

- **SonarQube**: Quality Gate passed, 0 new issues
- **Codecov**: all modified lines covered
- **Client bundle boundary**: 0 server leaks
- **Greptile**: never ran — trial credit limit reached
- **CodeRabbit**: never ran — free-tier review limit reached

So gitar-bot's was the only substantive review this PR received. Worth knowing that two of the four review bots were silently unavailable rather than clean.

kojiwakayama (approved):

PR #4242, opened 2026-08-26
fix(security): let an operator grant host execution to a shared runtime

kwakayama:
Fixes veryfront/veryfront-issue-inbox#848 — preview returns 503 on staging.

## What broke

#4166 (`d4c0ddb410`, 2026-08-26 13:14) changed `resolveHostProjectExecutionPosture`:

```ts
return {
  allowHostProjectCodeExecution: !options.sharedRuntime,   // grant no longer consulted
  overrideIgnored: options.overrideConfigured,
};
```

The operator grant is read, recorded as ignored, and discarded. Its stated rationale:

> Dedicated single-project runtimes already carry the host execution capability. The former override therefore grants no additional topology…

**That premise does not hold here.** `veryfront-server` is this platform's shared rendering and executor runtime, and no dedicated isolated runtime is provisioned for hosted project previews or agents. So the capability was withdrawn while the product still depended on it.

## Observed impact

Staging, `0.1.1253-rc.16336`:

```
GET https://<project>.preview.veryfront.org/?studio_embed=true…   →  503

{"title":"Project execution unavailable","status":503,
 "detail":"Shared runtimes do not execute tenant API modules in the host process
           or same-process Workers",
 "suggestion":"Route the project to a dedicated isolated runtime"}
```

In Studio this shows as the preview pane reporting **"refused to connect"**.

| | Framework | Grant honoured | Preview |
|---|---|---|---|
| production | `0.1.1252` | yes — predates #4166 | working |
| staging | `0.1.1253-rc.16336` | **no** | **broken** |

The chart sets `VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION: "1"` in **both** environments, so promoting the `0.1.1253` line would break production the same way.

## The change

```ts
allowHostProjectCodeExecution: !options.sharedRuntime || options.overrideConfigured,
overrideIgnored: options.overrideConfigured && !options.sharedRuntime,
```

- Shared runtime **with** a grant → executes. This is the case that was broken.
- Shared runtime **without** a grant → still fail-closed. Unchanged.
- Dedicated runtime → already had the capability; the grant is still reported as `overrideIgnored` so startup can surface stale operator configuration.

## This is a security posture decision, made deliberately

#4166 is a security change and the boundary it describes is real: tenant code runs in the shared process, and per-request separation there is source scoping, not a tenant boundary.

This PR **re-accepts that posture** rather than discovering it. It is what production runs today. The argument is one of sequencing: the capability was removed before the replacement (routing execution to a dedicated isolated runtime, #356) existed, which produced a broken product surface rather than a tightened boundary.

Retiring the grant should land **with** that routing, not ahead of it. #848 records this so it is not lost.

## Verification — red before green

```
without this change:  2 failed   ← shared-runtime-with-grant, and fail-closed
with this change:     1 passed (9 steps) | 0 failed
```

Consumers of the posture, all passing:

```
src/routing/api/handler.test.ts                          exit 0
src/server/handlers/request/api/api-handler-wrapper.test.ts  exit 0
src/agent/project/agent-runtime.test.ts                  exit 0
src/data/server-data-fetcher.test.ts                     exit 0
deno fmt --check / deno lint / deno check                exit 0
```

## Summary by CodeRabbit

* **New Features**
  * Operators can explicitly enable tenant API code execution on shared runtimes through a host configuration setting.
  * Dedicated runtimes continue to allow execution without requiring the setting.

* **Security**
  * Shared runtimes remain isolated by default and fail closed when no grant is configured.
  * Startup warnings now clearly indicate when shared-runtime execution is enabled or when a setting is unnecessary.

### DIFF
--- a/src/security/README.md
+++ b/src/security/README.md
@@ -419,8 +419,8 @@ A runtime that cannot honour a configured isolation flag never fakes it. A
 compiled binary cannot prepare isolated API route source
 (`security/sandbox/isolation-capability.ts`), so `WORKER_ISOLATION_API=1` in a
 compiled deployment keeps the requested isolation posture and API ownership
-returns the typed `project-execution-unavailable` 503 naming it. The deprecated
-`VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION` setting does not override the
+returns the typed `project-execution-unavailable` 503 naming it. The
+`VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION` grant does not override the
 API-specific isolation flag.
 
 OpenAPI metadata is currently attached to handler functions. Because reading
@@ -438,10 +438,20 @@ evaluating tenant modules.
 ## Shared execution grant
 
 `VERYFRONT_HOST_ALLOW_PROJECT_EXECUTION` grants host execution to a shared
-runtime. Without it a shared runtime rejects same-process tenant execution and
-routes to a dedicated isolated project runtime. A dedicated single-project
-runtime already has the capability, so there the setting changes no supported
-topology and is reported as ignored.
+runtime. Without it a shared runtime rejects same-process tenant execution:
+preview rendering and tenant `/api/*` return the typed
+`project-execution-unavailable` 503 response until routing to a genuinely
+external or dedicated isolated project runtime exists. A dedicated
+single-project runtime already has the capability, so there the setting changes
+no supported topology and is reported as ignored.
+
+Only the affirmative values `1`, `true`, `yes`, and `on` (trimmed,
+case-insensitive) enable the grant; every other value is treated as
+unconfigured and fails closed. The grant does not override requested worker
+isolation: with `WORKER_ISOLATION_API=1`, API execution keeps the isolation
+posture — in a compiled deployment, which cannot prepare isolated API route
+source, API ownership still returns the documented
+`project-execution-unavailable` 503 despite the grant.
 
 **Granting it is a deliberate reduction in isolation.** Tenant code then runs in
 the shared process, where per-request separation is source scoping and not a
@@ -467,7 +477,9 @@ is no host-rendering or remote-import fallback. The current HTTP renderer does
 not yet produce a generation-owned isolated page and layout graph, so remote
 SSR and server-executing RSC endpoints return `503 Service Unavailable` before
 resolving a renderer. API and data workers do not resolve or receive the
-renderer contract.
+renderer contract. On a shared runtime, requested SSR isolation also outranks
+the shared execution grant: the SSR handler fails closed with the typed 503
+rather than silently downgrading an isolation request into host rendering.
 
 Deno Workers share the host process. Worker retirement is lifecycle hygiene,
 not a hard per-worker memory or CPU boundary. They are therefore limited to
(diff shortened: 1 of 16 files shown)


kwakayama:
## All three Codex P1 findings addressed

**@codex #2 — "Restore the grant in the topology-aware execution predicate" — correct, and it meant this PR did nothing.**

`isHostProjectCodeExecutionAllowedForTopology` discarded the capability for any shared runtime:

```ts
return isExplicitlyLocalProject(value) ||
  (!sharedRuntime && isExplicitHostProjectCodeExecutionAllowed(value));
```

So the resolver set the capability and the predicate threw it away. The SSR guard and API handler would have kept returning the same 503 this PR exists to fix. **My verification was the failure here**: I tested the resolver in isolation and ran four consumer suites, but none of them exercised shared-runtime-with-grant, so "all green" meant nothing. Good catch — this would have merged as a no-op and then looked like the staging deploy had failed.

Fixed, and `project-locality.test.ts` updated (it asserted the denial). The topology parameter is now unused so it is gone; `requiresIsolatedProjectRuntime` still samples topology once for its own `&& sharedRuntime`, and the single-sample invariant test now uses the ungranted path where topology actually decides — otherwise that test was passing for the wrong reason.

**@codex #3 — "Warn when a shared runtime activates host execution" — correct.**

`production-server` warned only on `overrideIgnored`, so an active grant was silent, including a stale setting silently becoming effective on upgrade. Added a distinct warning; the ignored-setting warning stays for dedicated runtimes.

**@codex #1 — "Align security guidance with the restored grant" — correct.**

`security/README.md` said a shared runtime never executes tenant source and the setting is ignored, and the module header plus `production-server.ts` repeated it. Replaced the "Deprecated shared execution override" section with one that says plainly that granting **reduces isolation**, why this deployment needs it, and that it is retired by routing execution to a dedicated runtime (#356) rather than by dropping it while that routing does not exist.

I also found a fourth instance you did not flag: `worker-pool.ts` documented `hostExecutionGranted: false` as *"the former override no longer grants execution"*. That posture reports the sandbox worker isolation surfaces and does not resolve the grant, so I corrected the comment to say so rather than wire a value it has no input for — an auditor must not read that field as evidence the runtime is not executing tenant code.

### Verification
```
host-execution-policy.test.ts   exit 0
project-locality.test.ts        exit 0   (red before green on the predicate)
routing/api/handler.test.ts     exit 0
api-handler-wrapper.test.ts     exit 0
agent-runtime.test.ts           exit 0
server-data-fetcher.test.ts     exit 0
isolation-posture.test.ts       exit 0
fmt / lint / check              exit 0
```

**gitar-bot** approved with no issues; **greptile** never ran (trial credit limit).

kwakayama:
Superseded by #4243.

This attempted a targeted fix. Three iterations each looked complete and were not — the grant is discarded at two structurally independent points (`host-execution-policy.ts` zeroes it at startup, `project-locality.ts` discards it again at the `!sharedRuntime &&` conjunct), so fixing either alone changes nothing in the response.

#4243 reverts the 20-file host-execution set to its pre-#4166 state instead, which restores a posture production runs healthily today and passes all eight affected suites unmodified — including the three this PR kept breaking.

PR #4264, opened 2026-08-27
Keep merge-group validation anchored to the queue base

kojiwakayama:
Summary
- Require merge-group validation to use queueEntry.baseHeadSha as the serialized queue base.
- Update the workflow regression check so the source-head compatibility fallback cannot return.

Verification
- deno task test:file scripts/ci/automated-review-gate.test.ts

Closes veryfront/veryfront-issue-inbox#834


kojiwakayama:
@codex review

PR #4278, opened 2026-08-28
test: use a reserved mock resolver address

kwakayama:
## Summary

- replace the public mock resolver sentinel with RFC 5737 TEST-NET-1
- allow that exact resolved address only for installed test transports
- keep direct and production internal-address validation unchanged

## Validation

- deno task lint
- deno task typecheck
- deno task test:file src/testing/mock-fetch.test.ts
- deno task test:file src/security/http/outbound-fetch.test.ts
- deno task test:file src/security/sandbox/worker-egress-guard.test.ts
- deno task test:unit: 4,389 passed; one unrelated CLI project-creation test failed because npm is unavailable in the local environment and reproduced in isolation


kwakayama:
Addressed the Sonar S3776 annotation by removing the wrapper around guardedEgressFetch. The test-only resolved-address allowance now flows through the existing dependency object, so the original guard remains intact.

Rechecked formatting, lint, type checking, anti-slop, and the focused mock/outbound/egress tests.

PR #4289, opened 2026-08-29
fix(platform): classify bind hostname resolution failures at serve startup

kojiwakayama:
Refs https://github.com/veryfront/veryfront-issue-inbox/issues/810 (Sentry VERYFRONT-STUDIO-5J)

## Root cause

`createDenoServerWithRuntime` classifies only `AddrInUse` bind failures; every other synchronous `Deno.serve()` startup fault is rethrown raw (`src/platform/adapters/runtime/deno/http-server.ts`). A bind hostname that fails DNS resolution therefore surfaced as an unhandled `Error: failed to lookup address information: Name or service not known` — the exact Sentry signature — with no port, no classification, and no suggestion.

## Fix

- `isAddressResolutionError`: a deliberately narrow classifier keyed on the cross-platform getaddrinfo message prefix (`failed to lookup address`), matched via `isErrorAcrossRealms` like its `isAddressInUseError` sibling.
- The catch relabels the fault as `initialization-error` through the error registry, mirroring the #806 AddrInUse precedent:
  - the raw OS error is preserved as `cause` (Sentry is not suppressed; correlation with platform logs still works),
  - a private bind hostname never reaches the message — only loopback/wildcard literals are named (`describeBindHost`), the port is always reported,
  - the message deliberately avoids the canonical "address already in use" phrase, so the dev server's port fallback does not retry ports for a fault that retrying cannot fix.
- Successful startup, cancellation, address-in-use, address-family-unavailable, and unrelated-failure rethrow behavior are all unchanged (existing tests cover each).

## Red → green

Red first: both new tests failed on the unclassified rethrow before the fix —

```
deno task test:file src/platform/adapters/runtime/deno/http-server.test.ts
  classifies a bind hostname that fails DNS resolution ... FAILED
  names a loopback bind host that fails DNS resolution ... FAILED
```

Green after: `ok | 1 passed (18 steps) | 0 failed`, plus `deno task test:file src/platform/adapters/runtime/deno/adapter.test.ts src/platform/adapters/runtime/deno/index.test.ts` → `4 passed (73 steps)`, `deno fmt`/`lint`/`check` clean on both files.

Synthetic fixtures only — no Sentry payload values are copied into tests (both getaddrinfo message variants are the documented OS strings, which carry no hostname).

### DIFF
--- a/src/platform/adapters/runtime/deno/http-server.test.ts
+++ b/src/platform/adapters/runtime/deno/http-server.test.ts
@@ -344,6 +344,69 @@ describe("Deno HTTP server lifecycle", () => {
     assertEquals(error.message.toLowerCase().includes("already in use"), false);
   });
 
+  it("classifies a bind hostname that fails DNS resolution", async () => {
+    // A deployment bound through a DNS name that does not resolve made
+    // Deno.serve() throw the raw getaddrinfo error, which reached Sentry as an
+    // unhandled `Error: failed to lookup address information: Name or service
+    // not known` with no port and no classification
+    // (veryfront-issue-inbox#810, Sentry VERYFRONT-STUDIO-5J).
+    const lookupFailure = new Error(
+      "failed to lookup address information: Name or service not known",
+    );
+
+    const error = await assertRejects(
+      () =>
+        createDenoServerWithRuntime(
+          {
+            serve() {
+              throw lookupFailure;
+            },
+          },
+          () => new Response("ok"),
+          { hostname: "registry.internal", port: 4321 },
+        ),
+      Error,
+      "4321",
+    ) as Error & { slug?: string; cause?: unknown };
+
+    assertEquals(error.slug, "initialization-error");
+    // The bind host can be an internal DNS name, which must never reach an
+    // error message (AGENTS.md); the port is the reportable half.
+    assertEquals(error.message.includes("registry.internal"), false);
+    assertStringIncludes(error.message.toLowerCase(), "could not be resolved");
+    // The raw OS-level signature is preserved for Sentry correlation rather
+    // than swallowed.
+    assertStrictEquals(error.cause, lookupFailure);
+  });
+
+  it("names a loopback bind host that fails DNS resolution", async () => {
+    // `localhost` itself can fail to resolve on a broken NSS setup. It is on
+    // the loopback allowlist, so unlike an infrastructure hostname it is safe
+    // to name -- and naming it is what makes the fault actionable.
+    const lookupFailure = new Error(
+      "failed to lookup address information: nodename nor servname provided, or not known",
+    );
+
+    const error = await assertRejects(
+      () =>
+        createDenoServerWithRuntime(
+          {
+            serve() {
+              throw lookupFailure;
+            },
+          },
+          () => new Response("ok"),
+          { hostname: "localhost", port: 4321 },
+        ),
+      Error,
+      "localhost:4321",
+    ) as Error & { slug?: string; cause?: unknown };
+
+    assertEquals(error.slug, "initialization-error");
+    assertStringIncludes(error.message.toLowerCase(), "could not be resolved");
+    assertStrictEquals(error.cause, lookupFailure);
+  });
+
   it("treats a DNS name that merely starts with 127. as private", async () => {
     // `127.api.prod.internal` is a hostname, not a loopback literal. A `127.`
     // prefix check exposed it -- the same leak the redaction exists to prevent,
--- a/src/platform/adapters/runtime/deno/http-server.ts
+++ b/src/platform/adapters/runtime/deno/http-server.ts
@@ -40,6 +40,21 @@ function isAddressInUseError(error: unknown): boolean {
 }
 
 /**
+ * Whether `error` is a failure to resolve the bind hostname.
+ *
+ * Matched on the getaddrinfo message prefix that Deno surfaces on every
+ * platform ("Name or service not known" on Linux, "nodename nor servname
+ * provided, or not known" on macOS) because the error arrives as a plain
+ * `Error` with no name or code to key on. Deliberately narrow, like
+ * `isAddressInUseError` above: only the lookup signature is claimed, so an
+ * unrelated startup fault keeps its own identity.
+ */
+function isAddressResolutionError(error: unknown): boolean {
+  if (!isErrorAcrossRealms(error)) return false;
+  return error.message.toLowerCase().includes("failed to lookup address");
+}
+
+/**
  * The bind host rendered for an error message, or `undefined` to omit it.
  *
  * A deployment can bind through an internal DNS name, and AGENTS.md:124 lists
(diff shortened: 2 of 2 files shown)


kwakayama:
## Review: 88/100 — solid, well-tested fix; a couple of minor nits

Narrowly-scoped classification for a real Sentry-reported gap, following the existing `#806` AddrInUse precedent closely, with red→green tests and hostname redaction verified.

**Strengths**
- `isAddressResolutionError` is appropriately narrow (keys only on the getaddrinfo message prefix) and is checked *before* `isAddressInUseError`, so it can't misclassify an unrelated startup fault.
- Security: `describeBindHost` correctly keeps private/internal hostnames out of both the message and the error `context`, while still naming loopback/wildcard literals and always reporting the port — matches the AGENTS.md hostname-redaction policy and the sibling AddrInUse path.
- The raw OS error is preserved as `cause`, so Sentry correlation with platform logs is not lost.
- Deliberate avoidance of the "already in use" phrase in the new message is well-reasoned and tested — it prevents the dev server's port-fallback logic from retrying ports for a fault retrying can't fix.
- Good test coverage: an infra-hostname case (host redacted) and a loopback case (host named), both asserting `slug`, message content, and `cause` identity. PR description documents the red→green run.

**Minor concerns / suggestions (non-blocking)**
- `isAddressResolutionError`'s match string (`"failed to lookup address"`) is documented as covering the Linux and macOS getaddrinfo phrasing only; Windows' getaddrinfo wording (e.g. "No such host is known") wouldn't match, so on Windows this would silently fall through to the raw rethrow this PR is trying to eliminate. Worth a follow-up if Windows-hosted Deno serve is a supported target, or at least a comment noting the gap is intentional/scoped to the reported platforms.
- The two `INITIALIZATION_ERROR.create({...})` call sites (resolution-failure vs. address-in-use) duplicate the `safeHost === undefined ? ... : ...` message/context branching almost verbatim. A small shared helper (`buildBindFailureError(detail prefix, hostname, port, cause)`) would remove the duplication — the PR's own comment already explains why the *phrase* must differ, which doesn't require duplicating the surrounding structure too.
- The `suggestion` field on `INITIALIZATION_ERROR` stays generic ("Check initialization requirements and dependencies"); given the PR motivation calls out "no actionable suggestion" as part of the original problem, a DNS-specific detail already covers most of that, so this is a nice-to-have rather than a gap.

Given the fix is small, faithfully mirrors an established pattern in the same file, is well tested, and doesn't touch any other behavior, this is safe to merge once CI is green; the Windows-coverage note is the only thing I'd want acknowledged (even if just "out of scope for this fix").

---

kojiwakayama:
Thanks — acknowledging the one item you wanted addressed explicitly:

**Windows coverage: intentionally out of scope for this fix.** The classifier keys on the phrasing in the reported Sentry group (Linux production, `Name or service not known`) plus the macOS variant, both carried behind Rust's shared `failed to lookup address` prefix. If a Windows-hosted `Deno.serve()` surfaces resolution failures with different phrasing (e.g. `No such host is known`), the fault falls through to the raw rethrow — exactly today's behavior, so no regression, and the veryfront-server image the Sentry group comes from is Linux-only. If Windows-hosted serve becomes a supported deployment target, widening the classifier is a one-line follow-up plus a fixture.

On the two nits: agreed the `safeHost` branching at the two call sites is near-duplicate — deliberately kept inline here since a shared builder would couple two messages whose divergence (the canonical in-use phrase) is itself load-bearing and separately documented; happy to fold both into a helper in a follow-up if a third bind-failure class ever appears. The generic `suggestion` is a registry-level property shared by all `initialization-error` instances; the DNS-specific detail carries the actionable half, as you note.

kwakayama:
## Critical review: 88/100

**Verdict:** No confirmed correctness or security blocker. The reported Linux/macOS DNS-resolution failure is classified narrowly and hostname redaction is covered. Windows-specific resolver wording remains an acknowledged follow-up, not a blocker for the Linux-only deployment scope. Current head: `487a8825ff738a38fdb8a862c7d146fb4a3db814`.

**Validation:** focused Deno tests, format/lint/typecheck, required CI, Sonar, CodeQL, and quality gate pass. No blocker or severity label beyond `codex` is warranted.

kwakayama:
@codex review

kwakayama:
## Review  — Score: 88/100

**Verdict:** Correct and narrow — verified against real Deno 2.6 (`Deno.serve` on an unresolvable host throws synchronously as a plain `Error`, no `name`/`code`, message prefixed `failed to lookup address information:` by Deno's own `resolve_addr` format on every platform, so the prefix match is right and Windows is covered too). Deductions are for a duplicated error-construction block, one factually wrong justification comment, and a missing negative assertion.

### Majors
- `src/platform/adapters/runtime/deno/http-server.ts:280-319` — the two `INITIALIZATION_ERROR.create` blocks are byte-identical apart from the detail prefix (`describeBindHost` call, ternary address, spread-`hostname` context, `cause`), 12 duplicated lines. The stated reason ("divergence is load-bearing") is one string literal, which survives hoisting. Minimal fix: compute `const safeHost = describeBindHost(hostname)`, `const address = safeHost === undefined ? \`port ${port}\` : \`${safeHost}:${port}\`` and one shared `context` object once above both guards, then leave two one-expression `throw`s differing only in the detail prefix. Net diff drops ~10 lines; no behavior change.

### Minors
- `src/platform/adapters/runtime/deno/http-server.ts:277-279` — "the dev server's port fallback must not trigger" is not what happens. `findAvailablePort` (`cli/commands/dev/port-fallback.ts:190`) runs *before* `start()` and never sees this error; the only consumer is `cli/commands/dev/command.ts:126`, which relabels a matching start error to `PORT_IN_USE` and advises `--port`. Fix the sentence to name that actual consequence (misclassified as `PORT_IN_USE`, wrong `--port` suggestion) or delete it.
- `src/platform/adapters/runtime/deno/http-server.test.ts:376` — the invariant the comment above calls deliberate (message must not carry the canonical in-use phrase) is untested here, while the sibling case asserts exactly that at `http-server.test.ts:344`. Add `assertEquals(error.message.toLowerCase().includes("already in use"), false);`.

### Nits
- `src/platform/adapters/runtime/deno/http-server.ts:284` — for the redacted case the message is `Bind hostname could not be resolved: port 4321`; the port is the one input that did not fail to resolve, so it reads as if the port was looked up. `Bind hostname could not be resolved (bind port ${port})` says the same without the mis-parse.

### CI
All required checks green (lint/format/typecheck, 8 coverage shards, CodeQL, Sonar, codecov patch); only "Automated review" pending.

kwakayama:
## Merge gate: not merged — score below threshold

Final review score **88/100**, auto-merge requires **>= 90**. Nothing else blocks: `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, all required checks SUCCESS (rest skipped), 0 unresolved review threads, Codex/Sonar/CodeQL clean.
Remaining gap is the acknowledged Windows resolver-wording follow-up in `isAddressResolutionError` — non-blocking, but it keeps the score at 88.
Labels: `blocked` removed (no CI failure, no conflict), `needs-human-input` applied. A maintainer can merge manually or raise the score.

PR #4299, opened 2026-08-29
fix(agent): prefer the terminal-run rejection slug over its English detail

kojiwakayama:
Refs https://github.com/veryfront/veryfront-issue-inbox/issues/757

The runtime half of #757, shippable independently: `isTerminalRunConversationRunAppendError` and `isIgnorableConversationRunAppendError` now share one predicate that prefers `slug === "terminal-run-append-rejected"` (the identifier the issue's final analysis settled on — same pattern the 404 deleted-run branch already uses) and keeps the exact-detail match as a fallback. Until the veryfront-api change registers the slug, the new branch never matches and behavior is byte-identical; after it, a reworded detail no longer degrades the terminal classification to retries.

**Deploy-order safe by construction** — this is the "runtime prefers the slug, string as fallback" step the issue prescribes, and it works correctly both before and after the api-side registration (which lives in veryfront-api and is not part of this PR; noted on the issue).

Red → green: new test constructs a 400 with the distinct slug and a *reworded* detail (predicates must classify terminal by slug alone — red before: both returned false) and a 400 with only `validation-failed` + reworded detail (must miss, never misfire — green both before and after). `durable-append-errors.test.ts` 5 steps and consumer suite `durable.test.ts` 50 steps pass, exit 0; fmt/lint/check clean.


kwakayama:
**Score: 93/100 — excellent.** Small, correct, well-tested fix that follows the existing pattern in this file exactly.

- **Correctness**: `isTerminalRunAppendRejection` widens the 400 match from exact-detail to `slug === "terminal-run-append-rejected" || detail === TERMINAL_RUN_APPEND_REJECTION_DETAIL` — a strict superset of the old condition, so it's provably non-regressive, and it mirrors the 404 branch's existing slug-based match (`DELETED_RUN_APPEND_REJECTION_DETAIL`) instead of inventing a new idiom.
- **Test coverage**: the new case directly targets the misfire hazard — a reworded detail *with* the new slug must still classify terminal, and a reworded detail *without* it (different slug) must still miss rather than misclassify. That's exactly the red/green pair this change needs.
- **Deploy-order safety**: correctly argued and structurally true — the slug branch is inert until the API starts emitting it, so this ships ahead of the API-side change with byte-identical current behavior.
- **Documentation**: PR description, inline comments, and commit message are all consistent with the diff and explain *why*, not just *what* — good trail back to issue #757.
- Minor nit (non-blocking): `isTerminalRunAppendRejection` re-checks `status === 400` even when called from `isIgnorableConversationRunAppendError`, which already guards on that status a few lines up — harmless duplication, but could be trimmed if the helper is only ever called from 400-guarded contexts.

Nothing here blocks merge; the branch shows as blocked, presumably on review/approval rather than a code issue.


---

Issue #880, opened 2026-04-07
perf: agent perf summary extension


ariskemper:
## PR 4: Agent Perf Summary Extension

**Goal:** Extend `_perf_lib.ts` `summarizePerf()` to extract and report the new agent/streaming metrics so they appear in perf comparison reports.

**Files:**
- Modify: `benchmarks/_perf_lib.ts` (add agent metric extraction to `summarizePerf()` and `summarizePerfOverview()`)

## What to build

Add these metrics to `summarizePerf()`:
- `agent_single_ttfe_ms`, `agent_single_total_ms`, `agent_single_events_per_sec`
- `agent_tools_ttfe_ms`, `agent_tools_total_ms`, `agent_tools_p95_ms`
- `chat_stream_events_per_sec`, `chat_stream_p95_ms`
- `agent_concurrent_p99_ms`, `agent_concurrent_rps`

Add cold/warm improvement tracking to `summarizePerfOverview()`:
- `agent_single_ttfe_improvement_from_cold_ms` / `_pct`
- `agent_tools_total_improvement_from_cold_ms` / `_pct`
- `agent_concurrent_p99_improvement_from_cold_ms` / `_pct`

## Acceptance criteria

- [ ] `deno check benchmarks/_perf_lib.ts` passes
- [ ] Agent metrics appear in perf summary output when agent scenario results are present
- [ ] Cold/warm improvement percentages calculate correctly for agent metrics
- [ ] Existing HTTP/rendering metrics are unaffected

## Context

Depends on #879. See `plans/server-side-performance-implementation.md` PR 4.

kojiwakayama:
Closing as stale. This issue references the old benchmark implementation plan and benchmark paths (`benchmarks/`, `plans/server-side-performance-implementation.md`) that are not present on current main. Reopen or file a new issue with current paths and acceptance criteria if this work still applies.

PR #4308, opened 2026-08-29
fix(embedding): retry disappearing local-index leases

kojiwakayama:
## Root cause

The failed coverage log for inbox issue #880 shows `readObservation` listing a lease and then receiving `NotFound` when it calls `lstat` on that lease. This is an expected snapshot race when the current writer releases its lock directory, but the acquisition path converted it into a fatal `LocalJsonStoreLockError`.

## Change

- Treat a lease that disappears between `readDir` and `lstat` as an invalid observation.
- Let lock acquisition retry that observation.
- Keep ownership checks fail-closed because a null observation still fails the current-owner comparison.
- Add a deterministic regression that removes the lock directory at the observed `lstat` boundary.

## Verification

- `deno task test:file src/embedding/rag-store.test.ts`
- The same focused command repeated 10 times
- `deno fmt --check src/embedding/local-json-store-lock.ts src/embedding/rag-store.test.ts`
- `deno lint src/embedding/local-json-store-lock.ts src/embedding/rag-store.test.ts`
- `deno check src/embedding/local-json-store-lock.ts src/embedding/rag-store.test.ts`
- `git diff --check`

Refs https://github.com/veryfront/veryfront-issue-inbox/issues/880

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved lock handling when a lease disappears during verification.
  - Lock acquisition now retries instead of failing due to a transient, inconsistent lock state.
  - Ownership checks continue to fail safely when lock information cannot be verified.

- **Tests**
  - Added coverage for retry behavior when a lease disappears during lock inspection.

### DIFF
--- a/src/embedding/local-json-store-lock.ts
+++ b/src/embedding/local-json-store-lock.ts
@@ -199,10 +199,11 @@ async function readObservation(
         leaseInfo = await lstat(join(lockDirectory, entryName));
       } catch (error) {
         if (isCanonicalNotFoundError(error)) {
-          throw new LocalJsonStoreLockError(
-            "RAG store lock lease changed while ownership was being inspected",
-            { cause: error },
-          );
+          // The directory listing and lease metadata are one observation. If a
+          // listed lease disappears between them, discard the snapshot. Lock
+          // acquisition retries, while ownership checks still fail closed on
+          // the null observation.
+          return null;
         }
         throw error;
       }
--- a/src/embedding/rag-store.test.ts
+++ b/src/embedding/rag-store.test.ts
@@ -182,6 +182,41 @@ describe("ragStore", () => {
     });
   }
 
+  it("retries when a listed lock lease disappears before lstat", async () => {
+    await withTempDir(async (tempDir) => {
+      const storagePath = join(tempDir, "data", "index.json");
+      const lockDirectory = `${storagePath}.veryfront-rag.lock`;
+      const leasePath = join(lockDirectory, `${crypto.randomUUID()}.lease`);
+      await Deno.mkdir(lockDirectory, { recursive: true });
+      await Deno.writeTextFile(join(lockDirectory, "owner.json"), "{");
+      await Deno.writeTextFile(leasePath, "live\n");
+
+      const lstatDescriptor = Object.getOwnPropertyDescriptor(Deno, "lstat");
+      assert(lstatDescriptor !== undefined);
+      const originalLstat = Deno.lstat.bind(Deno);
+      let changed = false;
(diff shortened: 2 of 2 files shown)


kwakayama (commented):
## Review: 90/100 — solid, narrowly-scoped race fix with a deterministic regression test

**Summary:** Correctly converts a benign snapshot race (a listed lease disappearing between `readDir` and `lstat`) from a fatal `LocalJsonStoreLockError` into a `null` observation that the acquisition loop retries, while leaving ownership checks fail-closed.

**Strengths**
- The fix is minimal and precisely targeted: only the ownerless/ambiguous-ownership lease-discovery loop in `readObservation` changes, where a partial scan can't be trusted for staleness math anyway. It doesn't touch the sibling `owner !== null` branch, which is correctly asymmetric (a missing lease there just means "no lease," since ownership is already known).
- I traced every caller of `readObservation` (`clearInterruptedTransition`, `acquireNativeLock`'s stale-check, `tryRecoverStaleLock`, `releaseOwnedLock`, `assertOwned`) — `null` is handled safely everywhere: acquisition retries, and ownership-verifying call sites throw via `observed?.owner?.token !== token` (or equivalent), so the "fail closed" claim in the description checks out rather than being asserted.
- The regression test reproduces the exact race deterministically by monkey-patching `Deno.lstat` to throw `NotFound` once for the targeted lease path and removing the lock directory underneath it, then asserts the operation completes instead of throwing. That's a good way to pin down a timing bug without relying on real concurrency/flakiness.
- Comment added at the fix site explains *why* discarding the whole snapshot (vs. skipping just that one lease) is correct — the loop is computing a newest-lease timestamp for staleness, so a partial view would be misleading.
- Small, self-contained diff (2 files, +40/-4); verification steps in the description (fmt/lint/check + the focused test run 10x) are reasonable for a race-condition fix.

**Minor / non-blocking**
- CI (coverage shards, lint, typecheck, etc.) was still `in_progress` at review time — please confirm it's green before merge, and note the PR is currently `mergeable_state: blocked`, which may just be pending required checks/reviews rather than an actual conflict.
- PR title ("retry disappearing lock leases") and description heading ("retry disappearing local-index leases") differ slightly — cosmetic only.
- Worth a one-line callout (in the description or a code comment) on the extreme case: if this snapshot race were to repeat continuously, acquisition would still bail out via the existing `LOCK_ACQUIRE_TIMEOUT_MS` (10s) rather than retrying forever — worth confirming that's the intended fallback for a persistently racy directory, though this is likely a non-issue in practice.

Nothing here blocks merge; happy to approve once CI is green.

---

kojiwakayama:
@codex review

PR #4313, opened 2026-08-30
fix(security): break the log-injection taint path CodeQL still sees (alert 333)

kwakayama:
Fixes [code scanning alert 333](https://github.com/veryfront/veryfront-code/security/code-scanning/333) (`js/log-injection`, error severity) at `scripts/postinstall.js:140`.

## Why the alert was still open

#4266 already hardened this log line: it stripped every control character and line separator from `error.message` before logging. That is correct at runtime, but CodeQL never saw it. Its `js/log-injection` sanitizer is narrow:

```ql
class StringReplaceSanitizer extends Sanitizer {
  StringReplaceSanitizer() {
    exists(string s | this.(StringReplaceCall).replaces(s, "") and s.regexpMatch("\\n"))
  }
}
```

It only recognises a **global replacement of a literal CR/LF by the empty string**. A codepoint *range* (`[-...]`) and a *non-empty* replacement (`" "`) both fall outside it, so the taint path from the HTTPS error in `scripts/postinstall-lib.js` to `console.warn` stayed unbroken and the alert stayed open.

## The change

`sanitizeLogValue` in `scripts/postinstall-lib.js` now removes line terminators in the recognised form first, then keeps the existing broader hardening:

```js
String(value)
  .replace(/[\r\n]/g, "")                                       // recognised sanitizer
  .replace(/[--  ]+/g, " ")
  .replace(/\s{2,}/g, " ")
  .trim()
  .slice(0, 500);
```

Removing rather than substituting means a multi-line message loses its breaks. The only multi-line message that reaches here is the checksum mismatch, whose continuation lines are indented, so it still reads as separate fields — covered by a test.

Moving the sanitizer into `postinstall-lib.js` is what makes it testable: `postinstall.js` runs `install()` on import, so nothing defined in it can be unit tested.

## Evidence

CodeQL **2.26.4** — the exact version that raised alert 333 — run locally over `scripts/postinstall.js` + `scripts/postinstall-lib.js` with `javascript-security-and-quality.qls`:

| tree | `js/log-injection` results |
|---|---|
| `origin/main` (79a3c51) | **1** — `scripts/postinstall.js:140`, matching alert 333 exactly |
| this branch | **0** |

No new findings of any rule were introduced (3 total results before, 2 after — the 2 remaining are pre-existing `js/file-access-to-http`, untouched here).

7 new unit tests cover line-terminator removal, Unicode line/paragraph separators, terminal escapes and NUL, whitespace collapsing, the length cap, the checksum-mismatch message staying readable, and non-string input.

## `scripts/postinstall-lib.test.js` was not being run

The file existed with 15 tests and no runner: `scripts/` is excluded from the Deno suite planner (`UNPLANNABLE_UNIT_ROOTS`), and `deno task test:scripts` — the task that exclusion's comment points at — did not list it. So the regression test added here would never have executed. It is now registered in `test:scripts`, which needed `TMPDIR,TMP,TEMP` added to that task's `--allow-env` for the pre-existing `os.tmpdir()` calls.

## Scope note

`scripts/postinstall.js` is not currently shipped: `veryfront@0.1.1253` on npm declares no `postinstall` hook and contains no `scripts/` directory, despite `scripts/README.md` describing the file as "copied into npm package". Whether that wiring should be restored is a separate question and is not addressed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved post-installation error logging by preventing line breaks, control characters, and terminal escape sequences from disrupting output.
  * Ensured error messages remain readable and are limited to a safe length.

* **Tests**
  * Added coverage for log sanitization, including Unicode separators, whitespace, long messages, checksum errors, and non-string values.

* **Chores**
  * Updated test permissions to support temporary-directory environment variables.


kwakayama (commented):
## Score: 85/100 — good, one behavior-change worth a second look

Tight, well-evidenced fix for a real CodeQL false-negative-turned-negative, with strong tests and a nice side-catch (the test file wasn't even being run). Not blocking, but one readability regression deserves attention before merge.

**Strengths**
- Root-causes *why* #4266's fix didn't satisfy CodeQL's `js/log-injection` sanitizer (narrow `replace(/\n/, "")`-shaped pattern) instead of just silencing the alert, and backs it with a reproducible before/after CodeQL run (1 → 0 results, no new findings elsewhere).
- Moving `sanitizeLogValue` into `postinstall-lib.js` is the right call — `postinstall.js` runs `install()` on import, so this was the only way to make it unit-testable at all.
- 7 new tests cover the meaningful edge cases (CRLF, Unicode separators, terminal escapes/NUL, whitespace collapsing, length cap, non-string input, the checksum-mismatch message).
- Good catch that `scripts/postinstall-lib.test.js` (15 existing tests) had no runner — `scripts/` is excluded from the Deno suite planner and `test:scripts` didn't list the file, so this regression test would never have executed. Registering it (plus the required `TMPDIR,TMP,TEMP` env allowlist) is a real fix beyond the stated scope.
- Commit message and PR description are unusually thorough and honest, including the scope note that `postinstall.js` isn't currently shipped in the npm package.

**Concern — worth fixing before merge**
- The new order (`replace(/[\r\n]/g, "")` deleting line terminators outright, *then* replacing other control chars with `" "`) changes runtime behavior, not just satisfies CodeQL: a `\n`/`\r` adjacent to non-whitespace on both sides now vanishes with **no** separator, jamming words together. This is directly visible in your own first test: `"HTTP 500\n✅ Checksum verified\r\n[INFO] installed"` → `"HTTP 500✅ Checksum verified[INFO] installed"`. The PR body acknowledges this and argues the only multi-line message reaching this function today (checksum mismatch) has indented continuation lines that happen to save it — but that's a fragile invariant for a general-purpose `sanitizeLogValue` export, not a guarantee.
  - Consider satisfying CodeQL's literal-pattern sanitizer without losing the space: e.g. `.replace(/[\r\n]/g, " ").replace(/[\r\n]/g, "")` — the second call is a runtime no-op (nothing left to match) but is still the literal `replace(/\n/, "")`-shaped node CodeQL's `StringReplaceSanitizer` looks for, so the taint path still breaks while spacing is preserved. Worth a quick local CodeQL re-run to confirm before adopting.
- Minor: the fix is intentionally tuned to CodeQL 2.26.4's specific sanitizer AST shape rather than a more robust general pattern — reasonable given the constraint (a query you don't control), but worth a one-line comment noting this may need revisiting if the query changes in a future CodeQL version.

Neither point blocks merging as-is; the security property (no real newline reaches the log) holds either way. The suggested reordering above would remove the readability regression essentially for free.

---

kwakayama:
## Codex review

Reviewed the exact PR head `20322b9c2cfb3829bb0b7ec3913b00ac50a27e5c` against base `79a3c51d73b435f1a4944e38f5dbd632f27f5551`.

### Findings

None actionable. The follow-up preserves separators between adjacent words while retaining the CodeQL 2.26.4-recognized sanitizer shape. The call site remains sanitized before `console.warn`, and the test registration covers the new regression suite.

### Verification

- `node --test scripts/postinstall-lib.test.js`: 22 passed.
- Focused Deno script test: 22 steps passed.
- `git diff --check`: passed.
- The broader `deno task test:scripts` was attempted; 54 tests passed, while unrelated suites failed at the existing `node:util/types` compatibility guard because this checkout has Deno 2.9.4 and the repository pins Deno 2.7.7.
- CodeQL was not rerun locally because the CLI is unavailable; the PR's prior CodeQL 2.26.4 evidence remains the relevant security validation, and CI is running for this exact head.

Verdict: APPROVE

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 20322b9c2cfb3829bb0b7ec3913b00ac50a27e5c
Score: 95/100
Actionable-Findings: 0
Verdict: APPROVE

kwakayama:
## Claude review

Reviewed the exact PR head `20322b9c2cfb3829bb0b7ec3913b00ac50a27e5c` against base `79a3c51d73b435f1a4944e38f5dbd632f27f5551`.

### Findings

None actionable. The two-pass CR/LF design preserves word separators while retaining the CodeQL 2.26.4 sanitizer shape. Control-character coverage matches repository patterns and includes Unicode line and paragraph separators. The call site sanitizes remote-controlled error text before logging, and `deno.json` registers the script tests with the required temporary-directory environment permissions.

### Verification

Reviewed the changed implementation, call site, test suite, test registration, and existing sanitizer patterns. The seven focused cases cover log forging, Unicode separators, terminal escapes, whitespace, length limits, checksum readability, and non-string coercion.

Verdict: APPROVE

Review-Gate:
Reviewer: Claude
Reviewed-SHA: 20322b9c2cfb3829bb0b7ec3913b00ac50a27e5c
Score: 95/100
Actionable-Findings: 0
Verdict: APPROVE

kwakayama:
## Review follow-up

Addressed the readability concern from the earlier review in commit `20322b9c2cfb3829bb0b7ec3913b00ac50a27e5c`: CR/LF are first converted to spaces so adjacent words remain separated, then the CodeQL 2.26.4-recognized empty replacement pass preserves the taint barrier. The regression assertion now covers the adjacent-word case, and the focused Deno and Node tests pass.

The exact-head gate is clean: no unresolved review threads, CodeQL green, and all required CI checks passed.

PR #4315, opened 2026-08-30
ci: align format check with local task

kojiwakayama:
## Summary

- make the CI format matrix leg call the repository fmt:check task
- add a workflow contract assertion that prevents command drift

## Red-green TDD

RED:
- the setup-deno workflow contract failed because CI ran bare deno fmt --check

GREEN:
- focused workflow contract: 11 steps passed
- formatter task: all configured source, scripts, build, and codemod paths passed
- CI workflow contracts: 20 steps passed
- deno task lint:ci
- git diff --check

Closes veryfront/veryfront-issue-inbox#884

Parent: veryfront/veryfront-issue-inbox#881 item 1.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -78,7 +78,7 @@ jobs:
         run: |
           case "${{ matrix.check }}" in
             test-layout) deno task test:layout ;;
-            format) deno fmt --check ;;
+            format) deno task fmt:check ;;
             # The chain lives in the lint:ci task so a contributor can run this
             # shard locally instead of learning what it covers from a red CI run.
             lint) deno task lint:ci ;;
--- a/scripts/ci/setup-deno-workflow.test.ts
+++ b/scripts/ci/setup-deno-workflow.test.ts
@@ -850,6 +850,11 @@ jobs:
       /^\s*lint\) deno task lint:ci ;;$/m,
       "the lint shard must delegate to the lint:ci task",
     );
+    assertMatch(
+      String(ciRunStep.run),
+      /^\s*format\) deno task fmt:check ;;$/m,
+      "the format shard must delegate to the fmt:check task",
+    );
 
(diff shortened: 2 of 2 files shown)


kwakayama:
## Code Review — Score: 94/100 (Excellent)

Tiny, correct CI-config fix that closes a real coverage gap, with a regression test to prevent drift.

**Strengths:**
- The change is precise: swapping `deno fmt --check` for `deno task fmt:check` isn't just cosmetic — the top-level `deno.json` `fmt.include` only covers `src/`, `cli/`, `react/`, `templates/`, so the bare CLI invocation was silently skipping format checks on `scripts/test/`, `scripts/build/*`, and `scripts/codemods/` (which use separate `--config` files). This PR actually widens CI coverage, not just aligns naming.
- Includes a targeted regression test (`setup-deno-workflow.test.ts`) asserting the exact `format) deno task fmt:check ;;` line in the workflow YAML, consistent with the existing pattern used for the `lint` shard right above it — good consistency with established conventions.
- Minimal, single-purpose diff (1 line prod change, 5 line test), easy to review and low risk.
- PR description documents red/green TDD evidence and links the originating issue — good traceability.

**Minor notes (non-blocking):**
- The new assertion only checks that the workflow string matches the task name; it doesn't (and can't, at this test's level) verify that `fmt:check` itself actually covers the previously-missed paths — that's implicitly relying on `deno.json` staying correct, which is fine but worth knowing if `fmt:check` is ever redefined narrower in the future.
- Might be worth a one-line PR description callout that this change *increases* format-check coverage (scripts/build, scripts/test, scripts/codemods) rather than framing it purely as a drift-alignment/naming fix — reviewers skimming the diff could otherwise assume this is a no-op rename.

No functional, security, or test-coverage concerns. Good to merge once CI is green.

---

kojiwakayama:
@codex review

PR #4319, opened 2026-08-30
ci: check proxy lock freshness in lint

kojiwakayama:
## Summary

- regenerate the committed proxy dependency lock at the start of lint:ci
- fail lint:ci when the regenerated lock differs from the committed file
- add a CI task contract regression for both commands

## Red-green TDD

RED:
- setup-deno CI contract failed because lint:ci omitted proxy-lock regeneration

GREEN:
- focused workflow contract: 11 steps
- proxy-lock regeneration and scoped diff
- deno task lint:ci
- git diff --check

Closes veryfront/veryfront-issue-inbox#886

Parent: veryfront/veryfront-issue-inbox#881 item 17.


kwakayama:
## Code Review: 84/100 — Good, minor suggestions

Small, focused CI hardening change that closes a real gap: `lint:ci` previously didn't catch a stale `scripts/build/proxy-deno.lock`, so a drifted lock could pass lint and only fail later in the binary-build jobs.

**Strengths**
- Minimal, surgical diff (2 files, +11/-1) that does exactly what the title/description say — no scope creep.
- Reuses the existing `build:proxy-lock` task and the same `git diff --exit-code -- scripts/build/proxy-deno.lock` pattern already used in the proxy/binary CI jobs (`.github/workflows/cicd.yml`), so the check is consistent with established convention rather than inventing a new mechanism.
- Placed first in the `lint:ci` chain for fast failure, and comes with a regression test (`setup-deno-workflow.test.ts`) asserting the task string, matching this file's existing "contract test on the CI task definition" style.
- Clear PR description with explicit red/green TDD narrative and issue linkage.

**Concerns**
- The `Verify proxy dependency lock is current` steps in the binary-build jobs run **after** a `deno task build:prepare` step that retries 3x specifically because the registry is known to be flaky ("error reading a body from connection", per the comment in `cicd.yml`). The lint job has no equivalent warm-up/retry before `build:proxy-lock`, and its Deno cache is keyed off `deno.lock`/`deno.json`, not `proxy-deno.lock` — so on a cache miss this new step is a bare, non-retried network call inside a long lint chain, more exposed to that same flakiness than the jobs it borrows the pattern from. Consider wrapping it with the same retry helper, or confirming cache overlap is high enough in practice that this isn't a real risk.
- The added test only asserts that the `lint:ci` string contains the two substrings — it doesn't exercise actual staleness detection (e.g. a fixture where the lock is intentionally stale). That's consistent with the rest of the file's style, so not a blocker, but it means the "regression" is really just "the wiring is present," not "the check works."

Neither point is blocking; both are worth a quick look before merge.

---

kojiwakayama:
@codex review

PR #4329, opened 2026-08-30
ci: cross-compile the darwin binaries on Linux runners

kwakayama:
## Why

The two macOS legs of `build-binaries` bill at 10x the Linux per-minute rate (7,608 macOS minutes = $472 gross in August) and do nothing macOS-specific: no codesigning, no notarization, no macOS smoke test. `deno compile --target` cross-compiles by downloading the matching denort runtime and ad-hoc signing the output, so the same compile on `ubuntu-latest` costs ~10% as much.

## What

The `aarch64-apple-darwin` and `x86_64-apple-darwin` matrix legs move from `macos-latest` to `ubuntu-latest`. Nothing else changes: artifact names, release assets, SHA256SUMS inputs, and the Homebrew formula placeholders are byte-identical.

## Verification

- `build-binaries` is main-only, so the first push to main after merge is the proving run; I'll watch both darwin legs there and revert if either fails.
- `compile-binary.test.ts` (20 tests) passes; the compile script has no host-OS assumptions (no `Deno.build.os` branching).
- Windows keeps its own runner and `continue-on-error`, unchanged.

https://claude.ai/code/session_0145mkV9Y8k6cWjtQt5H676r

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -873,12 +873,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-latest
+          # The darwin binaries cross-compile on Linux: deno compile --target
+          # downloads the matching denort runtime and ad-hoc signs the output,
+          # and no step here codesigns, notarizes, or smoke-tests on macOS.
+          # macOS runners bill at 10x the Linux rate for the same compile.
+          - os: ubuntu-latest
             target: aarch64-apple-darwin
             name: veryfront-macos-arm64
             entrypoint: cli/main.ts
             profile: full
-          - os: macos-latest
+          - os: ubuntu-latest
             target: x86_64-apple-darwin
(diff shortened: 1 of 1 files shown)


kwakayama:
@codex review

kwakayama:
Reverted in #4337.

The post-merge proving run passed, but the artifact regressed: `veryfront-macos-arm64` grew 997,511,522 → 1,141,496,162 bytes (+14%) because `deno compile` embeds every locked npm package and the Linux host also resolves its own native ones (`@esbuild/linux-x64`, `lightningcss-linux-x64-gnu`/`-musl`, `@kreuzberg/node-linux-x64-gnu`). Control: `veryfront-linux-x64` is byte-identical across both runs, so no code change explains it.

Not broken (the darwin packages are still there), but 144 MB of dead weight for every macOS user. Reland behind a target-scoped lock for the full profile.

PR #4342, opened 2026-08-31
fix(ci): use HTTPS for Chromium apt mirrors

kwakayama:
## Summary

SonarCloud marks `main` as failing `githubactions:S5332` because the shared Chromium install action rewrites Ubuntu APT sources through clear-text HTTP URLs.

This PR matches either existing protocol and rewrites both Ubuntu mirrors to HTTPS, leaving no clear-text URL literal in the action. The CI contract test now locks that security invariant.

## Verification

- Deno 2.7.7: `scripts/ci/setup-deno-workflow.test.ts` passed, 2 tests and 11 steps
- Formatting and lint checks passed for the changed test
- Shell replay confirmed both HTTP and HTTPS source entries rewrite to `https://archive.ubuntu.com/ubuntu`
- `git diff --check` passed

### DIFF
--- a/.github/actions/install-chromium/action.yml
+++ b/.github/actions/install-chromium/action.yml
@@ -17,9 +17,9 @@ runs:
     - name: Configure apt sources, retries, and mirrors
       shell: bash
       run: |
-        sudo sed -i \
-          -e 's|http://azure.archive.ubuntu.com/ubuntu|http://archive.ubuntu.com/ubuntu|g' \
-          -e 's|http://security.ubuntu.com/ubuntu|http://archive.ubuntu.com/ubuntu|g' \
+        sudo sed -E -i \
+          -e 's|[[:alpha:]]*://azure.archive.ubuntu.com/ubuntu|https://archive.ubuntu.com/ubuntu|g' \
+          -e 's|[[:alpha:]]*://security.ubuntu.com/ubuntu|https://archive.ubuntu.com/ubuntu|g' \
           /etc/apt/sources.list /etc/apt/sources.list.d/ubuntu.sources 2>/dev/null || true
         {
           echo 'Acquire::ForceIPv4 "true";'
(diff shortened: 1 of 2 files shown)


kwakayama:

A minimal, well-targeted CI fix that closes the SonarCloud `S5332` clear-text-HTTP finding and locks the fix with a regression test.

**Strengths:**
- **Correctly fixes the actual bug, not just the symptom.** The old `sed` only normalized the `azure.archive.ubuntu.com` host to `archive.ubuntu.com` while leaving the scheme as `http://` — it never actually eliminated the clear-text URL Sonar was flagging. The new `[[:alpha:]]*://` → `https://...` pattern matches any scheme (http, https, or bare) and rewrites unconditionally to `https://`, which is the real fix.
- **Right layer for the fix.** Applied once in the shared `install-chromium` composite action used by all Chromium-dependent jobs, rather than patched per-caller.
- **Test coverage matches the change.** The new assertions pin both rewritten `-e` patterns verbatim and assert `http://` is absent from the resulting `run` string — a genuine regression guard against a Sonar-flagging string sneaking back in, not just a happy-path check.
- **Small, self-contained diff** (2 files, +24/-3) with a clear commit message and a PR description that states the problem, the fix, and manual verification steps.

**Minor observations (non-blocking):**
- `[[:alpha:]]*` will also match a bare `://` with no scheme prefix (zero occurrences of the character class), which is harmless here since real `sources.list` entries always carry a scheme, but worth knowing if this pattern is ever reused elsewhere.
- Several CI checks were still `in_progress` at review time (coverage shards, lint, integration tests) — worth confirming they land green before merge, though none are expected to be affected by a CI-config-only change.

No functional, security, or test-coverage concerns. Approve once outstanding checks finish.

---

kojiwakayama (approved):

PR #4345, opened 2026-08-31
release: v0.1.1255

kwakayama:
Promotes `main` from `0.1.1255-rc` to a clean `0.1.1255` so the release job publishes to npm `--tag latest`.

## Why now

veryfront-code#4340 (`documents.readonly` removal) and #4224 (full `drive` removal) are both merged but reach **no environment** while `main` carries the `-rc` suffix — `version-check` publishes suffixed versions as RC only. Both are prerequisites for the Google OAuth verification reply (veryfront/veryfront-issue-inbox#279).

## Files

Mirrors the v0.1.1254 release: `deno.json`, `src/utils/version-constant.ts`, and the regenerated `src/html/hydration-script-builder/hydration-runtime.generated.ts` (it embeds the version string). `templates/manifest.generated.ts` needed no change this time — confirmed by the staleness check below.

## Verified

- `0.1.1255` is above both `main` (`0.1.1255-rc`) and npm `latest` (`0.1.1254`), so the release job will not no-op
- `deno task typecheck` — exit 0, no `is stale` output

## Summary by CodeRabbit

* **Release**
  * Promoted the package from release candidate to stable version 0.1.1255.
  * Updated the displayed version information to consistently show 0.1.1255.


kwakayama:
## Code Review — Score: 96/100 (excellent)

Trivial, mechanical release PR — this is essentially config, not logic, so it's scored on correctness/process rather than design or test coverage.

**Strengths**
- Change is exactly what it claims to be: `deno.json`, `src/utils/version-constant.ts`, and the generated `hydration-runtime.generated.ts` all bump `0.1.1255-rc` → `0.1.1255` in lockstep — verified the embedded version string in the generated file directly and it matches.
- `templates/manifest.generated.ts` was correctly left untouched, and the PR body explains why (staleness check found no diff) rather than silently omitting it — good transparency for a generated-file release PR.
- PR description gives clear rationale (unblocks two merged PRs — `documents.readonly`/`drive` scope removal — from reaching npm `latest` because `main` is stuck on an `-rc` suffix) and links the downstream motivation (OAuth verification reply).
- Author states they ran `deno task typecheck` (exit 0, no "is stale" output) before opening — good practice for a PR that touches a generated file.
- Mirrors the prior `v0.1.1254` release pattern per the description, so it's following established repo convention rather than improvising.

- No independent way to verify "above npm `latest`" claim from the diff alone — that's an external-state assertion, not really reviewable in-PR; presumably enforced by the `version-check`/publish CI job.
- CI was still mostly `in_progress` at review time (coverage shards, tests, lint, typecheck) — worth confirming green before merge, standard for any PR but doubly so for a release cut.

Nothing here warrants a lower score — no code logic changed, diff is minimal and self-consistent, and the description does the legwork of explaining *why* each file did or didn't change.

---

kwakayama:
@codex review

PR #4360, opened 2026-09-01
fix(release): transform materialized project sources locally

kojiwakayama:
Fixes veryfront/veryfront-issue-inbox#924

## What changed

- use the native runtime filesystem for transforms after the release source is materialized
- retain the request-scoped adapter for control-plane source retrieval and publication
- cover an MDX route with an aliased project component import against a rejecting remote filesystem adapter

## RED-GREEN TDD

RED: the regression build reached ready state but attempted to read the materialized page and component through the remote adapter.

GREEN: the same build publishes both modules without any remote filesystem reads.

## Verification

- deno task test:file src/release-assets/scaffolded-project-build.test.ts
- deno task test:file src/release-assets
- deno task typecheck
- deno task lint:ci
- deno task build:npm
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release builds to correctly resolve aliases and relative imports from the materialized project files.
  * Prevented unnecessary remote file requests during release asset generation.
  * Ensured project dependencies are included correctly in the generated release manifest.

* **Tests**
  * Added coverage verifying builds succeed using local release files and include all expected project modules.

### DIFF
--- a/src/release-assets/build-executor.ts
+++ b/src/release-assets/build-executor.ts
@@ -26,6 +26,7 @@ import { VERSION } from "#veryfront/utils/version.ts";
 import { createFileSystem, isNotFoundError, realPath } from "#veryfront/platform/compat/fs.ts";
 import { getOsType } from "#veryfront/platform/compat/process.ts";
 import type { RuntimeAdapter } from "#veryfront/platform/adapters/base.ts";
+import { getLocalAdapter } from "#veryfront/platform/adapters/registry.ts";
 import { STAT_OPERATION_EXTENSION_PRIORITY } from "#veryfront/platform/adapters/fs/veryfront/extension-priority.ts";
 import {
   dirname,
@@ -864,6 +865,22 @@ function releaseLogicalPathFromMaterializedPath(
 export const releaseAssetBuildInternals = Object.freeze({
   releaseLogicalPathFromMaterializedPath,
 });
+
+/**
+ * Bind transforms to the immutable source snapshot materialized for this build.
+ *
+ * Hosted execution enters through a request-scoped Veryfront API adapter. That
+ * adapter remains the authority used to fetch the release once, but absolute
+ * paths beneath `tempDir` belong to the build-owned local snapshot. Sending
+ * those paths back to the project API returns 404 and leaves aliases or
+ * relative imports unresolved during finalization.
+ */
+async function createMaterializedReleaseTransformAdapter(
+  adapter: RuntimeAdapter,
+): Promise<RuntimeAdapter> {
+  const localAdapter = await getLocalAdapter();
+  return { ...adapter, fs: localAdapter.fs };
+}
 
 /**
  * Import edges of an App Router server module, read from its authored source.
@@ -3200,6 +3217,7 @@ async function runBuildInner(
   // Bytes are held per-hash only until uploaded, then dropped (M3).
   const pendingBytes = createPendingAssetStore();
   const knownPaths = new Set(sourceByPath.keys());
+  const transformAdapter = await createMaterializedReleaseTransformAdapter(input.adapter);
   const vendorDependencies = input.dependencyMode === "immutable";
   const configuredVendor = input.vendorHttpImports;
   const vendorHttpImports = vendorDependencies ? configuredVendor : undefined;
@@ -3329,7 +3347,7 @@ async function runBuildInner(
           tempDir,
           knownPaths,
           projectImportAliases,
-          input.adapter,
+          transformAdapter,
         );
         serverModuleImports.set(logicalPath, imports);
         // Everything a server module reaches is server code by default. A
@@ -3369,7 +3387,7 @@ async function runBuildInner(
     }
     let code: string;
     try {
-      code = await transform(source, sourceFile, tempDir, input.adapter, {
+      code = await transform(source, sourceFile, tempDir, transformAdapter, {
         projectId: input.projectId,
         dev: false,
         ssr: false,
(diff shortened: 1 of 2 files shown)


kojiwakayama:
@codex review

kwakayama:
## Automated review: 85/100 — Good, minor suggestions

Solid, well-scoped bugfix with a regression test that actually catches the reported failure mode.

**Strengths**
- Root cause is correctly targeted: transforms now read the materialized snapshot through the runtime's own local fs (`getLocalAdapter().fs`), while `client`/the request-scoped adapter stays the authority for release retrieval and manifest publish — matches the "materialize once, transform locally" description.
- `getLocalAdapter()` is an established pattern already used the same way in `mdx-module-loader.ts`, `module-loader/index.ts`, and `filesystem-store.ts` — this isn't introducing a new abstraction, just applying an existing one where it was missing.
- The added test (`resolves project imports from the materialized release instead of the remote adapter`) is a real regression guard: it makes the remote adapter's `readFile` reject and asserts `remoteReads` stays empty while the build still reaches `ready` and the manifest contains both the MDX route and its aliased component — this is exactly the RED/GREEN case described in the PR body.
- Diff is minimal and non-invasive: `createMaterializedReleaseTransformAdapter` only overrides `fs` via a shallow spread, so `env`/`server`/other adapter capabilities that `buildFrameworkDependencies` or `transform()` might still need stay intact.
- Doc comment on the new helper clearly explains *why* (remote adapter 404s on tempDir-local paths), which will save the next person a debugging session.

**Concerns**
- Worth confirming explicitly: does `runBuildInner` (or the `task:release-asset-build` handler) ever execute in a context where `detectRuntime()` resolves to `"cloudflare"` or `"unknown"`? `getLocalAdapter()` → `AdapterRegistry.doInitialize()` throws in both of those cases, so if hosted release builds can ever run directly in an edge/Workers context (vs. a Node/Deno/Bun build service), this change could turn a working build into a hard `INITIALIZATION_ERROR`/`PLATFORM_ERROR`. If builds only ever run in a Node/Deno/Bun-hosted build service, this is a non-issue — but it'd be good to state that assumption in the PR description or a comment, since it's the one place this fix could regress rather than fix.
- The mock `remoteAdapter` in the new test only overrides `readFile`; other `FileSystemAdapter` methods (`stat`, `exists`, `resolveFile`, etc.) still delegate to `fsAdapter`'s real fs. If any transform path in this flow uses one of those instead of `readFile`, it wouldn't be caught by this regression test. Not blocking, but consider also failing those methods on the mock for tighter coverage.
- Minor: no changelog/release-notes entry — flag only if this repo expects one for user-facing release-build fixes.

Nothing here blocks merge; the Cloudflare/edge-runtime question is the only item I'd want a explicit answer to before approving.


---

kojiwakayama:
Addressing the automated review notes:

- The release-asset task is intentionally a writable project-runtime surface. Its handler creates the build directory with `Deno.makeTempDir` before calling this executor, so it cannot execute in Cloudflare Workers or an unknown runtime today. `getLocalAdapter()` therefore matches the existing execution contract; the npm build also verifies the Node-compatible path.
- The rejecting test adapter does not delegate `stat`, `exists`, `resolveFile`, or other filesystem methods. The shared `fsAdapter` fixture only defines `readFile`, and the test spread preserves that narrow surface. Any unexpected read through the request adapter is therefore either recorded and rejected by `readFile`, or unavailable.
- No changelog entry is needed for this internal release-build correction; the linked issue and PR provide the release note trail.

No code change is required from these notes.

kojiwakayama:
Addressed the CodeRabbit containment warning in `309b8a96e`:

- the transform filesystem is now wrapped by Veryfront SecureFs and rooted to the immutable materialized release directory
- the wrapper uses module-loading policy, rejects traversal and paths outside that root, and advertises the symlink-free fixed-project semantics enforced by that policy
- the regression now creates an actual file outside the release root, deliberately attempts to read it through the injected transform adapter, and proves the read is blocked while the aliased MDX build still publishes both expected modules

RED: the prior native adapter read the outside marker successfully. GREEN: the secured adapter rejects it. The complete release-assets suite, typecheck, and full lint gate pass on this change.

kojiwakayama:
@codex review

kwakayama (commented):
## Findings

- **[P2] Keep dependency-hash reads inside the materialized release root.** The new secured adapter is passed to transforms at `src/release-assets/build-executor.ts:3233` and `:3403`, but `transformToESM` constructs its own reader (`src/transforms/pipeline/index.ts:450`). That reader bypasses the adapter with raw local filesystem access for an absolute path outside `projectDir` (`:497-503`). Dependency hashing recursively reads every accepted local import (`src/cache/dependency-graph.ts:350-373`), including `file://` specifiers (`:196-205`, `:230-232`). Consequently, tenant source containing `import "file:///host/path"` (or a relative path escaping the snapshot) can make the release build read host files before final import validation rejects the output. This defeats the containment this change introduces. Pass a root-confined `readFile` explicitly for materialized transforms, and add a regression test through the real pipeline; the current test only directly probes `adapter.fs.readFile` at `src/release-assets/scaffolded-project-build.test.ts:397-403`, so it misses this bypass.

## Score

- Correctness: 35/40
- Tests: 14/20
- Reliability/security: 9/15
- Maintainability: 14/15
- Scope/docs: 10/10

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 309b8a96ee35d6224a196ea436cb5df18440d3ab
Score: 82/100
Actionable-Findings: 1
Verdict: REQUEST_CHANGES

PR #4361, opened 2026-09-01
fix(cli): register bundler before deploy

kojiwakayama:
Fixes veryfront/veryfront-issue-inbox#925

## What changed

- register the built-in bundler contracts before deploy enters project configuration and build work
- preserve argument validation before extension setup
- add an order-sensitive deploy handler regression test

## RED-GREEN TDD

RED: the handler ignored the registration dependency and entered live deploy work, where the test failed on an attempted network request.

GREEN: the handler records bundler registration before deploy execution and performs no external work in the focused test.

## Verification

- deno task test:file cli/commands/deploy/handler.test.ts
- deno task typecheck
- deno task lint:ci
- source CLI staging dry-run completed through config and target resolution
- git diff --check

### DIFF
--- a/cli/commands/deploy/handler.test.ts
+++ b/cli/commands/deploy/handler.test.ts
@@ -26,6 +26,23 @@ describe("Deploy Handler", () => {
     it("accepts a single ParsedArgs parameter", () => {
       assertEquals(handleDeployCommand.length, 1);
     });
+
+    it("registers the builtin bundler before deploying", async () => {
+      const calls: string[] = [];
+
+      await handleDeployCommand(createArgs(), {
+        ensureBundlerContracts: () => {
+          calls.push("ensure");
+          return Promise.resolve();
+        },
+        deploy: () => {
+          calls.push("deploy");
+          return Promise.resolve(null);
+        },
+      });
+
+      assertEquals(calls, ["ensure", "deploy"]);
+    });
   });
 
   describe("parseDeployArgs via handler", () => {
--- a/cli/commands/deploy/handler.ts
+++ b/cli/commands/deploy/handler.ts
@@ -6,8 +6,24 @@ import { deployCommand, parseDeployArgs } from "./command.ts";
 import { showHeader } from "#cli/utils";
 import type { ParsedArgs } from "#cli/shared/types";
 import { parseArgsOrThrow } from "#cli/shared/args";
+import { ensureCliBundlerContracts } from "#cli/shared/default-contracts";
 
-export async function handleDeployCommand(args: ParsedArgs): Promise<void> {
+interface DeployHandlerDependencies {
+  ensureBundlerContracts: () => Promise<void>;
+  deploy: typeof deployCommand;
+}
+
+const defaultDependencies: DeployHandlerDependencies = {
+  ensureBundlerContracts: ensureCliBundlerContracts,
+  deploy: deployCommand,
+};
+
+export async function handleDeployCommand(
+  args: ParsedArgs,
+  dependencies: DeployHandlerDependencies = defaultDependencies,
+): Promise<void> {
   showHeader();
-  await deployCommand(parseArgsOrThrow(parseDeployArgs, "deploy", args));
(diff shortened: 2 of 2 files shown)


kojiwakayama:
@codex review

kwakayama:
## Code Review: 88/100 — Good, minor suggestions

Small, well-scoped fix that registers the built-in bundler contracts before deploy enters project/build work, matching the pattern already used by `build` and `serve` handlers.

**Strengths**
- Root cause matches the fix: `ensureCliBundlerContracts()` is now awaited before `deploy()`, and it's idempotent (`tryResolve` short-circuits), so the added call is a no-op once bundler contracts are already registered.
- Argument validation still runs first (`parseArgsOrThrow` before `ensureBundlerContracts`), which the PR calls out explicitly and the pre-existing "empty branch"/"empty env" tests continue to cover — good attention to not regressing that ordering.
- The new test verifies call *order* (`["ensure", "deploy"]`) via injected dependencies rather than asserting on side effects of the real bundler extension, which keeps the test fast and hermetic (no live import/network attempt in CI).
- Dependency-injection shape (`DeployHandlerDependencies`) is consistent with the existing `deployProject` override already supported by `command.ts`/`DeployOptions`, so it's an extension of an established testability pattern in this file rather than a one-off.
- Description follows RED→GREEN with concrete verification commands (typecheck, lint, focused test, staging dry-run), which is good practice for a fix this size.

- `build` and `serve` handlers call `ensureCliBundlerContracts()` directly rather than through an injectable dependency object — `deploy` is now the only handler using DI for this. Not wrong (deploy already had an override pattern via `deployProject`), but worth a one-line note in the PR description on why deploy diverges, for future readers comparing the three handlers.
- Only one new test covers the ordering fix. Given the bug was "handler entered live deploy work before registration," it might be worth also asserting that `ensureBundlerContracts` rejecting causes `deploy` to *not* be called (fail-fast), though this is a nice-to-have, not a must-fix.

Nothing here blocks merging — this is a solid, targeted regression fix.

---

kojiwakayama:
Addressing the automated review notes:

- The dependency seam exists only to make the deploy-handler ordering test hermetic. Calling the real deploy command crosses project-link and network boundaries, as the RED run demonstrated. The production default still calls `ensureCliBundlerContracts` and `deployCommand` directly through that fixed dependency object.
- A separate rejection test would duplicate the same control-flow guarantee: `ensureBundlerContracts` is awaited before the only call to `deploy`, so a rejection cannot reach deploy. The focused test pins the sequence, while the existing invalid-argument tests pin validation before registration.

No code change is required from these non-blocking notes.

kojiwakayama:
Merge-group diagnosis: coverage shard 1 failed only because the unrelated config-loader timing guard `bounds total prefix reconstruction for colon-bearing CSI` measured 13,889 ms against a 109 ms control under runner contention. The same exact reviewed head passed its pull-request coverage shard, and all #4361-focused and functional tests passed in the merge group. This is an infrastructure/performance flake, not a deploy-handler defect. Requeueing the unchanged reviewed head per merge-queue policy.

kwakayama (commented):
No actionable findings.

- Correctness: 40/40 — deploy arguments are parsed before initialization, then bundler contracts are awaited before deployment dispatch. [handler.ts:26](https://github.com/veryfront/veryfront-code/blob/55dd3fc3fa0e73dbd023bf0e02eb6cdf283c373d/cli/commands/deploy/handler.ts#L26-L28)
- Tests: 19/20 — the new dependency-seam test verifies the required awaited ordering. [handler.test.ts:30](https://github.com/veryfront/veryfront-code/blob/55dd3fc3fa0e73dbd023bf0e02eb6cdf283c373d/cli/commands/deploy/handler.test.ts#L30-L44)
- Reliability/security: 15/15
- Maintainability: 15/15
- Scope/docs: 10/10

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 55dd3fc3fa0e73dbd023bf0e02eb6cdf283c373d
Score: 99/100
Actionable-Findings: 0
Verdict: APPROVE

Issue #939, opened 2026-04-10
Add public runtime hooks for mid-run system/context refresh


kojiwakayama:
## Problem

`veryfront-agent` still owns project steering and dynamic project-context refresh logic because the framework runtime does not yet expose a clean public way to refresh system/context state across a long-lived hosted run.

Today the framework supports static `system`, startup `context`, and `beforeStream`, but that is not the same as a runtime-facing hook for:

- refreshing project/system context mid-run
- reacting to new project steering state without rebuilding host-local runtime semantics
- keeping refresh mechanics generic while leaving product policy in the host

That leaves another migration-critical runtime seam outside `veryfront/agent`.

## Desired outcome

Add a public `veryfront/agent` hook or runtime contract that lets downstream hosts refresh system/context state during a hosted run without forking the framework runtime loop.

This should make it possible for `veryfront-agent` to keep Veryfront-specific steering policy local while using framework-owned refresh mechanics.

## Acceptance shape

- public `veryfront/agent` API, not `src/internal-*`
- supports runtime-safe system/context refresh during a hosted run
- keeps host policy/data ownership outside the framework
- docs/examples for downstream hosts
- regression coverage for refresh behavior and non-refresh stability

## Why now

After AG-UI request/run-control seams and human-input waits, mid-run system/context refresh is one of the clearest remaining framework gaps in the `veryfront-agent` migration parity matrix.

kojiwakayama:
PR opened: #944

This adds the public `resolveRuntimeState` seam on `agent()` so hosted runtimes can refresh the active system prompt and host-owned runtime context at step boundaries without forking the framework runtime loop.

Verification:
- `deno test --allow-all src/agent/runtime-refresh.test.ts src/agent/provider-transport.test.ts src/agent/ag-ui-handler.test.ts src/agent/chat-handler.test.ts src/agent/runtime/index.test.ts`
- `deno task docs:validate`
- `deno task verify`

PR #4365, opened 2026-09-01
fix(cli): resolve relative deploy project directories

kojiwakayama:
## Summary

- normalize relative project directories once at the shared Deploy Execution boundary
- cover deploy, `up`, and MCP callers without duplicating path policy at each adapter
- keep the existing route-containment checks unchanged
- add RED-GREEN shared-boundary and `up --project-dir .` end-to-end coverage

## Verification

- `deno task test:file cli/commands/deploy cli/commands/up cli/shared/deployment/deploy-project.test.ts` (188 steps)
- `deno task typecheck`
- `deno task lint`
- `deno task build:npm`
- staging deploy dry run with `--project-dir .` completed config and target resolution

Closes veryfront/veryfront-issue-inbox#939

### DIFF
--- a/cli/commands/deploy/command.test.ts
+++ b/cli/commands/deploy/command.test.ts
@@ -19,6 +19,7 @@ import { DeployArgsSchema, parseDeployArgs } from "./index.ts";
 import { deployCommand } from "./command.ts";
 import { parseCliArgs } from "#cli/shared/args";
 import type { ParsedArgs } from "#cli/shared/types";
+import { resolve } from "veryfront/platform/path";
 
 // Never touched: the fake executor records the request without reading it.
 // Named rather than inlined so it cannot be mistaken for a real checkout.
@@ -43,6 +44,40 @@ async function captureConsole<T>(fn: () => Promise<T>): Promise<{ result: T; out
 }
 
 describe("deploy command adapters", () => {
+  it("normalizes relative project directories before deployment", async () => {
+    let observedRequest: DeployProjectRequest | undefined;
+    const deployProject: DeployProject = {
+      execute(request) {
+        observedRequest = request;
+        return Promise.resolve({
+          kind: "dry-run",
+          plan: {
+            branch: "main",
+            projectId: null,
+            projectSlug: "demo",
+            environment: "staging",
+            environmentId: null,
+            controlPlane: "https://control.example.test/api",
+            plannedActions: ["create-release", "deploy"],
+          },
+        });
+      },
+    };
+
+    await deployCommand({
+      projectDir: ".",
+      projectSlug: "demo",
+      branch: "main",
+      env: "staging",
+      dryRun: true,
+      force: false,
+      quiet: true,
+      deployProject,
+    });
+
+    assertEquals(observedRequest?.projectDir, resolve("."));
+  });
+
   it("render human and JSON output from one injected deployment executor", async () => {
     const sentinelResult: DeployResult = {
       projectId: "project-sentinel",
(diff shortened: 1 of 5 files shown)


kojiwakayama:
@codex review

kwakayama:
## Code Review: 42/100 — correct fix, but incomplete: the same bug is still reachable through `vf up`

**Summary:** The `resolve()` call added to `cli/commands/deploy/command.ts` is a correct, well-tested fix for `vf deploy --project-dir .`, but the identical bug pattern exists in at least one sibling caller of the same `DeployProject.execute()` that this PR leaves untouched, so the class of bug this PR claims to close isn't actually closed.

### Strengths
- The fix itself is minimal and correct: `resolve(options.projectDir ?? cwd())` in `toDeployRequest()` (`cli/commands/deploy/command.ts:98`) ensures `DeployProjectRequest.projectDir` is always absolute before it reaches `resolveProjectRouteDirectory()`'s `isWithinDirectory(projectRoot, routeRoot)` check in `cli/shared/deployment/deploy-project.ts:557` — which is exactly where a relative `projectRoot` (never resolved) compared against an always-absolute `routeRoot` (via `resolve(projectRoot, directory)`) would previously mismatch.
- Test coverage for the touched file is genuine RED-GREEN: `command.test.ts` adds a case asserting `observedRequest?.projectDir === resolve(".")`, and updates the existing `"adapter-must-not-read"` fixture to expect a resolved path too.

### Concerns
- **The same unresolved-relative-path pattern is left unpatched in `cli/commands/up/command.ts:148`**: `const { projectDir = cwd(), ... } = options;` is passed straight through to `deployProject.execute({ projectDir, ... })` at `cli/commands/up/command.ts:277`, with no `resolve()` call anywhere in between. `up` accepts `--project-dir`/`-d` via the same `CommonArgs.projectDir` (`cli/shared/args.ts:223`) as `deploy` does, and its request flows into the exact same `collectProjectPageRoutes` → `resolveProjectRouteDirectory` containment check this PR is fixing. `vf up --project-dir .` should reproduce the original bug (#939) today, after this PR merges.
- `cli/mcp/tools/deploy-tool.ts:66` has the identical `options.projectDir ?? cwd()` shape. It's not currently reachable with a relative path from the one caller in `standalone.ts` (which never sets `projectDir`, so it falls back to `cwd()`), but it's the same latent footgun and will bite the next caller that does pass one.
- Given two more call sites share this exact shape, the root cause is arguably that `DeployProjectRequest.projectDir` isn't normalized once at the `DeployProject.execute()` boundary in `deploy-project.ts` — patching each caller individually is how this PR itself already missed one. Worth considering resolving it there instead, so `deploy`, `up`, and the MCP tool can't regress independently again.
- Minor: the PR description says coverage was added to `cli/shared/deployment/deploy-project.test.ts`, but the diff only touches `cli/commands/deploy/command.test.ts` — worth fixing the description, or actually adding a request-boundary/route-containment test at the `deploy-project.ts` level as originally described, which would have caught the `up` command gap above.

### Suggested next step
Move the `resolve()` normalization into `createDeployProject().execute()` (or a shared helper called by all three entry points) so `deploy`, `up`, and `vf_trigger_deploy` can't drift again, and add/extend a test at the `deploy-project.ts` boundary (as the PR description originally claimed) covering `--project-dir .` end-to-end through the route-containment check, plus a regression test for `vf up --project-dir .`.


---

kojiwakayama:
Addressed the cross-caller review in 389ea25e2. Normalization now occurs once inside `DeployProject.execute()`, so deploy, up, and the MCP deploy tool receive the same absolute-root invariant. The command-specific normalization and test were removed. Added RED-GREEN coverage at the shared boundary and an `up` end-to-end case using a relative project directory. The combined deploy and up suites pass 188 steps, and the PR description now reflects the actual test scope.

kojiwakayama:
@codex review

kojiwakayama:
Fixed the CI-only semantic-audit failure in 19b43f8cb. The relative-path tests now derive the repository root from `import.meta.url` instead of reading shared process cwd. The focused tests, type check, semantic disposition audit, and complete `lint:ci` chain all pass locally.

kojiwakayama:
@codex review

PR #4369, opened 2026-09-02
fix(security): stop validating env and stored tokens against project veryfront.json apiUrl

kojiwakayama:
## Summary

`resolveApiCredentialCandidatesForAuth` (`cli/shared/config.ts`) read `veryfront.json` from the current project directory and built a single `validationEnv` whose `apiUrl` came from that file, then attached it to every credential candidate — shell-environment (`VERYFRONT_API_TOKEN`), `.env`-file, and token-store credentials included. `veryfront login` and `veryfront whoami` validate those candidates by fetching `${apiUrl}/me` (or `/projects`) with `Authorization: Bearer <token>`. A malicious repository could therefore check in `veryfront.json` with `"apiUrl": "https://attacker.example"`, and any developer with a stored Veryfront token or `VERYFRONT_API_TOKEN` who ran `veryfront login` or `veryfront whoami` inside that directory would send their credential to the attacker-controlled host.

## Finding

- Codex finding id: `dbaac2593a00819193938d6155b42c8f` (finding 36), severity: medium
- Introduced in 71356afcf94da86bc63997ef6eca397090f8a9d2

## Fix

`resolveApiCredentialCandidatesForAuth` now builds two validation environments:

- a baseline `validationEnv` derived from the environment only (`resolveCliApiUrl(env)` — `VERYFRONT_API_URL`, an explicit non-default `VERYFRONT_API_BASE_URL`, or the default production URL), used for shell-environment, `.env`-file, and token-store candidates; and
- a `configFileValidationEnv` that includes the project `veryfront.json` `apiUrl`, used only for the credential that itself came from that same config file.

A checked-in `veryfront.json` can no longer redirect credentials it did not supply. Operators who intentionally target a different API still can via `VERYFRONT_API_URL` / `VERYFRONT_API_BASE_URL`.

## Test evidence

- New regression tests in `cli/shared/config.test.ts` (`resolveApiCredentialCandidatesForAuth`): non-config candidates keep the default API URL even when `veryfront.json` sets `apiUrl: "https://attacker.example"`, while the config-file token still validates against its own `apiUrl`. The first test was verified to fail against the unfixed code.
- Updated two tests in `cli/auth/login.test.ts` that had encoded the vulnerable behavior (env/stored tokens validated against the `veryfront.json` apiUrl); they now assert validation goes to the default API URL and never touches the config host.
- Local runs, all passing: `deno task test:file cli/shared/config.test.ts` (7 passed, 42 steps), `deno task test:file cli/auth/login.test.ts` (94 steps), `deno task test:file cli/shared/constants.test.ts`, `deno task test:file cli/auth/auth.integration.test.ts`; `deno fmt --check` and `deno lint` clean on touched files; `deno check cli/shared/config.ts cli/shared/config.test.ts` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ versions are listed at
 
 ## Unreleased
 
+### Breaking: project API URLs no longer steer ambient credentials
+
+Veryfront no longer sends a shell, `.env`, or stored CLI credential to the
+`apiUrl` in a project's `veryfront.json` during login, `whoami`, or an
+authentication preflight. Those credentials use the API URL selected by your
+environment, or the default Veryfront API URL when you do not set one. A token
+defined in the same `veryfront.json` still uses that file's `apiUrl`.
+
+If you use a self-hosted control plane with an ambient credential, set
+`VERYFRONT_API_URL` or `VERYFRONT_API_BASE_URL` in your trusted environment.
+
 ### Breaking: Redis workflow `runTtl` no longer expires runs
 
 `RedisBackendConfig.runTtl` is now a deprecated no-op. It previously started a
--- a/cli/auth/login.test.ts
+++ b/cli/auth/login.test.ts
@@ -1608,7 +1608,7 @@ describe("Login Module", { sanitizeOps: false, sanitizeResources: false }, () =>
       }
     });
 
-    it("validates an environment token against the configured API URL in JSON mode", async () => {
+    it("validates an environment token against the default API URL, not a veryfront.json apiUrl, in JSON mode", async () => {
       const originalFetch = globalThis.fetch;
       const originalLog = console.log;
       const originalError = console.error;
@@ -1630,7 +1630,7 @@ describe("Login Module", { sanitizeOps: false, sanitizeResources: false }, () =>
         globalThis.fetch = ((input: string | URL | Request, init?: RequestInit) => {
           requestedUrls.push(String(input));
           requestedAuth.push(String(new Headers(init?.headers).get("authorization") ?? ""));
-          if (String(input) === "https://control.example.test/api/me") {
+          if (String(input) === "https://api.veryfront.com/me") {
             return Promise.resolve(
               Response.json({ id: "env-user", email: "env@example.com" }),
             );
@@ -1657,7 +1657,7 @@ describe("Login Module", { sanitizeOps: false, sanitizeResources: false }, () =>
           email: "env@example.com",
           source: "env",
         });
-        assertEquals(requestedUrls, ["https://control.example.test/api/me"]);
+        assertEquals(requestedUrls, ["https://api.veryfront.com/me"]);
         assertEquals(requestedAuth, ["Bearer env-valid-token"]);
         assertEquals(output.join("\n").includes("env-valid-token"), false);
         assertEquals(errors, []);
@@ -1673,7 +1673,7 @@ describe("Login Module", { sanitizeOps: false, sanitizeResources: false }, () =>
       }
     });
 
-    it("validates a stored token against the configured API URL in JSON mode", async () => {
+    it("validates a stored token against the default API URL, not a veryfront.json apiUrl, in JSON mode", async () => {
       const originalFetch = globalThis.fetch;
       const originalLog = console.log;
       const originalError = console.error;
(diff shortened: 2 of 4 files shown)


kwakayama:
## Code Review: 89/100 — Good, minor suggestions

Tight, well-targeted fix for a real credential-exfiltration issue. Ready to merge modulo one documentation gap noted below.

**Strengths**
- Correctly scoped root-cause fix: `resolveApiCredentialCandidates` now takes a separate `configFileValidationEnv` that only the `config-file` candidate uses, so shell-env/`.env`/token-store credentials can no longer be steered to an attacker-controlled `apiUrl` from a checked-in `veryfront.json`.
- I traced the fix through to the actual validation call sites (`describeExistingSession`, `whoami`, `ensureAuthenticated` in `cli/auth/login.ts`, all consuming `candidate.validationEnv`) and confirmed the new precedence in `resolveCliApiUrl` (env var > explicit base URL > config apiUrl > default) matches the PR description — this isn't just a superficial diff, the vulnerable data flow is actually closed.
- Good regression tests: the new `resolveApiCredentialCandidatesForAuth` tests directly reproduce the attack (malicious `apiUrl` in `veryfront.json`) and assert non-config candidates stay on the default URL while the config-file token still validates against its own URL — and the author states the first test was verified to fail pre-fix, which is exactly the right way to prove a security fix.
- Backward-compatible signature (`configFileValidationEnv` defaults to `validationEnv`), so the other caller of the shared helper (`resolveApiTokenForMode`, used by `resolveConfigBase` for normal command execution rather than login/whoami validation) is unaffected — appropriately out of scope since that path's `validationEnv` field is never consumed for a fetch.
- Small, self-contained diff (3 files) with a clear inline comment explaining the security rationale and a commit message that references the originating finding ID.

**Concerns**
- Behavior change not called out in `CHANGELOG.md`: operators who rely on a project `veryfront.json` `apiUrl` (without setting `VERYFRONT_API_URL`/`VERYFRONT_API_BASE_URL`) to point `login`/`whoami` at a self-hosted control plane will now have their env/stored token validated against the default production URL instead, which will fail. The PR body mentions the escape hatch, but anyone who hits this will see it as a regression, not a security fix, unless it's surfaced in the changelog for self-hosted users.
- Minor: worth a one-line note in the PR (or code comment) on why `resolveApiTokenForMode`'s `validationEnv` plumbing is left as-is — it's correct as analyzed, but a future reader modifying that path could reintroduce the same class of bug without realizing `validationEnv` there is currently dead for fetch purposes.

Nothing here blocks merging; the changelog note is the only thing I'd actually ask for before/shortly after merge.

---

kojiwakayama:
@codex review

kojiwakayama:
Addressed the changelog concern in 92b816011c. The Unreleased section now explains that login, whoami, and authentication preflights keep shell, .env, and stored credentials on the trusted environment-selected API URL, and documents the self-hosted control-plane migration path. Focused config and login suites, formatting, and diff checks pass.

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kwakayama:
@codex review

kojiwakayama:
@codex review

PR #4370, opened 2026-09-02
fix(security): never cache renders carrying an admitted application identity

kojiwakayama:
## Summary

The legacy `/_veryfront/data/*.json` endpoint builds an application request with the configured identity headers removed (`denyHeaders: ctx.applicationIdentityHeaderNames`) and then calls `renderer.renderPage(..., { applicationIdentity: ctx.applicationIdentity })` with no cache guard. For trusted-proxy authentication the user's identity exists only in those stripped headers (e.g. `x-auth-subject`), so `buildCacheKey`'s `requestHasCacheSensitiveState` check — which only looks for Authorization/Cookie/x-api-key on the application request — never fires. An identity-dependent page/`getServerData` result could therefore be persisted under the public slug-based render cache key and later served to a *different* authenticated trusted-proxy user requesting the same URL: a cross-user data leak. The adjacent SSR service (`ssr.service.ts` — `mustNotCache = ... || hasApplicationIdentity`) and page-data endpoint (`page-data-endpoint-handler.ts` — `!hasApplicationIdentity` gates the cache key) already guard against exactly this; the legacy data path missed it.

## Finding

- Codex finding id: `6af1308533bc8191a2608e50a4e76918`
- Severity: medium
- Introduced in: 05f9e36d3 (feat(auth): add provider-neutral application authentication)

## Fix

Fail closed at the cache-key level, covering the legacy data endpoint and every other `renderPage` caller uniformly: both `Renderer.buildCacheKey` (`src/rendering/renderer.ts`) and `RenderPipeline.buildCacheKey` (`src/rendering/orchestrator/pipeline.ts`) now return `null` whenever `options.applicationIdentity` is non-null (alongside the existing sensitive-header check, so the documented explicit `cacheKey` override semantics are unchanged). A `null` cache key disables both cache lookup and cache persistence, so identity-bearing renders never read from or write to the shared render cache.

## Test evidence

- New regression tests:
  - `src/rendering/orchestrator/pipeline.behavior.test.ts` — "never caches HTML rendered for an admitted application identity": renders with an `applicationIdentity` and a header-stripped request, asserts zero `checkCache`/`persistResult` calls.
  - `src/rendering/renderer.test.ts` — "returns no render cache key for an admitted application identity": asserts `buildCacheKey` yields `null` for an identity-bearing request with no sensitive headers.
- `deno task test:file src/rendering/orchestrator/pipeline.behavior.test.ts` — ok, 1 passed (56 steps), 0 failed
- `deno task test:file src/rendering/renderer.test.ts` — ok, 5 passed (66 steps), 0 failed
- `deno task test:file src/server/handlers/request/module/data-endpoint-handler.test.ts` — ok, 1 passed (18 steps), 0 failed
- `deno task test:file tests/integration/server/module-handler-cache.test.ts` — ok, 1 passed (4 steps), 0 failed
- `deno fmt --check` clean and `deno lint` clean on all touched files; `deno check src/rendering/index.ts src/server/index.ts` passes

### DIFF
--- a/src/rendering/orchestrator/pipeline.behavior.test.ts
+++ b/src/rendering/orchestrator/pipeline.behavior.test.ts
@@ -821,6 +821,50 @@ describe("RenderPipeline behavior", () => {
     assertEquals(persists, [], "personalized HTML must never be persisted");
   });
 
+  it("never caches HTML rendered for an admitted application identity", async () => {
+    const checks: Array<string | undefined> = [];
+    const persists: Array<string | undefined> = [];
+    const pipeline = createPipeline("/project/pages/index.mdx", {
+      cacheCoordinator: {
+        checkCache: async (slug, cacheKey) => {
+          checks.push(cacheKey);
+          return {
+            depAwareSlug: slug,
+            moduleCacheKey: cacheKey ?? slug,
+            cacheStatus: "miss",
+            lookupDurationMs: 0,
+          };
+        },
+        persistResult: async (_result, _slug, cacheKey) => {
+          persists.push(cacheKey);
+        },
+      },
+    } as Partial<RenderPipelineConfig>);
+
+    // Trusted-proxy auth strips identity headers from the application request,
+    // so the request itself carries no cache-sensitive state; only the
+    // admitted identity marks this render as personalized.
+    await pipeline.renderPage("", {
+      delivery: "string",
+      request: new Request("http://localhost/"),
+      applicationIdentity: {
+        issuer: "https://issuer.example.test",
+        subject: "user-123",
+        groups: ["engineering"],
+        roles: ["reader"],
+        groupsComplete: true,
+        claims: { sub: "user-123" },
+      },
+    });
+
+    assertEquals(
+      checks,
+      [],
+      "identity-scoped HTML must never be read from the shared render cache",
+    );
+    assertEquals(persists, [], "identity-scoped HTML must never be persisted");
+  });
+
   it("bounds the complete API render key for a flag-off override", () => {
     const cachePrefix = "project:preview:branch:v1";
     const pipeline = createPipeline("/project/pages/index.mdx", {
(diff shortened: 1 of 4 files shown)


kwakayama:
## Code Review: 90/100 — Solid, well-tested fail-closed fix for a real cross-user cache leak

**Summary:** Correctly closes a cross-user data leak in the legacy data endpoint's render cache and does so at the right layer (both cache-key builders) rather than patching the one call site.

**Strengths:**
- Root cause is verified: `data-endpoint-handler.ts` strips `ctx.applicationIdentityHeaderNames` from the application request before calling `renderPage(..., { applicationIdentity })`, so `requestHasCacheSensitiveState` (which only inspects Authorization/Cookie/x-api-key) never sees the identity, letting personalized output ride the shared slug-based cache key. Confirmed in the diff.
- The fix is applied uniformly to `Renderer.buildCacheKey` and `RenderPipeline.buildCacheKey`, matching the guard pattern already used in `ssr.service.ts` and `page-data-endpoint-handler.ts` — good consistency rather than a one-off patch.
- Fail-closed placement is correct: the `applicationIdentity` check sits after the explicit `options.cacheKey` override branch in both files, so documented override semantics are preserved (a caller that explicitly scopes its own key is still trusted).
- Regression tests added at both cache-key builders reproduce the exact reported scenario (header-stripped request + admitted identity → zero `checkCache`/`persistResult` calls, `buildCacheKey` returns `null`), which is the right place to pin this invariant since it's shared by every `renderPage` caller.
- Minimal, surgical diff (88/-3 across 4 files); clear commit message and PR description with root cause, fix, and test evidence.

**Minor points (non-blocking):**
- No new integration-level regression test at `data-endpoint-handler.ts` itself confirming the end-to-end path (the PR cites existing tests there passing, not a new one) — the two new unit tests cover the shared root cause directly, which is reasonable, but an endpoint-level test would guard against a future caller reintroducing the same bug via a different path.
- This disables caching entirely for any identity-bearing render rather than scoping the cache key per-identity; correct fail-closed choice for a security fix, but worth a follow-up if per-user caching at this endpoint matters for performance.
- CI was still in progress at review time — please confirm all checks (lint/format/typecheck/tests) are green before merge, though the PR's cited local `deno task test:file` / `deno fmt --check` / `deno lint` runs are a good sign.

Nothing here blocks merge; the two minor points are suggestions, not requirements.


---

kojiwakayama:
@codex review

PR #4374, opened 2026-09-02
Fix agent service heartbeat JSON body

kojiwakayama:
## Summary

- send the JSON object required by the strict control-plane heartbeat endpoint
- lock the wire contract in the hosted registration lifecycle test

## Why

The generic hosted agent replicas register successfully, then every heartbeat fails because the framework declares JSON without sending a body. The API deployment health gate correctly detects both registrations as stale.

## Verification

- red: focused registration test failed with an undefined heartbeat body
- green: `deno task test:file src/agent/service/registration.test.ts` (27 steps)
- `deno task typecheck`
- focused `deno fmt --check`
- focused `deno lint`
- repository pre-push suite: 129 batches, 2,012 steps, plus serial and working-directory exclusion suites

This fixes the client at the owning boundary. It does not relax the API parser or deployment health gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Heartbeat requests now include a valid empty JSON payload, improving compatibility with the control plane.

* **Tests**
  * Added validation to ensure heartbeat requests use the expected JSON body format.

### DIFF
--- a/src/agent/service/registration.test.ts
+++ b/src/agent/service/registration.test.ts
@@ -288,6 +288,11 @@ describe("agent/agent-service-registration", () => {
       "heartbeats must carry bearer auth",
     );
     assertEquals(calls[1]?.init?.method, "POST", "the heartbeat must be a POST");
+    assertEquals(
+      calls[1]?.init?.body,
+      "{}",
+      "the heartbeat must send the JSON object required by the control plane",
+    );
     assertEquals(JSON.parse(String(calls[0]?.init?.body)), {
       service_name: "docs-agent",
       service_key: "docs-agent:test",
--- a/src/agent/service/registration.ts
+++ b/src/agent/service/registration.ts
@@ -446,6 +446,7 @@ async function sendHeartbeatRequest(
     response = await fetchImpl(getHeartbeatEndpoint(input.apiUrl, input.serviceId), {
       method: "POST",
       headers: createHeaders(input.authToken),
+      body: JSON.stringify({}),
       signal: abortSignal,
     });
   } catch (cause) {


kojiwakayama:
@codex review exact head 17f34e6619969aa4b5b091cf9f3f7f4ff9327be2

kwakayama (commented):
## Review: 92/100 — correct, minimal fix with a regression test that pins the wire contract

**Strengths**
- Root cause is real and precisely diagnosed: `createHeaders()` sets `Content-Type: application/json` on every request, but `sendHeartbeatRequest` sent no body, so a strict control-plane parser would reject the heartbeat. `body: JSON.stringify({})` makes the request self-consistent with its own declared content type.
- Fix is applied at the right boundary (the client), not by loosening server-side/API parsing — matches the PR description's stated intent and is the correct fix for a shared contract.
- Test change is a good regression guard: it asserts `calls[1].init.body === "{}"` on the heartbeat call specifically (not the registration call), so it will actually catch a future regression rather than just re-asserting the currently-passing behavior.
- Change is tightly scoped (+6/-0 across 2 files, 1 commit) — no unrelated churn, no API surface change.
- PR description documents red/green verification, typecheck, fmt, lint, and the full pre-push suite, which is more rigor than this diff size would normally require.

**Minor/optional (not blocking)**
- `mergeable_state` is currently `blocked` — worth confirming that's just outstanding required reviews/CI (which were still queued at review time) and not an actual merge conflict.
- Very small nit, non-blocking: `body: JSON.stringify({})` is a magic literal repeated implicitly against the registration path's `JSON.stringify(buildRegistrationRequest(input))`. A named constant (e.g. `EMPTY_JSON_BODY = "{}"`) shared between the heartbeat call and its test assertion would make the "this must literally be `{}`" contract self-documenting, but this is style preference, not a defect.

This is a textbook small bugfix: clear root cause, minimal diff, and a test that would have failed before the fix and passes after. No functional, security, or design concerns.

---

kwakayama (commented):
No actionable findings.

The heartbeat now sends the required JSON object at `src/agent/service/registration.ts:449` while preserving its existing POST method, authentication headers, endpoint, and abort behavior (`src/agent/service/registration.ts:446-451`). The regression test pins the exact wire body to `"{}"` at `src/agent/service/registration.test.ts:291-295`.

Rubric: correctness 40/40, tests 20/20, reliability/security 15/15, maintainability 15/15, scope/docs 10/10.

Review-Gate:
Reviewer: Codex
Reviewed-SHA: 17f34e6619969aa4b5b091cf9f3f7f4ff9327be2
Score: 100/100
Actionable-Findings: 0
Verdict: APPROVE

PR #4375, opened 2026-09-02
fix(security): redact gate failure evidence in exported eval summaries

kojiwakayama:
## Summary

Eval report exporters are documented as receiving redacted reports by default, but `redactEvalReportForExport` only redacted record-level fields and `summary.metrics[].label` — `summary.gateFailures` passed through untouched. `createEvalReport` copies each failing gate/budget metric's `explanation` and `evidence` verbatim into `summary.gateFailures`, and the RAG citation metrics (`knowledge.citationPrecision` / `knowledge.citationRecall`) place citation labels and expected/retrieved source labels into `evidence`, where the expected-source label can fall back to quote/text/content-derived data. Since knowledge metrics are gated at `min: 1`, any failing citation metric leaked private RAG citation and source details to every configured third-party exporter (e.g. the bundled MLflow exporter) even with `includeCitations`, `includeRetrievedContext`, and `includeMetricEvidence` all at their false defaults.

- Codex finding id: 8f8421e22e948191a0882b08c437ffd6
- Severity: medium

## Fix

`redactMetricSummaries` now redacts `summary.gateFailures` on the same terms as record-level metric results: each entry's `evidence` is dropped unless `redaction.includeMetricEvidence` is set, and its `explanation` is dropped unless `redaction.includeMetricExplanations` is set. Opt-in behavior is unchanged.

## Test evidence

- Extended `src/extensions/eval/eval-report-exporter.test.ts`: the default-redaction test now seeds `summary.gateFailures` with a failing `knowledge.citationPrecision` entry carrying citation/source evidence and asserts both `evidence` and `explanation` are stripped from the exported report; the allow-all test asserts they survive when explicitly opted in.
- `deno task test:file src/extensions/eval/eval-report-exporter.test.ts`: 1 passed (10 steps), 0 failed. Verified the new assertions fail against the unpatched exporter.
- `deno task test:file src/eval/report.test.ts`: 1 passed (9 steps), 0 failed.
- `deno fmt --check`, `deno lint`, and `deno check` clean on both touched files.

## Summary by CodeRabbit

- **Improvements**
  - Exported evaluation reports now redact gate-failure explanations and supporting evidence by default.
  - Export options can preserve gate-failure explanations and evidence when detailed reporting is needed.
  - Metric summaries continue to omit sensitive labels when metric evidence is excluded.

### DIFF
--- a/src/extensions/eval/eval-report-exporter.test.ts
+++ b/src/extensions/eval/eval-report-exporter.test.ts
@@ -39,6 +39,21 @@ function createReport(): EvalReport {
           label: 'Answer contained "the private launch codename"',
         },
       ],
+      gateFailures: [
+        {
+          recordId: "q1:1",
+          exampleId: "q1",
+          repetition: 1,
+          name: "knowledge.citationPrecision",
+          family: "knowledge",
+          severity: "gate",
+          explanation: 'Citation "[1]" did not match an expected source.',
+          evidence: {
+            citations: ["internal-doc-1"],
+            expectedSources: ["secret roadmap passage"],
+          },
+        },
+      ],
     },
     records: [
       {
@@ -150,6 +165,13 @@ describe("EvalReportExporterRegistry", () => {
     // `evidence` carries, so it must not survive redaction either.
     assertEquals(exportedRecord.metrics?.[0]?.label, undefined);
     assertEquals(exportedReport.summary.metrics[0]?.label, undefined);
+    // Gate failures copy the failing metric's explanation and evidence (citation and source
+    // labels for RAG metrics), so default redaction must strip them from the summary too.
+    const exportedGateFailure = exportedReport.summary.gateFailures?.[0];
+    assert(exportedGateFailure);
+    assertEquals(exportedGateFailure.name, "knowledge.citationPrecision");
+    assertEquals(exportedGateFailure.explanation, undefined);
+    assertEquals(exportedGateFailure.evidence, undefined);
     assertEquals(exportedReport.dataset, {
       kind: "json",
       examples: 1,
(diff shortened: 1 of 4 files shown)


kwakayama:
## Review: 92/100 — solid, minimal, well-verified security fix

Real leak (default-redacted eval reports were shipping raw RAG citation/source evidence via `summary.gateFailures` to third-party exporters), fixed with a small, correctly-scoped change and tests that fail against the pre-patch code.

**Strengths**
- Root cause is correctly identified: `redactMetricSummaries` redacted `summary.metrics[].label` but never touched `summary.gateFailures`, even though `createEvalReport` copies `explanation`/`evidence` into it verbatim from the same failing metric results that are redacted at the record level.
- Fix applies the exact same policy gates (`includeMetricEvidence` / `includeMetricExplanations`) already used for record-level metrics, keeping the redaction contract consistent rather than inventing a new flag.
- `EvalGateFailureSummary` has no `label`-equivalent field, so nothing else needed stripping there — confirmed by checking `src/eval/types.ts`.
- Test additions are meaningful, not just coverage padding: they seed a realistic failing `knowledge.citationPrecision` gate failure with citation/source evidence, assert it's stripped under default redaction, and assert it survives when explicitly opted in — matching the two branches the fix touches.
- PR description is exemplary: root cause, blast radius, exact fix, and test evidence (including "verified new assertions fail against unpatched exporter") are all spelled out. Single, well-scoped commit.

- No test for the mixed case (`includeMetricEvidence: true` + `includeMetricExplanations: false`, or vice versa) on `gateFailures`, only the all-false/all-true cases. The logic is straightforward (two independent `if`s) so this is low risk, but a mixed-case assertion would fully close the gap now that the two fields are handled independently.
- The function's early-return-by-reference behavior changed slightly (previously returned the original `summary` object unmodified when `includeMetricEvidence` was true; now always returns a shallow copy). This is more correct/defensive, not a regression, but worth being aware of if anything downstream relied on reference identity (unlikely).

Given this is a scoped security patch with clean tests, passing format/lint/typecheck per the PR description, and no design or API surface changes, this looks mergeable as-is once CI (still running at review time) finishes green.

---

kojiwakayama:
Closing as fully superseded on main. PR #4386 already merged the gate-failure redaction with stronger public type documentation and focused tests, and PR #4399 already merged deterministic route ordering with total-order tie coverage. Merging this branch would only reintroduce alternate implementations of behavior now present on main.

PR #4386, opened 2026-09-02
fix(security): redact gate failure evidence in exported eval reports

kojiwakayama:
## Vulnerability

Eval report exporters are documented and implemented as receiving redacted reports by default, but `redactEvalReportForExport` only sanitized `records` and `summary.metrics[].label`. `createResultFailure` (`src/eval/report.ts:196-210`) copies each blocking result's `evidence` and `explanation` verbatim into `summary.gateFailures`, and the RAG citation metrics `knowledge.citationPrecision` / `knowledge.citationRecall` put citation labels plus expected/retrieved source labels into that evidence (with `expectedSourceLabel` falling back to `contentMatch` / `verificationQuote` / `content` / `text`, i.e. content-derived text). Because those metrics are gates at `min: 1`, any failing citation check pushed private RAG citation and source detail into `summary.gateFailures`, and the exporter registry then handed that report to every configured third-party exporter — even with `includeCitations`, `includeRetrievedContext`, `includeMetricEvidence`, and `includeMetricExplanations` all left at their fail-closed defaults. Impact: private evaluation content leaves the process to an external observability vendor precisely in the case default redaction is documented to protect.

- Codex finding id: `8f8421e22e948191a0882b08c437ffd6`
- Severity: medium
- Introduced in: `9604ba031` (Eval V2: add RAG citation metrics, #2905)

## Fix

`redactMetricSummaries` becomes `redactSummary`, which now also walks `summary.gateFailures` and applies the same policy the per-record results already follow: `evidence` survives only under `includeMetricEvidence`, `explanation` only under `includeMetricExplanations`. The failure entry itself (record id, example id, repetition, metric name, family, severity) still reaches exporters, so gate-failure counts and shapes in downstream exporters (for example `ext-eval-report-mlflow`, which reads `gateFailures.length`) are unchanged. CLI output and on-disk reports are untouched; this boundary governs only what reaches an exporter. The redaction option docs were updated to state that both flags cover summary gate failures.

## Tests

Two cases added to `src/extensions/eval/eval-report-exporter.test.ts`, built from a failing `knowledge.citationPrecision` record whose summary is produced by the real `summarizeEvalRecords`, so the report.ts → exporter copy path is exercised end to end:

- default redaction leaves the gate failure present but drops `evidence` and `explanation`, and no citation text or expected-source label appears anywhere in the serialized exported report;
- `includeMetricEvidence` + `includeMetricExplanations` still deliver both.

Verified in the worktree:

- `deno task test:file src/extensions/eval/eval-report-exporter.test.ts` — ok, 12 steps (the new default-redaction case fails against `origin/main`'s exporter, confirmed by reverting the source file and re-running).
- `deno task test:file src/eval/report.test.ts` — ok, 9 steps.
- `deno task test:file src/eval/runner.test.ts` — ok, 21 steps.
- `deno fmt --check`, `deno lint`, `deno check` on both touched files — clean.

### DIFF
--- a/src/extensions/eval/eval-report-exporter.test.ts
+++ b/src/extensions/eval/eval-report-exporter.test.ts
@@ -1,7 +1,8 @@
 import "#veryfront/schemas/_test-setup.ts";
 import { assert, assertEquals, assertThrows } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
-import type { EvalReport } from "veryfront/eval";
+import type { EvalMetricResult, EvalReport } from "veryfront/eval";
+import { summarizeEvalRecords } from "veryfront/eval";
 import {
   createEvalReportExporterRegistry,
   redactEvalReportForExport,
@@ -93,6 +94,34 @@ function createReport(): EvalReport {
       },
     ],
   };
+}
+
+const failingCitationMetric: EvalMetricResult = {
+  name: "knowledge.citationPrecision",
+  family: "knowledge",
+  severity: "gate",
+  score: 0,
+  pass: false,
+  explanation: "0 of 1 citations were supported.",
+  evidence: {
+    tool: "knowledge.search",
+    citations: ["[1] secret roadmap passage"],
+    expected: ["Private roadmap: secret roadmap passage"],
+    supported: [],
+    unsupported: ["[1] secret roadmap passage"],
+    supportedCount: 0,
+    citationCount: 1,
+  },
+};
+
+/** Report whose only metric is a failing citation gate, so the summary carries a gate failure. */
+function createCitationFailureReport(): EvalReport {
+  const report = createReport();
+  const record = report.records[0];
+  assert(record);
+  record.metrics = [failingCitationMetric];
+  report.summary = summarizeEvalRecords(report.records);
+  return report;
 }
 
 describe("EvalReportExporterRegistry", () => {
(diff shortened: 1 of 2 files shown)


kwakayama:
## Code Review: 92/100 — Excellent, minor nit only

Precise, well-tested fix that closes the exact leak path described (summary `gateFailures` bypassing export redaction for citation evidence/labels).

**Strengths:**
- Correctly identified that `createResultFailure` (report.ts) copies `evidence`/`explanation` verbatim into `summary.gateFailures`, bypassing the per-record redaction that `redactRecord`/`redactMetricResults` already enforced — root cause matches the stated finding exactly.
- New `redactGateFailures` mirrors the existing `redactMetricResults` pattern (same flag semantics: `includeMetricEvidence` / `includeMetricExplanations`), keeping the codebase's redaction policy consistent rather than inventing a parallel scheme.
- Test coverage is genuinely end-to-end: it builds the failing `knowledge.citationPrecision` record through the real `summarizeEvalRecords` (not a hand-built summary), then asserts the raw citation text and expected-source label are absent from the *serialized* exported report — a stronger check than just asserting the field is `undefined`. A second test confirms the opt-in (`includeMetricEvidence`/`includeMetricExplanations`) still delivers both.
- Blast radius is intentionally minimal: failure identity (record id, example id, metric name/family/severity) is preserved so downstream consumers like `ext-eval-report-mlflow` that read `gateFailures.length` are unaffected; CLI/on-disk output paths are untouched since this only touches the exporter boundary.
- Docstrings on `includeMetricEvidence`/`includeMetricExplanations` were updated to state they now also govern summary gate failures — keeps the public contract honest.

**Minor observations (non-blocking):**
- The redaction now also strips `explanation` from `record.error`-sourced gate failures (`createRecordFailure`, family `"check"`), not just metric-evidence failures, since they share the same `EvalGateFailureSummary` shape. That's a reasonable/consistent side effect but wasn't called out in the PR description — worth a one-line mention so reviewers don't have to trace it themselves.
- `redactGateFailures`/`redactSummary` always allocate new arrays/objects even when both include-flags are `true` (a no-op copy). Negligible on eval-sized reports; not worth changing.

No security, correctness, or test-coverage concerns. This is a small, self-contained security patch that plugs the described leak without touching unrelated behavior — good to merge once CI is green.


---

kojiwakayama:
@codex review

PR #4397, opened 2026-09-02
fix(security): keep eval data from restoring disabled agent-service scope

kojiwakayama:
## Summary

The agent-service eval adapter lets a harness disable project-scoped execution by passing an explicit `projectId: null` (the config type is `string | null`, and the same holds for `conversationId`, `branchId` and `model`). But `getRequestOverrides()` in `src/eval/agent-service.ts` resolved those four fields with nullish coalescing — `input.projectId ?? readString(record.projectId) ?? null` — where `record` is the eval example's own `input`. An explicit `null` therefore fell straight through to values read from the eval data, and the resulting scope was emitted in `forwardedProps.veryfront` and sent to the live agent service with the configured Authorization bearer token. A trusted CI/harness running attacker-controlled eval data while deliberately keeping the run unscoped could have that data reintroduce a project, conversation or branch id and steer the run into project-scoped tools, files and API state reachable by that token. The server still verifies token access, so this is not unauthenticated cross-tenant access, but it is a confused-deputy bypass of the caller's intended scoping restrictions.

## Finding

- Codex finding id: `c8c27e0d32988191ae21589e8bcae4bf` (finding 134)
- Severity: medium
- Path: `src/eval/agent-service.ts` (introduced in `e77ddd725a821901105731c5f553f45426128023`)

Commit `2888c1818` ("fix(security): prevent client-controlled AG-UI agent selection", #4199) removed exactly this eval-data fallback for `agentId`, but left `projectId`, `conversationId`, `branchId` and `model` in place.

## Fix

Added a small `resolveScopeOverride()` helper and used it for those four fields in `getRequestOverrides()`. Only an omitted (`undefined`) caller value now falls back to the eval example's data — the documented convenience for unconfigured adapters is preserved. An explicit `null` is treated as the caller disabling that scope and is never overridden by example data. `allowedTools`, `forceRuntimeOverrides` and `maxSteps` are unchanged: their config types do not admit `null`, so `??` already only fires on an omitted value.

No public API surface changed; `deno task docs:api-reference:check` reports the generated reference is still current.

## Test evidence

Two regression tests added to `src/eval/agent-service-hardening.test.ts`:

- `keeps eval example data from re-enabling explicitly disabled request scope` — `buildAgentServiceEvalRequestBody` with `projectId/conversationId/branchId/model: null` and an example supplying all four yields no `forwardedProps` at all, while omitting them still lets the example data through (documented behavior preserved).
- `does not let eval data restore project scope disabled in the adapter config` — an end-to-end `runEval` against `createAgentServiceEvalAdapter({ projectId: null, ... })` with a hostile example captures the outgoing AG-UI request body and asserts `forwardedProps` is absent.

Both tests fail on the pre-fix source (verified by reverting `src/eval/agent-service.ts` and re-running: `FAILED | 0 passed (30 steps) | 1 failed (2 steps)`) and pass with it.

Commands run locally in a worktree off `origin/main`:

- `deno fmt --check src/eval/agent-service.ts src/eval/agent-service-hardening.test.ts` — clean
- `deno lint src/eval/agent-service.ts src/eval/agent-service-hardening.test.ts` — clean
- `deno check src/eval/index.ts` — clean
- `deno task test:file src/eval/agent-service-hardening.test.ts` — ok, 32 steps
- `deno task test:file src/eval/agent-service.test.ts` — ok, 30 steps
- `deno task lint:anti-slop` — baseline ok
- `deno task docs:api-reference:check` — reference current

### DIFF
--- a/src/eval/agent-service-hardening.test.ts
+++ b/src/eval/agent-service-hardening.test.ts
@@ -6,6 +6,7 @@ import {
   assertThrows,
 } from "#veryfront/testing/assert.ts";
 import { describe, it } from "#veryfront/testing/bdd.ts";
+import { datasets, evalAgent, runEval } from "veryfront/eval";
 import {
   buildAgentServiceEvalRequestBody,
   buildLiveEvalCaseMetadata,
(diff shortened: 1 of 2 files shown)


kwakayama (commented):

Precise, minimal fix for a real confused-deputy bug, with verified regression coverage matching the exact attack scenario.

**Strengths**
- Correctly scoped: I checked every field in `AgentServiceEvalAdapterConfig` against `getRequestOverrides()`/`createVeryfrontForwardedProps()` — `projectId`, `conversationId`, `branchId`, `model` are the only ones with an eval-data fallback (`?? readString(record.x) ?? null`), so those are exactly the fields that needed `resolveScopeOverride()`. The header-only fields (`projectSlug`, `releaseId`, `branchName`, `environment*`, `forwardedHost/Proto`) never read from example data, so they're correctly untouched. `allowedTools`/`forceRuntimeOverrides`/`maxSteps` don't admit `null` in their types, so `??` was already safe there — also correctly left alone.
- Test coverage is exactly on-target: a unit test on `buildAgentServiceEvalRequestBody` proving `null` blocks example data while `undefined` still allows the documented fallback, plus an end-to-end `runEval`/`createAgentServiceEvalAdapter` test with a hostile example asserting the outgoing request body has no `forwardedProps`. PR description states both fail against the pre-fix source — good practice, confirms the tests actually catch the regression.
- Clear precedent and traceability: explicitly mirrors the `agentId` hardening from #4199, and the PR body identifies the introducing commit and the scanner finding id, which makes this easy to audit.
- `resolveScopeOverride()` is a tight, well-documented helper (JSDoc explains the `undefined` vs `null` semantics) rather than a sprawling refactor — keeps the diff small and reviewable.

**Minor nit (non-blocking)**
- `resolveScopeOverride()` isn't exported, so it's only exercised indirectly through `buildAgentServiceEvalRequestBody`. That's sufficient given the two new tests cover all four fields' null/undefined behavior, just flagging it as an option if you want a more direct unit test later.

No design, security, or standards concerns beyond the nit above. CI was still queued at review time — worth confirming green before merge, but nothing in the diff itself raises doubts about the result.

---

PR #4406, opened 2026-09-04
fix(ci): give the sentry runtime packages job headroom on its timeout

kwakayama:
## Summary
- main's CI/CD kept going red today: `tests-sentry-runtime-packages` has `timeout-minutes: 15`, but a cold-cache run takes ~12-15min. Both times it crept over 15min (22:17 and 22:39 UTC runs) it got timeout-cancelled — GitHub reports this as job conclusion "cancelled".
- That cascades: `prerelease` needs this job and auto-skips when a dependency doesn't succeed, then `quality-gate-registry` still runs (its `if` uses `always()` by design — see `tests/integration/ci/registry-release-workflow.test.ts`) and hard-fails with "exact version and GITHUB_SHA are required" since there's no RC_VERSION to check.
- Bump the timeout to 20min for headroom. One-line change, no logic touched.

## Test plan
- [x] YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] No test references this job's timeout value
- [ ] CI green on this PR

## Summary by CodeRabbit

* **Chores**
  * Increased the runtime package test workflow timeout to improve reliability for slower cold-cache runs.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -694,7 +694,8 @@ jobs:
   tests-sentry-runtime-packages:
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    # Cold-cache runs take ~12-15min; 15 was cutting it too close.
+    timeout-minutes: 20
     name: tests (sentry runtime packages)
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1


kwakayama:

Minimal, well-justified CI timeout bump that fixes a real, clearly-diagnosed problem.

**Strengths**
- Root cause is concrete and well-documented: cold-cache runs at ~12-15min were tripping the 15min timeout, and the PR description traces the exact cascade (job cancelled → `prerelease` skips → `quality-gate-registry` still runs via `always()` → hard-fails on missing `RC_VERSION`) with links to the two failed runs and the relevant test.
- Change is truly minimal (1 file, +2/-1) and scoped to exactly the value that needs to change — no unrelated edits.
- Added an inline comment explaining *why* 20 was chosen, which will save the next person from re-deriving this.
- Verified the diff — no other file hardcodes this job's timeout (`git grep` confirms), consistent with the PR's own test-plan claim.

**Minor observations (non-blocking)**
- 20 minutes is still a guess based on two data points; if cold-cache time is variable, consider whether a bit more headroom (or caching improvements for that job) would be worth a follow-up, but this is a reasonable stopgap.
- Test plan's last box ("CI green on this PR") is unchecked — worth confirming before merge that the fix actually resolves the observed failures, though that's expected to just be pending CI.

This is a low-risk, well-reasoned infra fix — straightforward to approve once CI confirms green.

---

kojiwakayama:
@codex review

PR #4410, opened 2026-09-04
test(agent): use hosted inference import alias

kojiwakayama:
## Summary

- use the canonical internal alias for the hosted inference credential test import
- address the remaining outside-diff CodeRabbit finding from #4407

## Verification

- hosted chat parser suite: 89 steps passed
- deno fmt --check and deno lint
- dependency and module boundary checks
- test typecheck baseline
- codex review --base origin/main: no actionable findings

### DIFF
--- a/src/agent/hosted/chat-request.test.ts
+++ b/src/agent/hosted/chat-request.test.ts
@@ -19,7 +19,7 @@ import {
   MAX_HOSTED_CHAT_REQUEST_MESSAGES,
 } from "./chat-request.ts";
 import { createHostedRunEventWriterCapabilityForRequest } from "./child-run-event-writer-token.ts";
-import { createHostedInferenceModelResolver } from "./inference-credential.ts";
+import { createHostedInferenceModelResolver } from "#veryfront/agent/hosted/inference-credential.ts";
 
 const conversationId = "10000000-1000-4000-8000-100000000001";
 const messageId = "10000000-1000-4000-8000-100000000002";


kojiwakayama:
@codex review

kwakayama:
## Review: 35/100 — significant concerns (change is directionally wrong)

This is a 1-line, single-file diff, but the direction of the change is the opposite of what this repo's own conventions call for, so it shouldn't merge as-is.

**Findings:**

- **The import change contradicts the project's documented and enforced convention.** `.claude/skills/vf-module-development/SKILL.md` states: *"Within the same module, use relative imports."* `inference-credential.ts` and `chat-request.test.ts` are both in `src/agent/hosted/` — same directory, same module. The pre-PR `./inference-credential.ts` was already correct.
- **The repo's actual lint rule agrees, and explicitly exempts this file twice over.** `scripts/lint/no-cross-boundary-relative-imports.ts` only flags `../`-style imports that *cross* a top-level module boundary (`src/agent/hosted/renderer.ts → ../../cache/keys.ts`), and it explicitly skips `.test.ts` files altogether (`shouldSkip`). `./inference-credential.ts` is neither a `../` import nor cross-module, and the file is a test file — this line was never going to trip that check.
- **This is inconsistent with 29 other test files in the same directory** (46 occurrences) that use same-directory relative imports (`./foo.ts`) for their sibling source modules — e.g. `chat-preparation.test.ts`, `veryfront-cloud-agent-service.test.ts`, `child-lifecycle.test.ts`. Merging this makes `chat-request.test.ts` the outlier, not the model to follow.
- **Root cause: this PR blindly applies a CodeRabbit "outside-diff" suggestion from #4407** whose stated rationale ("do not add relative internal imports outside `cli/`") is not this repo's actual guideline — it appears to be a stale/generic rule from CodeRabbit's own learnings, not sourced from this repo's skill docs or lint script. Worth double-checking/correcting that CodeRabbit learning so it stops generating this suggestion on future PRs.
- No test coverage or behavior change involved (it's an import path only), and `#veryfront/agent/hosted/inference-credential.ts` does resolve correctly via the `"#veryfront/": "./src/"` catch-all in `deno.json`, so this won't break the build — the issue is purely convention/consistency, not correctness.

**Recommendation:** close this PR (or revert to `./inference-credential.ts`) rather than merging. If the CodeRabbit finding on #4407 is going to keep recurring, it's worth telling CodeRabbit (via a learning/comment) that same-directory relative imports are intentional in this repo, rather than "fixing" each instance it flags.

---

kojiwakayama:
Thanks — this is actionable and correct. The source and test are in the same module, where this repository intentionally uses same-directory relative imports; the boundary lint also does not prohibit this case. I am closing this follow-up rather than merging a convention regression. The original outside-diff suggestion on #4407 should be treated as inapplicable.

PR #4411, opened 2026-09-04
fix(agent): stop a failed detached run from crashing the process

kwakayama:
## Summary

`DetachedRunTracker.registerExecution`'s `.finally()` derives a new promise (`trackedExecution`) that's only ever read by `waitForDrain`, which only runs during shutdown. During normal operation nothing attaches a rejection handler to it, so a failed detached run became an **unhandled promise rejection and crashed the whole process** instead of just failing that one run.

## How this was found

Staging's `veryfront-agent` pods were crash-looping — 6 restarts each in 16 minutes — once [code#4407](https://github.com/veryfront/veryfront-code/pull/4407) let run-scoped inference credentials actually reach `requireSecureInferenceApiBaseUrl` in `src/provider/veryfront-cloud/shared.ts` for the first time (that check only runs when a run-scoped credential is present, which #4407 fixed the binding for). That check currently rejects staging's `VERYFRONT_API_URL`, which is a legitimate internal-cluster address (`http://veryfront-api.veryfront-staging.svc.cluster.local`) but plain HTTP, not HTTPS or loopback.

**That URL-policy question is separate and intentionally not addressed by this PR.** Whether a trusted internal cluster hostname should be treated as safe for run-scoped credentials is a real security tradeoff, not something to silently loosen as a side effect of a crash fix. This PR only stops the crash: the underlying request will still fail cleanly with a 400 (`CONFIG_INVALID`) until that policy question is resolved separately — it just won't take the process down anymore.

## Verification

- New regression test using a real `unhandledrejection` listener on `globalThis`, negative-controlled: reverted the fix and confirmed the test fails (`unhandled` captures the rejection) before restoring it.
- Full `detached-run-tracker.test.ts` suite: 7/7 passing.
- `deno fmt` / `deno lint` / `deno check` clean on both changed files.
- Semantic-unit-boundary lint gate unaffected (no new global-mutation surface).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented rejected background executions from triggering unhandled promise rejection errors or crashing the process.
  * Preserved existing shutdown behavior while ensuring execution failures remain safely handled.

### DIFF
--- a/src/agent/service/detached-run-tracker.test.ts
+++ b/src/agent/service/detached-run-tracker.test.ts
@@ -20,6 +20,46 @@ function deferred(): { promise: Promise<void>; resolve: () => Promise<void> } {
   };
 }
 
+/** Subscribes to unhandled rejections across Deno (WHATWG events) and Node/Bun (`process.on`). */
+function onUnhandledRejectionForTest(onReason: (reason: unknown) => void): () => void {
+  const target = globalThis as {
+    addEventListener?: (
+      type: string,
+      listener: (event: { reason: unknown; preventDefault(): void }) => void,
+    ) => void;
+    removeEventListener?: (
+      type: string,
+      listener: (event: { reason: unknown; preventDefault(): void }) => void,
+    ) => void;
+    process?: {
+      on?: (event: string, listener: (reason: unknown) => void) => void;
+      off?: (event: string, listener: (reason: unknown) => void) => void;
+    };
+  };
+
+  if (
+    typeof target.addEventListener === "function" &&
+    typeof target.removeEventListener === "function"
+  ) {
+    const listener = (event: { reason: unknown; preventDefault(): void }) => {
+      onReason(event.reason);
+      event.preventDefault();
+    };
+    target.addEventListener("unhandledrejection", listener);
+    return () => target.removeEventListener?.("unhandledrejection", listener);
+  }
+
+  const targetProcess = target.process;
+  if (typeof targetProcess?.on === "function" && typeof targetProcess.off === "function") {
+    const off = targetProcess.off.bind(targetProcess);
+    const listener = (reason: unknown) => onReason(reason);
+    targetProcess.on("unhandledRejection", listener);
+    return () => off("unhandledRejection", listener);
+  }
+
+  throw new Error("No supported unhandled-rejection API in this runtime");
+}
+
 describe("agent/detached-run-tracker", () => {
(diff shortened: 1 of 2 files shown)


kwakayama:
## Automated Code Review

**Score: 93/100 — Excellent.** Minimal, well-tested fix for a real production crash, with the fix and the unrelated policy question cleanly separated.

**Strengths**
- Correct root cause and correct fix: `execution.finally(...)` in `registerExecution` produces a *new* promise (`trackedExecution`) that only `waitForDrain` ever reads; nothing else attaches a handler, so a failed detached run became an unhandled rejection that took the whole process down. Adding `trackedExecution.catch(() => {})` suppresses that without altering `waitForDrain`'s behavior — it still reads the same `trackedExecution` reference stored in `activeExecutions`, and `waitForDrain` uses `Promise.allSettled`, which never throws regardless.
- Confirmed the original rejection is still observed and logged: the caller (`durable-chat-run-start.ts`) `await`s `detachedExecution` directly and its `onError` handler logs the failure, so this change only removes the *duplicate*, unhandled rejection surface — it doesn't silently eat errors.
- Good regression test: uses a real `unhandledrejection` listener on `globalThis` rather than mocking, and the PR description states it was negative-controlled (reverts the fix, confirms the test fails) — that's the right way to validate a fix for this class of bug.
- Tight scope and honest PR description: staging's plain-HTTP `VERYFRONT_API_URL` triggering `requireSecureInferenceApiBaseUrl` is explicitly called out as a separate security/policy question and *not* addressed here, rather than quietly loosened as a side effect. That's the right call for a crash-fix PR.
- Commit message and PR body clearly explain the "how this was found" chain (#4407 → staging crash loop → this fix), which will help whoever reads this in git blame later.

**Minor / optional suggestions (non-blocking)**
- Consider a one-line comment in the test itself (not just the PR description) noting it was negative-controlled by reverting the fix — that provenance is easy to lose once the PR description is no longer next to the code.
- Not required for this fix, but worth a follow-up issue: since `trackedExecution`'s rejection is now fully suppressed, if `waitForDrain` is ever refactored to stop being the sole reader, there'd be no signal at all from this promise — the current design is fine because the real error is already logged upstream, but that invariant is easy to break silently in a future change without a comment near `activeExecutions.set(...)` too.

This is a textbook example of a scoped incident-response fix — small diff, clear regression test, and no scope creep into the adjacent (and genuinely separate) TLS/loopback policy decision.

---

kojiwakayama (approved):

PR #4437, opened 2026-09-07
ci: standardize Sonar quality gate check

kwakayama:
## Summary

- rename the diagnostic scanner check from the transitional "sonar" label
- keep "SonarQube Cloud quality gate" as the single required Sonar check
- update the workflow contract and quality-gate runbook

The active GitHub ruleset was switched to the canonical check only after that check had passed on main.

## Verification

- deno task test:file tests/integration/ci/merge-quality-gate-workflow.test.ts
- codex review --base main

## Summary by CodeRabbit

- **CI/CD**
  - Renamed the SonarQube scan check to **“SonarQube Cloud scan”** for clearer status reporting.
  - The **“SonarQube Cloud quality gate”** remains the required check and now explicitly depends on the scan result.

- **Documentation**
  - Updated merge-correctness guidance to reflect the current SonarQube Cloud check names.
  - Removed references to the legacy **“sonar”** check and migration-only instructions.

### DIFF
--- a/.github/QUALITY_GATES.md
+++ b/.github/QUALITY_GATES.md
@@ -22,10 +22,9 @@ cannot receive `SONAR_TOKEN`. Fork pull requests still skip other protected
 dependency jobs and therefore fail this aggregate gate closed. Codecov
 reporting remains advisory.
 
-The scanner still emits the legacy `sonar` check during the ruleset migration.
-`SonarQube Cloud quality gate` is the canonical final check and depends on that
-scanner result. Remove the legacy check name only after the ruleset requires
-the canonical name and the canonical check has passed on `main`.
+The scanner emits the diagnostic `SonarQube Cloud scan` check.
+`SonarQube Cloud quality gate` is the only Sonar check required by the ruleset
+and depends on that scanner result.
 
 The active merge queue ruleset gives required checks at least 70 minutes to
 report a conclusion. This covers the longest configured dependency path: 60
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -462,9 +462,7 @@ jobs:
   sonar:
     runs-on: ubuntu-latest
     timeout-minutes: 35
-    # Keep this legacy check name until the repository ruleset requires the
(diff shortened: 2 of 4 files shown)


kwakayama:
## Code Review — Score: 45/100 (significant concern, fix before merge)

Small, well-scoped CI/docs change, but it ships a test regression against its own diff.

**Strengths**
- Tight, single-purpose diff (3 files, +6/-9) that matches its stated intent: drop the transitional `sonar` check name now that the ruleset requires the canonical `SonarQube Cloud quality gate` check.
- `.github/QUALITY_GATES.md` is updated in the same commit as the workflow change, so the runbook doesn't go stale.
- `tests/integration/ci/merge-quality-gate-workflow.test.ts` was correctly updated to assert the new `SonarQube Cloud scan` display name instead of the legacy one.
- PR description explains *why* the migration is safe now (ruleset already requires the canonical check on `main`) rather than just *what* changed.

**Concern — breaks an existing test**
- `src/config/cicd-coverage-workflow.test.ts:154` still asserts `assertStringIncludes(workflow, "name: sonar")` against `.github/workflows/cicd.yml`. After this PR, that job's `name:` field is `SonarQube Cloud scan` (`.github/workflows/cicd.yml:465`), and `assertStringIncludes` is a case-sensitive substring check — `"name: SonarQube Cloud scan"` does not contain `"name: sonar"`. I confirmed with a plain `grep -n "name: sonar" .github/workflows/cicd.yml` on the PR branch: no match. This test will fail on this PR's own commit.
- The PR's "Verification" section only lists running `merge-quality-gate-workflow.test.ts`, which is why this wasn't caught — `cicd-coverage-workflow.test.ts` covers the same workflow file but wasn't run/updated.

**Actionable fix**
- Update `src/config/cicd-coverage-workflow.test.ts:154` to assert the new name (e.g. `assertStringIncludes(workflow, "name: SonarQube Cloud scan")`), and re-run the full `deno task test` (or at least both `*-workflow.test.ts` files touching `cicd.yml`) before pushing.

Once that assertion is fixed, this is a clean, low-risk change — happy to re-review.


---

PR #4444, opened 2026-09-07
fix(proxy): settle WebSocket cleanup before close events

kojiwakayama:
The upstream WebSocket client emitted `close` while reader cancellation and its receive pump were still running. Under CI coverage, the next-message operation could outlive the test and fail the leak check.

Join opening, cancellation, receive and queued-write cleanup, then release the writer before emitting one close event. The test server also waits for its owned upgraded sockets to close.

Callback failures also retain ownership: a throwing `onopen` now reports the error and completes original stream cancellation before emitting close. The new regression fails on the previous implementation and passes with the fix.

Validation: deterministic deferred-cancellation and late-opening regressions failed before the fix. Native Deno loopback tests verify both stream locks are released at the close callback; 51 proxy suites and 544 steps pass under parallel coverage with leak tracing. Focused tests on the current main base, types, formatting, generated-reference checks, and both required reviews pass.

Current-head validation additionally passes all 14 focused cases, type checks, lint, formatting and exact-head review. Review claimed a duplicate `closed` declaration; inspection found one declaration per block, confirmed by type checking, so no deletion was needed.

### DIFF
--- a/src/proxy/websocket-client.test.ts
+++ b/src/proxy/websocket-client.test.ts
@@ -124,6 +124,42 @@ function silentStream(cancel?: () => Promise<void>): {
 }
 
 describe("upstream WebSocket client", () => {
+  it("settles stream ownership when the open callback throws", async () => {
+    const cancelStarted = Promise.withResolvers<void>();
+    const canceled = Promise.withResolvers<void>();
+    const { stream } = silentStream(() => {
+      cancelStarted.resolve();
+      return canceled.promise;
+    });
+    const socket = new UpstreamWebSocket("ws://upstream.test/_ws", new Headers(), () => stream);
+    const errored = Promise.withResolvers<void>();
+    const closed = Promise.withResolvers<void>();
+    let closeCount = 0;
+    socket.onopen = () => {
+      throw new Error("synthetic open callback failure");
+    };
+    socket.onerror = () => errored.resolve();
+    socket.onclose = () => {
+      closeCount++;
+      closed.resolve();
+    };
+    try {
+      await errored.promise;
+      await cancelStarted.promise;
+      assertEquals(socket.readyState, WebSocket.CLOSING);
+      assertEquals(closeCount, 0);
+    } finally {
+      canceled.resolve();
+      stream.close();
+    }
+    await closed.promise;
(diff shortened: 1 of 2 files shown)


kwakayama:
## Review: 90/100 — solid, well-tested fix for a real flakiness/leak source

**Summary:** Correctly sequences reader cancellation, the receive pump, and queued writes before settling `close`, fixing a genuine race where `close` could fire while cleanup was still in flight. Backed by targeted regression tests.

**Strengths**
- Root-causes the actual race: `#close` previously fired `reader.cancel()` without awaiting it, so a pending `read()` could outlive the `close` event — exactly the leak the new tests target.
- `#opening`/`#reading` tracking correctly handles the "late open while already closing" case too (verified by the second new test), not just the common path.
- `#settled` is still set synchronously at the top of `#close` before any `await`, so concurrent `#fail`/`#close` calls can't double-fire `onclose` — the async rewrite didn't reintroduce a reentrancy bug.
- New tests are precise: two deterministic unit tests isolate the deferred-cancellation and late-open races without timing hacks, and the 32-iteration native-loopback test gives real confidence against flakiness. Test helper changes (`sockets` as `Map<WebSocket, Promise<void>>`, awaiting each socket's close in `server.close()`) are a sensible, minimal way to make the server-side teardown deterministic too.
- Style and structure match the file's existing conventions (`Promise.withResolvers`, `describe`/`it`, comments explaining *why* not *what*).

**Minor concerns (non-blocking)**
- Asymmetric cleanup on the "late opened while closing" branch in `#open`: `connection.readable.cancel()` is now awaited, but `connection.writable` is never touched (no `getWriter()`/`close()`/`abort()`). Probably harmless since the underlying `WebSocketStream.close()` already tears down the transport, but worth a one-line comment (or matching cleanup) so the asymmetry reads as intentional rather than missed.
- `close()` → `onclose` latency is now bounded by reader cancellation + pump drain + pending writes instead of firing immediately. Looks fine for this bridge's usage in `main.ts` (just forwards the close code once it arrives), but flag if any caller assumed near-synchronous close.

Nothing here blocks merging — the concerns are polish, not correctness issues.

---

PR #4446, opened 2026-09-07
Bind executor sessions to deployment owners

kojiwakayama:
Executor sessions now bind allocations to an explicit global service or project deployment owner. This prevents an application-selected project from changing the owner of a platform service's executor and rejects changed owners in allocation, observation, renewal and cleanup replies.

Owner data is validated and snapshotted before asynchronous allocation. Existing source, image, generation and session retirement checks remain in force.

Validation: session and channel suites pass 71 test steps; formatting, lint and type checks pass. This is a dependent foundation change; it does not enable hosted execution.


kojiwakayama:
@codex review

PR #4447, opened 2026-09-07
Add authenticated executor allocator client

kojiwakayama:
The broker needs an authenticated client for the Operator's dedicated executor control endpoint. This adds fixed-origin HTTPS with certificate verification, a rotated projected broker token per call, exact owner-bound request envelopes, bounded payloads and responses, and no redirects or automatic retries.

Session retirement includes the original token read, DNS lookup and socket cleanup. This prevents a canceled allocation from freeing process capacity while underlying work remains. Service names such as `@example/agent` retain their exact registered identity.

Native Node coverage is included in the existing Sonar gate using the reviewed source-map reporter. Unit thresholds and quality gates remain unchanged. Validation: 12 HTTPS cases on Node 22; 28 native cases on host and Linux CI setup; client coverage 93.24%, transport 91.01%; 72 mapping, workflow and session steps; types, lint, formatting and both required reviews pass. No consumer wiring or executor activation is added.


kojiwakayama:
@codex review

PR #4462, opened 2026-09-08
fix(test): isolate working-directory mutations across suites

kwakayama:
Parallel Deno test files share the process working directory. The login and embedded-build fixtures were missing from the cwd lane, and three integration fixtures changed cwd while unrelated tests traversed repository files or launched child processes.

Run cwd-mutating fixtures in separate processes in both the normal suite runner and CI coverage shards, retaining parallel execution for other files. Anchor error-docs reads to the module URL and give the dependency-installer probe an explicit working directory. The existing cross-file cwd-lock probes remain concurrent.

Fixes veryfront/veryfront-issue-inbox#1043.

Validation:
- Reproduced the error-docs filesystem failures with the scaffold fixtures running in parallel; the same four-file command now passes all 24 steps.
- New suite-planning regressions failed before the fix; focused planner/coverage checks now pass all 38 steps, including complete and disjoint suite inventories.
- Full local pre-push checks passed: formatting, lint, typecheck, and 4,553 unit suites / 38,899 steps.
- Script typechecking, cwd-read audit, semantic audit, test layout, and diff checks passed.
- Local review of the implementation and the final formatting-only update found no actionable regressions. Tests requiring temporary-directory writes were independently verified outside the review sandbox.
- CI caught the scripts-specific formatting config; `deno task fmt:check` now passes, and the full pre-push gate passed again after the correction.
- Two full default-parallel root-integration runs passed, each with 605 suites / 3,980 steps. An intervening run hit a timer leak in the unchanged MDX shared-realm cache test; that test passed three isolated reruns and the next full integration run.
- Installed the pinned local Chromium binary after the first attempt exposed the missing browser; all three browser hydration checks then passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test execution reliability by isolating tests that change the working directory into separate processes.
  - Coverage runs now distribute test suites across batches while preserving working-directory isolation.
  - Documentation and integration checks now resolve repository files consistently, regardless of the launch directory.

- **Documentation**
  - Added guidance for handling working-directory-sensitive tests, including process isolation and reliable path resolution practices.

### DIFF
--- a/scripts/test/coverage-ci.ts
+++ b/scripts/test/coverage-ci.ts
@@ -2,6 +2,7 @@ import { walk } from "#std/fs/walk";
 import {
   buildTestProcessEnv,
   LOOPBACK_TEST_PERMISSIONS,
+  partitionDenoSuiteFiles,
   UNIT_DENO_TEST_ENV,
 } from "./suites.ts";
 
@@ -150,10 +151,12 @@ async function runShard(args: string[]): Promise<void> {
   const { planSuiteFiles } = await import("./run-suite.ts");
   const { files } = await planSuiteFiles({ suite: "coverage:unit", shard });
 
-  await runDeno(
-    buildDenoTestCommandArgs({ coverageDir, files }),
-    { ...UNIT_COVERAGE_ENV },
-  );
+  for (const batch of partitionDenoSuiteFiles(files, null)) {
+    await runDeno(
+      buildDenoTestCommandArgs({ coverageDir, files: batch }),
+      { ...UNIT_COVERAGE_ENV },
+    );
+  }
 
   await clearEmptyCoverageProfileJson(coverageDir);
   const lcov = await captureDeno(buildCoverageCommandArgs([coverageDir]));
--- a/scripts/test/run-deno-suite.ts
+++ b/scripts/test/run-deno-suite.ts
@@ -5,6 +5,7 @@ import {
   DENO_TEST_ENV,
   hasDenoPermissionFlag,
   LOOPBACK_TEST_PERMISSIONS,
+  partitionDenoSuiteFiles,
   PROVIDER_EGRESS_DENY_NET,
   UNIT_DENO_TEST_ENV,
 } from "./suites.ts";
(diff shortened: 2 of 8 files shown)


kwakayama:
## Automated review: 72/100 — solid isolation fix, blocked by a formatting failure

**Summary:** The fix itself is well-designed and well-tested, but `ci (format)` is currently red, so this can't merge as-is.

### Strengths
- **Correct core fix.** `partitionDenoSuiteFiles` (relocated to `suites.ts`) now splits any file in `CWD_MUTATING_TEST_FILES` into its own single-file batch while still packing non-mutating files up to `maxFilesPerProcess`, including the `null` (unbounded) case used by `coverage:unit`/`coverage:integration`. I traced the batching logic by hand against the new `run-suite.test.ts` assertions (`limit` ∈ `{2, 50, null}`) and it produces the expected grouping in each case.
- **Closes the actual gaps described in the issue**: `cli/auth/login.test.ts` and `cli/commands/build/embedded-preset-flags.test.ts` were missing from `UNIT_CWD_FILES`, and the three integration fixtures (`shell-adapter`, `standalone-auth-scaffold`, `missing-parent-race`) are now isolated via the new partitioning rather than just documented.
- **Root-causes rather than papering over it**: anchoring `error-docs-links.test.ts` reads to `import.meta.url`/`REPOSITORY_ROOT` and giving the dependency-installer probe an explicit `cwd: parentDir` are real fixes to cwd-dependent behavior, not just re-serialization.
- **Good regression coverage and docs**: the new `run-suite.test.ts` case exercises the batching invariant directly, and `tests/README.md` is updated with the "why" (a `withCwd` lock only coordinates callers that take it) so the next person doesn't reintroduce the bug.
- **Coverage loop is sound**: looping `runDeno` per batch in `coverage-ci.ts` with `--coverage-raw-data-only` is safe to call repeatedly into the same `coverageDir` since profile files are uniquely named per run.

### Concerns
- **`ci (format)` is failing right now** — `deno fmt --check` flags `scripts/test/run-suite.test.ts` for the import line and a couple of the new assertions not being wrapped to the project's line width. This is a one-command fix (`deno task fmt` / `deno fmt scripts/test/run-suite.test.ts`) but it does mean the PR as pushed doesn't pass CI.
- Minor: `localFileForDocsBaseUrl` now returns a `URL` instead of a `string`; error messages that interpolate it (e.g. `` `no "#${fragment}" anchor in ${docFile}` ``) will now print a `file://...` absolute path instead of the readable relative `docs/...` path. Not a bug, just a slightly less readable assertion message — worth a `.pathname`/relative-format tweak if you want the old readability back.
- Several other checks (`tests (integration)`, `coverage shard 1-4/4`, `ci (lint)`, `tests (binary e2e)`) were still in progress at review time — worth confirming they land green given this PR is precisely about isolation/coverage plumbing.

### Before merge
1. Run `deno fmt` locally and push the formatting fix — this is the only blocking item.
2. Once CI is fully green (lint + all shards + integration + binary e2e), this looks mergeable as-is.


---

PR #4463, opened 2026-09-09
fix(ci): support complete large review timelines

kojiwakayama:
## Description

The automated review gate fails with `review status unavailable` after a PR accumulates more than 500 timeline events. This currently blocks #4454 even though the review/comment sources remain within their limits.

Allow up to 5,000 timeline events in both evidence collection and failure finalization. Keep the 500-item limit on other sources and reject incomplete, malformed, or over-limit pagination. Complete ordering evidence is retained, including findings beyond the first 500 events.

The workflow loads this script from the default branch, so this fix must land before #4454 can reconcile its review status.

## Related Issue(s)

Unblocks the review gate for #4454.

## Type of Change

- [x] Bug fix
- [x] Test update

## Validation

- `deno task test:file scripts/ci/automated-review-gate.test.ts`: 8 suites, 160 steps pass. Three new regressions failed before the fix.
- `deno check --config=scripts/test.deno.json scripts/ci/automated-review-gate.test.ts`: passes.
- `git diff --check`: passes.
- Coverage includes successful evidence after 500 events, a later same-second finding, rate-limit failure finalization, partial pagination, and both source limits.

## Checklist

- [x] No public documentation changes are needed.
- [x] Tests demonstrate the failure and the fix.

### DIFF
--- a/scripts/ci/automated-review-gate.mjs
+++ b/scripts/ci/automated-review-gate.mjs
@@ -19,6 +19,8 @@ const REVIEW_REQUEST_MARKER =
 const FULL_SHA = /^[0-9a-f]{40}$/i;
 const REQUEST_KEY = /^[a-z0-9-]{1,64}$/i;
 const MAX_ITEMS_PER_SOURCE = 500;
+// Timelines include commits and lifecycle events as well as review evidence.
+const MAX_TIMELINE_ITEMS = 5000;
 const CODEX_REACTION_RETRY_DELAYS_MS = [1000, 2000, 4000];
 const MAX_TIMEOUT_TARGETS_PER_RUN = 25;
 const MAX_TIMEOUT_DISCOVERY_PAGES = 20;
@@ -1125,7 +1127,13 @@ export async function findAutomatedReview(
   )?.proof;
 }
 
-async function collectAll(github, endpoint, parameters, source) {
+async function collectAll(
+  github,
+  endpoint,
+  parameters,
+  source,
+  maxItems = MAX_ITEMS_PER_SOURCE,
+) {
   const items = [];
   for await (
     const response of github.paginate.iterator(endpoint, {
@@ -1137,8 +1145,8 @@ async function collectAll(github, endpoint, parameters, source) {
       throw new Error(`${source} pagination returned malformed data`);
     }
     items.push(...response.data);
-    if (items.length > MAX_ITEMS_PER_SOURCE) {
-      throw new Error(`${source} exceeded ${MAX_ITEMS_PER_SOURCE} items`);
+    if (items.length > maxItems) {
+      throw new Error(`${source} exceeded ${maxItems} items`);
     }
   }
   return items;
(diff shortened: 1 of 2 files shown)


kwakayama:
## Automated Code Review

**Score: 87/100** — Solid, narrowly-scoped fix to the review-gate's timeline pagination cap, with targeted regression tests and no reduction in the gate's fail-closed guarantees.

**Strengths:**
- The fix is minimal and precise: only the `pull request timeline` source gets the raised cap (500 → 5,000 via a new `MAX_TIMELINE_ITEMS` constant and an optional `maxItems` param on `collectAll`); the other four sources (`reviews`, `comments`, `reactions`, `events`, `statuses`) keep their existing 500-item cap. This avoids a blanket loosening of the anti-DoS/anti-abuse limit.
- Applied consistently at both call sites that read the timeline (`collectAutomatedReviewEvidence` and `currentReviewPropagationRetry`), so evidence collection and failure-finalization can't disagree about how much timeline history is "complete."
- Good, behavior-level test coverage rather than just exercising the code path: a success case with >500 timeline events, a same-second finding that would previously be hidden past the 500-item boundary, and a rate-limit failure finalized past 500 timeline items. The existing fail-closed cap test was also extended with a case that now fails past the new 5,000-item boundary (51×100-page timeline), so the raised ceiling still fails closed rather than becoming effectively unbounded.
- Still throws on malformed/incomplete pagination data exactly as before — no relaxation of the malformed/partial-page checks that make this gate fail closed.
- Clear PR description: explains the concrete failure mode (`review status unavailable` past 500 timeline events), the blocked PR it unblocks, and lists the exact validation commands run locally (8 suites/160 steps, `deno check`, `git diff --check`).

**Minor concerns (non-blocking):**
- Doubling the timeline page count (up to 50 pages instead of 5) happens on *every* evaluation, and `collectAutomatedReviewEvidence`/`currentReviewPropagationRetry` both now pull up to 5,000 timeline items — for very large/active PRs this adds real GitHub API call volume and latency per run. Worth keeping an eye on secondary rate limits if this gate runs frequently across many large PRs, though not a reason to block this fix.
- `5000` is a bare literal; the file already uses unformatted small constants (`500`, `25`, `20`) so this is consistent with existing style, just flagging it as the one truly "magic" number introduced.
- Since this script is itself the gate that decides whether *other* PRs (including this repo's own) pass automated review, it'd be worth a second pair of eyes given the self-referential blast radius (per the PR body, this must land before #4454 can reconcile) — not a defect in the diff itself, just a reason for a human to double-check the merge timing.

Nothing here rises to "must fix" — this is a good, well-tested change I'd be comfortable approving conditional on normal review sign-off.

---

kojiwakayama:
Reviewed `0696bbfd0644ccea5bf0827bc1bb4654dc84fead` for the failure it fixes and the scope of the change. The two-file change is justified: complete review timelines need a larger bounded allowance, and both evidence collection and failure finalization must use the same limit. The other sources retain their limits. The tests check that later findings remain visible and that incomplete or excessive pagination still fails closed. I found no actionable code issue in this diff.

Validation: `deno task test:file scripts/ci/automated-review-gate.test.ts` passed (8 suites, 160 steps); the explicit script type check passed with Deno 2.7.7. A completed Codex review is still missing for this head.

@codex review

PR #4466, opened 2026-09-09
feat(sandbox): allow bounded background command timeouts

kwakayama:
## Problem

`start_background_command` is documented for long-running sandbox work, but its tool schema accepts only `command`. The sandbox therefore applies its 30-second default, so callers cannot run a bounded multi-minute command for rollout recovery verification.

## Change

- Add optional `timeoutSeconds` to the hosted background-command tool.
- Preserve the fixed `/workspace` and project scope while forwarding it as the existing sandbox `timeout_seconds` option.
- Keep omission behavior unchanged; the sandbox remains authoritative for its configured maximum.

This is the framework prerequisite for veryfront-agent draft PR #1857 and inbox issue veryfront/veryfront-issue-inbox#1024.

## Verification

- Red: the focused regression showed the sandbox command request omitted `timeout_seconds`.
- Green: pinned Deno 2.7.7 focused test passes all 7 steps.
- `deno check src/sandbox/agent-service-tools.ts` and focused formatting pass.
- Normal pre-push gate passed with pinned Deno: generation, formatting, lint, typecheck, and the complete unit suite.

No dependency, sandbox timeout-policy, API, or production configuration change.

## Summary by CodeRabbit

* **New Features**
  * Background commands can now accept an optional timeout.
  * Configured timeouts are forwarded to the execution service in seconds.

* **Bug Fixes**
  * Preserved caller-provided execution options when starting background commands.

### DIFF
--- a/src/sandbox/agent-service-tools.test.ts
+++ b/src/sandbox/agent-service-tools.test.ts
@@ -85,10 +85,11 @@ function createCommandPayload(overrides: Record<string, unknown> = {}) {
 async function executeStartBackgroundCommand(
   tools: SandboxShellToolSet,
   command: string,
+  timeoutSeconds?: number,
 ): Promise<unknown> {
   const execute = tools.start_background_command?.execute;
   assertExists(execute);
-  return await execute({ command });
+  return await execute({ command, ...(timeoutSeconds === undefined ? {} : { timeoutSeconds }) });
 }
 
 describe("sandbox/agent-service-tools", () => {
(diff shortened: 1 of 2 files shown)


kwakayama:
## Code Review: 88/100 — Good, minor suggestions

Small, well-scoped feature that plumbs an existing `ExecOptions.timeout_seconds` field through to the `start_background_command` tool. Clean diff, focused test, no unrelated churn.

**Strengths**
- Tight scope: only touches the background-command path, reuses the existing `ExecOptions.timeout_seconds` type rather than inventing a parallel concept.
- Security-conscious plumbing: `getBackgroundCommandExecOptions` spreads caller-supplied `options` *first*, then overwrites with the project-scoped `getExecOptions()` and fixed `cwd`, so a caller can't smuggle in a different `cwd`/project scope via the new parameter — matches the PR description's claim.
- Schema validation (`v.number().int().positive().optional()`) rejects non-integer/negative/zero values at the tool boundary before they ever reach the sandbox request.
- New test asserts the full request body (`command`, `cwd`, `projectReference`, `timeout_seconds`) round-trips correctly when a timeout is supplied, and the pre-existing tests already cover the omitted case (no `timeout_seconds` key sent).
- Backward compatible: `timeoutSeconds` is optional, `options` is optional on `startBackgroundCommand`, omission behavior unchanged.

**Minor suggestions (non-blocking)**
- No client-side upper bound on `timeoutSeconds` — a caller could request an enormous value (e.g. `Number.MAX_SAFE_INTEGER`). The PR notes "the sandbox remains authoritative for its configured maximum," which is a reasonable design choice, but it'd be worth a one-line comment near the schema confirming that's an intentional deferral rather than an oversight, since it's not obvious from this file alone.
- Consider a quick test for the schema rejecting an invalid `timeoutSeconds` (0, negative, non-integer) to lock in that boundary behavior explicitly, rather than relying on it as incidental.
- CI was still running at review time (typecheck in progress); worth confirming it lands green before merge.

Overall: solid, minimal, well-tested change appropriate for its stated purpose (unblocking veryfront-agent PR #1857 / issue #1024).


---

kwakayama:
Withdrawing this task-created draft after tracing the server contract. Current veryfront-sandbox background commands ignore request timeout fields and use the operator-owned fixed JOB_EXEC_TIMEOUT_MS, so this client option has no effect and also widened the programmatic client beyond the intended timeout-only scope. Issue #1024 will use the existing measured 600-second staging job budget with a 540-second fixture and a 60-second margin. The local worktree is retained as investigation evidence; no code from this draft should ship.

PR #4474, opened 2026-09-10
perf(runtime): reduce request instrumentation and path resolution overhead

kojiwakayama:
## Description

Warm requests repeatedly sanitize the same phase labels and resolve absolute filesystem paths through the working directory. Cache up to 128 phase names that sanitization leaves unchanged, and avoid working-directory reads and rebuilding already canonical absolute paths.

Relative paths still follow the current directory. Windows roots, traversal normalization, phase cardinality, redaction, and protection against replaced built-in methods retain regression coverage. No public API changes.

Earlier seven-trial measurements on Deno 2.7.7 / Apple M5 showed full synthetic HTTP latency reductions of 29% for API routes, 24% for cached pages, 15% for uncached production SSR, and 9% for development SSR. Those captures predate the current base; they are supporting evidence, not a claim about deployed traffic. Profiles and experiment notes remain untracked.

## Type of Change

- [x] Performance improvement
- [x] Test update

## Validation

- `deno task test:file src/observability/request-profiler.test.ts src/observability/telemetry-error.test.ts src/utils/logger/redact.test.ts src/platform/compat/path/`: 10 tests, 249 steps pass.
- `deno task test:file src/security/path-validation/ src/security/path-validation.test.ts src/security/secure-fs.test.ts src/platform/adapters/runtime/shared/node-filesystem-adapter.test.ts src/utils/path-utils.test.ts tests/integration/server/production/api.test.ts tests/integration/server/production/ssr.test.ts`: 22 tests, 404 steps pass.
- Focused Deno formatting and lint checks, plus `git diff --check`, pass.

## Checklist

- [x] Public behavior and API signatures remain compatible; no reference regeneration is needed.
- [x] Added focused regression tests and ran the surrounding security and HTTP integration suites.

The focused portable path suite also passes under Node 24 and CI-pinned Bun 1.3.6. The new working-directory mocks select the active runtime, fixing the Node and Bun CI failures observed on the previous head. Deno path tests, formatting, and lint pass after this test-only correction.

### DIFF
--- a/src/observability/request-profiler.test.ts
+++ b/src/observability/request-profiler.test.ts
@@ -251,6 +251,72 @@ describe("request profiler", () => {
     assertEquals(isRequestProfilingEnabled("/_veryfront/page-data/blog.json"), true);
   });
 
+  it("keeps repeated phase names sanitized and request durations independent", async () => {
+    const capture = (duration: number) =>
+      runWithRequestProfiling(
+        { category: "html", method: "GET", pathname: "/bench/phases" },
+        () => {
+          markRequestProfilePhase("render.ssr", duration);
+          markRequestProfilePhase("phase?token=<TOKEN>", duration);
+          markRequestProfilePhase("x".repeat(MAX_OBSERVABILITY_NAME_LENGTH + 1), duration);
+          return Promise.resolve(finalizeRequestProfiling(200));
+        },
+      );
+    const first = await capture(1);
+    // Exercise more distinct names than the per-request phase budget across sessions.
+    for (let i = 0; i < MAX_REQUEST_PROFILE_PHASES + 1; i++) {
+      await runWithRequestProfiling(
+        { category: "html", method: "GET", pathname: "/bench/phases" },
+        () => {
+          markRequestProfilePhase(`other.${i}`, 1);
+          finalizeRequestProfiling(200);
+          return Promise.resolve();
+        },
+      );
+    }
+    const second = await capture(2);
+    assertExists(first);
+    assertExists(second);
+    assertEquals(first.phases["render.ssr"], 1);
+    assertEquals(second.phases["render.ssr"], 2);
+    assertEquals(first.phases["phase?token=[REDACTED]"], 1);
+    assertEquals(second.phases["phase?token=[REDACTED]"], 2);
+    assertEquals(
+      Object.keys(second.phases).every((name) => name.length <= MAX_OBSERVABILITY_NAME_LENGTH),
+      true,
+    );
+    assertEquals(JSON.stringify(second).includes("<TOKEN>"), false);
+  });
+
+  it("does not expose raw phase names to replaced Set methods", async () => {
+    const originalHas = Set.prototype.has;
+    const originalAdd = Set.prototype.add;
+    const observed: unknown[] = [];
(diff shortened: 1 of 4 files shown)


kwakayama:
## Code Review: Score 88/100 — Good, minor suggestions

Solid, well-tested perf work in a security-sensitive area (path resolution, telemetry redaction). No correctness bugs found after tracing edge cases by hand; the changes carry their own regression tests including two adversarial ones.

**Strengths**
- `portableNormalize`'s POSIX fast-path (`src/platform/compat/path/portable.ts:204`) is a sound short-circuit: the guard regex (`[\\]|\/\/|\/\.{1,2}(?:\/|$)`) correctly forces the slow path for backslashes, doubled separators, `.`/`..` segments (including at the very start or end of the string), while leaving genuinely canonical names (dotfiles, `...foo`, unicode) untouched — I traced ~10 edge cases against the slow path and they agree.
- `portableResolve`'s right-to-left "find the last fully-qualified segment" pre-scan is a correct optimization, not just a rewrite: it only short-circuits on segments whose branch in the main loop fully replaces `resolved` with no dependency on prior state (POSIX absolute, Windows drive-absolute/UNC), and correctly excludes Windows root-relative (`\foo`) and drive-relative (`C:foo`) segments from being scan-stoppers since those need borrowed context from an earlier/cwd value. The new tests explicitly assert zero `Deno.cwd()` calls when a qualified path is present and fresh reads when it isn't.
- The `sanitizedPhaseNames` cache in `request-profiler.ts` is a valid memoization: `sanitizeTelemetryText` is pure/deterministic for a fixed `maxLength`, so caching "name → itself" for names redaction left unchanged is safe. Crucially, it never caches a name that required redaction (`sanitized === name` gate), so credential-bearing input is always recomputed and never retained as a cache key — matches the stated invariant.
- Good defense-in-depth: the new cache is accessed via saved `Reflect.apply`/prototype-method references rather than `set.has()`/`set.add()`, consistent with the existing primordial-safety pattern in `portable.ts`, and there's a dedicated test proving replaced `Set.prototype.has`/`add` never observe the raw (potentially sensitive) phase name.
- `resetRequestProfiles()` was correctly updated to clear the new cache, keeping test isolation intact for the module-level singleton.

**Concerns / suggestions (non-blocking)**
- The 29%/24%/15%/9% latency numbers in the description are asserted but not reproducible from anything in the PR (profiles/experiment notes are "untracked," per the author's own caveat). Fine as supporting color, but not verifiable — worth linking a benchmark script or gist next time.
- `sanitizedPhaseNames` is a process-wide, unbounded-lifetime cache (only cleared via the test-only `resetRequestProfiles`), capped at 128 entries for the whole process rather than per-tenant/session. Low risk given it only stores already-"clean" bounded strings, but worth a one-line comment noting it's intentionally global (it already has a comment about the redaction invariant, just not about scope).
- Minor: in `portableResolve`, the pre-scan calls `analyzeRoot` on Windows paths during the scan and then the main loop calls it again for the same segments once resolution starts — duplicate work on the (non-hot) Windows path. Not worth blocking on given the stated target is Linux/Deno prod, but could memoize if it matters later.

CI was still finishing several shards/lint at review time; nothing observed so far suggests a red build tied to this diff.

---

kojiwakayama:
@codex review

PR #4487, opened 2026-09-12
fix(agent): authenticate platform MCP tools with the run invocation credential

kwakayama:
## Problem

Hosted agent runs can read project files but cannot write them: every `create_file`/`update_file`/`delete_file` from a schedule-fired or control-plane-verified run fails with `[authorization-denied] Access denied`, while `get_file`/`list_files` succeed. This silently breaks every template that persists records from agents (verified live on `agentic-inbox-processing-outlook`; `agentic-job-submission-processing` last wrote successfully on 2026-09-02, before the current runtime rollout).

Full analysis: veryfront/veryfront-issue-inbox#1327

## Root cause

Since #4214 the hosted agent-stream path hands `withVeryfrontPlatformRemoteTools` the `projectRuntimeToken` (`ctx.proxyToken`) — the stream credential signed with `RUNTIME_SCOPES.runsStream`, which carries `files:read` but no `files:write`. The run invocation credential minted expressly for platform tool callbacks (`mintRuntimeInvocationAuthToken` → `runsExecute`, which does carry `files:write`) arrives as `payload.credentials.authToken` but only reached framework API calls, never the MCP tool source. veryfront-api's scope ladder then correctly denies the write.

## Fix

Pass the verified request credential (`apiAuthToken`) to the framework-owned platform MCP tool source. Project-code-visible surfaces are unchanged and keep the runtime's project credential: the agent environment (`VERYFRONT_API_TOKEN` env var), the sandbox (`projectAgentSandbox.authToken`), and Studio remote tools. This preserves #4214's isolation intent — project code and sandbox never see the run credential — while restoring the authority the invocation credential exists to provide ("The runtime calls integration tools back through the platform with this token").

`apiAuthToken` already falls back to `ctx.proxyToken` when the payload carries no credential, so behavior for token-less requests is unchanged.

## Red/green

The existing tests pinned the regressed routing. Red: flipping the platform-MCP authorization assertions to expect the request credential fails on main (3 tests: control-plane stream config, legacy stream policy identities, credential isolation). Green: the one-line token routing change. The isolation test's environment assertion (`VERYFRONT_API_TOKEN: "run-scoped-token"`) still passes untouched, demonstrating the env boundary holds.

`deno task test:file src/server/handlers/request/agent-stream.handler.test.ts` — 3 passed (67 steps), 0 failed. `deno check` + `deno lint` + `deno fmt --check` clean on touched files.

https://claude.ai/code/session_019B9SYdGjpN6sQ4424NLv91

## Summary by CodeRabbit

- **Bug Fixes**
  - Platform MCP requests now authenticate with the verified run invocation credential, improving authorization for platform tool callbacks.
  - Project-code-visible surfaces continue using the runtime’s project credential.
  - Updated request-handling tests to reflect the corrected authentication behavior.

### DIFF
--- a/src/server/handlers/request/agent-stream.handler.test.ts
+++ b/src/server/handlers/request/agent-stream.handler.test.ts
@@ -956,7 +956,7 @@ describe("server/handlers/request/agent-stream.handler", () => {
           platformMcpFetchCalls += 1;
           assertEquals(
             new Headers(observeFetchRequestInit(init).headers).get("authorization"),
-            "Bearer run-scoped-token",
+            "Bearer request-scoped-user-token",
           );
           return Promise.resolve(
             new Response(
@@ -2768,7 +2768,7 @@ describe("server/handlers/request/agent-stream.handler", () => {
         assertEquals(String(url), `${TEST_PUBLIC_API_ORIGIN}/mcp`);
         assertEquals(
           new Headers(observeFetchRequestInit(init).headers).get("authorization"),
-          "Bearer run-scoped-token",
+          "Bearer request-scoped-user-token",
         );
         const request = JSON.parse(String(observeFetchRequestInit(init).body)) as {
           id: string;
@@ -3063,7 +3063,7 @@ describe("server/handlers/request/agent-stream.handler", () => {
         }
 
         if (String(url) === `${TEST_PUBLIC_API_ORIGIN}/mcp`) {
-          assertEquals(authorization, "Bearer run-scoped-token");
+          assertEquals(authorization, "Bearer request-scoped-user-token");
           capturedMcpRequest = {
             url: String(url),
             authorization,
@@ -3131,9 +3131,12 @@ describe("server/handlers/request/agent-stream.handler", () => {
     assertStringIncludes(capturedSystem ?? "", "project_reference=support-agent-fork");
     assertStringIncludes(capturedSystem ?? "", '<project_context>\nproject_reference: "proj-1"');
     assertEquals(capturedProjectContextToken, "run-scoped-token");
+    // The platform MCP source is framework-owned and authenticates with the
+    // verified run invocation credential; only project-code-visible surfaces
+    // (the agent environment above) keep the runtime's project credential.
     assertEquals(capturedMcpRequest, {
       url: `${TEST_PUBLIC_API_ORIGIN}/mcp`,
-      authorization: "Bearer run-scoped-token",
+      authorization: "Bearer request-scoped-user-token",
     });
     assertEquals(capturedAllowedRemoteTools, ["list_projects", "search_knowledge"]);
     assertEquals(capturedRemoteToolNames, ["search_knowledge", "list_projects"]);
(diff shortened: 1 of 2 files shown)


kwakayama:
## Automated Code Review

**Score: 88/100** — Good: a tight, well-reasoned fix for a real authz bug, with the isolation boundary it depends on independently verified by tests.

### Strengths
- **Minimal, targeted diff**: exactly one production line changed (`token: projectRuntimeToken` → `token: apiAuthToken` at `agent-stream.handler.ts:1199`), scoped to the single call site of `withVeryfrontPlatformRemoteTools`. No other consumer of `projectRuntimeToken` (agent env vars, `projectAgentSandbox.authToken`, Studio remote tools) was touched, which is exactly what's needed to preserve the #4214 isolation contract.
- **Root cause is well-explained and checks out**: I traced `apiAuthToken` (`payload.credentials?.authToken || ctx.proxyToken || ""`, line 1023) vs `projectRuntimeToken` (`ctx.proxyToken || ""`, line 1038) — they're genuinely different credentials, and only the platform MCP source (a framework-internal fetch to `${apiUrl}/mcp`, never surfaced to project code) now gets the former.
- **The isolation invariant is tested, not just asserted in prose**: the "keeps request-scoped credentials out of project agent environments" test confirms `VERYFRONT_API_TOKEN` in the agent env still resolves to the project token while the platform MCP request now correctly carries the new credential — this is the test that actually matters for the security claim in the PR body, and it passes.
- **Good code comments**: the added block at `agent-stream.handler.ts:1033-1037` clearly documents *why* the platform MCP source gets different treatment than sandbox/env, which will save the next reader from re-deriving this.
- **Red/green discipline**: PR body documents the failing-on-main assertions and the passing fix, which is easy to verify against the diff (3 hunks in the test file, all flipping the expected bearer token for `/mcp` calls).

### Minor concerns (non-blocking)
- No new test exercises the actual previously-broken path (`create_file` succeeding end-to-end); the fix is validated via header/token assertions rather than a scoped-permission simulation. Given the test harness's mocked fetch, this may not be practical here, but worth confirming against a real `files:write`-scoped token in staging before/after rollout.
- The fallback semantics (`apiAuthToken` falling back to `ctx.proxyToken` when the payload has no credential) are correct but mean any upstream bug in how veryfront-api mints/attaches `payload.credentials.authToken` would silently degrade back to the old (broken-for-writes) behavior rather than fail loudly — that dependency lives outside this repo, so just flagging it as an integration risk to confirm once deployed.
- Test token naming (`"request-scoped-user-token"`) reads as if it's the raw end-user token; the PR body clarifies it's actually the minted run-invocation credential (`runsExecute` scope). A rename to something like `"run-invocation-token"` would make the test intent match the PR's own terminology, but this is cosmetic.
- CI was still running several checks (coverage shards, lint, JS/TS CodeQL) at review time — worth confirming full green before merge, though `typecheck`, `format`, and `test-layout` already passed.

Overall: sound, narrowly-scoped security fix with the right test backing its central claim.

---

kojiwakayama (approved):

PR #4494, opened 2026-09-14
fix(run-events): tool results carry parentMessageId like tool call starts

kojiwakayama:
## Why

The conversation run event encoder already names the assistant message on `TOOL_CALL_START` (`parentMessageId`). It did not do the same on `TOOL_CALL_RESULT`, so the API had to work out a tool result's turn from the made-up message id `<assistantMessageId>:tool:<toolCallId>` when it ingests the event. Recovering it that way is a stopgap. The encoder should state the parent directly, the same way it does for starts.

## What changed

- `getToolCallResultPayloadSchema` (`src/run-events/payload.ts`) now has an optional `parentMessageId` field. It must be a non-empty string, the same rule `TOOL_CALL_START` uses.
- `ConversationRunEventEncoder` (`src/agent/conversation/run-events.ts`) sets `parentMessageId` to the active assistant message id on all four places that emit a tool result: `tool-input-error`, `tool-output-available`, `tool-output-error` and `tool-output-denied`. When no message is open, the key is left out, which matches `TOOL_CALL_START`.
- The public JSDoc for `getToolCallResultPayloadSchema` and the generated `docs/api-reference/veryfront/run-events.md` now describe `parentMessageId`.
- The made-up `messageId` is unchanged, because consumers and the API's ingest-time recovery depend on its format.

## Contract fixture

No change. `tests/fixtures/contracts/native-run-events.json` only covers the native frame types (for tool calls, that is `TOOL_CALL_STATUS_CHANGED`) and has no `TOOL_CALL_RESULT` sample. The fixture and its SHA-256 hash stay as they are, so veryfront-api needs no companion fixture PR. The API's typed `TOOL_CALL_RESULT` schema on its main branch already accepts an optional `parentMessageId`.

## Tests

- `src/agent/conversation/run-events.test.ts`: for each of the four branches, one test checks that `parentMessageId` equals the active message id inside an open message, and another checks that the key is missing when no message is open. Both tests also pin the made-up `messageId`. These failed before the change and pass after it. The existing exact-shape test now includes the new key.
- `src/run-events/payload.test.ts`: the schema accepts a string `parentMessageId` and rejects a number or an empty string. This failed before the change, because the passthrough schema let a number through.
- Every other test that uses the encoder or the fixture passes: native-run-events, hosted-lifecycle, legacy-run-read-adapter, run-chunk-mirror, run-stream-mirror, durable-run-event-sink, request-protocol, vocabulary, chat/conversation, durable, run-event-normalization, lifecycle-run-event-adapter, sse-parser, api-contract-fixture, and the two fixture contract tests.

## Out of scope

`lifecycle-run-event-adapter.ts` builds its own tool records and uses a different shape: `messageId` on starts and no parent on results. This PR leaves it unchanged; any change there is a separate piece of work.

https://claude.ai/code/session_01CqY82FYm9aUkyawWmh4T9n


kojiwakayama:
@codex review

kwakayama:
## Review: 92/100 — excellent

Small, well-scoped bugfix that does exactly what it says: `TOOL_CALL_RESULT` now carries `parentMessageId` the same way `TOOL_CALL_START` already did, removing the API's need to reverse-engineer a result's turn from the synthetic `<assistantMessageId>:tool:<toolCallId>` message id.

**Strengths**
- Correctness verified by reading: all four result branches (`tool-input-error`, `tool-output-available`, `tool-output-error`, `tool-output-denied`) use `...(this.activeMessageId ? { parentMessageId: this.activeMessageId } : {})`, an exact mirror of the existing `TOOL_CALL_START` pattern a few lines up — no drift between the two.
- Test coverage is thorough: new tests cover both the "inside an open message" and "no message open" cases for all four branches, plus schema-level tests in `payload.test.ts` asserting the field accepts a string and rejects a number/empty string. The existing exact-shape test was updated rather than left stale.
- Backward-compatible: the field is optional and omitted (not `null`/`undefined`-valued) when absent, matching how `TOOL_CALL_START` behaves — no shape change for existing consumers.
- Good scope discipline: PR description explicitly calls out why the contract fixture doesn't need touching (no `TOOL_CALL_RESULT` sample in it) and explicitly excludes `lifecycle-run-event-adapter.ts`'s different shape as out of scope rather than silently leaving it inconsistent.
- Docs (JSDoc + generated `run-events.md`) updated alongside the schema change.

**Minor nit (non-blocking)**
- The `...(this.activeMessageId ? { parentMessageId: this.activeMessageId } : {})` spread is now duplicated 5 times (1 start + 4 result branches) in `run-events.ts`. Pre-existing style, not introduced by this PR, but now would be a reasonable time to extract a small `parentMessageIdField()` helper if you want to avoid a 6th copy-paste next time this pattern is touched. Not asking for a change here.

No security, design, or test-coverage concerns. CI was still running several jobs at review time (coverage shards, lint, typecheck, integration/e2e) with everything completed so far green — worth a final glance before merge, but nothing in the diff itself suggests risk there.

---

PR #4500, opened 2026-09-16
feat(agent): name the accepting replica on hosted run starts

kojiwakayama:
## Why

A hosted agent service keeps each detached run's session in the memory of the replica that accepted the start. Run control for that run (cancel, resume) only works on that replica. Starts usually arrive through a load-balanced address, so a later control request through the same address can land on a replica without the session and do nothing. With two replicas, a cancellation sent through the shared address missed its run in 2 of 5 cases.

The project agent-stream route already solves this: it names its replica in `x-veryfront-runtime-owner-invoke-url` so the caller can reach it again. The hosted default-chat start route (`POST /api/runs`) did not.

## What changes

- `createHostedAgentServiceRouteSet` accepts an optional `resolveRuntimeOwnerInvokeUrl(request)`. It defaults to the existing resolver in `src/internal-agents/runtime-owner.ts` (`VERYFRONT_RUNTIME_OWNER_HOST`, `POD_IP`, or a network interface).
- `handleDurableChatRunExecuteRequest` adds `x-veryfront-runtime-owner-invoke-url` when it accepts a start (202).
- Nothing changes for any other response. A resolver that throws or returns nothing also leaves the accepted response as it was, since the run is accepted either way and control keeps using the shared address.

The advertised URL keeps the resolver's existing shape, the same value the agent-stream route sends. A caller uses its origin to reach the replica.

## Tests

- an accepted default-chat start names the replica that accepted it
- a rejected start (401) names none
- an accepted start is unaffected when the replica address cannot be resolved

The route-set test factory defaults the resolver to `null`, so existing tests do not depend on the host's network interfaces. `deno check`, `deno fmt --check`, `deno lint`, every `src/agent/service` test suite, and `docs:api-reference:check` pass.

https://claude.ai/code/session_01NonPHDcbWsisd2GFB68ALo

## Summary by CodeRabbit

* **New Features**
  * Durable chat start responses can now include the runtime owner’s invocation URL when available.

* **Bug Fixes**
  * Unauthenticated requests and cases where the runtime owner URL cannot be resolved no longer advertise a replica URL.
  * Chat starts continue returning a successful response even when runtime owner resolution fails.


kojiwakayama:
@codex review

kojiwakayama:
@codex review

PR #4512, opened 2026-09-16
fix(agent): preserve provider stream failure at replay boundary

kwakayama:
Preserves the provider-stream failure when a replay turn cannot complete, instead of reporting a replay-boundary error.\n\nValidation: `deno task test:file src/internal-agents/run-stream.test.ts`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated streamed responses to report the clearer error message “Provider stream failed” when a provider stream fails.
  - Pending and subsequent completion requests now consistently receive the stream failure instead of the previous replay-turn error.


kwakayama:
Closing: this only changes the surfaced replay-boundary message and does not address the provider-stream failure. The underlying failure needs a separate root-cause fix.

PR #4518, opened 2026-09-17
chore: open the 0.1.1260-rc prerelease line so main publishes again

kojiwakayama:
## Why

`veryfront-code` publishes an npm build only when `version-check` reports `is_stable=false` (`cicd.yml`, the `prerelease` job). `deno.json` has held the **released** `0.1.1259` — a plain semver — since veryfront-code#4501, so `is_stable=true` and every merge to main since has logged:

```
::notice::Stable version unchanged: 0.1.1259 → skipping publication
```

and published nothing. That includes **veryfront-code#4514**, the fix for veryfront/veryfront-issue-inbox#1480, which is on main (`53b775c0a4`) but exists in no consumable build.

That matters for the 18 Sep customer demo: `veryfront-agent` pins `veryfront: "0.1.1259-rc.19435"`, an RC of the *previous* version, so the agent runtime still resolves an undated `anthropic/claude-haiku-4-5` to the 4096-token fallback and still produces `Provider replay turn failed before its boundary`.

## What this does

Moves the base to `0.1.1260-rc`, so CI publishes `0.1.1260-rc.<run>` per main build again — the same RC line `veryfront-agent` already consumes.

- `deno.json` → `0.1.1260-rc`
- `src/utils/version-constant.ts` → kept in sync, as `scripts/release.ts` does
- `src/html/hydration-script-builder/hydration-runtime.generated.ts` regenerated via `deno task generate` (it embeds the version)

`lint`, `lint:test-typecheck` and `docs:api-reference:check` pass.

## Why not `deno task release`

The release script commits, tags `v<version>` and pushes straight to `main` — it bypasses review, and a tag is wrong for opening a prerelease line rather than cutting a stable one.

## Still required after this merges

1. **Bump `veryfront` in `veryfront-agent/package.json`** to the `0.1.1260-rc.<run>` this publishes
2. **Build and promote the agent image** — staging, then production

Neither is automatic, and the demo runs on veryfront.com.

Refs veryfront/veryfront-issue-inbox#1480

## Summary by CodeRabbit

* **Chores**
  * Updated the application version to `0.1.1260-rc`.
  * Reformatted test task configuration without changing its behavior.


kojiwakayama:
@codex review

PR #4524, opened 2026-09-19
test(dev): stop the free-port cache guard test racing parallel jobs for its port

kwakayama:
## Problem

`cli/commands/dev/dev-cache-guard.integration.test.ts` → "still clears the shared ESM cache when the dev port is free and nothing persists it" failed on #4521 ([job](https://github.com/veryfront/veryfront-code/actions/runs/35449761506/job/105914828667)) with `a free dev port must still clear stale caches: false !== true`, on a change that doesn't touch the dev command.

The test binds an ephemeral port, releases it, then calls `clearLocalCachesIfPortFree(port)` with the real `isPortAvailable` probe. `tests (integration)` runs files in parallel, so another job can bind the just-released port before the probe runs. The probe then reports the port taken and the guard correctly refuses to clear: a time-of-check race in the test, not a product bug.

## Fix

The free-port case now injects a probe that reports the port free (and asserts the requested port was the one probed), so the test pins down the guard's decision: free port + non-persistent cache ⇒ clear.

- The real probe's free-port behaviour is already tested in `cli/commands/dev/port-fallback.test.ts` (`isPortAvailable`), which retries across several released ports because of this same race.
- The taken-port case keeps the real probe on a port the test still holds, which can't race.

## Verification

- `deno test cli/commands/dev/dev-cache-guard.integration.test.ts`: 6/6 steps pass
- `deno fmt`, `deno lint`, `deno check` clean
- `lint:test-semantic-dispositions`, `lint:anti-slop`, `lint:testing-front-door`, `lint:cwd-relative-test-reads`, `lint:test-typecheck`, `lint:ban-test-only` all pass


kwakayama:
## Code Review — Score: 85/100 (Good)

Small, well-targeted test-only fix for a genuine flaky-test root cause (time-of-check/time-of-use race on a released ephemeral port under parallel integration runs).

**Strengths**
- Correct diagnosis: the old test released a real port and then probed it for real, so a parallel job could grab it between release and probe — that's a test race, not a product bug, and the PR body explains this clearly.
- Fix stubs the probe instead of weakening the assertion, and *also* strengthens the test by asserting the requested port was the one actually probed (`assertEquals(probedPorts, [port], ...)`) — that's new coverage, not just a flake patch.
- Correctly leaves the taken-port case (`holdPort()` + real probe) alone, since a port the test itself holds can't race.
- Verified the parameter order against `clearLocalCachesIfPortFree`'s real signature (`port, clear, probe, persists`) — the stub is wired to the right slot.
- Confirmed the PR's claim that free-port probing behavior of the real `isPortAvailable` is covered elsewhere in `port-fallback.test.ts` (with its own retry-for-race logic) — so no coverage is lost by stubbing here.

**Minor nit (non-blocking)**
- `const port = 4173;` is a magic number with no comment on *why* that value (it's Vite's conventional preview port, but that's not stated). Since the probe is fully stubbed, the actual value is arbitrary — a one-line comment (e.g. "arbitrary; the probe below is stubbed so no real port is bound") or a clearly-named constant would save the next reader a moment of "why this number."

Nothing here blocks merge; the nit is cosmetic. CI was still running at review time (bot review comments from CodeRabbit/Codex are rate-limited/in-progress) — worth confirming green before merge, but not part of this review.

---

PR #4527, opened 2026-09-19
test(eval): allow for Node timer and Date.now granularity in progress-duration checks

kwakayama:
## Problem

`tests (node shard 2/2)` failed on main ([run 35466315411](https://github.com/veryfront/veryfront-code/actions/runs/35466315411/job/105959483969)):

```
✖ keeps the target duration when grading times out
  Error: progress duration 39ms covers the wait
    at src/eval/runner.test.ts:950
```

The test sets `recordTimeoutMs: 40` and asserts the `record-finished` progress duration is `>= 40`. "reports the whole case duration in progress" does the same with a 30ms metric delay and `>= 30`. Two effects make an exact bound race the clock:

- the runner measures with `Date.now()` (`src/eval/runner.ts`), which truncates to whole milliseconds;
- Node schedules timers against a loop time cached at the start of the tick, so a timer can fire about 1ms early. `src/eval/agent-service-hardening.test.ts` already allows for this with `NODE_TIMER_EARLY_FIRE_MS`.

## Fix

Both assertions now accept up to `TIMER_WAIT_TOLERANCE_MS = 2` below the delay, one millisecond for each effect, with a comment giving the reason. What the tests guard is that progress reports the whole case, grading wait included, rather than the target's own measure. Those are 1ms and 7ms in these tests, so 28ms and 38ms still tell them apart.

## Verification

- `deno task test:file src/eval/runner.test.ts`: 39/39 steps pass
- `deno fmt`, `deno lint`, `deno check` clean; `lint:anti-slop`, `lint:test-semantic-dispositions`, `lint:testing-front-door` pass
- the Node leg (`tests (node shard 1/2, 2/2)`) runs in this PR's CI


kwakayama:
## Code Review — Score: 92/100 (Excellent)

Minimal, well-justified test fix for a real timing flake, with the root cause correctly diagnosed and verified.

**Strengths:**
- Root cause is concrete and verified, not guessed: `runner.ts` measures with `Date.now()` (truncates to whole ms) and Node can fire a timer ~1ms early against its cached loop-tick time, matching the observed `39ms` vs. asserted `>= 40` failure from the linked CI run.
- Tolerance is documented at the definition site with the *why*, consistent with the existing `NODE_TIMER_EARLY_FIRE_MS` precedent in `agent-service-hardening.test.ts` — this PR reasonably extends it by 1ms to also cover `Date.now()` truncation, since that test measures differently.
- Margins are checked, not just asserted: 28ms/38ms vs. the 1ms/7ms target-only durations they must stay distinguishable from — the tolerance doesn't quietly widen the assertion into meaninglessness.
- Change is scoped to exactly the two flaky assertions, no unrelated cleanup; PR description and commit message both explain the problem, fix, and verification (local test run, fmt/lint/check, anti-slop lints) clearly.
- Test-only diff (no production code touched) keeps blast radius essentially zero.

**Minor nit (non-blocking):** `TIMER_WAIT_TOLERANCE_MS` and `NODE_TIMER_EARLY_FIRE_MS` are now two similarly-named constants encoding overlapping timer-slop reasoning in different test files — not worth blocking on, but could be worth a shared helper/comment cross-reference if a third test needs this pattern.

No functional, security, or coverage concerns. Approve once CI is green.

---

PR #4529, opened 2026-09-19
Use authorized local web fetch in hosted Claude runs

kwakayama:
Hosted Claude runs could select Anthropic native `web_fetch` even when an authorized local fetch tool was available. Vertex rejects that native tool type, breaking research requests.

Prefer the existing protected local fetch implementation for authorized web-fetch requests. Native web search remains available; explicit denials and empty allowlists retain their existing behavior.

Validation: reproduced the wrong native selection with a failing regression, then passed 574 hosted/runtime tests (1,059 steps), Deno type checking, formatting, and local diff reviews. Reviewer subprocesses reported read-only temporary-directory limitations; the full test commands passed in the writable worktree.


kwakayama:
## Automated code review — Score: 85/100 (Good)

Small, well-targeted fix that correctly removes a redundant guard rather than adding a workaround.

**Strengths**
- Root-cause fix, not a patch-over: `shouldIncludeHostedWebFetchFallback` no longer special-cases "provider already has a native `web_fetch`" to skip the local fallback. Downstream, `localProviderToolNames` (chat-runtime-tool-assembly.ts:664-671) already excludes any provider-native tool name that made it into `sortedLocalTools`, so removing the early guard doesn't reintroduce the duplicate-registration bug it used to prevent — it just moves the dedup to the one place that already owned it.
- Correctly scoped: `getProviderNativeToolNames` shows only Anthropic exposes `web_fetch` natively (OpenAI only has `web_search`), so this change is effectively an Anthropic/Vertex-only behavior fix, matching the stated bug (Vertex rejecting the native tool type).
- Net diff is tiny (-5/+0 in the implementation, 4-line test update) — low blast radius, easy to reason about and to revert if wrong.
- PR description is clear about intent, includes a regression-test-first validation story, and correctly flags the pre-existing "empty allowlist" and "explicit denial" paths as untouched (verified — those branches in `shouldIncludeHostedWebFetchFallback` are unchanged).
- CI is green on the checks that have completed so far (typecheck, format, lint-adjacent jobs); a few longer jobs (integration tests, coverage shards) were still in progress at review time.

**Minor concerns**
- Test coverage only updates the branch where `allowedProviderToolNames` is set explicitly (`["web_fetch"]`). The likely real-world regression path — no explicit `allowedToolNames`/`allowedProviderToolNames`, just `sourceProviderToolNames` including `web_fetch` on an Anthropic model — isn't directly exercised for Anthropic (only for OpenAI, where native `web_fetch` doesn't exist anyway, so that branch doesn't actually test the dedup interaction). Worth adding one more case mirroring the existing OpenAI "falls back to local web_fetch" test but for `anthropic/claude-*` with no explicit allow lists, to pin down the default-path behavior this PR is actually meant to fix.
- Behavior change is broader than "don't duplicate": an explicit `allowedProviderToolNames: ["web_fetch"]` request for the *native* tool will now always resolve to the local implementation instead when a local `web_fetch` tool is registered and authorized. That appears intentional per the PR description ("prefer the existing protected local fetch implementation"), but it's a real API/behavior change for any caller that was relying on getting the provider-native tool specifically — worth a one-line callout in the PR description or changelog if this project tracks one, since it's not obvious from the diff alone.

Trivial in size but touches hosted tool-selection/security-relevant logic, so the above is meant as tightening rather than a blocker — nothing here rises to a merge blocker at 85.

---

kojiwakayama:
@codex review

PR #4530, opened 2026-09-19
test(mdx): bound the scanner scaling check by exponent, not by a fixed time ratio

kwakayama:
## Problem

`tests (bun)` failed on main ([run 35476234783](https://github.com/veryfront/veryfront-code/actions/runs/35476234783/job/105986236484)):

```
error: side-effect scanner time grew 8.1x (45.2ms -> 364.6ms) for 4x the input
(fail) transforms/mdx/esm-module-loader/utils/source-spans > keeps large brace-heavy division scans linear across import scanners
```

#4525 replaced the wall-clock budget in these tests with a growth ratio, bounded at 8x for 4x the input. I set that bound from V8 measurements only. Under JavaScriptCore the same **linear** scan grows about 8.1x, so the bound failed a scan that has no regression.

## Fix

The bound is now the growth **exponent** `e` in `time ~ size^e`, which is what the test is actually about and is independent of both the input factor and the engine's constant factors:

| | exponent |
|---|---|
| linear scan, V8 (Deno, Node) | ~0.90 |
| linear scan, JavaScriptCore (Bun), from the failure above | ~1.51 |
| **bound** | **1.75** |
| quadratic scan | 2.0 |

## Verification

- `deno task test:file src/transforms/mdx/esm-module-loader/utils/source-spans.test.ts`: passes (241 steps)
- **Still catches the regression it exists for:** a throwaway quadratic scanner fed through `assertLinearScan` fails with `quadratic scanner time grew 46.4x (0.7ms -> 34.8ms) ... a size^2.77 curve`
- Bun's observed 8.1x ratio maps to 1.51, which now passes; the failure message reports the exponent alongside the ratio
- `deno fmt`, `deno lint`, `deno check` clean; `lint:anti-slop`, `lint:test-semantic-dispositions`, `lint:testing-front-door` pass


kwakayama:
## Code Review — Score: 92/100 (Excellent)

Small, well-reasoned test fix that replaces an engine-specific magic number with a principled measure.

**Strengths:**
- Root-causes the flake correctly: a fixed 8x time-ratio bound baked in V8's constant factor, and JavaScriptCore's linear scan legitimately grows faster (~8.1x). Measuring the growth *exponent* (`e` in `time ~ size^e`) makes the bound engine-independent, which is the right fix rather than just bumping the threshold.
- The new bound (1.75) is well-justified by data in the PR description: ~0.90 for V8's linear scan, ~1.51 for JSC's, comfortably below quadratic's 2.0. I verified the arithmetic (`ln(8.1)/ln(4) ≈ 1.51`) and it checks out.
- PR description documents verification directly, including that a throwaway quadratic scanner still fails the check (measured ~2.77) — good evidence the assertion still catches real regressions, not just a loosened test.
- Minimal, self-contained diff (1 file, +13/-7); no other references to the old `MAX_SCALING_TIME_FACTOR` constant were left behind.
- Failure message is improved (now reports the fitted exponent alongside the raw ratio), which will make future failures easier to diagnose.

**Minor notes (non-blocking):**
- `Math.log(growth)` is `-Infinity` when `largeMillis` rounds to 0 (fully JIT-optimized away), which trivially passes the assertion rather than erroring — same edge-case risk existed before this change via the ratio calc, so not a regression, just noting it's still there.
- PR is currently blocked (`mergeable_state: blocked`, likely on required review) rather than a code issue.

Nice, tightly-scoped fix with the reasoning shown in the description — nothing here needs to change before merge.


---

kwakayama:
@codex review

PR #4534, opened 2026-09-20
fix(observability): report accumulated usage on every agent.run exit

kojiwakayama:
## Root cause

`agent.run` spans reported spend only when a run finished cleanly.

1. **The failure path never had a usage source.** The runtime does not throw on a
   provider/stream error — it emits an SSE `error` frame and closes the stream, so
   `onFinish` never fires and `completedResponse` stays `null`
   (`src/internal-agents/run-stream.ts:1255`). The run then finalises through the
   normal branch with `final_status: "failed"` and
   `buildRuntimeUsageTraceAttributes(undefined)` === `{}`
   (`src/internal-agents/run-stream.ts:1547`). The cancelled
   (`run-stream.ts:1582`) and generic-error (`run-stream.ts:1613`) branches set no
   usage attributes at all.
2. **A second emitter could never report cost.** `src/agent/hosted/agent-run-lifecycle.ts:77`
   starts a span *also named* `agent.run` for `operationName: "chat"`, and its final
   state (`HostedAgentRunSpanFinalState.usage`) was a token-only shape, so
   `buildUsageTraceAttributes` (`src/agent/hosted/trace-attributes.ts:151`) emitted
   `gen_ai.usage.*` and nothing else. Any query over `{ name = "agent.run" }` sums both
   emitters, so every hosted chat run contributed `agent.usage.cost_credits = absent`
   on **every** status, completed included. The cost fields were already present at
   runtime — the finalizer passes a full `ChatMessageMetadata` — and were being dropped
   at the type boundary in `src/agent/hosted/lifecycle.ts:10`.

### Verified against staging, not assumed

The brief's verification run was executed against staging Tempo (`gcx traces`), not
simulated:

```
{ name = "agent.run" && span.run.id = "359dab21-0471-59c7-9ef2-5df586729617" }
-> trace 484778e8d5b812ac6f6f0264b8556442 (veryfront-server, staging)
   agent.run.final_status = failed
   error.type            = Error
   error.message         = "Provider replay turn failed before its boundary"
   (no gen_ai.usage.*, no agent.usage.* attributes at all)
```

The attribute fingerprint (`agent.id`, `run.id`, `thread.id`, `parent.run.id`,
`agent.run.status`, `gen_ai.operation.name`) is the one built by
`buildInternalAgentRunTraceAttributes`, so this run came from the internal-agent
emitter and died on its generic-failure branch — exactly the branch this PR fixes.
The expected 34.8974 credits / 244,019 tokens can only be confirmed after deploy; the
pre-fix ground truth (zero usage attributes) is confirmed.

## What changed

- `src/agent/runtime/index.ts`: new `onUsage` stream callback, fired after every model
  call with a **snapshot** (`{ ...totalUsage }`) of the run's running total.
- `src/internal-agents/run-stream.ts`: records that total and reports it on the
  completed, cancelled and failed exits, all three now reading one `resolveRunUsage()`
  so a future change to `completedResponse.usage` cannot apply to one exit only. Both
  dispatch branches now share **one** callbacks object.
- A truncated total is marked with a new attribute `agent.run.usage_is_floor: true`,
  emitted only when there is a total to qualify.
- `src/agent/hosted/{lifecycle,agent-run-lifecycle,trace-attributes}.ts`: the hosted
  `agent.run` emitter now carries cost. Both emitters build usage attributes with the
  same `buildRuntimeUsageTraceAttributes`.

## Review findings addressed

| Finding | Resolution |
| --- | --- |
| Production wiring had zero regression protection (only the injectable `override` branch was tested) | Both branches now pass the same `runtimeStreamCallbacks` object, so the framework-dispatch wiring cannot be deleted silently. Verified: deleting the `onUsage` entry fails 2 tests. |
| Second `agent.run` emitter (hosted chat) can never emit `cost_credits` | Fixed; new test `reports run cost, not only tokens, on the hosted agent.run span`. |
| `agent.usage.capture_status` overloaded, destroying the provider signal | Fixed; the run-level fact moved to `agent.run.usage_is_floor`. The failed-run test now asserts the provider's `"complete"` survives. |
| (minor) `capture_status: "partial"` set even with no usage | Fixed; the floor marker is emitted only when usage attributes are. New test: `omits the usage floor marker when a cancelled run never reached a model`. |
| (minor) Inconsistent `resolveRunUsage()` vs `accumulatedUsage` between exits | Fixed; all three exits use `resolveRunUsage()`. |
| (minor) Snapshot assertion was unfalsifiable (harness copied the value) | Fixed; the collector keeps the received reference. Verified: replacing `{ ...totalUsage }` with `totalUsage` now fails the test. |

Nothing was declined.

## Tests, seen failing first

Each new assertion was confirmed to fail against the unfixed production code by
mutating it and re-running:

- remove the `onUsage` entry from the shared callbacks → `records usage accumulated
  before a terminal runtime error…` and `…before cancellation…` FAIL.
- `usage: buildHostedAgentRunSpanUsage(metadata)` → `metadata?.usage` → `reports run
  cost, not only tokens, on the hosted agent.run span` FAILS.
- `callbacks?.onUsage?.({ ...totalUsage })` → `(totalUsage)` → `reports the cumulative
  run usage to onUsage after every model call` FAILS.

No existing test was weakened; the deep-equality assertion in `creates a terminal
adapter and finalizes the span` still holds unchanged.

## Gates run

| Gate | Result |
| --- | --- |
| `deno task test:file` on run-stream, agent-run-lifecycle, runtime-stream-usage-callback, trace-attributes, trace-usage, lifecycle, hosted-chat-finalization, chat-execution-runtime, prepared-chat-execution, hosted-terminal | all pass |
| `deno check --no-lock` on all changed files | pass |
| `deno task typecheck` | pass |
| `deno fmt --check` on changed files | pass |
| `deno lint` (src/agent/hosted, src/agent/runtime, src/internal-agents) | pass, 385 files |
| `lint:test-typecheck`, `lint:style`, `lint:anti-slop`, `lint:barrel-jsdoc`, `lint:testing-front-door`, `lint:module-boundaries`, `lint:dependency-boundaries` | pass |
| `docs:api-reference:check` | current |

A whole-directory `deno test` over `src/agent/hosted src/internal-agents
src/agent/conversation` reports 17 failures — the identical 17 on clean `origin/main`
in a scratch worktree (they are `Deno.env.get` reassignment tests that need the repo
suite's isolation, not this change). This branch: 543 passed / 1079 steps;
`origin/main`: 543 passed / 1077 steps.

## What this does NOT close

Definition-of-done items 2 and 3 are **not** met, and #1500 should stay open:

- The ~20% shortfall on *completed* runs is still unexplained. Sampled completed
  `agent.run` spans in staging (3 of 20 in a 36h window) do carry
  `agent.usage.cost_credits` with `capture_status: "complete"`, so the shortfall is not
  "completed spans carry no cost". It needs a per-run comparison against
  `ai_usage.cost_credits`, which needs DB access this branch did not have.
- Runs rejected before the stream opens still emit no span at all.
- The detached hosted chat path (`src/agent/hosted/hosted-chat-finalization.ts`,
  `kind: "detached"`) passes `metadata: undefined`, so its span still reports neither
  tokens nor cost. Fixing it needs a usage extractor over the opaque `finalStep`.

Refs veryfront/veryfront-issue-inbox#1500


kojiwakayama:
CI on `1e21a7bb70`: 44 checks pass, 0 fail. Only `Automated review` is outstanding, and it reports `PR#4534 draft waits for review` — it is gated on the draft status, not on this branch.

One iteration was needed: the first push failed the SonarQube quality gate on **new-code coverage (73.8% < 80%)**, caused by this PR itself — `buildHostedAgentRunSpanUsage` was written as 14 hand-rolled `!== undefined` ternaries and the test only supplied 8 of the fields, so 16 of its 45 new lines never executed. Fixed by making the mapping data-driven over a key list (`satisfies` keeps it tied to both types) and by having the test supply every billing field it claims to map. New-code coverage is now **96.4%** and every quality-gate condition is OK.

No test was weakened to get there — the extra fields are additional assertions, and the fail-first check was re-run after the rewrite (reverting `usage:` to `metadata?.usage` still fails `reports run cost, not only tokens, on the hosted agent.run span`).

kojiwakayama:
## Independent review: 🟠 needs work

**Risk if merged as-is:** medium  |  **Reviewer confidence:** high

### Look at this first
Run this one command before reading anything else: `cd /Users/agent1/code/veryfront/veryfront-code && git grep -n 'operationName: \"chat\"' origin/main -- src | grep -v '\\.test\\.'`. It returns only two TYPE DECLARATIONS (agent-run-lifecycle.ts:56, trace-attributes.ts:182) and zero call sites. That single fact invalidates the stated rationale for roughly half this diff and for the PR's explanation of the remaining gap: the hosted lifecycle never emits a span named `agent.run` in production (no caller passes `operationName: \"chat\"`, none supplies `spanName`, so agent-run-lifecycle.ts:74-77 always yields `invoke_agent <agentName>`), so hosted chat runs were never in the `{ name = \"agent.run\" }` sum the reporter queried and were never diluting it. The PR's new hosted test hand-constructs the non-production `operationName: \"chat\"` configuration and asserts `spanName === \"agent.run\"`, which is precisely why the in-pipeline review did not catch this. Decide from there whether to fix the hosted path properly (span name + the cost-stripping projection in hosted-terminal.ts) or cut it from this PR and merge the internal-agent half alone — that half is genuinely correct and covers the failed (-100%) and cancelled (-92%) buckets, about 314 of the 389 missing credits.

### Blocking
1. The entire hosted half of the diff is dead code, and it is dead for TWO independent reasons — I verified both against origin/main. (a) WRONG SPAN: `git grep -n 'operationName: "chat"' origin/main -- src | grep -v test` returns zero call sites. The only production caller of `createHostedAgentRunSpanController` is chat-execution-runtime.ts:446 with `operationName: "invoke_agent"`, and `spanName` is never supplied (createVeryfrontCloudPreparedHostedChatExecutionRuntimeOptions, cloud-prepared-chat-execution-runtime.ts:83-116, omits it; prepared-chat-execution.ts:107 forwards that undefined). So agent-run-lifecycle.ts:74-77 always resolves to `invoke_agent <agentName>`. Hosted chat spans were NEVER in the `{ name = "agent.run" }` query the issue measures. (b) FIELDS STRIPPED: even if the name were right, `buildHostedAgentRunSpanUsage()` reads billing fields off terminal metadata, but every production dispatch first passes through `toConversationHostedTerminalState()` in src/agent/conversation/hosted-terminal.ts:187-212 — a file this PR does not touch — which rebuilds metadata as `{modelId, usage:{inputTokens,outputTokens,cachedInputTokens}, usageCaptureStatus}` and discards costCredits/costUsd/veryfrontChargeUsd/costSource/billingMode. Both the adapter's `toTerminalState` and its `dispatch` (hosted-terminal.ts:261, :272) go through it. Fix: either delete the hosted half and say so, or fix hosted-terminal.ts AND the span name — a type widening alone changes nothing in production.
2. The commit message and PR body assert a false causal story that will outlive the PR. The code comment now shipped at agent-run-lifecycle.ts:32-37 says 'This span is named \"agent.run\" — the same name the internal-agent path emits — so a spend query over that name sums both emitters.' That is false; only src/internal-agents/run-stream.ts:1270 emits `agent.run` in production. The issue's own counts corroborate it (104 agent_run rows vs 101 agent.run spans — roughly 1:1 with internal-agent runs). Consequence: the PR's explanation for the reporter's completed-run 20% shortfall (74.94 of the 388.72 missing credits) is wrong, so that item is not merely unverified — it is actively misdiagnosed, and the next person will start from a false premise. Rewrite the body/commit message before merge and re-open the completed-run gap as genuinely un-root-caused.
3. The two new hosted tests are false green — triple-insulated from production, textbook coverage padding. `reports run cost, not only tokens, on the hosted agent.run span` (agent-run-lifecycle.test.ts:163) (1) passes `operationName: "chat"`, a value no production call site uses, then asserts `spanName === "agent.run"`; (2) injects `createTerminalAdapter: (input) => createRecordingTerminalAdapter(input)`, whose `toTerminalState` copies `state.metadata` through verbatim (test file lines ~35-51) — a pass-through stand-in for the real projection that strips the cost fields; (3) calls `adapter.terminal.onTerminalState(...)` directly with a metadata shape `toConversationHostedTerminalState` cannot produce, so `dispatchConversationHostedTerminalState` is never exercised. It asserts properties of three test doubles, not of shipped code. Same applies to its sibling `leaves hosted agent.run usage attributes absent...`. Re-point both at `dispatchConversationHostedTerminalState` with the real adapter, or they stay green through any future breakage.
4. `agent.run.usage_is_floor` is gated on the wrong condition, and it is wrong in both directions. The gate should be `completedResponse === null` (i.e. resolveRunUsage() actually fell back to the accumulator), not `state.sawTerminalError`. Confirmed-reachable false POSITIVE: finalizeAgUiEventsUnstamped (src/agent/ag-ui/encoder.ts:1064-1070) sets `sawTerminalError = true` for EMPTY_ASSISTANT_OUTPUT on a run where onFinish already fired with the complete total (a run that produced only tool calls, no visible text) — `finalizeRunEvents` runs at run-stream.ts:1526, before the attribute block, so the run is stamped failed + usage_is_floor on an exact total. Same over-marking on the cancelled (1586) and failed (1617) branches, which mark unconditionally even when `resolveRunUsage()` returned `completedResponse.usage`. One-line change fixes four findings raised across all three reviewers. Worth doing pre-merge because this attribute IS the trust filter the reconciliation work in #1500 is supposed to be built on.

### Non-blocking
- Issue body suggested fix #4 ('Make agent.run.final_status agree with agent_run.status') is untouched AND is the one residual item missing from the PR's otherwise honest 'What this does NOT close' section. It matters for the issue's own verification procedure: run 359dab21 is `cancelled` in agent_run but the PR's new test pins its span to `failed`, so its 34.8974 credits land in the span side's failed bucket against the DB's cancelled bucket — the per-status tables in repro steps 2 vs 3 still cannot reconcile even after this fix.
- Issue body 'Secondary: runs missing entirely' also lists a cancelled/CANCELLED classification-agent run (run_6cf18fa8…) with no span at all, which the PR's 'rejected before the stream opens' (START_REQUEST_REJECTED / PERMISSION_DENIED) explanation does not cover and never mentions.
- Scope widening not mentioned in the PR body: `buildUsageTraceAttributes` now delegates to `buildRuntimeUsageTraceAttributes`, and it is called from the shared `buildFinalizedAgentRunTraceAttributes` (trace-attributes.ts:306), so all 13 `agent.usage.*` billing keys now land on every hosted span finalised through that path — in practice the `invoke_agent <agentName>` spans, including nested delegated sub-agent runs. Not a new exposure class (these fields already reach the chat client via ChatMessageMetadata), but any future trace-level spend roll-up that sums `agent.usage.cost_credits` without filtering by span name will now double count parent + delegated children.
- Hosted runs that die mid-stream (kind: "response") still report no spend at all: `buildChatStreamChunkMessageMetadata` (src/chat/chat-ui-message-helpers.ts:337) attaches usage only when it sees a `finish` part with totalUsage, so `dispatchFailedTerminalError` arrives with usage-free metadata. This is the hosted counterpart of the exact defect the PR fixes for internal agents, and it is not in the not-closed list.
- Efficiency/divergence: each of the three exits calls `resolveRunUsage()` twice and builds the full usage attribute map twice — once for `buildRuntimeUsageTraceAttributes(...)` and again inside `buildRunUsageFloorAttributes(...)`, which calls the builder only to count keys. Hoist `const usage = resolveRunUsage(); const usageAttributes = buildRuntimeUsageTraceAttributes(usage);` and make the marker `Object.keys(usageAttributes).length > 0`.
- src/agent/runtime/runtime-stream-usage-callback.test.ts repeats `assertEquals(seen[0]?.costCredits, 2)` and `assertEquals(seen[0]?.totalTokens, 15)` verbatim before and after the seen[1] block; only `assertEquals(seen[0] === seen[1], false)` is load-bearing. Literal duplication in a repo with a SonarCloud duplication gate.

### Where the reviewers disagreed
Four real conflicts; I checked each against origin/main rather than averaging.

(1) THE BIG ONE — is the hosted half broken? Reviewer 1 says blocker (cost fields stripped at runtime by toConversationHostedTerminalState); Reviewer 2 says major (the span isn't even named `agent.run`); Reviewer 3 saw neither and returned wouldMerge: true. Reviewers 1 and 2 are BOTH right, about two independent defects that compound — I verified each separately. R2's is the more fundamental: fixing R1's stripping alone would put cost on `invoke_agent` spans that the issue's query never selects. R3's clean bill here should be discarded: its lens (contract compatibility) genuinely did not cover call-site reachability, and its other conclusions — attribute-parity superset, no docs:api-reference churn, no testing-front-door baseline changes, no double counting from accumulateUsage — I spot-checked and accept.

(2) THE FLOOR MARKER — R2 calls it major in one direction (completedResponse null + no terminal error ⇒ truncated total published unmarked on a 'completed' run); R1 calls it minor in the OPPOSITE direction (sawTerminalError set by the EMPTY_ASSISTANT_OUTPUT finalizer ⇒ exact total marked as a floor). R1's direction is demonstrably reachable — I confirmed encoder.ts:1064-1070 sets the flag after onFinish already delivered the complete total. R2's direction is NOT proven reachable: R3 is right that the real runtime calls `callbacks?.onFinish?.(response)` unconditionally on the success path (runtime/index.ts:2693), so a genuinely completed in-process run always sets completedResponse. So R2 over-rated its severity, and R3 is right that the accompanying test models a state production cannot reach. But all three are chasing one bad gate: changing `state.sawTerminalError` to `completedResponse === null` fixes R1's minor, R2's major, R1's nit and R3's nit at once.

(3) IS DoD ITEM 2 MET? R1 asserts the PR body contradicts itself by listing item 2 as unmet. All three reviewers independently reached the same conclusion and I confirmed it: `totalUsage` is zero-seeded at runtime/index.ts:3496 and returned as `usage: totalUsage` on both exits (4521, 4538), so `completedResponse.usage` IS the run's accumulated cross-turn total. Item 2 was already satisfied on main; the issue's automated triage brief misdiagnosed it. This matters — it removes the PR's stated reason for the completed-run shortfall and leaves that 20% with no explanation at all.

(4) FAIL-FIRST — R1 verified it empirically by reverting the five non-test source files to origin/main in a scratch worktree and re-running the suites (four tests go red, the two the body honestly labels as guards stay green). R2 and R3 both reported 'could-not-check'. R1's evidence is the only direct one and I credit it: the fail-first claim holds, and the PR is honest about which of its tests are guards rather than regression proofs. Note separately that the body's '17 identical pre-existing failures on origin/main' claim is the known local-dependency-skew shape from this pipeline and remains unverified by anyone.

---
<sub>**Provenance:** independent automated review (Claude Code) via @kojiwakayama's token. Tooling output, not a human approval. Three reviewers examined this diff through separate lenses (correctness, security/authz, contract & test quality); a synthesis pass adjudicated their disagreements against `origin/main` and dropped findings that did not survive. Verify before acting.</sub>

kojiwakayama:
Worked through the audit's four blocking items. I re-verified each against `origin/main`
before touching anything; three were right as stated, one was right about the defect and
wrong about how to reach it, and one proposed remedy I've declined with reasoning.

Pushed as `6efdad2`. PR body rewritten — the old one asserted a causal story that is
false, which is the finding I'd most want a reviewer to see corrected.

---

### 1. The hosted half was dead code — **confirmed, both reasons, fixed**

Verified independently:

```
$ git grep -n 'operationName: "chat"' origin/main -- src | grep -v '\.test\.'
src/agent/hosted/agent-run-lifecycle.ts:56:  operationName: "chat" | "invoke_agent";
src/agent/hosted/trace-attributes.ts:182:  operationName: "chat" | "invoke_agent";
```

Two type declarations, zero call sites. And `createVeryfrontCloudPreparedHostedChatExecutionRuntimeOptions`
never sets `spanName`, which `prepared-chat-execution.ts:107` forwards as `undefined`, so
`agent-run-lifecycle.ts:74-77` always yields `invoke_agent <agentName>`. Confirmed the
stripping too: `toConversationHostedTerminalState` rebuilds metadata as
`{modelId, usage, usageCaptureStatus}` and drops every billing field before it reaches
the span.

**Fixed the stripping** (`hosted-terminal.ts` now carries the billing fields), which is
what made the rest of the hosted diff do anything at all.

**Declined the span rename.** The audit offered "fix the span name" as the other half.
The same controller also spans delegated sub-agent runs, so renaming it to `agent.run`
would make a `{ name = "agent.run" }` spend sum count a parent and its children — worse
for the reconciliation this issue exists to enable than the gap it would close. The
hosted fix is therefore scoped honestly: it makes hosted spend queryable, it does not
move the issue's numbers, and the PR body says exactly that.

I also didn't take the "delete the hosted half" option: the gap it closes is real
(hosted runs reported tokens and no spend on every status), it's just not the gap #1500
measures.

### 2. False causal story in the body and comments — **confirmed, corrected**

`src/internal-agents/run-stream.ts:1270` is the only production emitter of `agent.run`.
The issue's own counts agree (104 `agent_run` rows vs 101 spans, ~1:1 with internal-agent
runs; every run named in the issue is an internal agent).

Corrected in the code comments (`agent-run-lifecycle.ts`, `trace-attributes.ts`,
`lifecycle.ts`) and in the PR body, and the consequence is stated rather than buried:
**the completed-run ~20% shortfall now has no root cause.** The previous explanation is
withdrawn, not downgraded to "unverified". #1500 stays open on it.

I also removed the body's "17 identical pre-existing failures on `origin/main`" claim.
It came from an ad-hoc `deno test` sweep that isn't a CI gate and that nobody reproduced,
so restating it would just launder it.

### 3. Hosted tests were false green — **confirmed, re-pointed at the real path**

Both now drive `dispatchConversationHostedTerminalState` over the **real**
`createConversationHostedTerminalAdapter`, with `operationName: "invoke_agent"` (the
value `chat-execution-runtime.ts` actually passes), and assert the real span name
`invoke_agent Ops Agent`.

Fail-first proof that they now reach production code:

```
$ git show origin/main:src/agent/conversation/hosted-terminal.ts > src/agent/conversation/hosted-terminal.ts
$ deno task test:file src/agent/hosted/agent-run-lifecycle.test.ts
reports run cost, not only tokens, on the hosted agent run span => FAILED
  -   undefined
  +   34.8974
```

That's the check the stubbed version structurally could not make.

### 4. `usage_is_floor` gated on the wrong condition — **defect confirmed, repro corrected**

Gate is now `completedResponse?.usage == null` — "the figure came from the accumulator" —
which is the question the attribute answers. Fixes both directions in one change, and
the hoist means each exit resolves usage and builds the attribute map once instead of
twice.

One correction to the audit's reproduction: it describes the false positive as *"a run
that produced only tool calls, no visible text"*. That case doesn't reach the guard —
`encoder.ts:913` sets `sawVisibleOutput = true` on `tool-input-start`. The reachable case
is an **empty assistant turn**: `onFinish` delivers the exact total, then
`finalizeAgUiEventsUnstamped` raises `EMPTY_ASSISTANT_OUTPUT` and sets `sawTerminalError`,
so the run was stamped failed *and* `usage_is_floor` on a complete figure. The mechanism
and the severity stand; only the illustration was off. New fail-first test
`does not mark a complete total as a floor on a run that failed after finishing` models
that case and is red against `1e21a7b`.

Deliberately **not** changed while I was in there: the earlier commit chose to give this
fact its own attribute rather than overwrite `agent.usage.capture_status`, because that
field is the provider's per-call verdict and has consumers reading it that way
(`src/eval/agent-service.ts`, `src/runtime/runtime-bridge.ts`, `src/chat/protocol.ts`).
That reasoning is sound and survives the gate change, so I kept it.

---

### Non-blocking items

All five are now either fixed or written into "What this does NOT close" rather than left
implicit:

- Efficiency/divergence (double `resolveRunUsage()` + double map build) — **fixed**, same
  change as #4.
- Scope widening / double-count risk on nested delegated sub-agent spans — **documented**
  in the PR body and in `trace-attributes.ts`, so a future roll-up author hits the warning.
- Issue fix #4 (`final_status` vs `agent_run.status`) — added to the not-closed list, with
  the concrete consequence: run `359dab21` is `cancelled` in the DB and `failed` on its
  span, so the issue's own per-status reconciliation still can't be made to agree.
- The cancelled classification-agent run with no span and no start-rejection code — added
  as a *third* distinct gap, not folded into the START_REQUEST_REJECTED explanation.
- Hosted runs dying mid-stream (`kind: "response"`) reporting no spend — added; it's the
  hosted counterpart of exactly what this PR fixes for internal agents.

### Nothing was weakened

No test deleted, no assertion widened, no code removed to satisfy a finding.
`hosted-terminal.test.ts`'s deep-equality assertion on the projected metadata is
unchanged and still green — the projection adds billing keys only when the input carries
them.

Gates: `deno task typecheck`, `deno check --no-lock` on every changed file,
`deno fmt --check`, `deno lint` (424 files), `lint:style`, `lint:anti-slop`,
`lint:barrel-jsdoc`, `lint:testing-front-door`, `lint:module-boundaries`,
`lint:dependency-boundaries`, `lint:test-typecheck`, `lint:test-semantic-dispositions`,
`docs:api-reference:check` — all pass, plus `test:file` on the twelve affected suites.

kwakayama:
## Review score correction: 93/100

Reviewed head: `6efdad2b315d465f50ea0c464d59de1bfabca9b8`

The hosted run finalization path carries complete terminal usage metadata, including provider cost, Veryfront charges, credits, billing mode, and capture status. It reuses the shared runtime usage attribute builder, preserves the production span name `invoke_agent <agentName>`, and keeps absent usage distinguishable from zero usage.

Validation:
- Focused hosted lifecycle, usage-callback, and internal-agent stream tests pass: 3 suites, 94 steps
- CI, coverage, Sonar, CodeQL, lint, format, and typecheck pass
- No unresolved review threads remain
- PR is still draft, so Automated review is waiting on readiness

The PR scores 93/100. Remaining work is readiness and the exact-head review gate, not a source defect.

PR #4537, opened 2026-09-20
ci: do not fail the automated review gate on a security-review quota notice

kojiwakayama:
## Problem

The review bot runs two products on separate quotas: the code review, which this gate exists to require, and an optional security review add-on. When only the security-review quota is exhausted, the bot posts

> You have reached your Codex usage limits for security reviews. Please try again later.

and still delivers the code review for the exact head.

`CODEX_USAGE_LIMIT` in `scripts/ci/automated-review-gate.mjs` matched that notice, so `latestCodexUsageLimit` turned it into a terminal `automated review rate limited` failure. Observed on PR #4536: the code review for the exact head was submitted at 15:41:05Z, and the gate published "rate limited" at 15:41:56Z on the strength of the security-quota comment posted at 15:37:21Z.

## Change

`latestCodexUsageLimit` now skips a usage-limit notice that names security reviews. It is neither a failure nor evidence of anything, so the gate keeps waiting for the code review.

The change is one added regex and one added clause in the existing filter, placed after the pinned-bot identity check.

## What stays fail-closed

- The generic notice (no `for ...` clause) and every other `for ...` clause, including code reviews, stay terminal rate-limit failures.
- The gate still requires the pinned bot's exact-head review evidence.
- Pending reviews still time out into a terminal failure, including behind a security-quota notice.
- Edited, unpinned, and stale comments are still ignored. The pinned bot identity checks, boundary/epoch logic, merge-group logic, and workflow YAML are untouched.

Because `terminalStatusHasBoundaryProof` delegates to the same helper, an already-published terminal "rate limited" status whose proof comment is a security-quota notice stops counting as boundary proof and is no longer preserved.

## Recovery for PRs already stuck

A stuck PR recovers on its own at the next scheduled evaluation whenever the gate can establish an evidence boundary, which is the normal case: the gate's own `@codex review` request comment for the head supplies one, as does a reset status or a lifecycle event. If a PR has no request comment, no reset status, and no lifecycle event, the boundary is undefined, the terminal status is preserved without consulting its proof comment, and that PR needs a new push or a fresh `@codex review`.

Note also that an exact-head code review already recovers a rate-limited status today; the harm this fixes is the gate declaring a terminal failure while the code review is still on its way.

## Tests

`scripts/ci/automated-review-gate.test.ts`: the `codexRateLimitComment` fixture used the security text for every rate-limit test, so it now uses the generic text and every existing fail-closed test keeps proving what it proved. Added, using a new `codexSecurityReviewRateLimitComment` fixture:

- security-quota notice, no review yet, stays pending (was: rate-limited failure)
- security-quota notice as the triggering comment, stays pending (was: rate-limited failure)
- security-quota notice plus valid exact-head proof, success
- security-quota notice plus a pending review older than the timeout, timed-out failure (was: rate-limited failure)
- a `for code reviews` notice, still a terminal rate-limited failure
- a published terminal "rate limited" status proven only by a security-quota notice, no longer terminal (was: preserved as a failure)

Four of these failed against the unmodified script before the fix.

## Security check

**Can anyone other than the pinned bot cause a notice to be ignored, or a review to be accepted?** No. The new regex is reached only after `isPinnedBot(comment?.user, CODEX_LOGIN)` has already passed, so a comment from anyone else never reaches it and was already skipped. The change touches no code in the review-proof path.

**Can an edited comment?** An edited comment can only cause a notice to be ignored, never a review to be accepted, and being ignored means the gate keeps waiting or times out rather than passing. The unedited-comment check still guards the direction that produces a terminal failure.

**Does ignoring the notice open any path from "no code review" to "success"?** No. The usage-limit check runs only inside the `if (!review)` branch, so it can only turn "no proof" into a terminal failure. Removing a reason to fail leaves a pending required status, which blocks merging exactly like a failure, and the 30-minute timeout still converts it into a terminal timed-out failure.

## Verification

| Command | Exit |
| --- | --- |
| `deno task fmt:check` | 0 |
| `deno task lint` | 0 |
| `deno check --config=scripts/test.deno.json scripts/ci/automated-review-gate.test.ts scripts/ci/automated-review-gate.mjs` | 0 |
| `deno task test:file scripts/ci/automated-review-gate.test.ts` | 0 (8 passed, 166 steps) |

The test file carries the workflow contract tests for this gate, and they pass. `scripts/ci/` is excluded from both the repo's fmt and lint configs, so the two touched files are not reformatted here; added lines stay within the repo's 100-column style.

Automated tooling output (Claude Code) posted via the repository owner's token — not the owner speaking.


kwakayama:
## Code Review: 90/100 — Excellent, narrowly-scoped fix

Correctly stops a false "rate limited" failure from a security-review quota notice, while keeping every other failure path intact.

**Strengths**
- The fix is minimal and surgical: one new regex, one added exclusion clause in the existing `if` filter in `latestCodexUsageLimit`, placed after the `isPinnedBot` check so it's unreachable by anyone but the pinned bot account — a forged comment from another user still can't influence the gate.
- I traced the boolean logic in `scripts/ci/automated-review-gate.mjs`: `CODEX_SECURITY_REVIEW_USAGE_LIMIT` is a strict subset of `CODEX_USAGE_LIMIT` (`for security reviews?` vs. the generic `for [^.]+`), so excluding it only removes evidence, it can never flip a "no review" state into "success" — the usage-limit check only runs inside the `if (!review)` branch, matching the PR's own security analysis.
- Because `terminalStatusHasBoundaryProof` reuses the same helper, a previously-published terminal failure whose only proof is a security-quota comment correctly stops being "sticky," letting a stuck PR self-heal at the next scheduled evaluation — a good side benefit of not duplicating the matching logic.
- Test coverage is thorough and specifically targets the regression: pending-vs-security-notice, exact-head proof alongside the notice, timeout still firing behind the notice, a `for code reviews` notice still failing closed, and the terminal-status boundary-proof case. The PR states 4 of the 6 new tests fail against the pre-fix script, which is good evidence the tests are meaningful rather than tautological.
- Commit message and PR description are unusually clear for a CI script change — the "what stays fail-closed" and "Security check" sections preemptively answer the obvious bypass questions (can a non-bot trigger this, can an edited comment, does this open a no-review→success path), and my read of the code confirms those answers.

- I couldn't execute `deno task test:file` myself in this environment (no `deno` binary available) to independently confirm the verification table; the reasoning checks out from static reading, but an independent CI run confirming green is worth waiting for before merge given this touches a security-relevant gate.
- `scripts/ci/` is excluded from fmt/lint per the PR description — worth double-checking that's still intentional/documented elsewhere, since exclusions on CI-critical scripts are easy to forget about later.

No actual weakening of the review-gate's security guarantees found — this narrows a false-positive, not the acceptance criteria for a passing review.


---

kojiwakayama:
@codex review

